PR #14, opened 2015-12-21
Update torture failures

jfbastien:
The following wasmate fix:
  https://github.com/WebAssembly/experimental/pull/82
Makes (import ...) use 'result' instead of 'return', fixing a few failures (some of them were hiding further failures).


binji:
lgtm

PR #19, opened 2016-01-11
Update JS name from "WASM" to "_WASMEXP_"

titzer:
This depends on https://codereview.chromium.org/1577803002/


binji:
v8 needs to be rolled into sexpr-wasm, so better to land that first then roll v8 with this fix.

binji:
I updated v8 and fixed this in bee3f91170d529889dd4d6a561c80abee21db7de. Closing.

PR #21, opened 2016-01-18
Change the memory sizes to be in units of pages rather than powers of two.

JSStats:
This patch would update sexpr-wasm-prototype for the following change to v8:
https://codereview.chromium.org/1597163002/
Issue 1597163002: wasm: change the module memory size to be multiples of the page size, 64k. 

There appears to be consensus to do this and this is one attempt, see:
https://github.com/WebAssembly/design/issues/497


binji:
I'd suggest landing this change in the spec repo first. This repo borrows tests from the spec repo, so those tests will fail if the meaning of memory size is changed.

JSStats:
Shall leave the wast using bytes to avoid a spec change. PR https://github.com/WebAssembly/spec/pull/209 changes the tests to use page aligned memory sizes, and with the v8 patch this passes the checks.

JSStats:
Reworked to use units of page for the wast text too. It's passing most of the tests, just a few from the spec fail but they do not appear to be the responsibility of this patch. The corresponding change has landed in v8 https://codereview.chromium.org/1597163002/

titzer:
lgtm

binji:
Thanks, lgtm!

binji:
https://github.com/WebAssembly/spec/pull/209 hasn't landed yet, but the page-size change landed in v8 before the signed LEB128 i32.const change, so I can't update that here without updating this first. So we'll just put up with some spec breakage until spec 209 lands.

PR #22, opened 2016-01-18
Tables can include imports and their index space includes imports.

JSStats:
If I understand the following issue correctly then the sexpr-wasm-prototype has a bug in the generation of tables, see https://github.com/WebAssembly/binaryen/pull/115 This is an attempt to correct this issue, and includes a small example test.


binji:
Commented in https://github.com/WebAssembly/binaryen/pull/115. The spec interpreter doesn't allow imports in the function table. There is a bug, though: sexpr-wasm should output the function index in the shared import/function index space that v8 expects.

JSStats:
Thank you, I did not realise the spec had this restriction. It's not even tested in the spec testsuite. 
https://github.com/WebAssembly/design/issues/517

sunfishcode:
See discussion in https://github.com/WebAssembly/design/issues/421 and https://github.com/WebAssembly/design/issues/522.

PR #37, opened 2016-03-01
Build on Windows

Cellule:
I made some changes to make this compatible on Windows.
I haven't tested on linux to make sure it still works (don't have a vm at hand right now).
I don't know if we want to support Windows as well, but I figured I could toss this here and see if this is something we want.

The main changes are
- getopt: added a project that builds getopt for Windows
- alloca: simply alias `_alloca` from `malloc.h`
- unistd: Added a file that replaces unistd on Windows
- struct initialization: `MyStruct myStruct = {};` doesn't compile on MSVC. Had to do it the C way ie: `memset(&myStruct, 0, sizeof(MyStruct));`
- `void *` math is not supported, since we go back and forth between `void *` and `size_t`, I made the required changes to use `size_t` instead.


jfbastien:
Hi Michael! Could you join the W3C CG?
https://www.w3.org/community/webassembly/
Thanks :-)

getopt/getopt.c
```diff
@@ -0,0 +1,973 @@
...
+This code is a modification of the Free Software Foundation, Inc.
+Getopt library for parsing command line argument the purpose was
+to provide a Microsoft Visual C friendly derivative. This code
+provides functionality for both Unicode and Multibyte builds.
+
+Date: 02/03/2011 - Ludvik Jerabek - Initial Release
+Version: 1.0
+Comment: Supports getopt, getopt_long, and getopt_long_only
+and POSIXLY_CORRECT environment flag
+License: LGPL
```
jfbastien: I don't think we can compile in something that's under the LGPL :-)

jfbastien:
Maybe the code could move away from getopt instead? I'm biased, but I like binaryen's support/command-line.cpp ;-)

binji:
Well, I'm not too excited about adding C++ code to a C codebase, but yeah, maybe moving away from getopt is a good idea.

jfbastien:
@Cellule is in the CG now.

binji:
OK, well the rest of the changes (aside from getopt) seem good. Perhaps we can land those first and work on an alternate getopt solution in a follow-up CL?

Cellule:
So I removed `getopt` and made a small custom parser for the time being for windows.
There might still be some build errors on Linux. I'll wait for the CI build and verify afterward.

README.rst
```diff
@@ -41,9 +41,9 @@ well. On Debian-based systems::
 
   $ sudo apt-get install bison
   ...
-  $ touch src/wasm-parser.y
+  $ touch src/wasm-bison-parser.y
```
Cellule: There were some errors in the README file that I discovered as I was doing the port. I figured I might as well put the change in this PR since it's a small change.

binji: Thanks, I forgot to fix the README when I made this change. :)

src/sexpr-wasm.c
```diff
@@ -1,269 +1,358 @@
...
+static const char* s_infile;
+static const char* s_outfile;
+static int s_dump_module;
+static int s_verbose;
+static int s_spec;
+static int s_spec_verbose;
+static int s_use_libc_allocator;
+
+#ifndef __GNUC__
+struct option
```
binji: nit: opening brace on this line

src/sexpr-wasm.c
```diff
@@ -1,269 +1,358 @@
...
+{
+    const char* name;
+    int has_arg;
+    int* flag;
+    int val;
+};
+#define	null_argument		0	/*Argument Null*/
```
binji: nit: space after define

binji: nit: no tabs please!

src/sexpr-wasm.c
```diff
@@ -1,269 +1,358 @@
...
+    option_index = -1;
+    const char* in_opt = argv[optind++];
+    i = 0;
+    for (; i < OPTIONS_LENGTH; i++)
    {
        const struct OptionsText* optText = &optionsText[i];
+        const struct option* opt = &s_long_options[i];
+        if (
+            strcmp(in_opt, optText->long_name) == 0 ||
+            (opt->val && strcmp(in_opt, optText->short_name) == 0)
```
binji: nit: opt_text

src/sexpr-wasm.c
```diff
@@ -1,269 +1,358 @@
...
+  exit(0);
+}
+
+static void parse_options(int argc, char** argv) {
+  int c;
+  int option_index = 0;
+#ifdef USE_MIN_PARSER
+  const char* optarg = NULL;
+  int optind = 1;
+  OptionsText optionsText[OPTIONS_LENGTH];
```
binji: nit: lowercase_with_underscores for variable names

src/sexpr-wasm.c
```diff
@@ -1,269 +1,358 @@
...
    {
        const struct OptionsText* optText = &optionsText[i];
+        const struct option* opt = &s_long_options[i];
+        if (
+            strcmp(in_opt, optText->long_name) == 0 ||
+            (opt->val && strcmp(in_opt, optText->short_name) == 0)
+        )
+        {
+            option_index = i;
+            c = opt->val;
```
binji: this won't pass the tests, a number of them use the compressed flag format: "-d -v" => "-dv". But I guess getting the test-runner to work can be another CL.

src/unistd.h
```diff
@@ -0,0 +1,49 @@
+#ifndef _UNISTD_H
```
binji: Add WebAssembly copyright header

src/sexpr-wasm.c
```diff
@@ -1,269 +1,358 @@
...
+
+  if (result == WASM_OK) {
+    result = wasm_check_script(lexer, &script);
+    if (result == WASM_OK) {
+      WasmMemoryWriter writer;
+      memset(&writer, 0, sizeof(WasmMemoryWriter));
+
+      if (wasm_init_mem_writer(&g_wasm_libc_allocator, &writer) != WASM_OK)
+        FATAL("unable to open memory writer for writing\n");
+
```
binji: Can you add a ZERO_MEMORY to wasm-internal.h instead?

src/unistd.h
```diff
@@ -0,0 +1,49 @@
...
+#define lseek _lseek
+/* read, write, and close are NOT being #defined here, because while there are file handle specific versions for Windows, they probably don't work for sockets. You need to look at your app and consider whether to call e.g. closesocket(). */
+
+#define ssize_t int
+
+#define STDIN_FILENO 0
+#define STDOUT_FILENO 1
+#define STDERR_FILENO 2
+/* should be in some equivalent to <sys/types.h> */
+typedef __int8            int8_t;
```
binji: shouldn't be needed; aren't these supplied by stddint.h in Windows?

src/wasm-binary-writer.c
```diff
@@ -1435,7 +1450,7 @@ static void write_module_spec(WasmWriteContext* ctx, WasmModule* module) {
   out_printf(&ctx->spec_writer_state, "};\nvar m = createModule([\n");
   WasmOutputBuffer* module_buf = &ctx->mem_writer->buf;
   const uint8_t* p = module_buf->start;
-  const uint8_t* end = module_buf->start + module_buf->size;
+  const uint8_t* end = (uint8_t*)((size_t)module_buf->start + module_buf->size);
```
binji: just `p + module_buf->size` is fine here

src/wasm-stack-allocator.c
```diff
@@ -45,9 +45,9 @@ static WasmStackAllocatorChunk* allocate_chunk(
       stack_allocator->fallback, real_size, CHUNK_ALIGN);
   if (!chunk)
     return NULL;
-  void* start = (void*)chunk + sizeof(WasmStackAllocatorChunk);
+  void* start = chunk + sizeof(WasmStackAllocatorChunk);
```
binji: should be `chunk + 1`

src/wasm-stack-allocator.c
```diff
@@ -171,7 +171,7 @@ void wasm_reset_stack_allocator(WasmStackAllocator* stack_allocator) {
     chunk = prev;
   }
   stack_allocator->first->current =
-      (void*)stack_allocator->first + sizeof(WasmStackAllocatorChunk);
+      stack_allocator->first + sizeof(WasmStackAllocatorChunk);
```
binji: should be ` + 1`

src/wasm.c
```diff
@@ -120,7 +120,7 @@ void wasm_vfprint_error(FILE* error_file,
         next_line_offset--;
       }
 
-      const size_t max_line = 80;
+#define max_line 80
```
binji: change to `MAX_LINE` for macro

src/alloca.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
```
binji: copyright header

src/unistd.h
Cellule: This code comes from 
http://stackoverflow.com/questions/341817/is-there-a-replacement-for-unistd-h-for-windows-visual-c
I'm not sure how to proceed since it doesn't include a licence

src/sexpr-wasm.c
Cellule: Yes I was aware of this case, but since I don't know if this parser will be the final thing, I figured I wouldn't support the compressed format for the time being.

src/unistd.h
binji: Looks like this is only needed for `isatty` in `wasm-flex-lexer.c`, is that right? If so, just remove this file and do the `#define` there.

src/alloca.h
```diff
@@ -0,0 +1,27 @@
...
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+#ifndef WASM_ALLOCA_H_
+#define WASM_ALLOCA_H_
+
+#ifdef __GNUC__
```
binji: Actually, rather than hiding a system header, could you move this definition and other platform-specific stuff to `wasm-platform.h`?

src/wasm-common.h
```diff
@@ -29,7 +29,15 @@
...
 
+#if defined(__GNUC__) && (__GNUC__ >= 4)
 #define WARN_UNUSED __attribute__ ((warn_unused_result))
+#elif defined(_MSC_VER) && (_MSC_VER >= 1700)
+#define WARN_UNUSED _Check_return_
+#else
+#define WARN_UNUSED
+#endif
+
+#define ZERO_MEMORY(var, type) memset((void*)&var, 0, sizeof(type))
```
binji: this can just be `ZERO_MEMORY(var) memset((void*)&(var), 0, sizeof(var))`

src/alloca.h
Cellule: I don't know if we can. Either I duplicate this where we include `alloca.h` (which is only 2 places) or we include `alloca.h` in `wasm-platform.h` which will include it for everyone.

binji: I'd prefer just including it everywhere, rather than shadowing a system include.

binji:
lgtm, thanks!

PR #40, opened 2016-03-04
Fix a test expected output after the memory-in-pages change.

JSStats:
Thank you for merging the memory-in-pages change. Retesting shows one failure in the expected output when using the pending spec tests.


binji:
lgtm

PR #41, opened 2016-03-07
Local decls rewrite

titzer:
PTAL. The dump tests don't pass, since they assert on bytes. I wasn't sure how to update them. Depends on V8 CL: https://codereview.chromium.org/1763433002/


binji:
Cool, thanks. You can run the tests locally with `test/run-tests.py -e <path to d8>`. The test runner will automatically rebaseline tests for you if you pass `-r`. So if you want to rebaseline the dump tests, just run `test/run-tests.py -e <path to d8> -r dump`.

titzer:
With the rebase of dumps, those tests pass. There are still a couple failures on d8, but those are good ole compiler bugs that i'm puzzled weren't hit before...oh...i know...they are DCHECKs and you run in release mode :-)

PR #46, opened 2016-03-08
Emit explicit arities for calls, breaks, & return

rossberg-chromium:
To be landed in sync with matching changes to V8, see:

https://codereview.chromium.org/1749993003/
https://codereview.chromium.org/1777513002/


binji:
nice, lgtm

binji:
Since we rolled this into the 0xb version, I've since implemented it there: https://github.com/WebAssembly/sexpr-wasm-prototype/commit/0eaa9963f314426acca33987e778ba9ad0b03008

PR #55, opened 2016-03-14
Windows: vsnprintf

Cellule:
Implementation of vsnprintf was under the wrong check on Windows.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 #Visual Studio Products
 Debug
 Release
+build/
 *.VC.opendb
 *.sdf
 *.vcxproj.user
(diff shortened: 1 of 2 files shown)


binji:
Yeah, I wasn't exactly sure about what was broken/fixed. My understanding is that `snprintf`/`vsnprintf` do not null-terminate in older vs2013 and earlier, but that `snprintf` was added in vs2015, and `vsnprintf` was updated to null-terminate as well. So on vs2015, `wasm_sprintf` and `wasm_vsnprint` could just be `#defined` to  `snprintf` and `vsnprintf`. But maybe it's just best to always wrap them on msvc.

binji:
oh, and lgtm :)

PR #56, opened 2016-03-18
Rename the 'not' operator to 'i32.eqz' and add 'i64.eqz'.

JSStats:
It fitted the 'convert' type of operator so this was used - not sure if it's what you would want.


binji:
Looks like a good start, but the build is failing.

The convert type is probably the best bet, yeah -- it looks like the spec repo added a new `TEST` type, but I'd rather not add a new expression type for just this.

Also, I have a CL locally that removes all the operator type stuff and just uses the opcode, FYI

JSStats:
Rebased. The failure looks like the d8 version might be out of sync, and it works locally, or do some 64 bit operators not yet work on d8-32bit and is this used in checks here?

binji:
Oh yeah, the version of d8 is pinned, so it probably doesn't have `i64.eqz`. I'll update it. Also, it looks like you'll need to update the lexer to get `i64.eqz` as well, that's failing on the gcc bot.

binji:
OK, updated v8.

JSStats:
Thank you. Updated. ptal.

src/wasm-common.h
```diff
@@ -142,7 +142,7 @@ enum { WASM_USE_NATURAL_ALIGNMENT = 0xFFFFFFFF };
   V(I32, ___, 0, 0x57, I32_CLZ)             \
   V(I32, ___, 0, 0x58, I32_CTZ)             \
   V(I32, ___, 0, 0x59, I32_POPCNT)          \
-  V(I32, ___, 0, 0x5a, I32_EQZ)             \
+  V(I32, ___, 0, 0x5a, I32_EQZ_I32)         \
```
binji: I'd rather keep the names consistent with the spec names, so `I32_EQZ` and `I64_EQZ`

binji: I'm surprised this works, the convert should use the second column as the type to convert from.

JSStats:
Thank you. Updated. ptal.

binji:
lgtm, thanks

PR #61, opened 2016-03-29
Update README.rst

ossy-szeged:
Readme isn't up-to-date, examples should be updated to work with the recent V8.


binji:
lgtm, thanks!

PR #78, opened 2016-05-26
fix bugs found by afl and asan

ddcc:
There were a couple of integer underflows and overflows that could result in out-of-bounds memory accesses, and some uninitialized memory that could result in bad calls to the free pointer of the memory allocator.

### DIFF
--- a/src/wasm-ast-parser.y
+++ b/src/wasm-ast-parser.y
@@ -1158,7 +1158,7 @@ static void dup_text_list(WasmAllocator* allocator,
      * iterate through the string once. */
     const char* src = node->text.start + 1;
     const char* end = node->text.start + node->text.length - 1;
-    size_t size = end - src;
+    size_t size = (end > src) ? (end - src) : 0;
     total_size += size;
   }
   char* result = wasm_alloc(allocator, total_size, 1);
--- a/src/wasm-binary-reader-interpreter.c
+++ b/src/wasm-binary-reader-interpreter.c
@@ -270,6 +270,8 @@ static WasmResult on_data_segment(uint32_t index,
   Context* ctx = user_data;
   WasmInterpreterMemory* memory = &ctx->module->memory;
   uint8_t* dst_data = memory->data;
+  if ((uint64_t)address + (uint64_t)size > memory->byte_size)
+    return WASM_ERROR;
   memcpy(&dst_data[address], src_data, size);
   return WASM_OK;
 }
@@ -527,15 +529,18 @@ static WasmResult begin_function_body(uint32_t index, void* user_data) {
 
 static WasmResult end_function_body(uint32_t index, void* user_data) {
   Context* ctx = user_data;
-
-  /* discard everything except the last expr */
-  uint32_t discard_max = ctx->expr_stack.size - 1;
+  uint32_t discard_max = 0;
 
   WasmInterpreterFuncSignature* sig =
       get_func_signature(ctx, ctx->current_func);
   if (sig->result_type == WASM_TYPE_VOID) {
-    /* discard last expr too */
-    discard_max++;
+    /* discard all expr */
+    discard_max = ctx->expr_stack.size;
+  } else if (ctx->expr_stack.size > 0) {
+    /* discard everything except the last expr */
+    discard_max = ctx->expr_stack.size - 1;
+  } else {
+    return WASM_ERROR;
   }
 
   uint32_t i;
@@ -1850,6 +1855,8 @@ WasmResult wasm_read_binary_interpreter(WasmAllocator* allocator,
     wasm_steal_mem_writer_output_buffer(&ctx.istream_writer,
                                         &out_module->istream);
     out_module->istream.size = ctx.istream_offset;
+  } else {
+    wasm_close_mem_writer(&ctx.istream_writer);
   }
   destroy_context(&ctx);
   return result;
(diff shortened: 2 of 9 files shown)


src/wasm-ast-parser.y
```diff
@@ -1158,7 +1158,7 @@ static void dup_text_list(WasmAllocator* allocator,
      * iterate through the string once. */
     const char* src = node->text.start + 1;
     const char* end = node->text.start + node->text.length - 1;
-    size_t size = end - src;
+    size_t size = (end > src) ? (end - src) : 0;
```
binji: You'll have to rebuild the prebuilt parser for this change; you can run `make update-bison` to do it.

src/wasm-binary-reader-interpreter.c
```diff
@@ -270,6 +270,7 @@ static WasmResult on_data_segment(uint32_t index,
   Context* ctx = user_data;
   WasmInterpreterMemory* memory = &ctx->module->memory;
   uint8_t* dst_data = memory->data;
+  assert((uint64_t) address + (uint64_t) size <= memory->byte_size);
```
binji: nit: no space between cast and value

src/wasm-binary-reader-interpreter.c
```diff
@@ -528,20 +529,22 @@ static WasmResult begin_function_body(uint32_t index, void* user_data) {
 static WasmResult end_function_body(uint32_t index, void* user_data) {
   Context* ctx = user_data;
 
-  /* discard everything except the last expr */
-  uint32_t discard_max = ctx->expr_stack.size - 1;
+  if (ctx->expr_stack.size > 0) {
```
binji: This is probably an error case, do you have the test that causes this?

binji:
Thanks for this! A few things: have you joined the [community group](https://www.w3.org/community/webassembly/)? Also, you mention that you found these via AFL, do you have the test cases that were produced that caused these failures? It would be nice to add them as regression tests.

It looks like the Travis failure you have is due to not updating the prebuilt Bison parser, AFAICT.

ddcc:
Yeah, I've joined the community group through Google. I have a couple of the test cases, but they are WASM files, and it looks like the current test infrastructure is only set up to run WAST. Additionally, they fail validation by the disassembler/assembler.

binji:
You can kinda write wasm files by hand (see the `test/binary/*` tests). It's an ad-hoc format designed to be just one step up from writing binary data directly, but you should be able to write bytes directly too.

No worries if this is too tedious to do, the fixes you have here look good already. I just wouldn't want them to break again. :) Fortunately, it looks like the new asserts should help w/ that.

test/binary/bad-function-body-size.txt
```diff
@@ -0,0 +1,18 @@
...
+;;; TOOL: run-gen-wasm-interp
+magic
+version
+section("type") { count[1] function params[0] results[1] i32 }
+section("function") { count[1] sig[0] }
+section("code") {
+  0x01 ;; 1 function body
+  0x01 ;; malformed 1 byte function body size (should be 5)
+  0x00 ;; 0 local entries
+  0x10 0x2A ;; 1 int32 with value 42
```
binji: you can use the standard mnemonics here, `i32.const 42` is fine

test/binary/bad-function-body-size.txt
```diff
@@ -0,0 +1,18 @@
+;;; ERROR: 1
+;;; TOOL: run-gen-wasm-interp
+magic
+version
+section("type") { count[1] function params[0] results[1] i32 }
+section("function") { count[1] sig[0] }
+section("code") {
+  0x01 ;; 1 function body
```
binji: I like the comments, but you can also use an arbitrary string as a comment: `count[1]` is what I've done in other files

src/wasm-binary-reader-interpreter.c
```diff
@@ -527,15 +529,17 @@ static WasmResult begin_function_body(uint32_t index, void* user_data) {
...
   Context* ctx = user_data;
-
-  /* discard everything except the last expr */
-  uint32_t discard_max = ctx->expr_stack.size - 1;
+  uint32_t discard_max = 0;
 
   WasmInterpreterFuncSignature* sig =
       get_func_signature(ctx, ctx->current_func);
   if (sig->result_type == WASM_TYPE_VOID) {
     /* discard last expr too */
```
binji: This comment makes less sense with the other comment removed. Can you either add new comments or remove this one?

test/run-gen-wasm-interp.py
```diff
@@ -0,0 +1,78 @@
...
+
+SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
+GEN_WASM_PY = os.path.join(SCRIPT_DIR, 'gen-wasm.py')
+
+
+def main(args):
+  parser = argparse.ArgumentParser()
+  parser.add_argument('-o', '--out-dir', metavar='PATH',
+                      help='output directory for files.')
+  parser.add_argument('-e', '--executable', metavar='PATH',
```
binji: remove, this doesn't use sexpr-wasm

binji:
OK, I took a look at the memory leak. It looks like the `WasmMemoryWriter` allocated at `wasm-binary-reader-interpreter.c:1846` is leaking when the binary reading fails. It's because the writer is normally given to the output module on success via `wasm_steal_mem_writer_output_buffer`. So the solution is to add this to the else branch:

```
  } else {
    wasm_close_mem_writer(&ctx.istream_writer);
  }
```

Also, it looks like you broke `run-tests.py` by removing the `--out-dir` option in `run-gen-wasm-interp.py`.

binji:
Awesome, lgtm

PR #84, opened 2016-06-07
Make return arity 1 only if function return type is not void

### LINKED ISSUE #82: Return nop arity
For the following function 
`(func $return-nop (return (nop)))` (https://github.com/WebAssembly/testsuite/blob/master/functions.wast)

There is a conflict between the function type and return arity.
The type of the function is 
```
; type 0
000000f: 40                                         ; function form
0000010: 00                                         ; num params
0000011: 00                                         ; num results
```
And the function body is the following
```
; function body 6
00001a0: 00                                         ; func body size (guess)
00001a1: 00                                         ; local decl count
00001a2: 00                                         ; OPCODE_NOP
00001a3: 09                                         ; OPCODE_RETURN
00001a4: 01                                         ; return arity
00001a0: 04                                         ; FIXUP func body size
```

Since `nop` is: [an empty operator that does not yield a value](https://github.com/WebAssembly/design/blob/master/AstSemantics.md#control-flow-structures)
then the arity of the return statement should be 0 thus having a valid type.

I found this by running the tests, I am surprised this didn't come up in the CI

Cellule:
Fixes #82 
Determine the arity of the return opcode from the return type instead of the presence of an expression next to it.
Fix the faulty baseline for spec/functions.txt

### DIFF
--- a/src/wasm-binary-writer.c
+++ b/src/wasm-binary-writer.c
@@ -650,7 +650,8 @@ static void write_expr(WasmContext* ctx,
     case WASM_EXPR_TYPE_RETURN:
       write_expr_opt(ctx, module, func, expr->return_.expr);
       write_opcode(&ctx->stream, WASM_OPCODE_RETURN);
-      wasm_write_u8(&ctx->stream, expr->return_.expr ? 1 : 0, "return arity");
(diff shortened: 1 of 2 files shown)


Cellule:
No longer needed by https://github.com/WebAssembly/design/pull/741

PR #87, opened 2016-06-08
Teach test scripts about ch.exe

Cellule:
Added support for ch.exe (run-js.py and polyfills in js scripts)
Todo, currently no script to build/download ch.exe it has to be built separately.


binji:
lgtm, thanks!

Issue #42, opened 2016-03-07
Add flag to display commands used to run d8 via `run-d8.py`


binji:

binji:
Not needed, now that d8 tests have been removed from the repo.

PR #88, opened 2016-06-08
Tests verbosity

### LINKED ISSUE #44: Easier method to debug broken d8 tests
Currently if d8 is broken, I just run manually:
```
out/sexpr-wasm <broken test> -o foo.wasm
gdb --args <d8> test/wasm.js -- foo.wasm
```

Probably should describe this process in the README. Maybe make it more convenient?

Cellule:
Implements #42 
Closes #44 ?

Added an option to run tests in series.
Added an option to pause at first error and prompts to retry the command with verbose (in series only) then prompts if should continue testing.
Prints the last command executed on failing tests in the summary to help reproduce the error.

Theses options helps to reproduced and see what is causing a test to fail

### DIFF
--- a/test/run-tests.py
+++ b/test/run-tests.py
@@ -46,15 +46,23 @@ SLOW_TIMEOUT_MULTIPLIER = 2
 # default configurations for tests
 TOOLS = {
   'sexpr-wasm': {
-    'EXE': '%(sexpr-wasm)s'
+    'EXE': '%(sexpr-wasm)s',
+    'VERBOSE-FLAGS': ['-v']
   },
   'run-js': {
     'EXE': 'test/run-js.py',
     'FLAGS': ' '.join([
       '-e', '%(sexpr-wasm)s',
       '--js-executable=%(js)s',
+      '-o', '%(out_dir)s',
       '--no-error-cmdline',
-    ])
+    ]),
+    'VERBOSE-FLAGS': [
+      ' '.join([
+        '--print-cmd',
+      ]),
+      '-v'
+    ]
   },
   'run-js-spec': {
     'EXE': 'test/run-js.py',
@@ -63,7 +71,14 @@ TOOLS = {
       '--js-executable=%(js)s',
       '--spec',
       '--no-error-cmdline',
-    ])
+      '-o', '%(out_dir)s',
+    ]),
+    'VERBOSE-FLAGS': [
+      ' '.join([
+        '--print-cmd',
+      ]),
+      '-v'
+    ]
   },
   'run-roundtrip': {
     'EXE': 'test/run-roundtrip.py',
@@ -72,7 +87,14 @@ TOOLS = {
       '-e', '%(sexpr-wasm)s',
       '--wasm-wast-executable=%(wasm-wast)s',
       '--no-error-cmdline',
-    ])
+      '-o', '%(out_dir)s',
+    ]),
+    'VERBOSE-FLAGS': [
+      ' '.join([
+        '--print-cmd',
+      ]),
+      '-v'
+    ]
   },
   'run-interp': {
     'EXE': 'test/run-interp.py',
@@ -81,7 +103,14 @@ TOOLS = {
       '--wasm-interp-executable=%(wasm-interp)s',
       '--run-all-exports',
       '--no-error-cmdline',
-    ])
+      '-o', '%(out_dir)s',
+    ]),
+    'VERBOSE-FLAGS': [
+      ' '.join([
+        '--print-cmd',
+      ]),
+      '-v'
+    ]
   },
   'run-interp-spec': {
     'EXE': 'test/run-interp.py',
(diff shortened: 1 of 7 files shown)


test/run-tests.py
```diff
@@ -533,6 +592,10 @@ def main(args):
                       help='override wasm-interp executable.')
   parser.add_argument('-v', '--verbose', help='print more diagnotic messages.',
                       action='store_true')
+  parser.add_argument('--no-multithreaded', help='Run all tests in series',
```
binji: I think it would be nicer to just use `options.jobs == 1` instead of this.

test/run-tests.py
```diff
@@ -587,22 +650,19 @@ def main(args):
 
   status = Status(options.verbose)
   infos = GetAllTestInfo(test_names, status)
-
-  inq = multiprocessing.Queue()
-  test_count = 0
+  infosToRun = []
```
binji: nit: snake_case for all variables

test/run-tests.py
```diff
@@ -615,28 +675,66 @@ def main(args):
...
+          continuedErrors += 1
+    else:
+      for info in infosToRun:
+        result = RunTest(info, options, variables, out_dir)
+        HandleTestResult(status, info, result, options.rebase)
+        if status.failed > continuedErrors:
+          if options.stop:
+            rerunVerbose = raw_input('Rerun with verbose option? (yes/no): ')
+            if len(rerunVerbose) != 0 and rerunVerbose[0] != 'n':
+              RunTest(info, options, variables, out_dir, 2)
```
binji: last arg (`2`) is not clear without looking at the definition. Usually I make it explicit in that case:
```RunTest(info, options, variables, out_dir, verbose_level=2)```

Or maybe use constants? `VERBOSE_2` or something.

File: test/run-tests.py [hunk 4]
```diff
@@ -615,28 +675,66 @@ def main(args):
...
-    while finished_tests < test_count:
-      try:
-        info, result = outq.get(True, 0.01)
-      except Queue.Empty:
-        status.UpdateTimer()
-        continue
+        finished_tests += 1
+        HandleTestResult(status, info, result, options.rebase)
+        if options.stop and status.failed > continuedErrors:
+          shouldContinue = raw_input('Continue testing? (yes/no): ')
```
binji: Not a big fan of interactive scripts, but if you find it useful, OK. Would like this to be refactored into a `YesNo` function returning a bool though.

test/run-tests.py
```diff
@@ -615,28 +675,66 @@ def main(args):
...
-    for i, p in enumerate(range(num_proc)):
-      args = (i, options, variables, inq, outq, out_dir)
-      proc = multiprocessing.Process(target=ProcessWorker, args=args)
-      proc.start()
+    if not options.no_multithreaded:
+      inq = multiprocessing.Queue()
+      outq = multiprocessing.Queue()
+      for info in infosToRun:
+        inq.put_nowait(info)
+      runWorkers = multiprocessing.Value('i', 1)
```
binji: This confused me at first, how about `running` or `is_running` instead?

test/run-tests.py
```diff
@@ -615,28 +675,66 @@ def main(args):
...
+      for i, p in enumerate(range(num_proc)):
+        args = (i, options, variables, inq, outq, out_dir, runWorkers)
+        proc = multiprocessing.Process(target=ProcessWorker, args=args)
+        allProcs.append(proc)
+        proc.start()
+      inq.close();
```
binji: nit: remove ;

test/run-tests.py
```diff
@@ -615,28 +675,66 @@ def main(args):
...
 
-      finished_tests += 1
-      HandleTestResult(status, info, result, options.rebase)
   except KeyboardInterrupt:
     while multiprocessing.active_children():
       time.sleep(0.1)
   finally:
-    while multiprocessing.active_children():
-      time.sleep(0.1)
+    for proc in allProcs:
```
binji: Need the same for `KeyboardInterrupt` above, probably should refactor it into a function.

test/run-tests.py
```diff
@@ -533,6 +592,10 @@ def main(args):
                       help='override wasm-interp executable.')
   parser.add_argument('-v', '--verbose', help='print more diagnotic messages.',
                       action='store_true')
+  parser.add_argument('--no-multithreaded', help='Run all tests in series',
+                      action='store_true')
+  parser.add_argument('--stop', help='Stop on first error',
```
binji: Mention in the help that on error this enters an interactive mode. I almost want to call it `--stop-interactive`... :)

test/run-tests.py
```diff
@@ -615,28 +675,66 @@ def main(args):
...
+  variables['out_dir'] = os.path.abspath(out_dir)
+
+  allProcs = []
+  continuedErrors = 0
   try:
-    for i, p in enumerate(range(num_proc)):
-      args = (i, options, variables, inq, outq, out_dir)
-      proc = multiprocessing.Process(target=ProcessWorker, args=args)
-      proc.start()
+    if not options.no_multithreaded:
```
binji: This is complex enough now that I'd prefer it the two cases pulled into their own functions. Maybe `RunMultiprocess` and `RunSingleProcess`?

test/run-tests.py
```diff
@@ -615,28 +675,66 @@ def main(args):
...
     while multiprocessing.active_children():
       time.sleep(0.1)
   finally:
-    while multiprocessing.active_children():
-      time.sleep(0.1)
+    for proc in allProcs:
+      if proc.is_alive():
+        proc.join(5)
+        if proc.is_alive():
+          proc.terminate();
```
binji: nit: remove ;

binji:
Thanks, a few suggestions but otherwise looks good. Looking forward to having `--print-cmd` :)

test/run-tests.py
Cellule: I think running is misleading, this doesn't tell if the worker are running, but tells if the workers should run.
How about `do_work` ?

test/run-tests.py
Cellule: Good idea

test/run-tests.py
binji: Hm, maybe this is just a common thing I've done then:
```
bool running = true;
while (running) {
  ...
  if (something) running = false;
  ...
}
```
Seems like the same idea.

I don't really like `do_work` because it sounds like a function (same issue I have with `runWorkers`). Maybe `should_continue` or `should_run`?

Cellule: I like `should_run`

binji:
Sorry, should have reviewed this before. :-}
lgtm

PR #89, opened 2016-06-13
remove JavaScript engine testing

### LINKED ISSUE #79: Non-community group submodules
One thing we've recently been thinking a lot about on Chakra is our testing story for WebAssembly. I think it would be ideal if we could use this as a submodule, and have our testing build wasts using sexpr-wasm on the fly. This would allow us to run WebAssembly tests through our CI without needing to check in a whole bunch of binary files.

However, what makes me hesitant to do this is that V8 is a submodule of sexpr-wasm, and (for obvious reasons) I'd rather not have V8 as a submodule of ChakraCore. So what do you guys think? Feasible to remove that dependency? Also, I'm wondering how V8 handles wasm testing? Presumably you don't have this as a submodule either (that seems like there would be some weird recursion), so do you check in .wasm files, or something else?

binji:
It's not really the appropriate place to handle it. Now that we have
wasm-wast and wasm-interp, we can at least verify that the tools are
internally consistent.

Resolves #79.


.gitmodules
```diff
@@ -1,9 +1,6 @@
 [submodule "third_party/testsuite"]
 	path = third_party/testsuite
 	url = https://github.com/WebAssembly/testsuite
-[submodule "third_party/v8/v8"]
-	path = third_party/v8/v8
-	url = https://chromium.googlesource.com/v8/v8.git
 [submodule "third_party/gtest"]
```
Cellule: I don't fully understand the purpose of `gtest`.
It looks like it is used to build `hexfloat_test`, which I can't find what this is used for.

binji: My understanding is that older versions of MSVC (even 2013, I think?) don't support parsing hexadecimal floats via `strtod`. I implemented this (see `wasm-literal.c`), and wanted to add tests for it. `hexfloat_test` does exhaustive roundtrip testing of floats, and tests 4 billion doubles too. They don't run on Travis because they're kind of slow.

Cellule: I see, thanks for the clarification

Cellule:
This looks good to me, pretty much the same changes I made when testing on my side.
:+1:

PR #90, opened 2016-06-17
Fix tests after testsuite update

binji:
* Modify AST parser to disallow any other order than `param`, `result`,
  `local`.
* Allow parsing all var integers as int64 (the lookups will fail later.)
* Fix interpreter logic determining whether to adjust the stack; we were
  correctly handling the case where `unreachable` was being used in an
  expression (the value stack is adjusted "up" as if a value was
  actually produced), but the same behavior is required for `br`,
  `br_table` and `return`.


binji:
Can't land this yet, need to wait for https://github.com/WebAssembly/testsuite to update.

binji:
testsuite updated, landing.

Cellule:
The testsuite has been updated today

PR #97, opened 2016-08-23
Fix error message in the generated invoke function

gahaas:
The error message wanted to print some file name and line number, but these values are not passed to invoke(). Additionally I propose to increase the passed and failed counters in invoke.


binji:
lgtm, thanks!

PR #102, opened 2016-09-20
PR for Rename sexpr-wasm binaries #100

citisolo:
Hi , I have renamed the binaries  to the suggested names and created some symlinks that keep the testsuite (sexpr-wasm -> wast2wasm and wasm-wast -> wasm2wast) in-tact

PR #104, opened 2016-09-22
Fix broken 'run-tests' target

sbc100:
This is followup to (#102) which renamed the executable
files.  It propagated the rename the tools and test
runner.

It also fixes the 'run-tests' target which was passing
the wrong argument name to the run_tests.py script.

### DIFF
--- a/test/run-tests.py
+++ b/test/run-tests.py
@@ -50,16 +50,16 @@ SLOW_TIMEOUT_MULTIPLIER = 2
 
 # default configurations for tests
 TOOLS = {
-  'sexpr-wasm': {
-    'EXE': '%(sexpr-wasm)s',
+  'wast2wasm': {
+    'EXE': '%(wast2wasm)s',
     'VERBOSE-FLAGS': ['-v']
   },
   'run-roundtrip': {
     'EXE': 'test/run-roundtrip.py',
     'FLAGS': ' '.join([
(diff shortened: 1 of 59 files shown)


test/run-tests.py
```diff
@@ -73,7 +73,7 @@
   'run-interp': {
     'EXE': 'test/run-interp.py',
     'FLAGS': ' '.join([
-      '-e', '%(sexpr-wasm)s',
+      '-e', '%(wast2wasm)s',
```
binji: you could change this `-e` as well

test/run-tests.py
```diff
@@ -89,7 +89,7 @@
   'run-interp-spec': {
     'EXE': 'test/run-interp.py',
     'FLAGS': ' '.join([
-      '-e', '%(sexpr-wasm)s',
+      '-e', '%(wast2wasm)s',
```
binji: this one too

sbc100: Done

binji:
Seems broken on Travis: `scripts/travis-test.sh: line 56: SEXPR2WASM: unbound variable`

sbc100:
Good to go now.  Can you take one more look since I made a few more changed to make the bots happy.

binji:
lgtm!

PR #122, opened 2016-09-28
update spec tests to HEAD

binji:


sbc100:
I assume this is HEAD of the binary_0xc branch of https://github.com/WebAssembly/testsuite which in turn matches the contents of the HEAD of the binary_0xc branch of the spec repo?

binji:
Yep, I'll update the comment to make that clearer.

PR #124, opened 2016-09-29
fix ASAN and LSAN builds

binji:


binji:
@sbc100 @KarlSchimpf

src/wasm-binary-writer-spec.c
```diff
@@ -284,7 +284,8 @@ static WasmFunc* append_nullary_func(WasmAllocator* allocator,
   WasmModuleField* func_field =
       append_module_field(allocator, module, WASM_MODULE_FIELD_TYPE_FUNC);
   WasmFunc* func = &func_field->func;
-  func->decl.flags = WASM_FUNC_DECLARATION_FLAG_HAS_FUNC_TYPE;
+  func->decl.flags = WASM_FUNC_DECLARATION_FLAG_HAS_FUNC_TYPE |
+                     WASM_FUNC_DECLARATION_FLAG_SHARED_SIGNATURE;
```
sbc100: Looks like most of this change is about freeing memory.  How is this related?

src/wasm-ast-parser.y
```diff
@@ -1242,21 +1249,6 @@ raw_module :
...
-            is_empty_signature(&import->func.decl.sig)) {
-          WasmFuncType* func_type =
-              wasm_get_func_type_by_var($4, &import->func.decl.type_var);
-          if (func_type) {
-            import->func.decl.sig = func_type->sig;
-            import->func.decl.flags |=
-                WASM_FUNC_DECLARATION_FLAG_SHARED_SIGNATURE;
-          }
-        }
-      }
```
sbc100: What is this code not needed anymore?

binji: This is pretty subtle; the functions in the import array are actually already included in the funcs array, so this behavior is actually duplicated.

src/wasm-binary-writer-spec.c
binji: the shared flag is set whenever the signature is owned by the module. This is true most of the time; the only time it's not is when the signature is first being created inside the parser, and has been added to the module yet. It's a bit of a hack, honestly, and I've refactored it a few times but haven't really made it much cleaner. I think I know how to clean it up, but it would be a bigger change, and probably shouldn't be tackled until the tests all pass.

PR #125, opened 2016-09-29
Fix initialization errors in Travis for read_table and read_memory.

KarlSchimpf:


sbc100:
What tool is it that finds these?  Are they actually used before they are written to?

KarlSchimpf:
The gcc tools (on Travis) generate these warnings, which are then converted to errors.

sbc100:
Interesting, are they just being conservative and saying that *all* locals must be initialized?   I don't know what ben's policy has been so far for stuff like this.  @binji?

KarlSchimpf:
I looked a little harder.

I just realized that in_u32_leb128 has an else clause RAISE_ERROR() that doesn't assign out_value. Hence the problem. I will fix by fixing that case instead!

KarlSchimpf:
Ok, fixing the else case in in_u32_leb128 to assign the output argument fixes the problem. Please review again.

KarlSchimpf:
The remaining problems are more "import" tests that are failing.

binji:
lgtm

PR #128, opened 2016-09-29
fix interp tests

binji:
* elem segment entries should be func_index not sig_index
* change the call_stack and value_stack tops to pointers instead of
  integers; this is more convenient for the interpreter, and means we
  don't have to convert back and forth between the two when calling to
  helper functions (e.g. `call_import`)
* pass dummy results `WasmInterpreterTypedValueVector` to
  `run_export_wrapper`, it's required now
* rename all uses of `vs_` -> `value_stack_`, `cs_` -> `call_stack_`
* destroy the func_imports array when destroying the
  `WasmInterpreterModule`.

### DIFF
--- a/src/wasm-binary-reader-interpreter.c
+++ b/src/wasm-binary-reader-interpreter.c
@@ -133,23 +133,13 @@ typedef struct Label {
 } Label;
 WASM_DEFINE_VECTOR(label, Label);
 
-typedef struct InterpreterFunc {
-  uint32_t sig_index;
-  uint32_t offset;
-  uint32_t local_decl_count;
-  uint32_t local_count;
-  WasmTypeVector param_and_local_types;
-} InterpreterFunc;
-WASM_DEFINE_ARRAY(interpreter_func, InterpreterFunc);
-
 typedef struct Context {
   WasmAllocator* allocator;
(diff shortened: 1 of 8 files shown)


binji:
Ah, just realized I merged some other stuff into this commit, so the description doesn't have all the changes.

binji:
@sbc100 @KarlSchimpf take a look at the most recent commit description for more info

binji:
ping

src/wasm-binary-reader-interpreter.c
```diff
@@ -454,10 +448,69 @@ static WasmResult on_import_global(uint32_t index,
...
+
+  for (i = 0; i < ctx->module->imports.size; ++i) {
+    WasmInterpreterImport* import = &ctx->module->imports.data[i];
+    switch (import->kind) {
+      case WASM_EXTERNAL_KIND_FUNC:
+        ctx->module->func_imports.data[i] = import;
+        break;
+
+      case WASM_EXTERNAL_KIND_TABLE:
+        /* TODO */
```
sbc100: assert?

binji: Done.

PR #130, opened 2016-09-29
fix a bug when checking whether a segment is valid

binji:
A segment has to fit in the initial size, not the max size. Also, when
building on a 32-bit arch, the offset must be cast to a 64-bit int so
the overflow check works properly.

### DIFF
--- a/src/wasm-ast-checker.c
+++ b/src/wasm-ast-checker.c
@@ -1059,12 +1059,10 @@ static void check_elem_segments(Context* ctx, const WasmModule* module) {
             offset, last_end);
       }
 
-      uint64_t max =
-          table->elem_limits.has_max ? table->elem_limits.initial : UINT32_MAX;
-
-      if (offset + elem_segment->vars.size > max) {
+      uint64_t max = table->elem_limits.initial;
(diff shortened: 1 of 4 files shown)


src/wasm-ast-checker.c
```diff
@@ -1059,12 +1059,11 @@ static void check_elem_segments(Context* ctx, const WasmModule* module) {
...
 
-      uint64_t max =
-          table->elem_limits.has_max ? table->elem_limits.initial : UINT32_MAX;
-
-      if (offset + elem_segment->vars.size > max) {
+      uint64_t max = table->elem_limits.initial;
+      if ((uint64_t)offset + elem_segment->vars.size > max) {
         print_error(ctx, CHECK_STYLE_FULL, &field->loc,
-                    "segment ends past the end of table max size (%" PRIu64 ")",
+                    "segment ends past the end of table initial size (%" PRIu64
```
sbc100: The grammar seems a little off to me here.  How about just `segment ends past the end of the table`?   `segment ends outside the initial table table size`?

binji: How about this?

PR #135, opened 2016-09-30
fix typechecking bug in `check_n_types`

binji:
Given a type stack and a block signature , check_n_types should check
that the signature is at the top of the stack. For example:

type stack:       [... , i32, f32, i32]
block signature:  [f32, i32]

A block signature with > 1 types is not currently allowed, but is
supported in the code.

The previous code was wrong, and only checked that the signature was at
the bottom of the type stack, in reverse.


binji:
@sbc100 @KarlSchimpf

PR #141, opened 2016-10-02
fix displayed index when writing out wast files

binji:
Also:

* change block comment to "label" instead of "exit"
* remove text parameter to write_end_block, it's always "end"
* write decimal float values for float constants (it's easier to read
  than hex floats)
* fix assertion in begin_global_init_expr, it wasn't taking into account
  imported globals
* add some new tests that show the correct indexing behavior when
  importing funcs, tables, memories and globals

### DIFF
--- a/src/wasm-ast-writer.c
+++ b/src/wasm-ast-writer.c
@@ -288,32 +288,29 @@ static void write_expr(Context* ctx, const WasmExpr* expr);
 static void write_begin_block(Context* ctx,
                               const WasmBlock* block,
                               const char* text) {
-  if (text)
-    write_puts_space(ctx, text);
+  write_puts_space(ctx, text);
   WasmBool has_label =
       write_string_slice_opt(ctx, &block->label, NEXT_CHAR_SPACE);
   write_types(ctx, &block->sig, NULL);
   if (!has_label)
-    writef(ctx, " ;; exit = @%d", ctx->depth);
+    writef(ctx, " ;; label = @%d", ctx->depth);
   write_newline(ctx, FORCE_NEWLINE);
   ctx->depth++;
   indent(ctx);
 }
 
-static void write_end_block(Context* ctx, const char* text) {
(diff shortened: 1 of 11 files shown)


src/wasm-ast-writer.c
```diff
@@ -335,6 +332,9 @@ static void write_const(Context* ctx, const WasmConst* const_) {
       char buffer[128];
       wasm_write_float_hex(buffer, 128, const_->f32_bits);
       write_puts_space(ctx, buffer);
+      float f32;
```
KarlSchimpf: Why not just use a cast, such as the following:
float* f2 = (float*)&const_->f32_bits;

and then writef(..., *f2)

This would avoid the memcpy.

src/wasm-ast-writer.c
```diff
@@ -344,6 +344,9 @@ static void write_const(Context* ctx, const WasmConst* const_) {
       char buffer[128];
       wasm_write_double_hex(buffer, 128, const_->f64_bits);
       write_puts_space(ctx, buffer);
+      double f64;
+      memcpy(&f64, &const_->f64_bits, sizeof(f64));
+      writef(ctx, "(;=%g;)", f64);
```
KarlSchimpf: Same here.

src/wasm-ast-writer.c
binji: Type-punning always makes me a bit nervous, even though it's probably OK. Anyway, taking a look at the disassembly, gcc (at -O1) and clang (at -O0) optimize away the memcpy here already.

KarlSchimpf: By using memcpy, aren't you already type-punning (by converting FOO* to void* implicitly). Anyway, I don't consider this a show-stopper.

lgtm.

binji: Yeah, I suppose so. But my understanding is that this is the only legitimate way to do so without invoking UB (summoning @jfbastien).

PR #142, opened 2016-10-03
implement assert_malformed

binji:
Checking for malformed modules is almost the same as checking for
invalid modules, the difference being that assert_malformed only tries
to decode the binary format, but doesn't do any other validation.

* Renamed check_assert_invalid -> check_assert_invalid_and_malformed
* Add wasm_default_assert_malformed_source_error_callback, which is the
  same as the assert_invalid one, but prints out "assert_malformed"
  instead
* Add some basic tests for assert_malformed

### DIFF
--- a/test/help/sexpr-wasm.txt
+++ b/test/help/sexpr-wasm.txt
@@ -22,14 +22,14 @@ examples:
   $ wast2wasm spec-test.wast --spec -o spec-test.json
 
 options:
-  -v, --verbose                        use multiple times for more info
-  -h, --help                           print this help message
-  -d, --dump-module                    print a hexdump of the module to stdout
-  -o, --output=FILE                    output file for the generated binary format
-      --spec                           parse a file with multiple modules and assertions, like the spec tests
-      --use-libc-allocator             use malloc, free, etc. instead of stack allocator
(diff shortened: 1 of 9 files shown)


test/help/sexpr-wasm.txt
```diff
@@ -22,14 +22,14 @@ examples:
...
+  -v, --verbose                                      use multiple times for more info
+  -h, --help                                         print this help message
+  -d, --dump-module                                  print a hexdump of the module to stdout
+  -o, --output=FILE                                  output file for the generated binary format
+      --spec                                         parse a file with multiple modules and assertions, like the spec tests
+      --use-libc-allocator                           use malloc, free, etc. instead of stack allocator
+      --no-canonicalize-leb128s                      Write all LEB128 sizes as 5-bytes instead of their minimal size
+      --debug-names                                  Write debug names to the generated binary file
+      --no-check                                     Don't check for invalid modules
+      --no-check-assert-invalid-and-malformed        Don't run the assert_invalid or assert_malformed checks
```
sbc100: This argument name seems excessively long..

binji: yeah, but like getopt_long you can abbreviate arguments as long as they're not ambiguous, so `--no-check-assert` (or even `no-check-`) works too.

PR #144, opened 2016-10-04
fix bug defining a global by an imported global

### LINKED ISSUE #143: Global usage error
I tried to build [emscripten's hello world wast](https://gist.github.com/kripken/661d24692e183f423390e39263607157) into wasm, but I get errors like

````
a.out.wast:65:3: global can only be defined in terms of a previously defined global.
  (global $STACKTOP (mut i32) (get_global $STACKTOP$asm2wasm$import))
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
````

The relevant lines are 

````
  (import "env" "STACKTOP" (global $STACKTOP$asm2wasm$import i32))
[..]
  (global $STACKTOP (mut i32) (get_global $STACKTOP$asm2wasm$import))
````

and that error was on the last line there.

I think that wast code should be valid? We import `$STACKTOP$asm2wasm$import`, then use it to init a mutable global (asm2wasm needs mutable imports, so first we import, then assign to a mutable since imports can't be mutable).

binji:
Because the index space of globals merges imported and defined locals,
the checker's Context::current_global_index must be incremented for when
checking both.

Fixes #143.

### DIFF
--- a/src/wasm-ast-checker.c
+++ b/src/wasm-ast-checker.c
@@ -1121,6 +1121,7 @@ static void check_import(Context* ctx,
       break;
     case WASM_EXTERNAL_KIND_GLOBAL:
       check_global(ctx, loc, &import->global);
+      ctx->current_global_index++;
       break;
     case WASM_NUM_EXTERNAL_KINDS:
       assert(0);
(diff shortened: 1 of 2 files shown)


binji:
@sbc100 @KarlSchimpf

PR #145, opened 2016-10-05
Fix crash bug when -d specified without -v

sbc100:
Initialize the s_log_stream in all cases since its
used with -v and -d both.

Add a test for -d on its own.


binji:
oops! lgtm

PR #147, opened 2016-10-05
Add WasmBinaryReaderContext type which is passed to reader callback

sbc100:
This includes the user_data pointer that was previously
passed but also includes information about the current
reader context such as the offset with in the binary.

### DIFF
--- a/src/wasm-binary-reader.c
+++ b/src/wasm-binary-reader.c
@@ -43,6 +43,23 @@ typedef uint32_t Uint32;
 WASM_DEFINE_VECTOR(type, WasmType)
 WASM_DEFINE_VECTOR(uint32, Uint32);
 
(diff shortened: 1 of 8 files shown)


sbc100:
@binji does this approach look reasonable?  I've only changed on_error and begin_signature_section callbacks so far.. what to check I'm on the right track.  (doing this without copying the data would require a bigger refactor I think).

sbc100:
@binji PTAL

src/wasm-binary-reader.c
```diff
@@ -116,11 +146,19 @@ typedef struct LoggingContext {
   int indent;
 } LoggingContext;
 
+static WasmBinaryReaderContext* GetUserContext(Context* ctx) {
```
binji: get_user_context

PR #148, opened 2016-10-06
Fix most of the spec tests

binji:
The only ones that are failing are imports, exports, linking and start.

Fixes:
* Make sure to assign loc_ for WasmVar
* Imports must occur before any definition, not just a definition of the
  same kind
* Check that load/store alignment is not larger than natural alignment
* Always check type stack at the end of the check_block; that way an
  empty block is still validated
* Only allow one memory or table, including imports
* Don't allow importing or exporting mutable globals
* Loop label signature is for the fallthrough at the bottom, not the
  branch target. This was implemented properly in the AST checker, but
  not in binary-reader-interpreter
* `top_type_is_any` will check if there is ANY anywhere on the type
  stack; previously this check did not look past the top label's type
  stack limit
* `drop_types_for_return` may be called without having enough values on
  the type stack; for example, at the end of a function the ends with
  return
* properly handle cleaning up the type stack for the interpreter when
  branching to the implicit function label
* rename invoke -> action a few places


binji:
@KarlSchimpf @sbc100 

This is a pretty big PR, but a lot of the bulk is just rebasing the tests. I can split it up if that would make it easier to review, though.

src/prebuilt/wasm-ast-parser-gen.c
```diff
@@ -146,6 +146,7 @@
       APPEND_FIELD_TO_LIST(module, export_field, EXPORT, export_, loc_, \
```
KarlSchimpf: I'm curious why this file is checked in (it is a generated source file)?

binji: Yeah, I'd rather not, but it makes it easier for developers who don't have bison + re2c to compile. By default it is not used.

PR #150, opened 2016-10-06
Refactor all uses of s_opcode_ to use functions

binji:
@KarlSchimpf


src/wasm-binary-writer.c
```diff
@@ -48,11 +48,6 @@ static const char* s_type_names[] = {
 };
 WASM_STATIC_ASSERT(WASM_ARRAY_SIZE(s_type_names) == WASM_NUM_TYPES);
 
-#define V(rtype, type1, type2, mem_size, code, NAME, text) \
-  [code] = "OPCODE_" #NAME,
-static const char* s_opcode_name[] = {WASM_FOREACH_OPCODE(V)};
-#undef V
-
 #define V(NAME, code) [code] = #NAME,
 static const char* s_section_name[] = {WASM_FOREACH_BINARY_SECTION(V)};
```
KarlSchimpf: Did you forget this one?

binji: No, that defines the binary section names. I think that is only used in this file.

PR #151, opened 2016-10-06
Allow adding elem segments to imported table

binji:
This was allowed by the AST checker, but not the binary reader.

This fixes another bug parsing emscripten's hello world wast (see issue #143).


binji:
So fast. :-) Thanks!

PR #155, opened 2016-10-07
Fix start spec test

binji:
* Rename call_import -> call_host
* Move host func callback to WasmInterpreterThread
* Export `call_host` as `wasm_call_host`
* In general, use function indexes instead of function offsets into the
  instruction stream
* modify `run_function` to use a function index, so it works with host
  functions or defined functions


binji:
Yeah, a little better... :-)

PR #157, opened 2016-10-07
Fix emscripten build to work with 0xc

binji:
Also remove all the interpreter stuff. It's still in flux.


binji:
@KarlSchimpf @sbc100

binji:
ping

PR #163, opened 2016-10-13
Update testsuite and fix tests

binji:
* `offset=` and `align=` can now specify hex values
* data and elem segment offsets are now not checked; they can be out of
  order or overlapping, and even out-of-bounds if the size is zero
* added the new tests `skip-stack-guard-page` and `unwind`


binji:
@sbc100 @KarlSchimpf

PR #165, opened 2016-10-13
Allow failures on the OSX build

binji:
They OSX builds are very slow to run; it's good to test them, but
they're unlikely to differ from the Linux builds significantly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,3 +41,10 @@ matrix:
       env: SANITIZER=ubsan
     - os: osx
       compiler: clang
+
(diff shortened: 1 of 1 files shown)


sbc100:
Why did you close this?

binji:
I pushed this manually here: https://github.com/WebAssembly/wabt/commit/7d9d479a2c9fc3eceb8ce8fd2c0636772eb68dac. But I think I rebased it, so it has a different SHA so GH didn't realize they're the same.

PR #178, opened 2016-10-21
Refactor interpreter for linking support

binji:
* Create WasmInterpreterEnvironment to hold all modules, functions,
  globals, tables and memories
* There are now three index spaces: "module", "environment" and
  "defined".
  - The "module" index space matches up w/ the normal WebAssembly index
    space, which is distinct for functions, globals, tables, etc.
  - The "environment" index space is a combined index space for all loaded
    modules.
  - The "defined" index space only includes defined objects, not
    imported objects. This is used, for example, when iterating over the
    code section; the function bodies are only specified for defined
    functions.
* A thread is implicitly associated with a module and environment, so
  simplify many function signatures to remove those additional arguments
* Any importable kind can be imported from a host module. Unfortunately,
  since the spec tests require polymorphic imports, the importing
  mechanism must be callback-based
* When interpreting spec tests, the environment and all modules must be
  retained throughout
* Move the binding hash code out of wasm-ast.{c,h} =>
  wasm-binding-hash.{c,h}
* Add wasm_init_output_buffer, for initializing the environment's
  instruction stream independently from a WasmMemoryWriter
* Add wasm_init_mem_writer_existing for initializing a WasmMemoryWriter
  given an existing WasmOutputBuffer
* Add predefined "spectest" module, with a generic function import.
  Still need to implement the spectest table, memory and global imports

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -205,9 +205,10 @@ add_library(libwasm STATIC
   src/wasm-binary-writer.c
   src/wasm-binary-writer-spec.c
   src/wasm-binary-reader-ast.c
+  src/wasm-binding-hash.c
   src/wasm-ast-writer.c
-  src/wasm-binary-reader-interpreter.c
   src/wasm-interpreter.c
+  src/wasm-binary-reader-interpreter.c
   src/wasm-apply-names.c
   src/wasm-generate-names.c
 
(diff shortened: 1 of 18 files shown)


binji:
@sbc100 @KarlSchimpf

PR #179, opened 2016-10-21
Create module_bindings hash in WasmScript

binji:
Actions have an optional module_var, which can reference any module, not
just the last one.


binji:
this depends on https://github.com/WebAssembly/wabt/pull/178

src/tools/wasm-interp.c
```diff
@@ -441,86 +401,148 @@ static void run_all_exports(WasmAllocator* allocator,
...
+                                       WasmInterpreterFuncSignature* sig,
+                                       void* user_data) {
+  func->host.callback = default_host_callback;
+  return WASM_OK;
+}
+
+static WasmResult spectest_import_table(WasmInterpreterImport* import,
+                                        WasmInterpreterTable* table,
+                                        void* user_data) {
+  return WASM_ERROR;
```
KarlSchimpf: I assume that the next few functions are not yet implemented (and hence return WASM_ERROR)?

src/wasm-binding-hash.c
```diff
@@ -0,0 +1,175 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "wasm-binding-hash.h"
+
+#define INITIAL_HASH_CAPACITY 8
+
+static size_t hash_name(const WasmStringSlice* name) {
```
KarlSchimpf: Too bad we aren't using C++. Hash functions/tables would already be implemented!

binji: :-)

PR #180, opened 2016-10-22
Update tests syntax to new flat syntax

citisolo:
more tests

### DIFF
--- a/test/interp/nested-if.txt
+++ b/test/interp/nested-if.txt
@@ -9,8 +9,8 @@
   (func (export "f") (result i32)
     block $exit
       i32.const 1
-      if 
-        i32.const 2           ;; outer if
+      if                     ;; outer if
+        i32.const 2          
         drop
         i32.const 3
         if                   ;; inner if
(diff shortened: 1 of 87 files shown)


test/interp/nested-if.txt
```diff
@@ -7,14 +7,18 @@
...
-        (i32.const 1)
-        (drop (i32.const 2))
-        (if                   ;; inner if
-          (i32.const 3)
-          (br $exit))))
-    (i32.const 4)))
+    block $exit
+      i32.const 1
+      if 
+        i32.const 2           ;; outer if
```
binji: move comment up to if

binji:
lgtm. By the way, next time you push a change, can you update it so it doesn't include the previous commits? It would make reviewing a bit easier. Thanks!

citisolo:
Sorry I didn't realize I was doing that.  How would I update it , what I've been doing is just pushing to the branch?

binji:
I think you just need to update your branch to `wabt:master`.

PR #183, opened 2016-10-24
Add reserved immediate to three operators

binji:
`call_indirect`, `grow_memory` and `current_memory` all take an additional
immediate which must be zero. After the MVP, this will be used to allow
specifying a table or memory index.


binji:
Depends on #182. Take a look a the last commit for now.

PR #190, opened 2016-10-26
Update tests syntax to new flat syntax

citisolo:
converted tests from parse/asserts


test/parse/assert/assertinvalid.txt
```diff
@@ -1,14 +1,14 @@
 ;;; FLAGS: --spec
 (assert_invalid
   (module
-    (func
-      (drop (i32.const 1)))
+    (func (i32.const 1)
```
binji: should be:
i32.const 1
drop

test/parse/assert/assertreturn.txt
```diff
@@ -1,8 +1,10 @@
 ;;; FLAGS: --spec
 (module
-  (func $foo (result i32) (i32.const 0))
+  (func $foo (result i32)
```
binji: this is fine, but it's OK to keep these as a single line too:
(func $foo (result i32) i32.const 0)

binji:
thanks!

PR #192, opened 2016-10-28
wasmdump improvments

sbc100:
- Split into three different passes: headers, details, and raw
- Nest imports correctly
- Show element count as part of section header rather than on its
  own line

### DIFF
--- a/src/tools/wasmdump.c
+++ b/src/tools/wasmdump.c
@@ -141,30 +141,30 @@ int main(int argc, char** argv) {
 
   // Pass 1: Print the section headers
   if (s_objdump_options.headers) {
-    s_objdump_options.mode = DUMP_HEADERS;
+    s_objdump_options.mode = WASM_DUMP_HEADERS;
     result = wasm_read_binary_objdump(allocator, data, size, &s_objdump_options);
-    if (result)
+    if (WASM_FAILED(result))
       goto done;
     s_objdump_options.print_header = 0;
   }
   // Pass 2: Print extra information based on section type
   if (s_objdump_options.verbose || s_objdump_options.section_name) {
-    s_objdump_options.mode = DUMP_DETAILS;
+    s_objdump_options.mode = WASM_DUMP_DETAILS;
     result = wasm_read_binary_objdump(allocator, data, size, &s_objdump_options);
-    if (result)
+    if (WASM_FAILED(result))
       goto done;
     s_objdump_options.print_header = 0;
   }
   if (s_objdump_options.disassemble) {
-    s_objdump_options.mode = DUMP_DISASSEMBLE;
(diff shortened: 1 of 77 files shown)


binji:
Hm, might need to come up with a flag to prevent printing the filename or something; the dump tests are all broken.

src/tools/wasmdump.c
```diff
@@ -127,11 +130,45 @@ int main(int argc, char** argv) {
...
+  // Perform serveral passed over the binary in order to print out different
+  // types of information.
+
+  s_objdump_options.print_header = 1;
+
+  // Pass 1: Print the section headers
+  if (s_objdump_options.headers) {
+    s_objdump_options.mode = DUMP_HEADERS;
+    result = wasm_read_binary_objdump(allocator, data, size, &s_objdump_options);
+    if (result)
```
binji: use this instead:
```
if (WASM_FAILED(result))
```

src/wasm-binary-reader-objdump.c
```diff
@@ -31,11 +32,21 @@ typedef struct Context {
   size_t current_opcode_offset;
   size_t last_opcode_end;
   int indent_level;
+  int print_details;
```
binji: I usually use `WasmBool` (and `WASM_{TRUE,FALSE}`) instead of int for boolean values

src/wasm-binary-reader-objdump.h
```diff
@@ -24,15 +24,27 @@ struct WasmAllocator;
 struct WasmModule;
 struct WasmReadBinaryOptions;
 
+typedef enum WasmObjdumpMode {
+  DUMP_HEADERS,
```
binji: prefix with `WASM_` when in a header.

src/wasm-binary-reader-objdump.c
sbc100: Done

src/wasm-binary-reader-objdump.h
sbc100: Done

src/tools/wasmdump.c
sbc100: Done

sbc100:
Ended up only printing the basename to avoid the testing issue.

binji:
Looks like there is a compile failure in the format string for 32-bit:
```
/home/travis/build/WebAssembly/wabt/src/wasm-binary-reader-objdump.c: In function ‘on_import_table’:
/home/travis/build/WebAssembly/wabt/src/wasm-binary-reader-objdump.c:380:7: error: format ‘%zx’ expects argument of type ‘size_t’, but argument 4 has type ‘uint64_t’ [-Werror=format=]
       elem_limits->max);
       ^
```

PR #197, opened 2016-10-31
Convert WasmType into an enum

sbc100:
Fix the bug in logging_begin_global that was exposed by the
using an explictly type here.


binji:
lgtm

PR #199, opened 2016-11-01
Update tests syntax to new flat syntax more tests

citisolo:
another batch of tests converted to the new flat syntax

### DIFF
--- a/test/interp/callimport-zero-args.txt
+++ b/test/interp/callimport-zero-args.txt
@@ -5,7 +5,7 @@
   (func (export "f") (result i32)
     i32.const 13
     call $imported
-    i32.add)
+    i32.add))
 (;; STDOUT ;;;
 called host spectest.print() => (i32:0)
 f() => i32:13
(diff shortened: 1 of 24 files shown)


binji:
Sorry to bug you about this, but it still is showing previous commits. I think you'll need to do something like this (assuming your upstream remote points to webassembly/wabt):
```
git checkout master
git reset upstream/master --hard
git checkout update_tests_syntax_to_new_flat_syntax_more_tests
git rebase master
```

Not 100% sure about that, but the basic idea should get you to a state where your master branch is aligned w/ wabt/master.

citisolo:
Terribly sorry about this, I'll do that right away, how do I check if everything is straight ?

binji:
No problem, thanks again for rewriting all these tests!

I guess there are a few things you can check. If you run `git log upstream/master...` it shouldn't display anything. This means that your master is the same as the upstream master. Then when you run `git log master...` from your test branch, you should just see the two commits you added most recently. Finally, when you upload the PR, it should show just those two commits, and none of the previous ones you already landed. Hope that helps!

citisolo:
seems like somethings still wrong bare with me , trying to sort it out

PR #200, opened 2016-11-02
Use a new format for the spec JSON writer/parser

binji:
The previous spec JSON format was defined around modules. This is
because the previous spec tests would only run assertions on the most
recently read module. In addition, the previous spec writer would write
the assertions as new exported functions in the module, and run those.
The primary reason for doing this was to allow for passing/returning i64
values, which was necessary to test in a JavaScript host.

Now that the primary host for running the spec tests is wasm-interp, we
no longer need do bundle assertions into the module. Also, some of the
new spec tests allow running exported functions on a module that is not
the most-recently-read module.

The new spec test format is now defined around commands. The commands
map directly to the spec format commands, e.g. `module`,
`assert_invalid`, `assert_trap`, etc.


binji:
@sbc100 @KarlSchimpf

binji:
rebased to master

binji:
I believe Sam and Karl aren't in today. I'll land, and fix issues in later PRs.

PR #202, opened 2016-11-03
more parse tests converted #139

citisolo:
deleted the old branch and started afresh with this new branch , a few tests to start off with


binji:
Thanks, lgtm!

PR #205, opened 2016-11-07
Convert tests to new flat syntax

citisolo:
submitting more tests, tried to make it as clean as possible , please highlight any issues with the commit


binji:
lgtm

PR #206, opened 2016-11-07
Simplify the default error handler

binji:
These changes are useful for handling errors via the default binary
error handler in wasm-interp.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -142,47 +142,53 @@ static void print_source_error(FILE* out,
   }
 }
 
+static void print_error_header(FILE* out, WasmDefaultErrorHandlerInfo* info) {
+  if (info && info->header) {
+    switch (info->print_header) {
+      case WASM_PRINT_ERROR_HEADER_NEVER:
+        break;
+
+      case WASM_PRINT_ERROR_HEADER_ONCE:
+        info->print_header = WASM_PRINT_ERROR_HEADER_NEVER;
+        /* Fallthrough. */
+
+      case WASM_PRINT_ERROR_HEADER_ALWAYS:
+        fprintf(out, "%s:\n", info->header);
+        break;
+    }
+    /* If there's a header, indent the following message. */
+    fprintf(out, "  ");
+  }
+}
+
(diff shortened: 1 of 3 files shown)


src/common.c
```diff
@@ -142,47 +142,48 @@ static void print_source_error(FILE* out,
...
-    const WasmLocation* loc,
-    const char* error,
-    const char* source_line,
-    size_t source_line_length,
-    size_t source_line_column_offset,
-    void* user_data) {
-  fprintf(stdout, "assert_malformed error:\n  ");
-  print_source_error(stdout, loc, error, source_line, source_line_length,
+  WasmDefaultErrorHandlerInfo* info = user_data;
+  FILE* out = info && info->out_file ? info->out_file : stderr;
```
KarlSchimpf: Is it possible to enclose this in a macro/function, so that the (conditional) code doesn't have to be duplicated on every use.

src/common.c
```diff
@@ -142,47 +142,48 @@ static void print_source_error(FILE* out,
...
+  print_error_header(out, info);
+  print_source_error(out, loc, error, source_line, source_line_length,
                      source_line_column_offset);
 }
 
 void wasm_default_binary_error_callback(uint32_t offset,
                                         const char* error,
                                         void* user_data) {
+  WasmDefaultErrorHandlerInfo* info = user_data;
+  FILE* out = info && info->out_file ? info->out_file : stderr;
```
KarlSchimpf: Like here!

src/common.c
```diff
@@ -142,47 +142,53 @@ static void print_source_error(FILE* out,
...
-    const WasmLocation* loc,
-    const char* error,
-    const char* source_line,
-    size_t source_line_length,
-    size_t source_line_column_offset,
-    void* user_data) {
-  fprintf(stdout, "assert_malformed error:\n  ");
-  print_source_error(stdout, loc, error, source_line, source_line_length,
+  WasmDefaultErrorHandlerInfo* info = user_data;
+  FILE* out = get_default_error_handler_info_output_file(info);
```
KarlSchimpf: Looks good!

PR #208, opened 2016-11-07
Fix spec imports test

binji:
* Properly handle registering an unnamed module (most recently read) in
  AST parser
* register command has optional name in spec JSON format
* Implement spectest_import_{table,memory,global}
* Implement on_register_command in wasm-interp (wires up module binding)
* Implement assert_unlinkable
* Set module's table_index and memory_index when importing table/memory
* Translate {get,set}_global index from module -> environment index
  space
* Don't close the environment's istream writer if the module read fails;
  this will destroy the output buffer. Just leave it as is, with extra
  unused data in the environment
* Add registered_module_bindings hash; this has the registered name of
  the module (not to be confused with the module name, which is internal
  only)
* Add {GET,SET}_GLOBAL opcodes to wasm_trace_pc and wasm_disassemble
* Add new trap when calling a function via call_indirect, when the table
  element hasn't been initialized (via an elem segment)
* Check the limits of tables + memories when importing
* Check the function signature when importing
* Use callback for printing errors from a host import; this lets us
  redirect the output the same way other errors are
* Add environment "marks". Similar to allocator marks, allows you to
  backup to a previous state when an operation fails (such as
  successfully reading a module that was supposed to fail)

### DIFF
--- a/src/tools/wasm-interp.c
+++ b/src/tools/wasm-interp.c
@@ -470,18 +470,22 @@ static void run_all_exports(WasmAllocator* allocator,
 static WasmResult read_module(WasmAllocator* allocator,
                               const char* module_filename,
                               WasmInterpreterEnvironment* env,
+                              WasmBinaryErrorHandler* error_handler,
                               WasmInterpreterModule** out_module,
                               WasmInterpreterThread* out_thread) {
   WasmResult result;
   void* data;
   size_t size;
+
   WASM_ZERO_MEMORY(*out_thread);
+  *out_module = NULL;
+
   result = wasm_read_file(allocator, module_filename, &data, &size);
   if (WASM_SUCCEEDED(result)) {
     WasmAllocator* memory_allocator = &g_wasm_libc_allocator;
     result = wasm_read_binary_interpreter(allocator, memory_allocator, env,
                                           data, size, &s_read_binary_options,
-                                          &s_error_handler, out_module);
+                                          error_handler, out_module);
 
     if (WASM_SUCCEEDED(result)) {
       if (s_verbose)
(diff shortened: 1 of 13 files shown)


src/tools/wasm-interp.c
```diff
@@ -521,30 +525,112 @@ static WasmResult default_host_callback(const WasmInterpreterFunc* func,
   return WASM_OK;
 }
 
+static WasmBool string_slice_equals_str(const WasmStringSlice* ss,
+                                        const char* s) {
+  return strncmp(ss->start, s, ss->length) == 0;
```
KarlSchimpf: Can a WasmStringSlice be NULL?

src/tools/wasm-interp.c
```diff
@@ -939,26 +1037,38 @@ static WasmResult on_module_command(Context* ctx,
...
-    ctx->last_module->name = dup_name;
-    WasmBinding* binding = wasm_insert_binding(
-        ctx->allocator, &ctx->env.module_bindings, &dup_name);
-    binding->index = ctx->env.modules.size - 1;
+  WasmInterpreterThread thread;
+  WasmInterpreterEnvironmentMark mark =
+      wasm_mark_interpreter_environment(&ctx->env);
+  WasmResult result = read_module(ctx->allocator, path, &ctx->env,
+                                  &s_error_handler, &ctx->last_module, &thread);
+  wasm_free(ctx->allocator, path);
```
KarlSchimpf: Is it safe to free? It appears to be used on line 1048.

src/tools/wasm-interp.c
```diff
@@ -1037,13 +1147,78 @@ static WasmResult on_assert_malformed_command(Context* ctx,
...
+
+  ctx->total++;
+  char* path = create_module_path(ctx, filename);
+  WasmInterpreterThread thread;
+  WasmInterpreterEnvironmentMark mark =
+      wasm_mark_interpreter_environment(&ctx->env);
+  WasmResult result = read_module(ctx->allocator, path, &ctx->env,
+                                  &error_handler, &ctx->last_module, &thread);
+  wasm_reset_interpreter_environment_to_mark(ctx->allocator, &ctx->env, mark);
+  wasm_free(ctx->allocator, path);
```
KarlSchimpf: Again, is it safe to free path since it is used on line 1213?

src/tools/wasm-interp.c
binji: No, I added an assert.

src/tools/wasm-interp.c
binji: Whoops :)

src/tools/wasm-interp.c
binji: :)

binji:
Fixed all bot failures and rebased, PTAL.

PR #211, opened 2016-11-08
Convert tests to new flat syntax

citisolo:
submitting more tests, I can see old commits in the history , just wondering  if they end up syncing when the pr is merged or if I should squash these ?


test/parse/expr/load.txt
```diff
@@ -1,16 +1,47 @@
...
+    i32.const 0
+    i64.load32_u
+    drop
+    i32.const 0
+    f32.load
+    drop
+    i32.const 0
+    f64.load
+    drop))
+
```
binji: please remove empty lines at the end of the file

test/parse/expr/return-if.txt
```diff
@@ -1,7 +1,9 @@
 (module
   (func (result i32)
-    (return
-      (if i32
-        (i32.const 1)
-        (i32.const 2)
-        (i32.const 3)))))
+    (i32.const 1)
```
binji: remove parens

PR #218, opened 2016-11-14
Add wast_resolve_names_* as a separate pass

binji:
This used to be handled by wasm_check_ast, but is clunkier than having a
separate pass.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -212,6 +212,7 @@ add_library(libwasm STATIC
   src/binary-reader-interpreter.c
   src/apply-names.c
   src/generate-names.c
+  src/resolve-names.c
 
   src/allocator.c
   src/common.c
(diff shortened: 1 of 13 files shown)


binji:
@sbc100 @KarlSchimpf ptal (looks like the failure on travis is a flake)

binji:
Yeah, I think so. ast-validator? or maybe just ast-validate? I've been naming the files as nouns (reader/writer/etc.) but perhaps it makes less sense in this case.

sbc100:
Why the ast- prefix on the filename?  alternatively, why isn't the new file called ast-resolve-names?

binji:
Hm, good point. I guess I added it originally for the parser and lexer. But yeah, probably not needed for anything else, it's reasonable to assume that they work on the AST.

sbc100:
Happy for this to land as-is, with any renaming as followup.

PR #226, opened 2016-11-23
Wasmdump improvments

### LINKED ISSUE #167: wasmdump cleanup
dschuff's feedback:

```
if (ctx->options->headers) {
+    printf("%-12s start=%#010" PRIzx " end=%#010" PRIzx " (size=%#010" PRIzx
```

I think the size would probably be more readable in decimal than hex.
Actually, maybe start and end, too? IIRC the JS engines report byte offsets in Decimal in error messages stacktraces.

```
+static WasmResult on_import_func(uint32_t index,
+                                 uint32_t sig_index,
```

Might be nice to print the index for imports like the rest of the things.
Or maybe even better, the combined index space (imports+defined) for globals and functions?

```
+static WasmResult on_import_global(uint32_t index,
```

Might be nice to print the type here.

```
+static WasmResult on_export(uint32_t index,
+                            WasmExternalKind kind,
+                            uint32_t item_index,
```

it's not really clear what this index means.

sbc100:
fixes #167


src/binary-reader-objdump.c
```diff
@@ -366,16 +390,22 @@ static WasmResult on_import(uint32_t index,
...
 
 static WasmResult on_import_func(uint32_t index,
                                  uint32_t sig_index,
                                  void* user_data) {
-  print_details(user_data, " - func sig=%d\n", sig_index);
+  Context* ctx = user_data;
+  print_details(
+      user_data, " - func[%d] sig=%d " PRIstringslice "." PRIstringslice "\n",
+      ctx->function_index, sig_index,
+      WASM_PRINTF_STRING_SLICE_ARG(ctx->import_module_name),
```
binji: Actually, the previous way is better; we should print the module and field names for all imported types, not just functions.

src/binary-reader-objdump.c
```diff
@@ -402,7 +432,7 @@ static WasmResult on_import_global(uint32_t index,
                                    WasmType type,
                                    WasmBool mutable_,
                                    void* user_data) {
-  print_details(user_data, " - global\n");
+  print_details(user_data, " - global mutable=%d type=%#x\n", mutable_, type);
```
binji: may as well use `wasm_get_type_name` here.

src/binary-reader-objdump.c
```diff
@@ -608,7 +640,10 @@ WasmResult wasm_read_binary_objdump(struct WasmAllocator* allocator,
     reader.on_opcode_bare = on_opcode_bare;
     reader.on_opcode_uint32 = on_opcode_uint32;
     reader.on_opcode_uint32_uint32 = on_opcode_uint32_uint32;
+    reader.on_opcode_uint32_uint32 = on_opcode_uint32_uint32;
```
binji: dupe

src/binary-reader-objdump.c
```diff
@@ -265,6 +270,22 @@ static WasmResult on_opcode_uint64(WasmBinaryReaderContext* ctx,
   return WASM_OK;
 }
 
+static WasmResult on_opcode_f32(WasmBinaryReaderContext* ctx,
+                                float value) {
+  Context* context = ctx->user_data;
+  size_t immediate_len = ctx->offset - context->current_opcode_offset;
+  log_opcode(context, ctx->data, immediate_len, "%f", value);
```
binji: maybe use %g to get nicer-looking values here (or "%.19g" to get full precision; not sure if we care about that here).

src/binary-reader-objdump.c
```diff
@@ -35,6 +35,11 @@ typedef struct Context {
   WasmBool print_details;
   WasmBool header_printed;
   int section_found;
+
+  WasmStringSlice import_module_name;
+  WasmStringSlice import_field_name;
+
+  int function_index;
```
binji: OK for now, but this will be needed for other imported types too (globals, tables, memories).

binji:
Hm, looks like OS/X doesn't print negative nan; maybe it's being canonicalized?

src/binary-reader-objdump.c
sbc100: I updated all the import callback to format in this way

src/binary-reader-objdump.c
sbc100: Done

src/binary-reader-objdump.c
sbc100: Done

src/binary-reader-objdump.c
sbc100: Ack

src/binary-reader-objdump.c
sbc100: Done

binji:
To handle the OS/X nan canonicalization, you can use `wasm_write_float_hex` and `wasm_write_double_hex` from literal.h instead. Or you could check for the values being nan (x != x) and special case that.

sbc100:
Switching to %g looks like it made travis happy.. unless I'm missing something?

sbc100:
OIC travis can be green even when OSX is not yet passing

binji:
Yeah, it's because OSX build is slow... :-|

src/binary-reader-objdump.c
```diff
@@ -271,18 +272,22 @@ static WasmResult on_opcode_uint64(WasmBinaryReaderContext* ctx,
...
   return WASM_OK;
 }
 
 static WasmResult on_opcode_f64(WasmBinaryReaderContext* ctx,
-                                double value) {
+                                uint64_t value) {
   Context* context = ctx->user_data;
   size_t immediate_len = ctx->offset - context->current_opcode_offset;
-  log_opcode(context, ctx->data, immediate_len, "%g", value);
+  char buffer[20];
```
binji: looks like this can actually be 25 characters, including null terminator. (see comment at the top of wasm_write_double_hex). I thought I added a #define for this in literal.h, but I don't see it. It's probably worth adding one, if you don't mind.

sbc100: Done

PR #229, opened 2016-11-30
Fix gen-spec-js.py

binji:
This now (mostly) matches the output from the spec interpreter.


binji:
@sbc100 ptal

src/binary-writer-spec.c
```diff
@@ -162,6 +162,19 @@ static void write_var(Context* ctx, const WasmVar* var) {
...
 
+static void write_type_object(Context* ctx, WasmType type) {
+  wasm_writef(&ctx->json_stream, "{");
+  write_key(ctx, "type");
+  switch (type) {
+    case WASM_TYPE_I32: write_string(ctx, "i32"); break;
+    case WASM_TYPE_I64: write_string(ctx, "i64"); break;
+    case WASM_TYPE_F32: write_string(ctx, "f32"); break;
+    case WASM_TYPE_F64: write_string(ctx, "f64"); break;
+    default: assert(0); break;
```
sbc100: Why not use wasm_get_type_name() here?

test/run-gen-spec-js.py
```diff
@@ -0,0 +1,97 @@
...
+    gen_spec_js.verbose = options.print_cmd
+
+    json_file = utils.ChangeDir(utils.ChangeExt(options.file, '.json'), out_dir)
+    js_file = utils.ChangeExt(json_file, '.js')
+    wast2wasm.RunWithArgs(options.file, '-o', json_file)
+
+    if options.js:
+      js = utils.Executable(options.js, *options.js_flags)
+      js.RunWithArgs(js_file)
+      gen_spec_js.RunWithArgs(json_file, '-o', js_file)
```
sbc100: Looks like this execution of these two backwards.. which probably means the --js flag is not used .. is it needed?

test/run-gen-spec-js.py
```diff
@@ -0,0 +1,97 @@
...
+def main(args):
+  parser = argparse.ArgumentParser()
+  parser.add_argument('-o', '--out-dir', metavar='PATH',
+                      help='output directory for files.')
+  parser.add_argument('--wast2wasm', metavar='PATH',
+                      help='override wast2wasm executable.')
+  parser.add_argument('--wasm2wast', metavar='PATH',
+                      help='override wasm2wast executable.')
+  parser.add_argument('--js', metavar='PATH',
+                      help='specify JavaScript compiler to use.')
```
sbc100: Would it be better to called this "JavaScript interpreter" or "JavaScript runtime".   How about "the name of a JavaScript execution engine with which to run the generated JavaScript file.  If not specified the JavaScript will be written to stdout"?

PR #231, opened 2016-12-02
Rename `user` sections to `custom` sections

sbc100:
As per https://github.com/WebAssembly/design/pull/87


src/binary-reader.h
```diff
@@ -50,10 +50,10 @@ typedef struct WasmBinaryReader {
   WasmResult (*end_module)(void* user_data);
 
   /* user section */
```
binji: here too

sbc100: Done

PR #235, opened 2016-12-06
Escape source filename when writing to json

Cellule:
`\` are used in paths on Windows and need to be escaped to be valid json


binji:
lgtm, thanks!

PR #239, opened 2016-12-09
Update testsuite

binji:
* Non-imported globals can not be used in initializer expressions
* block/loop/if labels can be repeated at the end
* get_global index in initializer expression should reference use module
  index space, not "defined" global index space


test/dump/global.txt
```diff
@@ -1,6 +1,11 @@
 ;;; TOOL: run-wasmdump
 ;;; FLAGS: -v --dump-verbose
 (module
+  (import "foo" "i32" (global i32))
```
sbc100: Maybe give these names that are not identical to the type symbols?  maybe i32_import?

binji: Done.

PR #241, opened 2016-12-09
Adjust asserts to allow writing invalid modules

Cellule:
I was able to convert all assert_invalid tests in the spec repository using this change


Cellule:
It seems there is some more work to do in wasm-interp to support assert_invalid case

Cellule:
I think I have fixed the error in wasm-interp.
However, I wasn't able to make running the assert_invalid scenario correctly in the interpreter.
I think it's better to leave the implementation to another PR and let this move on to at least generate invalid cases for now

PR #245, opened 2016-12-13
Make some binary reading/writing utilities public

sbc100:
This is mostly a refactor to make some of the internal
binary reading/writing utilities public and available
to the wasm linker prototype that I'm working on.

### DIFF
--- a/src/binary-reader-objdump.c
+++ b/src/binary-reader-objdump.c
@@ -61,16 +61,10 @@ static void WASM_PRINTF_FORMAT(2, 3)
   va_end(args);
 }
 
-#define SEGSTART(segname, name)                                             \
-  static WasmResult begin_##segname##_section(WasmBinaryReaderContext* ctx, \
-                                              uint32_t size) {              \
-    return begin_section(ctx->user_data, name, ctx->offset, size);          \
-  }
-
-static WasmResult begin_section(Context* ctx,
-                                const char* name,
-                                size_t offset,
-                                size_t size) {
+static WasmResult do_begin_section(Context* ctx,
+                                   const char* name,
+                                   size_t offset,
+                                   size_t size) {
   switch (ctx->options->mode) {
     case WASM_DUMP_HEADERS:
       printf("%9s start=%#010" PRIzx " end=%#010" PRIzx " (size=%#010" PRIzx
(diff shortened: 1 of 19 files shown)


src/binary-reader-objdump.c
```diff
@@ -608,16 +595,16 @@ static WasmBinaryReader s_binary_reader = {
     .end_module = end_module,
     .on_error = on_error,
 
+    .begin_section = begin_section,
```
binji: seems weird to have a begin_section without an end_section

src/binary-writer.c
```diff
@@ -90,31 +91,46 @@ static uint32_t write_u32_leb128_at(WasmStream* stream,
...
-static void write_u32_leb128(WasmStream* stream,
-                             uint32_t value,
-                             const char* desc) {
-  uint32_t length = write_u32_leb128_at(stream, stream->offset, value, desc);
+uint32_t wasm_write_fixed_u32_leb128_at(WasmStream* stream,
+                                        uint32_t offset,
+                                        uint32_t value,
+                                        const char* desc) {
+  uint8_t data[MAX_U32_LEB128_BYTES];
+  uint32_t rtn = wasm_write_fixed_u32_leb128_raw(data, data+MAX_U32_LEB128_BYTES, value);
```
binji: nit: better name than `rtn`? How about `length`?

src/binary-writer.c
```diff
@@ -180,38 +196,38 @@ static void write_fixup_u32_leb128_size(Context* ctx,
...
 }
 
-static void write_str(WasmStream* stream,
-                      const char* s,
-                      size_t length,
-                      WasmPrintChars print_chars,
-                      const char* desc) {
-  write_u32_leb128(stream, length, "string length");
+void wasm_write_str(WasmStream* stream,
+                           const char* s,
```
binji: nit: format

src/binary-writer.c
```diff
@@ -272,176 +288,177 @@ static void write_expr(Context* ctx,
...
     }
     case WASM_EXPR_TYPE_STORE: {
-      write_opcode(&ctx->stream, expr->store.opcode);
+      wasm_write_opcode(&ctx->stream, expr->store.opcode);
       uint32_t align =
           wasm_get_opcode_alignment(expr->store.opcode, expr->store.align);
       wasm_write_u8(&ctx->stream, log2_u32(align), "alignment");
-      write_u32_leb128(&ctx->stream, (uint32_t)expr->store.offset,
+      wasm_write_u32_leb128(&ctx->stream, (uint32_t)expr->store.offset,
                        "store offset");
```
binji: nit: format
other places in this file too

src/binary-writer.h
```diff
@@ -18,6 +18,7 @@
 #define WASM_BINARY_WRITER_H_
 
 #include "common.h"
+#include "stream.h"
```
binji: Can you forward declare WasmStream instead?

sbc100: Interestingly it was an enum I needed.. and I didn't realize I could forward declare them.

Done.

src/binary-reader-objdump.c
sbc100: Should I add that then, even though there are no consumers for it yet?

binji: I'd prefer it, but maybe that's just my OCD... :)

PR #248, opened 2016-12-16
Rewrite the binary section reader code

binji:
The previous way took advantage of the fact that WebAssembly sections
have a required order, but ultimately it made the code more complicated
and fragile.

This change makes parsing more natural; read each section and dispatch
to a function that knows how to parse it.

### DIFF
--- a/src/binary-reader.c
+++ b/src/binary-reader.c
@@ -105,22 +105,23 @@ WASM_DEFINE_VECTOR(uint32, Uint32);
   if (!(cond))                        \
     RAISE_ERROR(__VA_ARGS__);
 
-#define V(NAME, code) [code] = #NAME,
+#define V(NAME, name, code) [code] = #NAME,
 static const char* s_section_name[] = {WASM_FOREACH_BINARY_SECTION(V)};
 #undef V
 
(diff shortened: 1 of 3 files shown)


src/binary-reader.c
```diff
@@ -1724,382 +1636,444 @@ static void read_function_body(Context* ctx,
...
-  in_u32(ctx, &magic, "magic");
-  RAISE_ERROR_UNLESS(magic == WASM_BINARY_MAGIC, "bad magic value");
-  uint32_t version;
-  in_u32(ctx, &version, "version");
-  RAISE_ERROR_UNLESS(version == WASM_BINARY_VERSION,
-                     "bad wasm file version: %#x (expected %#x)", version,
-                     WASM_BINARY_VERSION);
-
-  CALLBACK(begin_module, version);
+static void read_type_section(Context* ctx, uint32_t section_size) {
```
sbc100: Looks like only read_custom_section() needs the second `section_size` argument.  Could we special case that?

binji: It gets secretly passed through in the `CALLBACK_SECTION` macro as well, so they all need it.

PR #250, opened 2016-12-19
Validate function vars in elem segments

binji:
They must reference a valid function.

### DIFF
--- a/test/parse/module/bad-table-invalid-function.txt
+++ b/test/parse/module/bad-table-invalid-function.txt
@@ -1,0 +1,10 @@
+;;; ERROR: 1
+(module
+  (table 1 anyfunc)
+  (func $f)
+  (elem (i32.const 0) $f 1))
+(;; STDERR ;;;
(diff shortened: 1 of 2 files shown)


test/parse/module/bad-table-invalid-function.txt
```diff
@@ -0,0 +1,10 @@
+;;; ERROR: 1
+(module
+  (table 1 anyfunc)
+  (func $f)
+  (elem (i32.const 0) $f 1))
+(;; STDERR ;;;
+parse/module/bad-table-invalid-function.txt:5:26: function variable out of range (max 1)
+  (elem (i32.const 0) $f 1))
```
sbc100: Seem a little but odd that the error message is displaying the max in 1-based index space but the error its pointing to is 0-base.   Seems to me that 1 is a valid index if max=1.

binji: Hm, good point. Probably better to fix in another PR, because it will affect a lot of tests.

sbc100: sgtm

PR #251, opened 2016-12-19
 Added symlinks to 'make clean' rule 

citisolo:
Added symlinks to 'make clean' rule so no broken symlinks are left hanging around after a clean


binji:
lgtm, thanks!

PR #252, opened 2016-12-20
Add a couple of string slice utility functions

sbc100:
These are needed to support the wasm-link utility
that will hopefully be landing soon.

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -56,6 +56,23 @@ WasmStringSlice wasm_empty_string_slice(void) {
   return result;
 }
 
+WasmBool wasm_string_slice_eq_cstr(const WasmStringSlice* s1, const char* s2) {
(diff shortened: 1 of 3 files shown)


src/common.c
```diff
@@ -56,6 +56,23 @@ WasmStringSlice wasm_empty_string_slice(void) {
   return result;
 }
 
+WasmBool wasm_string_slice_eq_cstr(const WasmStringSlice* s1, const char* s2) {
+  size_t s2_len = strlen(s2);
+  if (s2_len != s1->length)
+    return WASM_FALSE;
+
+  return strncmp(s1->start, s2, s2_len) == 0 ? WASM_TRUE : WASM_FALSE;
```
binji: `return strncmp(...) == 0;` is fine here.

PR #254, opened 2016-12-22
"Copied from" is so inelegant

jfbastien:
Wording is better this way.


binji:
Hm, yes it is a bit weird. I copied the template from there, but then that gets put into the generated output, which makes the sentence inaccurate. That said, "as generated by" isn't quite right either. I guess I would prefer just being explicit and saying the template is from <link>, generated by <tool>, where the tool is `gen-spec-js.py`.

jfbastien:
@binji gotcha. How about this?

binji:
Perfect :)

Issue #259, opened 2016-12-27
assert_unlinkable should check for LinkError instead of type error


saambarati:
Now that WebAssembly.LinkError is a thing:
https://github.com/WebAssembly/design/blob/master/JS.md#constructor-properties-of-the-webassembly-object

this code should probably go from:

```
function assert_unlinkable(bytes) {
  let mod = module(bytes);
  try { new WebAssembly.Instance(mod, registry) } catch (e) {
    if (e instanceof TypeError) return;
  }
  throw new Error("Wasm linking failure expected");
}
```

to
```
function assert_unlinkable(bytes) {
  let mod = module(bytes);
  try { new WebAssembly.Instance(mod, registry) } catch (e) {
    if (e instanceof WebAssembly.LinkError) return;
  }
  throw new Error("Wasm linking failure expected");
}
```

PR #260, opened 2016-12-27
Update gen-spec-prefix.js

saambarati:
See: https://github.com/WebAssembly/wabt/issues/259


binji:
lgtm, though I should mention that this code is [copied from the spec repo](https://github.com/WebAssembly/spec/blob/master/interpreter/host/js.ml#L68), so you'll likely want to update that too.

PR #271, opened 2017-01-11
Add symbol resolution to wasm-link

sbc100:
Symbols exported by one module can now be consumed
by another module at link time.  Imports that can
be resolved at link time are removed the output.
Any unresolved imported remaining in the output
binary are considered and error unless the -r flag
is passed.

Also split binary reading logic out of wasm-link.c
into its own binary-reader-linker file.

### DIFF
--- a/src/binary-reader-linker.c
+++ b/src/binary-reader-linker.c
@@ -38,7 +38,7 @@ static WasmResult on_reloc(WasmRelocType type,
   if (section_index >= binary->sections.size) {
     WASM_FATAL("invalid section index: %d\n", section_index);
   }
-  Section* sec = &binary->sections.data[section_index - 1];
+  Section* sec = &binary->sections.data[section_index];
   if (offset + RELOC_SIZE > sec->size) {
     WASM_FATAL("invalid relocation offset: %#x\n", offset);
   }
--- a/src/binary-reader-linker.c
+++ b/src/binary-reader-linker.c
@@ -23,26 +23,26 @@
 
 typedef struct Context {
   WasmAllocator* allocator;
-  InputBinary* binary;
+  WasmLinkerInputBinary* binary;
 
-  Section* reloc_section;
+  WasmSection* reloc_section;
 
   WasmStringSlice import_name;
-  Section* current_section;
+  WasmSection* current_section;
 } Context;
 
 static WasmResult on_reloc_count(uint32_t count, WasmBinarySection section_code,
                                  WasmStringSlice section_name,
                                  void* user_data) {
   Context* ctx = user_data;
-  InputBinary* binary = ctx->binary;
+  WasmLinkerInputBinary* binary = ctx->binary;
   if (section_code == WASM_BINARY_SECTION_CUSTOM) {
     WASM_FATAL("relocation for custom sections not yet supported\n");
   }
 
   uint32_t i;
   for (i = 0; i < binary->sections.size; i++) {
-    Section* sec = &binary->sections.data[i];
+    WasmSection* sec = &binary->sections.data[i];
     if (sec->section_code != section_code)
       continue;
     ctx->reloc_section = sec;
@@ -62,7 +62,8 @@ static WasmResult on_reloc(WasmRelocType type,
     WASM_FATAL("invalid relocation offset: %#x\n", offset);
   }
 
-  Reloc* reloc = wasm_append_reloc(ctx->allocator, &ctx->reloc_section->relocations);
+  WasmReloc* reloc =
+      wasm_append_reloc(ctx->allocator, &ctx->reloc_section->relocations);
   reloc->type = type;
   reloc->offset = offset;
 
@@ -86,8 +87,8 @@ static WasmResult on_import_func(uint32_t index,
                                  uint32_t sig_index,
                                  void* user_data) {
   Context* ctx = user_data;
-  FunctionImport* import =
-      wasm_append_function_import(ctx->allocator, &ctx->binary->function_imports);
+  WasmFunctionImport* import = wasm_append_function_import(
+      ctx->allocator, &ctx->binary->function_imports);
   import->name = ctx->import_name;
   import->sig_index = sig_index;
   import->active = WASM_TRUE;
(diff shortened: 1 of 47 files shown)


sbc100:
@binji ptal.  Let me know if you want to to split this up.  I could for example split out the refactor of the reader, or the change the wasmdump formatting for relocations.

src/tools/wasm-link.c
```diff
@@ -588,26 +330,85 @@ static void write_combined_memory_section(Context* ctx,
...
+    FunctionImportVector* imports = &binary->function_imports;
+    for (j = 0; j < imports->size; j++) {
+      FunctionImport* import = &imports->data[j];
+      if (import->active)
+        write_function_import(ctx, import, binary->type_index_offset);
+    }
+
+    GlobalImportVector* globals = &binary->global_imports;
+    for (j = 0; j < globals->size; j++) {
+      write_global_import(ctx, &globals->data[j]);
```
binji: Why are function imports treated differently than global imports?

src/tools/wasm-link.c
```diff
@@ -759,61 +563,130 @@ static WasmBool write_combined_section(Context* ctx,
...
+  size_t i, j;
+  for (i = 0; i < ctx->inputs.size; i++) {
+    InputBinary* binary = &ctx->inputs.data[i];
+    for (j = 0; j < binary->exports.size; j++) {
+      Export* export = &binary->exports.data[j];
+      ExportInfo* info = wasm_append_export_info(ctx->allocator, &export_list);
+      info->export = export;
+      info->binary = binary;
+      WasmStringSlice name = wasm_dup_string_slice(ctx->allocator, export->name);
+      WasmBinding* binding = wasm_insert_binding(ctx->allocator, &export_map, &name);
```
binji: Might be a good idea to check for dupes, I don't think `wasm_insert_binding` does (though maybe it should)

src/tools/wasm-link.c
```diff
@@ -759,61 +563,130 @@ static WasmBool write_combined_section(Context* ctx,
...
+  uint32_t total_global_imports = 0;
+  for (i = 0; i < ctx->inputs.size; i++) {
+    InputBinary* binary = &ctx->inputs.data[i];
+    /* The imported_function_index_offset is the sum of all the function
+     * imports from objects that precede this one.  i.e. the current running
+     * total */
+    binary->imported_function_index_offset = total_function_imports;
+    binary->imported_global_index_offset = total_global_imports;
+    binary->memory_page_offset = memory_page_offset;
+    memory_page_offset += binary->memory_page_count;
```
binji: You'll have to handle tables too, later PR?

src/binary-reader-linker.c
```diff
@@ -0,0 +1,280 @@
+/*
+ * Copyright 2016 WebAssembly Community Group participants
```
sbc100: Done

src/binary-reader-linker.c
```diff
@@ -0,0 +1,280 @@
...
+static WasmResult on_reloc(WasmRelocType type,
+                           uint32_t section_index,
+                           uint32_t offset,
+                           void* user_data) {
+  Context* ctx = user_data;
+  InputBinary* binary = ctx->binary;
+  if (section_index >= binary->sections.size) {
+    WASM_FATAL("invalid section index: %d\n", section_index);
+  }
+  Section* sec = &binary->sections.data[section_index];
```
sbc100: I changed it from a one-based to zero-based index.   I got feedback from dan that maybe we should use the section_code instead though so this might change again.

src/wasm-link.h
```diff
@@ -0,0 +1,131 @@
...
+ */
+
+#ifndef WASM_LINK_H_
+#define WASM_LINK_H_
+
+#include "binary.h"
+#include "common.h"
+#include "vector.h"
+
+#define LINKER_IMPORT_MODULE_NAME "linker"
```
sbc100: Done.  Although do you really want me to change all the types in the this file.  It would be quite a big change :(

src/tools/wasm-link.c
sbc100: I added a TODO for this.

src/tools/wasm-link.c
sbc100: Currently we generate relocation for the ELEM section, but I don't have any way to tagging data which refers to addresses of functions.   Such references would generate relocation of type WASM_RELOC_TABLE_INDEX_SLEB/I32 which would then needs to be correctly handled.  So yes, we don't support that in wabt yet.

src/tools/wasm-link.c
```diff
@@ -75,15 +90,15 @@ static void on_option(struct WasmOptionParser* parser,
 }
 
 static void on_argument(struct WasmOptionParser* parser, const char* argument) {
-  wasm_append_string_value(s_allocator, &s_infiles, &argument);
+  wasm_append_string_value(parser->user_data, &s_infiles, &argument);
```
sbc100: Done

test/run-wasm-link.py
```diff
@@ -92,13 +101,27 @@ def main(args):
...
       wasmdump.RunWithArgs('-d', '-v', '-h', output)
     else:
       wasm_link.RunWithArgs('-o', output, *wasm_files)
       wasmdump.RunWithArgs('-d', '-h', '-v', output)
 
+    if options.spec:
+      with open(out_file) as json_file:
+        spec = json.load(json_file, object_pairs_hook=OrderedDict)
+        spec['commands'] = [c for c in spec['commands'] if c['type'] != 'module']
+        module = OrderedDict([('type', 'module'),
```
sbc100: Created a separate PR to start using yapf for formatting.

src/binary-reader.c
```diff
@@ -1694,12 +1705,15 @@ static void read_custom_section(Context* ctx, uint32_t section_size) {
     uint32_t i, num_relocs, section;
     in_u32_leb128(ctx, &section, "section");
     in_u32_leb128(ctx, &num_relocs, "relocation count count");
-    CALLBACK(on_reloc_count, num_relocs);
+    WASM_ZERO_MEMORY(section_name);
+    if (section == WASM_BINARY_SECTION_CUSTOM)
+      in_str(ctx, &section_name, "section name");
+    CALLBACK(on_reloc_count, num_relocs, section, section_name);
```
sunfishcode: Is it intended to encode the section string after num_relocs? It would seem more natural to encode the section string immediately after the section code, which would also keep num_relocs immediately adjacent to the reloc array.

sbc100: Agreed.  It should be other away around as the "spec" says: https://github.com/WebAssembly/tool-conventions/blob/master/Linking.md.

I will fix

PR #273, opened 2017-01-11
Pass --bindir to test helpers rather than full path to each tool

sbc100:
This simplifies the invocations and aids maintainability.


binji:
Awesome! lgtm

PR #276, opened 2017-01-17
Add support for yapf python formatting tool

sbc100:
This changs adds .style.yapf to define the python style
we are using.

I also ran yapf over all the python files:
 $ yapf -i `git ls-files *.py`

Going forward, we should probably add a travis test to
prevent regressions.  We should probably also switch to
more conventional 4-space indentation, which is used by
almost everybody outside of Google.

### DIFF
--- a/.style.yapf
+++ b/.style.yapf
@@ -1,3 +1,4 @@
 [style]
 split_before_named_assigns = False
 based_on_style = chromium
+column_limit = 79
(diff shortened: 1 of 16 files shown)


binji:
hm, yapf seems to be fighting the flake8 plugin on appveyor

binji:
Other than that, lgtm. Some of the edits I don't like that much, but I can deal. :)

Issue #277, opened 2017-01-17
function-section index versus function-index-space index


sunfishcode:
wasmdump -v prints functions with their function-index-space indices, which includes imports:
```
IMPORT:
 - func[0] sig=8 <- env.__syscall6
 - func[1] sig=8 <- env.__syscall140
 - func[2] sig=12 <- env.sbrk
 - func[3] sig=13 <- env.abort
FUNCTION:
 - func[4] sig=0
 - func[5] sig=1
 - func[6] sig=1
```

However, wasmdump -d prints functions with their function-section index.
```
a.out.wasm:     file format wasm 0x00000d

Code Disassembly:
func 0
 00018c: 02 7f                      | block i32
 00018e: 41 00                      |   i32.const 0
...
```
Note the "func 0". It's confusing to see two different index spaces for functions.

As far as I'm aware, function-section indices aren't used for anything in wasm, so my suggestion would be to print function-index-space indices everywhere, because that's what someone debugging a wasm binary is likely to need.

binji:
Agreed, WebAssembly already has enough index spaces, we shouldn't add more. :-)

PR #278, opened 2017-01-17
Always use function index space in wasmdump

binji:
This fixes issue #277.


src/binary-reader-objdump.c
```diff
@@ -370,14 +372,16 @@ static WasmResult on_function_signature(uint32_t index,
...
-  ctx->function_index++;
+  print_details(user_data, " - func[%d] sig=%d\n",
+                ctx->num_func_imports + index, sig_index);
   return WASM_OK;
 }
 
 static WasmResult begin_function_body(uint32_t index, void* user_data) {
   Context* ctx = user_data;
 
+  index += ctx->num_func_imports;
```
sunfishcode: Would it make sense to change the index passed into begin_function_body (and end_function_body, I suppose) to be the function-index-space index? From a quick scan, it seems like that's what most of the places that use that argument ultimately want.

The one thing that a quick scan of the code turns up that needs the function-section index is this, in binary-reader-interpreter.c's begin_function_body:
```
  Uint32Vector* fixups = &ctx->func_fixups.data[index];
```
but if that's all, then perhaps it could be handled by adding a function like `get_func_by_env_index` but for reading fixups.

binji: Hm, interesting idea. I'll investigate.

binji: Thanks, this is nicer!

sbc100:
Great! lgtm

PR #280, opened 2017-01-18
Use sudo: false in travis builds

sbc100:
This should allow the use of faster, container based
linux VMs.


sbc100:
Abandoning because we need sudo to get a 32-bit build environment running.

PR #282, opened 2017-01-18
Allow more names than function in binary-reader

sbc100:
ast-binary-reader will still fail if there are too
many names but the binary-reader itself will allow this
since the spec allows it.


binji:
Probably should fix the binary-reader-ast too, but I guess that's fine for another PR

PR #285, opened 2017-01-20
Add function names and relocations to wasmdump disassembly

sbc100:
Relocations are only shown if -r is passed.

### DIFF
--- a/src/binary-reader-objdump.c
+++ b/src/binary-reader-objdump.c
@@ -29,7 +29,7 @@ typedef uint32_t Uint32;
 WASM_DEFINE_VECTOR(uint32, Uint32);
 
 typedef struct Context {
-  const WasmObjdumpOptions* options;
+  WasmObjdumpOptions* options;
   WasmAllocator* allocator;
   WasmStream* out_stream;
   const uint8_t* data;
@@ -47,6 +47,8 @@ typedef struct Context {
 
   WasmStringSlice import_module_name;
   WasmStringSlice import_field_name;
+
+  uint32_t next_reloc;
 } Context;
 
 static WasmBool should_print_details(Context* ctx) {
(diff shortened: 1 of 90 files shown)


src/binary-reader-objdump.c
```diff
@@ -65,34 +67,40 @@ static void WASM_PRINTF_FORMAT(2, 3)
   va_end(args);
 }
 
-static WasmResult do_begin_section(Context* ctx,
-                                   const char* name,
-                                   size_t offset,
-                                   size_t size) {
-  WasmBool section_match = !ctx->options->section_name ||
-                           !strcasecmp(ctx->options->section_name, name);
+static WasmResult begin_section(WasmBinaryReaderContext* ctx,
```
binji: Probably better to be consistent with ctx -> Context, context -> WasmBinaryReaderContext.

src/binary-reader-objdump.h
```diff
@@ -37,18 +47,21 @@ typedef struct WasmObjdumpOptions {
   WasmBool raw;
   WasmBool disassemble;
   WasmBool debug;
+  WasmBool relocs;
   WasmObjdumpMode mode;
   const char* infile;
   const char* section_name;
   WasmBool print_header;
+  WasmStringSliceVector function_names;
```
binji: Kinda weird to hang this on the Options, rather than the Context.

binji:
lgtm

src/binary-reader-objdump.h
sbc100: its because it needs to persist between passes.  I though about creating another "state/context" object that could be passed in along with options.

src/binary-reader-objdump.c
sbc100: I agree it should be consistent but almost all the WasmBinaryReaderContext in this file are currently called 'ctx'.   Perhaps a cleanup CL should follow

PR #289, opened 2017-01-24
Fix reading of custom section relocations

sbc100:
The string name of the target section comes before
the relocation count.

This was not tested because we don't currently generate
any relocations for custom sections, so I also added
a binary/relocs.txt test to generate a binary that
does.


binji:
lgtm

PR #290, opened 2017-01-24
Assert exhaustion

Cellule:
Added support for `assert_exhaustion` type which was added in https://github.com/WebAssembly/spec/pull/388

note: I wasn't able to install `re2c 0.16` on my machine so the generated code might differ more than it should


binji:
Nice, lgtm!

PR #294, opened 2017-01-26
Implement assert_invalid for wasm-interp

binji:


binji:
@sbc100 ptal

src/binary-reader-interpreter.c
```diff
@@ -894,6 +940,13 @@ static WasmResult on_elem_segment_function_index_check(uint32_t index,
     return WASM_ERROR;
   }
 
+  uint32_t max_func_index = (ctx)->func_index_mapping.size;
```
sbc100: Why the braces here?

binji: Whoops, fixed!

PR #253, opened 2016-12-21
wasm2wast: parse the name section by default

binji:
You can now disable this behavior by passing `--no-debug-names`.

PR #295, opened 2017-01-27
Update README now that wasm2wast defaults to including debug names

jayphelps:
As of #253, wasm2wast now includes debug-names by default, letting you opt-out via `--no-debug-names`.

I also noticed that now wasm2wast and wast2wasm are inconsistent in that regard. The later still defaulting to excluding debug names. Is that intentional? Happy to PR if it too should be reversed. There's mostly a tradition of toolchains requiring opt-in for optimizations like excluding debug names, but I'm new around here 🤓

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,8 +117,8 @@ Some examples:
 # parse binary file test.wasm and write s-expression file test.wast
 $ out/wasm2wast test.wasm -o test.wast
 
-# parse test.wasm and write test.wast, using the debug names, if any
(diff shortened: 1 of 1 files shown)


binji:
Thanks! My thoughts as to difference between wast2wasm and wasm2wast: when you're converting from binary to text, it is generally useful to use the names section if it exists. When you're converting from text to binary, you only want to add the names section for debugging purposes.
I think of it like passing -g to gcc. :)

PR #296, opened 2017-01-28
Remove assert_{invalid,malformed} checks from wast2wast

binji:
Now that assert_invalid is checked by wasm-interp, all spec assertions
can be handled by the interpreter. Having a subset that are also handled
by wast2wasm is not useful.

Also, there was a bug in the option parser where passing an option that
didn't match could still work if a prefix was valid.


binji:
I took another look at the tests, and I'm pretty sure I'm not removing any coverage. Always good to double check :)

PR #298, opened 2017-01-31
Rename all wasm prefixes to wabt

binji:


src/ast.c
```diff
@@ -213,549 +213,549 @@ WasmModuleField* wasm_append_module_field(struct WasmAllocator* allocator,
...
       break;
 
-    case WASM_EXPR_TYPE_SET_LOCAL:
+    case WABT_EXPR_TYPE_SET_LOCAL:
       CALLBACK(on_set_local_expr);
       break;
 
-    case WASM_EXPR_TYPE_STORE:
+    case WABT_EXPR_TYPE_STORE:
       CALLBACK(on_store_expr);
```
KarlSchimpf: Why not WABT_CALLBACK to avoid using what may be a common macro name. If you are doing it for constants, why not the (non-constant) macro?

binji:
Yes, `CALLBACK` is local to ast.c. I was trying to be careful about using the prefix only for external names. Probably made a few mistakes, though :)

PR #302, opened 2017-02-11
Full typechecking for unreachable code

binji:
The typechecking is now shared between the validator and
binary-reader-interpreter as well.

* Shared the various LabelType enumerations -> WabtLabelType.
* Fixed the "invalid depth" errors, the max value was incorrect.
* Removed some tests that aren't useful anymore:
  - interp/if-then-br hasn't been useful for a while, but now is even
    less so because it doesn't validate without dropping the i32.const
    from the true branch
  - typecheck/bad-br-multi-type isn't possible because the block must
    specify the signature for the br.
  - typecheck/bad-label-multi-type used to test matching signature for
    fallthrough and br, but isn't possible because of block signatures.


src/type-checker.c
```diff
@@ -0,0 +1,523 @@
...
+
+#define CHECK_RESULT(expr) \
+  do {                     \
+    if (WABT_FAILED(expr)) \
+      return WABT_ERROR;   \
+  } while (0)
+
+#define COMBINE_RESULT(result_var, result) \
+  do {                                     \
+    result_var |= (result);                \
```
KarlSchimpf: Why not also parenthesize "result_var"?

src/type-checker.c
```diff
@@ -0,0 +1,523 @@
...
+  }
+}
+
+WabtResult wabt_typechecker_get_label(WabtTypeChecker* tc,
+                                      size_t depth,
+                                      WabtTypeCheckerLabel** out_label) {
+  if (depth >= tc->label_stack.size) {
+    assert(tc->label_stack.size > 0);
+    print_error(tc, "invalid depth: %" PRIzd " (max %" PRIzd ")",
+                depth, tc->label_stack.size - 1);
```
KarlSchimpf: Would it make sense to set out_label to NULL here? That would guarantee that if the caller inadvertently used the pointer, it would fail (rather than potentially using some out of date value).

src/type-checker.c
binji: Done.

src/type-checker.c
binji: Done.

PR #303, opened 2017-02-14
Add coverage build

binji:
This should be helpful in determining which (of any) of the wabt-specific tests we can remove in favor of the spec tests.


binji:
Maybe so, that'd be pretty cool. :)

PR #312, opened 2017-02-23
Use multiprocessing.pool

binji:
This replaces the ad-hoc mechanism we used before with threads. We'll
have to keep an eye on the Travis build, as we switched away from
multiprocessing before because it would flake occasionally. Hopefully
multiprocessing.pool doesn't have the same issue.

This also removes should_stop_on_error from the RunMultiThreaded path,
as it isn't allowed anyway (it early-outs when parsing arguments).

### DIFF
--- a/test/run-tests.py
+++ b/test/run-tests.py
@@ -21,11 +21,6 @@ import difflib
 import fnmatch
 import multiprocessing
 import os
-try:
-  import Queue
-except ImportError:
-  # Python3?
-  import queue as Queue
 import re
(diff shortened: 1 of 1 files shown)


binji:
Yeah, not sure why I wasn't using pool in the first place. :roll_eyes:

sbc100:
Was there any noticeable change in execution speed?

binji:
Just tested on my desktop, speed seems to be basically the same.

PR #313, opened 2017-02-23
Update gen-spec-js to support assert_exhaustion

binji:


binji:
FYI @jfbastien

binji:
Added test.

binji:
Anyone wanna take a look? :-D

binji:
That magical, huh?

Issue #308, opened 2017-02-17
Move to C++


binji:
I've been thinking about it, and pretty much everybody has said it is a hassle using C. I like it, but that's a silly reason to keep it this way, as maintainability suffers greatly when I'm the only one who wants to hack on the code :)

I'd like to start moving toward using C++ incrementally. Here are some thoughts on how it could progress:

- [x] Rename .c files to .cc files
- [x] Switch to using a C++ compiler
- [x] Use nullptr intead of NULL
- [x] Use C++ casts instead of C-style casts
- [x] Use enum classes
- [x] Use wabt namespace instead of prefix
- [x] Use function overloading
- [x] Switch init/destroy functions into constructors/destructors
- [x] Use new/delete instead of malloc/free (maybe not everywhere at first, since we use realloc)
- [x] Use std::unordered_map instead of WabtBindingHash
- [x] Use std::vector instead of WabtVector
- [x] Slowly remove some of the manual memory management (i.e. use unique_ptr instead, etc.)
- [x] Use virtual functions instead of function pointers (in binary-reader and elsewhere)
- [x] Use the C++ named headers (e.g. cstddef instead of stddef.h)
- [x] Move variables definitions into for loops (e.g. `for (int i = 0; ... )`)

KarlSchimpf:
Sounds good! However, I also assume that you will use virtual functions
instead of a struct of function pointers?

Karl Schimpf

binji:
Yeah, probably. This list is incomplete :)

dschuff:
I think moving to C++ is reasonable, but we should also consider if we want to continue exporting a C API, and what that means exactly. Currently someone who wants to use WABT as a parser will need to use the various entry point functions and the AST data structures directly, which is fairly convenient but obviously would mean compromises between keeping the core data structures as C vs hiding the AST from C users.

Binaryen has a [C API](https://github.com/WebAssembly/binaryen/blob/master/src/binaryen-c.h) which hides the IR, redefining core types (Type, Literal, etc) exporting constructor functions for all the IR nodes, and exposing functionality such as validation, interpretation, and optimization. But it doesn't allow e.g. traversal of the IR after it is created, which seems like something a user of wabt as a parser would want. OTOH an API which just exported a bunch of complex C structs would probably not be particularly suitable for users of C APIs from languages other than C or C++ so maybe we want to rethink that anyway (and maybe not unless/until some use case appears?)

binji:
[...]

PR #329, opened 2017-03-01
Make most enumerations into enum classes

binji:


src/tools/wast2wasm.cc
```diff
@@ -238,5 +238,5 @@ int main(int argc, char** argv) {
 
   wabt_destroy_ast_lexer(lexer);
   wabt_destroy_script(&script);
-  return result;
+  return static_cast<int>(result);
```
sbc100: Return result != OK ?  Or are there different result values you want to return from main?

binji:
> Seems like chromium style says it still uses MACRO_STYLE for enum values

True, but it sounds like that was for legacy reasons, the Google C++ style guide says to use `kMixedCase`. Could do that here instead, I suppose.

> Seems like going with that style would result in a much smaller change as none of the test output would change? (maybe?)

Yes, but the names match better with the binary document this way, so... :-) Do you have a preference for `MACRO_STYLE`? I don't like that it looks like it could be a macro. And it's kind of ugly, especially for enum class, where it would be `EnumType::ENUM_VALUE`.

src/tools/wast2wasm.cc
binji: Done.

binji:
Gonna land this because I think having enum classes is better overall. Let's continue discussion of style on #308.

PR #330, opened 2017-03-01
Update CMake and Makefile to use .hh instead of .h

binji:
The built-in CMake rules look for an .hh file instead of a .h when
building for C++. Without it, the build will always think that the
generated parser is stale and rebuild it.

### DIFF
--- a/src/prebuilt/ast-parser-gen.cc
+++ b/src/prebuilt/ast-parser-gen.cc
@@ -1,8 +1,8 @@
-/* A Bison parser, made by GNU Bison 3.0.4.  */
+/* A Bison parser, made by GNU Bison 3.0.2.  */
 
 /* Bison implementation for Yacc-like parsers in C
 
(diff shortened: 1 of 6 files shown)


src/prebuilt/ast-parser-gen.cc
```diff
@@ -1,8 +1,8 @@
-/* A Bison parser, made by GNU Bison 3.0.4.  */
+/* A Bison parser, made by GNU Bison 3.0.2.  */
```
KarlSchimpf: Why did the Bison version regress?

binji: I think I have a newer version on my laptop. It would be nice to have it built automatically, but I figure as long as the travis bot that tests the prebuilt parser passes, everything should be OK.

right? :-)

KarlSchimpf: It should. Just wanted to make sure you were aware of the issue.

PR #331, opened 2017-03-01
Replace the Wabt/wabt prefix with a C++ namespace

binji:


src/vector.cc
```diff
@@ -18,27 +18,29 @@
...
+                     size_t* capacity,
+                     size_t desired_size,
+                     size_t elt_byte_size) {
   if (desired_size > *capacity) {
     size_t new_capacity = *capacity ? *capacity * 2 : INITIAL_VECTOR_CAPACITY;
     while (new_capacity < desired_size)
       new_capacity *= 2;
     size_t new_byte_size = new_capacity * elt_byte_size;
-    *data = wabt_realloc(*data, new_byte_size);
+    *data = realloc(*data, new_byte_size);
```
KarlSchimpf: Is this a "C" realloc, or a wabt::realloc?

src/writer.cc
```diff
@@ -70,60 +72,60 @@ void wabt_init_file_writer_existing(WabtFileWriter* writer, FILE* file) {
...
-void wabt_close_file_writer(WabtFileWriter* writer) {
+void close_file_writer(FileWriter* writer) {
   fclose(writer->file);
 }
 
-void wabt_init_output_buffer(WabtOutputBuffer* buf, size_t initial_capacity) {
+void init_output_buffer(OutputBuffer* buf, size_t initial_capacity) {
   assert(initial_capacity != 0);
-  buf->start = wabt_alloc(initial_capacity);
+  buf->start = alloc(initial_capacity);
```
KarlSchimpf: Again, just questioning whether a wabt:: prefix is appropriate so that it doesn't accidentally use "C" alloc  (which could happen by including another standard C header).

src/writer.cc
```diff
@@ -70,60 +72,60 @@ void wabt_init_file_writer_existing(WabtFileWriter* writer, FILE* file) {
...
-static void ensure_output_buffer_capacity(WabtOutputBuffer* buf,
+static void ensure_output_buffer_capacity(OutputBuffer* buf,
                                           size_t ensure_capacity) {
   if (ensure_capacity > buf->capacity) {
     assert(buf->capacity != 0);
     size_t new_capacity = buf->capacity * 2;
     while (new_capacity < ensure_capacity)
       new_capacity *= 2;
-    buf->start = wabt_realloc(buf->start, new_capacity);
+    buf->start = realloc(buf->start, new_capacity);
```
KarlSchimpf: Same question here.

src/apply-names.h
```diff
@@ -36,8 +38,8 @@ struct WabtModule;
  *    ...
  *    (call $foo ...)
  */
-WABT_EXTERN_C_BEGIN
-WabtResult wabt_apply_names(struct WabtModule*);
-WABT_EXTERN_C_END
+Result apply_names(struct Module*);
```
binji: I was thinking since we're using C++ features in the headers now, I'd drop all of the extern "C" stuff. When we get back to place where we're happy with the C++ API, we can add back in a nice C wrapper API

src/ast-lexer.cc
```diff
@@ -114,11 +116,11 @@ static WabtResult fill(WabtLocation* loc,
     /* TODO(binji): could just alloc instead, because we know we'll need to
      * memmove below */
     char* new_buffer =
-        static_cast<char*>(wabt_realloc(lexer->buffer, new_buffer_size));
+        static_cast<char*>(realloc(lexer->buffer, new_buffer_size));
```
binji: Not stupid, I was being lazy and just relying on the compiler to tell me when it was ambiguous. But you're right, it's better to give it a different name so it's clear. Ultimately we'll remove these functions for new/delete, but better to have it clear now.

src/ast.cc
```diff
@@ -19,153 +19,145 @@
 #include <assert.h>
 #include <stddef.h>
```
binji: Probably should use the C++ headers, yeah. I'll add that to the checklist. :-)

binji:
@KarlSchimpf I've changed alloc/realloc/free -> wabt_alloc/wabt_realloc/wabt_free. This is a bit clunky, but as I mentioned above, it will likely all be removed soon anyway. PTAL

src/apply-names.h
KarlSchimpf: OK.

PR #332, opened 2017-03-03
Use new/delete instead of malloc/free

binji:
Also switch some void* -> char*, because it removes some unnecessary
casts. C++ does not like void*.

### DIFF
--- a/src/tools/wasmopcodecnt.cc
+++ b/src/tools/wasmopcodecnt.cc
@@ -334,13 +334,13 @@ int main(int argc, char** argv) {
   init_stdio();
   parse_options(argc, argv);
 
-  void* data;
+  char* data;
   size_t size;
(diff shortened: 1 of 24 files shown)


src/tools/wasmopcodecnt.cc
```diff
@@ -354,21 +354,27 @@ int main(int argc, char** argv) {
     init_opcnt_data(&opcnt_data);
     result = read_binary_opcnt(data, size, &s_read_binary_options, &opcnt_data);
     if (WABT_SUCCEEDED(result)) {
+      fprintf(stderr, "here1\n");
```
KarlSchimpf: I assume these fprintf's are for debugging? Should they be removed (Or at least protected via condition, or hidden behind a macro)?

binji: Oops! Nice catch :)

Issue #334, opened 2017-03-05
new name section support


pipcet:
We should support the new extensible "name" section format, as specified in https://github.com/WebAssembly/design/commit/cd2af4d8149b23be87cc9eabbc4a97c7029178ad.

pipcet:
https://github.com/WebAssembly/wabt/compare/master...pipcet:new-name-section?expand=1
has some code, which seems to work if I'm reading the spec correctly.

binji:
Thanks for looking into this! Feel free to open a PR. Taking a look at what you have there, it looks pretty good. We'll want to support the other name sections for wasm2wast (binary-reader-ast) and wasmdump (binary-reader-objdump), but that seems reasonable to do in a separate PR.

sbc100:
This is  basically done (at least for now right?) isn't it?

binji:
I think so. There's more work to do if we add names for other types, but that isn't spec'd yet I think.

PR #337, opened 2017-03-06
Extensible name section support

pipcet:
This PR should change the code to support the new extensible name section rather than the previous now-obsolete layout.

However, I've so far been unable to find any code (except for my own) that produces or reads the new format, so it's not as well-tested as it should be; it might be worth waiting for adoption in other tools, in any case.

See https://github.com/WebAssembly/wabt/issues/334.

### DIFF
--- a/src/binary-writer.cc
+++ b/src/binary-writer.cc
@@ -874,19 +874,6 @@ static Result write_module(Context* ctx, const Module* module) {
     char desc[100];
     begin_custom_section(ctx, WABT_BINARY_SECTION_NAME, LEB_SECTION_SIZE_GUESS);
     write_u32_leb128(&ctx->stream, 1, "function name type");
-
-    begin_subsection(ctx, "function name subsection", LEB_SECTION_SIZE_GUESS);
-    write_u32_leb128(&ctx->stream, module->funcs.size, "num functions");
-    for (i = 0; i < module->funcs.size; ++i) {
-      const Func* func = module->funcs.data[i];
-      write_u32_leb128(&ctx->stream, i, "function index");
-      snprintf(desc, sizeof(desc), "func name %" PRIzd, i);
-      write_str(&ctx->stream, func->name.start, func->name.length,
-                PrintChars::Yes, desc);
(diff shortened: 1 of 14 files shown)


pipcet:
@binji:
> We'll want to support the other name sections for wasm2wast (binary-reader-ast) and wasmdump (binary-reader-objdump), but that seems reasonable to do in a separate PR.

I'm not sure I understand; are you saying you want to keep support for the old pre-MVP format?

pipcet:
Oops, thought it was running the tests locally, but it wasn't. I'll fix those.

binji:
> I'm not sure I understand; are you saying you want to keep support for the old pre-MVP format?

Sorry, I said we want to support the other name sections but I meant we want to support reading the new kinds of names in those files (e.g. names of labels, globals, memories, etc.)

pipcet:
> I meant we want to support reading the new kinds of names in those files (e.g. names of labels, globals, memories, etc.)

I may have missed something (again), but the spec doesn't mention those yet. I'm in favor of adding those as soon as they appear in the spec.

(There are quite a few commits now, so let me know if and when you want to merge so I can squash them into a single commit to keep the history readable.)

binji:
You're right, I assumed they had been added since that was the purpose of restructuring the name section. :)

src/binary-writer.cc
```diff
@@ -300,6 +304,26 @@ static void end_section(Context* ctx) {
   ctx->last_section_leb_size_guess = 0;
 }
 
+static void begin_subsection(Context* ctx,
+                             const char* name,
+                             size_t leb_size_guess) {
+  assert(ctx->last_subsection_leb_size_guess == 0);
+  char desc[100];
+  snprintf(desc, sizeof(desc), "subsection \"%s\"", name);
```
binji: use wabt_snprintf instead

pipcet:
Thanks! I've addressed those two, do let me know if you spot anything else.

I've joined the CG now.

binji:
Great!

PR #338, opened 2017-03-06
Add --debug flag to wasmdump and wasm-link

sbc100:
Enable logging in the binary reader when the --debug
flag is passed to wasmdump and wasm-link.

Also don't wrap on_error when logging is enabled otherwise
tools that don't set on_error such as wasmdump won't see
any errors (since the reader will think a handler is installed).

### DIFF
--- a/src/tools/wasmdump.cc
+++ b/src/tools/wasmdump.cc
@@ -57,7 +57,7 @@ static Option s_options[] = {
      "print raw section contents"},
     {FLAG_DISASSEMBLE, 'd', "disassemble", nullptr, NOPE,
      "disassemble function bodies"},
-    {FLAG_DEBUG, '\0', "debug", nullptr, NOPE, "disassemble function bodies"},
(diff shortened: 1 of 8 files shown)


src/tools/wasmdump.cc
```diff
@@ -177,22 +183,25 @@ int main(int argc, char** argv) {
     result = read_binary_objdump(data, size, &s_objdump_options);
     if (WABT_FAILED(result))
       goto done;
-    s_objdump_options.print_header = 0;
+    s_objdump_options.print_header = false;
+    s_objdump_options.log_stream = nullptr;
```
pipcet: I don't understand why you set `.log_stream = nullptr` only after the second pass through `read_binary_objdump`; that way, some output is duplicated when calling `wabtdump -h --debug hw.wasm`

sbc100: I have no idea why I did that.  Will fix.

sbc100: https://github.com/WebAssembly/wabt/pull/340

PR #342, opened 2017-03-08
Remove src/array.h and use std::vector instead

binji:
This caused the IntepreterModule destructor to become deleted, since the
import array is now a std::vector, in a union member.

This change also fixes the fallout from that by separating
InterpreterModule into HostInterpreterModule and
DefinedInterpreterModule, both derived from InterpreterModule, instead
of using a union.


src/binary-reader-interpreter.cc
```diff
@@ -395,8 +395,7 @@ static Result on_import(uint32_t index,
                         StringSlice field_name,
                         void* user_data) {
   Context* ctx = static_cast<Context*>(user_data);
-  assert(index < ctx->module->defined.imports.size);
-  InterpreterImport* import = &ctx->module->defined.imports.data[index];
+  InterpreterImport* import = &ctx->module->imports[index];
```
KarlSchimpf: Did you mean to drop the assert? If not, imports[index] does not assert check. However, imports.at(index) does.

src/binary-reader-interpreter.cc
```diff
@@ -569,29 +559,26 @@ static Result on_import_table(uint32_t import_index,
     return Result::Error;
   }
 
-  assert(import_index < ctx->module->defined.imports.size);
-  InterpreterImport* import = &ctx->module->defined.imports.data[import_index];
+  InterpreterImport* import = &ctx->module->imports[import_index];
```
KarlSchimpf: again, did you mean operator[] or method at()

src/interpreter.h
```diff
@@ -227,25 +227,42 @@ struct InterpreterHostImportDelegate {
...
+  size_t istream_start;
+  size_t istream_end;
+};
+struct HostInterpreterModule : InterpreterModule {
+  HostInterpreterModule(const StringSlice& name);
+
+  InterpreterHostImportDelegate import_delegate;
+};
+
+inline DefinedInterpreterModule* InterpreterModule::as_defined() {
```
KarlSchimpf: FYI, I think the inline need not be repeated here, since it was declared inline.

src/binary-reader-interpreter.cc
binji: Yeah, I figured the assert is kind of nice, but it's pretty rare to see people use .at() over operator[].

src/interpreter.h
binji: good point

src/interpreter.h
```diff
@@ -227,25 +227,42 @@ struct InterpreterHostImportDelegate {
...
+  HostInterpreterModule(const StringSlice& name);
+
+  InterpreterHostImportDelegate import_delegate;
+};
+
+DefinedInterpreterModule* InterpreterModule::as_defined() {
+  assert(!is_host);
+  return static_cast<DefinedInterpreterModule*>(this);
+}
+HostInterpreterModule* InterpreterModule::as_host() {
```
sbc100: Empty line between methods?

src/interpreter.h
```diff
@@ -227,25 +227,42 @@ struct InterpreterHostImportDelegate {
...
-      size_t istream_start;
-      size_t istream_end;
-    } defined;
-    struct {
-      InterpreterHostImportDelegate import_delegate;
-    } host;
-  };
 };
-WABT_DEFINE_VECTOR(interpreter_module, InterpreterModule);
+struct DefinedInterpreterModule : InterpreterModule {
```
sbc100: Does it make sense to 'class' for these, given we have constructors, destructors and inheritance?   Or are you avoiding 'class' for now?

binji: For now it feels like it will be visual noise, since all the user code expects the members to be public anyway.

binji:
Hm, the mingw build is still failing. Looks like as soon as you include `<vector>` and `<stdio.h>` it switches to mingw_printf and the format specifier checks stop working. Really tempted to just turn these off for mingw.

src/interpreter.h
binji: done

src/binary-reader-interpreter.cc
KarlSchimpf: Ok. I can agree to that.

Issue #324, opened 2017-02-28
MSYS2 6.3.0 build fails


binji:
The MSYS2 Mingw64 build fails here. Looks like it doesn't like `PRIu64`

```
C:/projects/wabt/src/validator.cc: In function 'void check_limits(Context*, const WabtLocation*, const WabtLimits*, uint64_t, const char*)':
C:/projects/wabt/src/validator.cc:627:52: error: unknown conversion type character 'l' in format [-Werror=format=]
                 desc, limits->initial, absolute_max);
                                                    ^
C:/projects/wabt/src/validator.cc:627:52: error: unknown conversion type character 'l' in format [-Werror=format=]
C:/projects/wabt/src/validator.cc:627:52: error: too many arguments for format [-Werror=format-extra-args]
C:/projects/wabt/src/validator.cc:633:50: error: unknown conversion type character 'l' in format [-Werror=format=]
                   desc, limits->max, absolute_max);
                                                  ^
C:/projects/wabt/src/validator.cc:633:50: error: unknown conversion type character 'l' in format [-Werror=format=]
C:/projects/wabt/src/validator.cc:633:50: error: too many arguments for format [-Werror=format-extra-args]
C:/projects/wabt/src/validator.cc:639:59: error: unknown conversion type character 'l' in format [-Werror=format=]
                   desc, limits->max, desc, limits->initial);
                                                           ^
C:/projects/wabt/src/validator.cc:639:59: error: format '%s' expects argument of type 'char*', but argument 5 has type 'uint64_t {aka long long unsigned int}' [-Werror=format=]
C:/projects/wabt/src/validator.cc:639:59: error: unknown conversion type character 'l' in format [-Werror=format=]
C:/projects/wabt/src/validator.cc:639:59: error: too many arguments for format [-Werror=format-extra-args]
cc1plus.exe: all warnings being treated as errors
```

rongjiecomputer:
I added this to `config.h`

```cpp
#ifdef __MINGW32__
#include <inttypes.h>

#undef PRIu64
#undef PRId64
#undef PRIx64
#define PRIu64 "I64u"
#define PRId64 "I64d"
#define PRIx64 "I64x"

#endif
```

[...]

PR #343, opened 2017-03-09
Disable print checking on mingw64, roll appveyor

binji:
It's not clear why, but it appears that the printf format checking isn't
correct on mingw64, see issue #324.


binji:
This is causing failures on 5.3.0 as well, see https://ci.appveyor.com/project/WebAssembly/wabt/build/1.0.689.

PR #344, opened 2017-03-09
[Mingw64] fix print format warning

rongjiecomputer:
Partial fix for #324 

In my previous `test.cpp` example in #324:

- Change both or just one of `stdio.h` and `stdint.h` to `cstdio` and `cstdint`: warnings.
- Add `stdlib.h`: warnings.

After checking `stdlib.h`, I found out that it includes `cstdlib` for C++ code and this (for some reason I do not know) is the culprit of the print format warnings. Adding `-D_GLIBCXX_INCLUDE_NEXT_C_HEADERS=1` solves this issue.

This only fixes print format warnings `-Wformat=`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,6 +96,7 @@ else ()
   if (MINGW)
     # Mingw won't define format (e.g. PRIu64) or limit (e.g. UINT32_MAX) macros
     # in C++ without these.
+    # See issue:324 and PR:344 for -D_GLIBCXX_INCLUDE_NEXT_C_HEADERS=1
     add_definitions(-D__STDC_LIMIT_MACROS=1 -D__STDC_FORMAT_MACROS=1 -D_GLIBCXX_INCLUDE_NEXT_C_HEADERS=1)
   endif ()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,7 +96,7 @@ else ()
   if (MINGW)
     # Mingw won't define format (e.g. PRIu64) or limit (e.g. UINT32_MAX) macros
     # in C++ without these.
-    add_definitions(-D__STDC_LIMIT_MACROS=1 -D__STDC_FORMAT_MACROS=1)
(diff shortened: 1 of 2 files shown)


rongjiecomputer:
First commit passes GCC 5.3.0, let's try GCC 6.3.0.

rongjiecomputer:
I was expecting some `-Wodr` warnings (appear in local build), but looks like they are gone on appveyor. Strange.

@binji Good enough to merge?

binji:
Cool, I was looking into this too. I found that

```
#define __STDC_FORMAT_MACROS 1
#include <inttypes.h>
#include <stdio.h>
#include <stdlib.h>

int main() {
  uint64_t i = 0;
  printf("foo %" PRIu64 "\n", i);
  return 0;
}
```

Will use the VC runtime printf and `%I64u`, whereas

```
#define __STDC_FORMAT_MACROS 1
#include <inttypes.h>
#include <stdlib.h>
#include <stdio.h>

int main() {
  uint64_t i = 0;
  printf("foo %" PRIu64 "\n", i);
  return 0;
}
```

(swapping order of stdlib.h and stdio.h) will use mingw_vprintf and `%llu`.

Defining `_GLIBCXX_INCLUDE_NEXT_C_HEADERS=1` first seems to fix both of them to use VC runtime printf. I'm not super happy about having to define this, but it seems like something is busted with the headers anyway, so I'm OK to merge.

CMakeLists.txt
```diff
@@ -96,7 +96,7 @@ else ()
   if (MINGW)
     # Mingw won't define format (e.g. PRIu64) or limit (e.g. UINT32_MAX) macros
     # in C++ without these.
-    add_definitions(-D__STDC_LIMIT_MACROS=1 -D__STDC_FORMAT_MACROS=1)
+    add_definitions(-D__STDC_LIMIT_MACROS=1 -D__STDC_FORMAT_MACROS=1 -D_GLIBCXX_INCLUDE_NEXT_C_HEADERS=1)
```
binji: Do you mind adding more to the comment here explaining why this define is needed? Maybe just reference the GH issue #344 and #324.

rongjiecomputer: Done!

binji:
Thanks!

PR #345, opened 2017-03-09
use array of chars instead of pointer to chars so sizeof is correct

jayphelps:
In #318 I incorrectly used `sizeof` on a pointer :man_facepalming: which obviously is not correct.


binji:
Oops, missed that myself. Looks like msvc won't let you use strlen here, but you can change the string to `const char format[] = "unable to read file %s";` and then sizeof will do the right thing.

jayphelps:
@binji done. Seems good.

binji:
Thanks!

PR #346, opened 2017-03-09
Fix mingw build by defining _POSIX

binji:
mingw32 tries to support both VC runtime printf and a POSIX-compliant
printf. It seems to randomly choose between the two, depending on
arbitrary factors like header include order. This change forces us to
always use the POSIX-compliant printf on mingw, and also changes the
WABT_PRINTF_FORMAT attribute to use the POSIX-compliant tag too
(gnu_printf, since printf defaults to ms_printf).

This seems to resolve the issues for #324, as well as the ones I found
in #342 as well.


sbc100:
lgtm

PR #348, opened 2017-03-10
binary reader API: pass module/field names to on_import_*

pipcet:
I think our binary reader API can be improved by passing module/field names to the specific on_import_{func,table,...} methods rather than requiring users to store the names in their private context.

There's no real reason to think of a function import, say, as two events: it's a single event, with many parameters, but that's not a problem for the current C++ code. Particularly for simple translating binary readers, saving temporary data in between calls that you're assuming are always paired makes the code slightly less readable.

### DIFF
--- a/src/binary-reader-objdump.cc
+++ b/src/binary-reader-objdump.cc
@@ -46,9 +46,6 @@ struct Context {
   uint32_t section_starts[kBinarySectionCount];
   BinarySection reloc_section;
 
-  StringSlice import_module_name;
-  StringSlice import_field_name;
-
   uint32_t next_reloc;
 };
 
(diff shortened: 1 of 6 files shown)


src/binary-reader-objdump.cc
```diff
@@ -414,26 +411,27 @@ static Result on_import(uint32_t index,
                         StringSlice module_name,
                         StringSlice field_name,
                         void* user_data) {
```
sbc100: Can we just remove this function completely?

pipcet: We could, yes. No strong opinion either way, but I left it in because there might be unknown new imports one day (which we couldn't deal with without code changes anyway, so...remove?)

sbc100: Yeah, I would remove it.  Otherwise its just noise.

binji:
Looks like a good change, thanks!

PR #349, opened 2017-03-10
Remove all uses of WABT_VECTOR from interpreter

binji:


src/binary-reader-interpreter.cc
```diff
@@ -890,8 +868,6 @@ static Result on_elem_segment_function_index_check(uint32_t index,
     return Result::Error;
   }
 
-  table->func_indexes[ctx->table_offset++] =
```
binji: This was actually a bug that was caught by the transition -- it shouldn't be updating the table in the "check" callback.

src/binary-reader.cc
```diff
@@ -466,6 +466,17 @@ static void write_indent(LoggingContext* ctx) {
     LOGF_NOINDENT(__VA_ARGS__); \
   } while (0)
 
+static void logging_on_error(BinaryReaderContext* context,
```
binji: This was a bug, basically the reader would crash when you turn on logging and there is an error.

sbc100: Hmm.. i just removed this didn't I?  Why does it crash? 

Perhaps we need more tests with logging enabled.   Can you make this a separate PR maybe?

binji: Good idea.

PR #351, opened 2017-03-11
Fix crash using binary reader logging with error

binji:
When the binary logging is turned on, it can't use the
`reader->on_error` directly since that function expects that the
user_data is the BinaryReaderInterpreter context (for example), but it
is actually the LoggingContext.

Some additional changes:

* Add a test that uses logging and succeeds
* Add a test that uses logging and has an error
* Change the `run-*-interp.py` scripts to use `-t` for tracing and `-v`
  for logging

### DIFF
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -149,11 +149,14 @@ static BinaryReaderContext* get_user_context(Context* ctx) {
 static void WABT_PRINTF_FORMAT(2, 3)
     raise_error(Context* ctx, const char* format, ...) {
   WABT_SNPRINTF_ALLOCA(buffer, length, format);
+  bool handled = false;
   if (ctx->reader->on_error) {
-    ctx->reader->on_error(get_user_context(ctx), buffer);
-  } else {
+    handled = ctx->reader->on_error(get_user_context(ctx), buffer);
+  }
+
+  if (!handled) {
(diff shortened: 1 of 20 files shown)


src/binary-reader.cc
```diff
@@ -2028,7 +2039,7 @@ Result read_binary(const void* data,
   WABT_ZERO_MEMORY(logging_reader);
   logging_reader.user_data = &logging_context;
 
-  logging_reader.on_error = reader->on_error;
+  logging_reader.on_error = logging_on_error;
```
sbc100: The reason I did this is because some tools don't register an on_error handler and what to default to default one.   If logging is enabled the default handler is not used, so the error gets swallowed.  Try wasmdump or wasm-link with --debug on a file with an error to see this.

binji: OK, I've reworked the on_error handlers to return a bool which says whether the error was handled.

CMakeLists.txt
```diff
@@ -106,10 +106,6 @@ else ()
     add_definitions(-Wno-clobbered)
   endif ()
 
-  if (COMPILER_IS_CLANG)
-    add_definitions(-fcolor-diagnostics)
-  endif ()
```
sbc100: Not part of this change?

binji: Whoops, done

PR #355, opened 2017-03-15
Add install target to Makefile

binji:
This means running `make` will generate a `bin` dir and put all the binaries in there. Similarly, `make clang-debug-asan` will install those binaries to the `bin` dir as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,15 +25,12 @@ GCC_FUZZ_CC := ${FUZZ_BIN_DIR}/afl-gcc
 GCC_FUZZ_CXX := ${FUZZ_BIN_DIR}/afl-g++
 EMSCRIPTEN_DIR ?= ${ROOT_DIR}/emscripten
 
-DEFAULT_COMPILER = CLANG
-DEFAULT_BUILD_TYPE = DEBUG
+DEFAULT_SUFFIX = clang-debug
 
 COMPILERS := GCC GCC_I686 GCC_FUZZ CLANG EMSCRIPTEN
 BUILD_TYPES := DEBUG RELEASE
 SANITIZERS := ASAN MSAN LSAN UBSAN
 CONFIGS := NORMAL $(SANITIZERS) COV NO_RE2C_BISON NO_TESTS
-EXECUTABLES := wast2wasm wasm2wast wasm-interp wasmopcodecnt hexfloat_test \
-	wasmdump wast-desugar wasm-link
 
 # directory names
 GCC_DIR := gcc/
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -88,51 +85,54 @@ NO_RE2C_BISON_PREFIX := -no-re2c-bison
...
 $(call TEST_TARGET,$(1),$(2),$(3)): $(call CMAKE_DIR,$(1),$(2),$(3))$$(BUILD_FILE)
-	$$(BUILD) -C $(call CMAKE_DIR,$(1),$(2),$(3)) run-tests
+	$$(BUILD_CMD) -C $(call CMAKE_DIR,$(1),$(2),$(3)) run-tests
+test-everything: $(CALL TEST_TARGET,$(1),$(2),$(3))
 endef
 
-.PHONY: all
-all: ${EXECUTABLES}
+.PHONY: all install
+all: install$(DEFAULT_SUFFIX)
```
sbc100: Why do you want install to be part of 'all'?  Why not leave install as a separate target?

binji: Good point, I guess it is convenient for me, but it's probably confusing to everyone else who expect `make install` to install the binaries to /usr/local or something.

binji:
Changed so `all` target doesn't `install` by default.

Makefile
```diff
@@ -88,51 +85,52 @@ NO_RE2C_BISON_PREFIX := -no-re2c-bison
...
 .PHONY: $(call TEST_TARGET,$(1),$(2),$(3))
 $(call TEST_TARGET,$(1),$(2),$(3)): $(call CMAKE_DIR,$(1),$(2),$(3))$$(BUILD_FILE)
-	$$(BUILD) -C $(call CMAKE_DIR,$(1),$(2),$(3)) run-tests
+	$$(BUILD_CMD) -C $(call CMAKE_DIR,$(1),$(2),$(3)) run-tests
+test-everything: $(CALL TEST_TARGET,$(1),$(2),$(3))
 endef
 
-.PHONY: all
-all: ${EXECUTABLES}
+.PHONY: all install test
```
sbc100: Where was the 'test' target being define previously?

binji: Took me a while to remember; marked the line with a comment.

Makefile
```diff
@@ -151,12 +149,6 @@ src/prebuilt/ast-parser-gen.cc: src/ast-parser.y
 src/prebuilt/ast-lexer-gen.cc: src/ast-lexer.cc
 	re2c --no-generation-date -bc -o $@ $<
 
-# defaults with simple names
-$(foreach EXECUTABLE,$(EXECUTABLES), \
-	$(eval $(call DEFAULT,$(DEFAULT_COMPILER),$(DEFAULT_BUILD_TYPE),$(EXECUTABLE),NORMAL)) \
```
binji: $4 ends up being NORMAL here

Makefile
```diff
@@ -88,51 +85,52 @@ NO_RE2C_BISON_PREFIX := -no-re2c-bison
...
 endif
 
 CMAKE_DIR = out/$($(1)_DIR)$($(2)_DIR)$($(3)_DIR)
-EXE_TARGET = $($(1)_PREFIX)$($(2)_PREFIX)$($(3)_PREFIX)
+BUILD_TARGET = $($(1)_PREFIX)$($(2)_PREFIX)$($(3)_PREFIX)
+INSTALL_TARGET = install-$($(1)_PREFIX)$($(2)_PREFIX)$($(3)_PREFIX)
 TEST_TARGET = test-$($(1)_PREFIX)$($(2)_PREFIX)$($(3)_PREFIX)
 
-define DEFAULT
-.PHONY: $(3)$($(4)_PREFIX) test$($(4)_PREFIX)
```
binji: And NORMAL_PREFIX is empty here

Makefile
```diff
@@ -88,51 +85,52 @@ NO_RE2C_BISON_PREFIX := -no-re2c-bison
...
-$(3)$($(4)_PREFIX): $(call EXE_TARGET,$(1),$(2),$(4))
-	ln -sf ../$(call CMAKE_DIR,$(1),$(2),$(4))$(3) out/$(3)$($(4)_PREFIX)
```
jayphelps: After this change, the symlinks are no longer available in `out/*` as the README describes. Is the README now out of date (instead `bin/*`) or should the symlinks be added back?

binji: Yeah, I've noticed this too. I guess the README should be updated, and the test runner should use out/clang/Debug too.

PR #356, opened 2017-03-15
Fix potential one-definition-rule breaking structs

binji:
Context is the most common potential problem here, but I noticed that
wasm-link.h had some others too (DataSegment and Export).

These probably aren't an issue yet, since they all have trivial
constructors. As soon as we add non-trivial types to these they break
though.

### DIFF
--- a/src/ast-writer.cc
+++ b/src/ast-writer.cc
@@ -33,6 +33,8 @@
 
 namespace wabt {
 
+namespace {
+
 static const uint8_t s_is_char_escaped[] = {
     1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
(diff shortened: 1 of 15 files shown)


src/ast-writer.cc
```diff
@@ -69,6 +71,8 @@ struct Context {
   int func_type_index;
 };
 
+}  // namespace
```
sbc100: Is there a reason not to put these outside the wabt namespace?

binji: Could do, shouldn't really make a difference either way. Is there a benefit to not having it nested?

binji:
Looks like the Travis error is a flaky git fetch.

src/ast-writer.cc
KarlSchimpf: Using the anonymous namespace is standard practice in C++ for static variables, since all variables in the anonymous namespace is automatically static.

However, I forgot to notice you should remove the "static" because the name is in the anonymous namespace (again, standard C++ practice).

binji: Agreed, though that's a bigger change. Mind if I do it in another PR?

KarlSchimpf: No problem commit as is.

PR #357, opened 2017-03-16
Use std::unordered_multimap for BindingHash

binji:
This change can't really be done in isolation, since once we add members
with constructors to a struct, it is no longer has a trivial
constructor. This propagates through all types that use it, etc.

There are a number of changes that are ugly, but hopefully reduced the
amount of code that needed to change. In particular, I changed some
union members to pointers so they would stay trivially constructable.

Another tricky change is the handling of duplicate bindings: previously
we just relied on the fact that our hash implementation would be
consistent. A nicer solution is to display the duplicated bindings in
source order. There's probably a nicer way to do this; bikeshedding
welcome. :-)

### DIFF
--- a/src/ast.cc
+++ b/src/ast.cc
@@ -23,12 +23,12 @@ namespace wabt {
 
 int get_index_from_var(const BindingHash* hash, const Var* var) {
   if (var->type == VarType::Name)
-    return find_binding_index_by_name(*hash, var->name);
+    return hash->find_index(var->name);
   return static_cast<int>(var->index);
 }
 
 ExportPtr get_export_by_name(const Module* module, const StringSlice* name) {
-  int index = find_binding_index_by_name(module->export_bindings, *name);
+  int index = module->export_bindings.find_index(*name);
   if (index == -1)
     return nullptr;
   return module->exports.data[index];
@@ -58,11 +58,11 @@ int get_local_index_by_var(const Func* func, const Var* var) {
   if (var->type == VarType::Index)
     return static_cast<int>(var->index);
 
-  int result = find_binding_index_by_name(func->param_bindings, var->name);
+  int result = func->param_bindings.find_index(var->name);
   if (result != -1)
     return result;
 
-  result = find_binding_index_by_name(func->local_bindings, var->name);
+  result = func->local_bindings.find_index(var->name);
   if (result == -1)
     return result;
 
@@ -321,11 +321,11 @@ Func::Func() : first_expr(nullptr) {
   WABT_ZERO_MEMORY(local_types);
 }
 
-void destroy_func(Func* func) {
-  destroy_string_slice(&func->name);
-  destroy_func_declaration(&func->decl);
-  destroy_type_vector(&func->local_types);
-  destroy_expr_list(func->first_expr);
+Func::~Func() {
+  destroy_string_slice(&name);
+  destroy_func_declaration(&decl);
+  destroy_type_vector(&local_types);
+  destroy_expr_list(first_expr);
 }
 
 void destroy_global(Global* global) {
@@ -338,7 +338,6 @@ void destroy_import(Import* import) {
   destroy_string_slice(&import->field_name);
   switch (import->kind) {
     case ExternalKind::Func:
-      destroy_func(import->func);
       delete import->func;
       break;
     case ExternalKind::Table:
@@ -380,7 +379,6 @@ void destroy_table(Table* table) {
 static void destroy_module_field(ModuleField* field) {
   switch (field->type) {
     case ModuleFieldType::Func:
-      destroy_func(field->func);
       delete field->func;
       break;
     case ModuleFieldType::Global:
@@ -435,10 +433,10 @@ Module::Module()
   WABT_ZERO_MEMORY(data_segments);
 }
 
-void destroy_module(Module* module) {
-  destroy_string_slice(&module->name);
+Module::~Module() {
+  destroy_string_slice(&name);
 
-  ModuleField* field = module->first_field;
+  ModuleField* field = first_field;
   while (field) {
     ModuleField* next_field = field->next;
     destroy_module_field(field);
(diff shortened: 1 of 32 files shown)


src/ast.cc
```diff
@@ -403,13 +380,15 @@ void destroy_table(Table* table) {
 static void destroy_module_field(ModuleField* field) {
   switch (field->type) {
     case ModuleFieldType::Func:
-      destroy_func(&field->func);
+      destroy_func(field->func);
+      delete field->func;
```
KarlSchimpf: Is "*field->func" a class object? If so, the destructor can call "destroy_func" and this can be simplified to just: "delete field->func".

src/ast.cc
```diff
@@ -435,6 +414,27 @@ static void destroy_module_field(ModuleField* field) {
...
+  WABT_ZERO_MEMORY(imports);
+  WABT_ZERO_MEMORY(exports);
+  WABT_ZERO_MEMORY(func_types);
+  WABT_ZERO_MEMORY(tables);
+  WABT_ZERO_MEMORY(elem_segments);
+  WABT_ZERO_MEMORY(memories);
+  WABT_ZERO_MEMORY(data_segments);
+}
+
 void destroy_module(Module* module) {
```
KarlSchimpf: Not knowing your plans, shouldn't this be converted to a destructor for class Module? This would clean up destruction. It would also simplify destruction of pointers to just a call to "delete".

Note: I understand you are in the process of (slowly) changing things. Therefore this may already be on your road map. Just bringing it up because the code now is somewhat "ugly" and confusing. That is, it isn't immediately obvious when to use:

      destroy_FOO(&F...)

and when to use:

    destroy_FOO(F...); delete F;

The current set up is forcing you to "copy" (non-obvious) destructor sequences to everything going out of scope, rather then letting C++ destructor rules automatically handle it.

src/ast.h
```diff
@@ -140,6 +144,8 @@ struct FuncDeclaration {
 };
 
 struct Func {
+  Func();
+
```
KarlSchimpf: I see that you have mulltiple cases where a Func instance is created with a name, or a name and flags. Why not define constructors for those cases so that you don't duplicate code to initiate such cases?

Again, I don't know if you expect to do that as a separate step.

src/binding-hash.h
```diff
@@ -18,34 +18,38 @@
...
-  BindingHashEntry* free_head;
-};
-
-Binding* insert_binding(BindingHash*, const StringSlice*);
-void remove_binding(BindingHash*, const StringSlice*);
-bool hash_entry_is_free(const BindingHashEntry*);
-/* returns -1 if the name is not in the hash */
-int find_binding_index_by_name(const BindingHash*, const StringSlice* name);
-void destroy_binding_hash(BindingHash*);
+typedef std::unordered_multimap<std::string, Binding> BindingHash;
```
KarlSchimpf: Somehow, this code seems to be missing a class that cleanly captures the API  better than this.

src/binding-hash.cc
```diff
@@ -16,164 +16,77 @@
...
-      /* yes, so add this new entry to the chain, even if it is already there */
-      /* add as the second entry in the chain */
-      free_entry->next = entry->next;
-      entry->next = free_entry;
-      entry = free_entry;
+namespace {
+
+typedef std::vector<const BindingHash::value_type*> BindingHashValueTypeVector;
+
+BindingHashValueTypeVector create_binding_duplicates_vector(
```
KarlSchimpf: Why not have this get a (reference or pointer) argument that it initializes, so that it doesn't have to copy the local variable "duplicates" back to the caller.

src/binding-hash.cc
```diff
@@ -16,164 +16,77 @@
...
+    const BindingHash& bindings) {
+  // This relies on the fact that in an unordered_multimap, all values with the
+  // same key are adjacent in iteration order.
+  std::vector<const BindingHash::value_type*> duplicates;
```
KarlSchimpf: Isn't this of type BindingHashValueTypeVector?

binji:
Agreed, I was trying to reduce the amount of stuff that needed to change, but maybe I erred on too little. I'll see if I can clean up some things.

src/ast.cc
binji: Done.

src/ast.cc
binji: Done.

src/ast.h
binji: Yeah, I think this would be better as a separate step.

src/binding-hash.cc
binji: I figured this would do RVO. But I can do it that way too.

src/binding-hash.cc
binji: Done.

src/binding-hash.h
```diff
@@ -18,34 +18,47 @@
...
   int index;
 };
 
-struct BindingHashEntry {
-  Binding binding;
-  struct BindingHashEntry* next;
-  struct BindingHashEntry* prev; /* only valid when this entry is unused */
-};
-WABT_DEFINE_VECTOR(binding_hash_entry, BindingHashEntry);
+class BindingHash : public std::unordered_multimap<std::string, Binding> {
```
KarlSchimpf: I'm a bit hesitant about using C++ containers as a base class, because it doesn't have a virtual destructor. However, this is only important if you are going to cast pointers to a BindingHash to its base class, and then delete it.

Considering how you are using this class, it should be acceptable. However, I might add a comment that one shouldn't cast pointers of this class to the C++ container base.

KarlSchimpf: BTW, I like this better than what you had. I might have moved the base class into a field of the class, but don't know it that would have introduced the need for helper accessors.

binji: Good idea.

binji: I considered it, and it shouldn't be too bad because I think the only function I currently use from `unordered_multimap` is `emplace`. Should be a relatively simple refactor if we want to do that later.

src/ast.cc
```diff
@@ -403,13 +379,14 @@ void destroy_table(Table* table) {
...
     case ModuleFieldType::Func:
-      destroy_func(&field->func);
+      delete field->func;
       break;
     case ModuleFieldType::Global:
       destroy_global(&field->global);
       break;
     case ModuleFieldType::Import:
-      destroy_import(&field->import);
+      destroy_import(field->import);
```
binji: whoops, missed this one.

PR #359, opened 2017-03-17
Add wabt::string_view, based on C++17 string_view

binji:
This change also adds `wabt-unittests` with `string_view` tests, and runs them on appveyor and travis.

### DIFF
--- a/scripts/travis-test.sh
+++ b/scripts/travis-test.sh
@@ -27,7 +27,8 @@ log_and_run() {
 }
 
 run_tests() {
-  (cd ${ROOT_DIR} && log_and_run test/run-tests.py ${RUN_TEST_ARGS} $* --timeout=10)
+  (cd ${ROOT_DIR} && log_and_run test/run-tests.py --bindir ${BINDIR} $* --timeout=10) && true
+  (log_and_run ${BINDIR}/wabt-unittests) && true
 }
 
 set_run_test_args() {
@@ -35,7 +36,7 @@ set_run_test_args() {
   local BUILD_TYPE=$2
   local CONFIG=${3:-}
 
-  RUN_TEST_ARGS="--bindir out/${COMPILER}/${BUILD_TYPE}/${CONFIG}"
+  BINDIR="out/${COMPILER}/${BUILD_TYPE}/${CONFIG}"
 }
 
 if [ ${CC} = "gcc" ]; then
--- a/src/string-view.h
+++ b/src/string-view.h
@@ -1,0 +1,288 @@
+/*
+ * Copyright 2017 WebAssembly Community Group participants
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef WABT_STRING_VIEW_H_
+#define WABT_STRING_VIEW_H_
+
+#include <cassert>
+#include <functional>
+#include <iterator>
+#include <string>
+
+#include "hash-util.h"
+
+namespace wabt {
+
+// This is a simplified implemention of C++17's basic_string_view template.
+//
+// Missing features:
+// * Not a template
+// * No allocator support
+// * Some functions are not implemented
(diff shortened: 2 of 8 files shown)


src/test-string-view.cc
```diff
@@ -0,0 +1,437 @@
...
+#include "string-view.h"
+
+using namespace wabt;
+
+static string_view::size_type npos = string_view::npos;
+
+TEST(string_view, default_constructor) {
+  string_view sv;
+  ASSERT_EQ(nullptr, sv.data());
+  ASSERT_EQ(0U, sv.size());
```
KarlSchimpf: Wouldn't this be safer?

ASSERT(size_t(0), sv.size());

src/test-string-view.cc
```diff
@@ -0,0 +1,437 @@
...
+}
+
+TEST(string_view, copy_constructor) {
+  string_view sv1("foo");
+  string_view sv2(sv1);
+
+  ASSERT_EQ(3U, sv2.size());
+  ASSERT_EQ('f', sv2[0]);
```
KarlSchimpf: How about writing a helper function that takes a "const char*" and a string_view, and iterates of the strlen() of the first argument and tests if the contents of the string_view matches (via ASSERT_EQ). Then most of these tests can just call that helper function to test contents.

It will also allow you to test more strings than just "foo",

src/string-view.h
```diff
@@ -0,0 +1,259 @@
...
+  void remove_prefix(size_type n);
+  void remove_suffix(size_type n);
+  void swap(string_view& s) noexcept;
+
+  // string operations
+  explicit operator std::string() const;
+  std::string to_string() const;
+  size_type copy(char* s, size_type n, size_type pos = 0) const;
+  string_view substr(size_type pos = 0, size_type n = npos) const;
+  int compare(string_view s) const noexcept;
```
binji: This matches the definition here: http://en.cppreference.com/w/cpp/header/string_view

src/string-view.h
```diff
@@ -0,0 +1,259 @@
...
+  size_type find_first_of(char c, size_type pos = 0) const noexcept;
+  size_type find_first_of(const char* s, size_type pos, size_type n) const;
+  size_type find_first_of(const char* s, size_type pos = 0) const;
+  size_type find_last_of(string_view s, size_type pos = npos) const noexcept;
+  size_type find_last_of(char c, size_type pos = npos) const noexcept;
+  size_type find_last_of(const char* s, size_type pos, size_type n) const;
+  size_type find_last_of(const char* s, size_type pos = npos) const;
+
+  // TODO(binji): implement
+#if 0
```
binji: I figured it was useful to see what is not implemented from C++17 basic_string_view. I can remove it if you feel strongly

src/string-view.h
```diff
@@ -0,0 +1,259 @@
...
+
+inline string_view::const_reverse_iterator string_view::crend() const noexcept {
+  return const_reverse_iterator(cbegin());
+}
+
+inline string_view::size_type string_view::size() const noexcept {
+  return size_;
+}
+
+inline string_view::size_type string_view::length() const noexcept {
```
binji: historical, since std::string does the same, I guess.

src/test-string-view.cc
binji: How so?

src/test-string-view.cc
binji: Good idea.

src/string-view.h
KarlSchimpf: If you leave in, add comment why

src/string-view.h
KarlSchimpf: Ok.

src/test-string-view.cc
KarlSchimpf: My suggestion works independent of the number of bytes (32/64), and how ints are defined.

src/string-view.h
binji: Already done. :-D

scripts/travis-test.sh
```diff
@@ -27,15 +27,16 @@ log_and_run() {
 }
 
 run_tests() {
-  (cd ${ROOT_DIR} && log_and_run test/run-tests.py ${RUN_TEST_ARGS} $* --timeout=10)
+  (cd ${ROOT_DIR} && log_and_run test/run-tests.py --bindir ${BINDIR} $* --timeout=10) && true
+  (log_and_run ${BINDIR}/wabt-unittests) && true
```
sbc100: Do you need the && true here?

Also do you need the brackes around this line (its not doing a chdir, like the one above).

binji: Not sure the right way to do this stuff in shell script. But the errexit was making it blow up without running the other tests, which I didn't want. And I thought the subshell would override that, but it doesn't seem to do it without `(...) && true`

Maybe better to just disable errexit here?

PR #360, opened 2017-03-17
Define __STDC_*_MACROS for all non-MSVC compilers

jgravelle-google:
This should make the waterfall's Travis build happy


jgravelle-google:
More context: http://en.cppreference.com/w/cpp/types/integer#Notes

PR #361, opened 2017-03-20
Fix lexing from binary buffer

Cellule:
Fix bug when lexing from buffer where subsequent read from the buffer overwrites bytes still available.
Once the lexing reaches the end of the current read buffer, it currently copies the remaining bytes at the beginning of the read buffer instead of the end. This means it overwrites bytes that hasn't been read yet.

ie:
```cpp
// full wast  `(module (func $dummy))`
// initial buffer size = 12
// Before Fill()
lexer->buffer = "(module (fun";
lexer->token = "(fun";
// calling fill because there is not enough content left in the read buffer

// after fill now
lexer->token = lexer->buffer = " $dummy))";

// after fill with this fix
lexer->token = lexer->buffer = "(func $dummy))"
```

### DIFF
--- a/src/ast-lexer.cc
+++ b/src/ast-lexer.cc
@@ -149,7 +149,7 @@ static Result fill(Location* loc,
     size_t bytes_left = lexer->source.buffer.size - offset;
     if (read_size > bytes_left)
       read_size = bytes_left;
-    memcpy(lexer->buffer,
+    memcpy(lexer->limit,
            static_cast<const char*>(lexer->source.buffer.data) + offset,
            read_size);
     lexer->source.buffer.read_offset += read_size;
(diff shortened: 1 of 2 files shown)


binji:
Thanks for this! Do you know of a repro for this bug? I'd like to prevent a regression here.

Cellule:
I don't know if there is a binary in wabt that uses binary buffer as input.
I am using wabt in chakra to parse the spec test. One in particular that failed parsing because of this is `br_table.wast`, but any wast bigger than ~60kb will have an issue because of this.
Maybe you can make wast2wasm to fully read and cache the .wast input and use the lexer with the cached buffer to test this code path.

binji:
OK, makes sense. I opened an issue to create a regression test, but let's land this for now.

PR #362, opened 2017-03-21
MSVC++ 12.0 float hex string parse workaround

Cellule:
Allow parse_double_hex and parse_float_hex to parse strings without a power.

In MSVC++ 12.0 (and previous versions ?) the float parsing doesn't support hex strings.
Not totally sure if it's a bug or just not supported, but this change is a workaround it for older MSVC compilers.

I tried to minimize changes to not affect modern compilers as `strtof` and `strtod` give the right result.
I used `parse_float_hex` and `parse_double_hex` to parse hex strings and fallback on normal implementation if not an hex string.

### DIFF
--- a/src/literal.cc
+++ b/src/literal.cc
@@ -283,37 +283,37 @@ static void parse_float_hex(const char* s,
   if (!seen_dot)
     significand_exponent += significand_shift;
 
-  assert(s < end && *s == 'p');
-  s++;
-
   if (significand == 0) {
     /* 0 or -0 */
     *out_bits = make_float(is_neg, F32_MIN_EXP, 0);
     return;
   }
 
-  assert(s < end);
   int exponent = 0;
   bool exponent_is_neg = false;
-  /* exponent is always positive, but significand_exponent is signed.
-   significand_exponent_add is negated if exponent will be negative, so it  can
-   be easily summed to see if the exponent is too large (see below) */
-  int significand_exponent_add = 0;
-  if (*s == '-') {
-    exponent_is_neg = true;
-    significand_exponent_add = -significand_exponent;
+  if (s < end) {
+    assert(*s == 'p');
     s++;
-  } else if (*s == '+') {
-    s++;
-    significand_exponent_add = significand_exponent;
-  }
+    /* exponent is always positive, but significand_exponent is signed.
+     significand_exponent_add is negated if exponent will be negative, so it  can
+     be easily summed to see if the exponent is too large (see below) */
+    int significand_exponent_add = 0;
+    if (*s == '-') {
+      exponent_is_neg = true;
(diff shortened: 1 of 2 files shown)


binji:
Is this affecting some new code? I checked to see where `strtof` and `strtod` are used, and it's currently only in `literal.cc`.

src/literal.cc
```diff
@@ -842,4 +843,37 @@ void write_double_hex(char* out, size_t size, uint64_t bits) {
   out[len] = '\0';
 }
 
+#if COMPILER_IS_MSVC
+#if _MSC_VER <= 1800
+float strtof(const char *nptr, char **endptr) {
+  const char* end = nptr + strlen(nptr);
+  // review:: should we check for leading whitespaces ?
```
binji: Yeah, probably better to match the behavior of `strtof` and slip whitespace.

src/literal.cc
```diff
@@ -842,4 +843,37 @@ void write_double_hex(char* out, size_t size, uint64_t bits) {
...
+  if (string_starts_with(nptr, end, "0x")) {
```
binji: It can start with a leading '+' or '-' too.

Cellule:
> Is this affecting some new code? I checked to see where strtof and strtod are used, and it's currently only in literal.cc, where it's only used when it's known that there are no hex floats.

It is not affecting new code. Only older MSVC compiler have bad float parsing implementation.
Right now `(f32.const 0xf32)` is a valid float format and we'll parse it in literal.cc `parse_float` with `LiteralType::Int`

I noticed that some recent changes prevent building in VS 2013. Notably the use of `constexpr` and `noexcept` keywords.
I wonder if we want to keep supporting VS2013, initially I was doing this work because we were supporting it in Chakra, but we are talking about dropping support soon https://github.com/Microsoft/ChakraCore/issues/2549

binji:
> It is not affecting new code. Only older MSVC compiler have bad float parsing implementation.
Right now (f32.const 0xf32) is a valid float format and we'll parse it in literal.cc parse_float with LiteralType::Int

That seems to be the correct behavior (checked against spec interpreter). Makes sense, I think, since we don't want to make the lexer context-sensitive.

> I noticed that some recent changes prevent building in VS 2013. Notably the use of constexpr and noexcept keywords.
I wonder if we want to keep supporting VS2013, initially I was doing this work because we were supporting it in Chakra, but we are talking about dropping support soon Microsoft/ChakraCore#2549

I'm not opposed to supporting VS2013 for now (though I know some others are eager to drop it). If we decide to keep it, we'll have to add an appveyor build so it doesn't regress.

Cellule:
I will close this PR for now.
I found out that some other compiler also have similar issue. I have to investigate what to do about it. I will probably just make a patch for windows

PR #364, opened 2017-03-21
wasmdump: Improve printing on relocation information

sbc100:
Print the addend and make it more clear what the
different offsets mean.

### DIFF
--- a/src/binary-reader-objdump.cc
+++ b/src/binary-reader-objdump.cc
@@ -253,8 +253,18 @@ static void log_opcode(Context* ctx,
           ctx->section_starts[static_cast<size_t>(BinarySection::Code)];
       size_t abs_offset = code_start + reloc->offset;
       if (ctx->last_opcode_end > abs_offset) {
-        printf("           %06" PRIzx ": %s\t%d\n", abs_offset,
+        printf("           %06" PRIzx ": %s\t%d", abs_offset,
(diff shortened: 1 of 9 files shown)


pipcet:
This is horribly nitpicky, but is there an official line on whether it's `R_WEBASSEMBLY_FUNC_INDEX_LEB` or `R_FUNC_INDEX_LEB`?

src/binary-reader-objdump.cc
```diff
@@ -633,8 +643,8 @@ Result on_reloc(RelocType type,
   Context* ctx = static_cast<Context*>(user_data);
   uint32_t total_offset =
       ctx->section_starts[static_cast<size_t>(ctx->reloc_section)] + offset;
-  print_details(ctx, "   - %-18s offset=%#x (%#x)\n", get_reloc_type_name(type),
-                total_offset, offset);
+  print_details(ctx, "   - %-18s idx=%#-4x addend=%#-4x offset=%#x(file=%#x)\n",
```
sbc100: Yes

sbc100:
I think I used R_FUNC_INDEX_LEB so save on horizontal space and fit more one line.  Its also somewhat redundant since all relocations are by definition web assembly.  Perhaps I should also drop the R_ to avoid confusion with the names in the Linking.md document?

PR #366, opened 2017-03-22
Replace wabt::*vector with std::vector

binji:
This propagates through a lot of code since std::vector will call
constructors and destructors. In particular, this CL adds many default
constructors and destructors to previously POD types. Many of them are
only there to construct `Var` and `StringSlice` types, so they likely
can be removed when those have their own constructors.

Since unions members cannot contain constructors or destructors
(without additional implementation), this CL changes those members to
pointers instead. (Perhaps in a future CL these will be std::variant
instead of union, so the members can be value types again.)


src/wasm-link.h
```diff
@@ -94,24 +94,22 @@ struct Section {
    * section. */
   size_t output_payload_offset;
 };
-WABT_DEFINE_VECTOR(section, Section);
-
-typedef Section* SectionPtr;
-WABT_DEFINE_VECTOR(section_ptr, SectionPtr);
-
-WABT_DEFINE_VECTOR(string_slice, StringSlice);
+typedef std::vector<Section*> SectionPtrVector;
```
sbc100: Looks like there is a newline missing above this line, although unrelated to this change.

src/tools/wasm-link.cc
```diff
@@ -50,22 +52,26 @@ static Option s_options[] = {
...
+
   Stream stream;
-  LinkerInputBinaryVector inputs;
-  ssize_t current_section_payload_offset;
+  std::vector<std::unique_ptr<LinkerInputBinary>> inputs;
+  ssize_t current_section_payload_offset = 0;
 };
 
+Context::Context() {
+  WABT_ZERO_MEMORY(stream);
```
sbc100: Why not just inline this?

src/tools/wasm-link.cc
```diff
@@ -231,6 +253,12 @@ static void write_slice(Stream* stream, StringSlice str, const char* desc) {
   write_str(stream, str.start, str.length, PrintChars::Yes, desc);
 }
 
+static void write_string(Stream* stream,
+                         const std::string& str,
+                         const char* desc) {
+  write_str(stream, str.data(), str.length(), PrintChars::Yes, desc);
+}
```
sbc100: Do other tools don't also need this too?

src/ast-parser.y
```diff
@@ -592,35 +597,37 @@ block_instr :
...
+      CHECK_END_LABEL(@8, $$->if_.true_->label, $8);
     }
 ;
 block :
     value_type_list instr_list {
-      WABT_ZERO_MEMORY($$);
-      $$.sig = $1;
-      $$.first = $2.first;
+      $$ = new Block();
+      $$->sig = std::move(*$1);
```
KarlSchimpf: Nit: Should the "delete $1" immediately follow the move, to make it clearer that a copy was constructed, and the old value should be deleted?

Note: Not that I asking for more changes, but at some point, you may want to change field "sig" to be a unique_ptr so that you don't have to build copies (just take over ownership), and let the destructor of what the unique_ptr appears in automatically delete the object.

src/ast-parser.y
```diff
@@ -635,55 +642,65 @@ expr1 :
...
+      expr->if_.true_->first = $2.first;
       expr->if_.false_ = $3.first;
       $$ = join_exprs2(&@1, &$1, expr);
     }
   | expr expr {
       Expr* expr = new_if_expr();
-      expr->if_.true_.first = $2.first;
+      expr->if_.true_ = new Block();
+      expr->if_.true_->first = $2.first;
+      expr->if_.false_ = nullptr;
```
KarlSchimpf: Should the initialization of field false_ start with the nullptr, removing the need for this explicit assignment?

src/ast-parser.y
```diff
@@ -1112,12 +1149,15 @@ inline_export :
...
-      WABT_ZERO_MEMORY($$);
-      $$.sig = $3;
+      $$ = new FuncType();
+      $$->sig = std::move(*$3);
+      delete $3;
     }
   | LPAR TYPE bind_var func_type RPAR {
-      $$.name = $3;
-      $$.sig = $4;
+      $$ = new FuncType();
```
KarlSchimpf: nit: Alternatively, you could define separate constructors that take these arguments explicitly, reducing the code to

  $$ = new FuncType($3, $4)

This will also allow the delete to be inside the construtor.

src/ast-parser.y
```diff
@@ -1324,66 +1363,68 @@ script_var_opt :
 
 action :
     LPAR INVOKE script_var_opt quoted_text const_list RPAR {
-      WABT_ZERO_MEMORY($$);
-      $$.loc = @2;
-      $$.module_var = $3;
-      $$.type = ActionType::Invoke;
-      $$.invoke.name = $4;
-      $$.invoke.args = $5;
+      $$ = new Action();
```
KarlSchimpf: Again, you might consider some explicit constructors for Action, to match the cases of the parser. This could also be applied to most parsing constructs.

Note: This doesn't have to be in this PR, just a suggestion on further cleanup.

src/binary-reader-linker.cc
```diff
@@ -125,7 +124,8 @@ static Result begin_section(BinaryReaderContext* ctx,
                             uint32_t size) {
   Context* context = static_cast<Context*>(ctx->user_data);
   LinkerInputBinary* binary = context->binary;
-  Section* sec = append_section(&binary->sections);
+  Section* sec = new Section();
+  binary->sections.emplace_back(sec);
```
binji: This case is a bit weird since sections is a vector of std::unique_ptr, so it is creating a new object, just not a Section. :-)

src/wasm-link.h
binji: Done.

src/tools/wasm-link.cc
binji: I was trying to avoid inlining constructors unless they were trivial (including members that would be default constructed). But yeah, this one looks reasonable to inline.

src/apply-names.cc
```diff
@@ -43,35 +42,33 @@ struct Context {
...
-      Label* label = ctx->labels.data[i];
+    for (int i = ctx->labels.size() - 1; i >= 0; --i) {
+      Label* label = ctx->labels[i];
       if (string_slices_are_equal(label, &var->name))
         return label;
     }
     return nullptr;
   } else {
-    if (var->index < 0 || static_cast<size_t>(var->index) >= ctx->labels.size)
+    if (var->index < 0 || static_cast<size_t>(var->index) >= ctx->labels.size())
```
binji: Hm, true. Though I kind of like that it's easier to grep for static_cast. In reality, it might be best to change var->index to a size_t since this ends up happening all over.

src/ast-parser.y
binji: Good idea, I'll move the deletes.

Agreed, it will be better to use unique_ptr more places. Can't do it here as easily since the bison token is a union, so no constructors. :-|

src/ast-parser.y
binji: Yeah, this is pretty clunky and error-prone. Let me see if I can do something nicer.

src/binary-reader-opcnt.cc
```diff
@@ -35,49 +35,37 @@ struct Context {
...
 
 static Result add_int_pair_counter_value(IntPairCounterVector* vec,
-                                             intmax_t first,
-                                             intmax_t second) {
-  for (size_t i = 0; i < vec->size; ++i) {
-    if (vec->data[i].first == first && vec->data[i].second == second) {
-      ++vec->data[i].count;
+                                         intmax_t first,
+                                         intmax_t second) {
+  for (size_t i = 0; i < vec->size(); ++i) {
```
KarlSchimpf: Couldn't you use a simpler for loop here, since only one element is being accessed at a time.

for (IntPairCounter& counter : vec) { ... }

src/binary-writer.cc
```diff
@@ -389,11 +394,11 @@ static void write_expr(Context* ctx,
       break;
     case ExprType::BrTable: {
       write_opcode(&ctx->stream, Opcode::BrTable);
-      write_u32_leb128(&ctx->stream, expr->br_table.targets.size,
+      write_u32_leb128(&ctx->stream, expr->br_table.targets->size(),
                        "num targets");
       uint32_t depth;
-      for (size_t i = 0; i < expr->br_table.targets.size; ++i) {
-        depth = get_label_var_depth(ctx, &expr->br_table.targets.data[i]);
+      for (size_t i = 0; i < expr->br_table.targets->size(); ++i) {
```
KarlSchimpf: How about:
    for (...& target : expr->br_table.targets) { depth = ...(ctx, *target); ....}

src/binary-writer.cc
```diff
@@ -631,18 +636,18 @@ static void write_reloc_section(Context* ctx, RelocSection* reloc_section) {
...
-
-  for (size_t i = 0; i < relocs->size; i++) {
-    write_u32_leb128_enum(&ctx->stream, relocs->data[i].type, "reloc type");
-    write_u32_leb128(&ctx->stream, relocs->data[i].offset, "reloc offset");
-    write_u32_leb128(&ctx->stream, relocs->data[i].index, "reloc index");
-    switch (relocs->data[i].type) {
+  std::vector<Reloc>& relocs = reloc_section->relocations;
+  write_u32_leb128(&ctx->stream, relocs.size(), "num relocs");
+
+  for (size_t i = 0; i < relocs.size(); i++) {
```
KarlSchimpf: Might consider a for range here too.

src/binary-writer.cc
```diff
@@ -656,34 +661,34 @@ static Result write_module(Context* ctx, const Module* module) {
...
       const FuncSignature* sig = &func_type->sig;
       write_header(ctx, "type", i);
       write_type(&ctx->stream, Type::Func);
 
-      uint32_t num_params = sig->param_types.size;
-      uint32_t num_results = sig->result_types.size;
+      uint32_t num_params = sig->param_types.size();
+      uint32_t num_results = sig->result_types.size();
       write_u32_leb128(&ctx->stream, num_params, "num params");
       for (size_t j = 0; j < num_params; ++j)
```
KarlSchimpf: Use for range here?

src/binary-writer.cc
```diff
@@ -656,34 +661,34 @@ static Result write_module(Context* ctx, const Module* module) {
...
 
-  if (module->imports.size) {
+  if (module->imports.size()) {
     begin_known_section(ctx, BinarySection::Import, LEB_SECTION_SIZE_GUESS);
-    write_u32_leb128(&ctx->stream, module->imports.size, "num imports");
+    write_u32_leb128(&ctx->stream, module->imports.size(), "num imports");
 
-    for (size_t i = 0; i < module->imports.size; ++i) {
-      const Import* import = module->imports.data[i];
+    for (size_t i = 0; i < module->imports.size(); ++i) {
```
KarlSchimpf: Same here.

src/binary-writer.cc
```diff
@@ -729,54 +734,52 @@ static Result write_module(Context* ctx, const Module* module) {
...
 
-  if (module->exports.size) {
+  if (module->exports.size()) {
     begin_known_section(ctx, BinarySection::Export, LEB_SECTION_SIZE_GUESS);
-    write_u32_leb128(&ctx->stream, module->exports.size, "num exports");
+    write_u32_leb128(&ctx->stream, module->exports.size(), "num exports");
 
-    for (size_t i = 0; i < module->exports.size; ++i) {
-      const Export* export_ = module->exports.data[i];
+    for (size_t i = 0; i < module->exports.size(); ++i) {
```
KarlSchimpf: Same here.

src/binary-writer.cc
```diff
@@ -928,10 +931,9 @@ static Result write_module(Context* ctx, const Module* module) {
   }
 
   if (ctx->options->relocatable) {
-    for (size_t i = 0; i < ctx->reloc_sections.size; i++) {
-      write_reloc_section(ctx, &ctx->reloc_sections.data[i]);
+    for (size_t i = 0; i < ctx->reloc_sections.size(); i++) {
```
KarlSchimpf: Same here.

src/resolve-names.cc
```diff
@@ -198,9 +202,9 @@ static Result on_br_if_expr(Expr* expr, void* user_data) {
 
 static Result on_br_table_expr(Expr* expr, void* user_data) {
   Context* ctx = static_cast<Context*>(user_data);
-  VarVector* targets = &expr->br_table.targets;
-  for (size_t i = 0; i < targets->size; ++i) {
-    Var* target = &targets->data[i];
+  VarVector& targets = *expr->br_table.targets;
+  for (size_t i = 0; i < targets.size(); ++i) {
```
KarlSchimpf: For range loop?

src/resolve-names.cc
```diff
@@ -301,8 +305,8 @@ static void visit_global(Context* ctx, Global* global) {
 static void visit_elem_segment(Context* ctx, ElemSegment* segment) {
   resolve_table_var(ctx, &segment->table_var);
   visit_expr_list(segment->offset, &ctx->visitor);
-  for (size_t i = 0; i < segment->vars.size; ++i)
-    resolve_func_var(ctx, &segment->vars.data[i]);
+  for (size_t i = 0; i < segment->vars.size(); ++i)
```
KarlSchimpf: Same here.

src/resolve-names.cc
```diff
@@ -318,16 +322,16 @@ static void visit_module(Context* ctx, Module* module) {
...
-    visit_func(ctx, module->funcs.data[i]);
-  for (size_t i = 0; i < module->exports.size; ++i)
-    visit_export(ctx, module->exports.data[i]);
-  for (size_t i = 0; i < module->globals.size; ++i)
-    visit_global(ctx, module->globals.data[i]);
-  for (size_t i = 0; i < module->elem_segments.size; ++i)
-    visit_elem_segment(ctx, module->elem_segments.data[i]);
-  for (size_t i = 0; i < module->data_segments.size; ++i)
-    visit_data_segment(ctx, module->data_segments.data[i]);
+  for (size_t i = 0; i < module->funcs.size(); ++i)
```
KarlSchimpf: Same here.

src/resolve-names.cc
```diff
@@ -318,16 +322,16 @@ static void visit_module(Context* ctx, Module* module) {
...
+    visit_func(ctx, module->funcs[i]);
+  for (size_t i = 0; i < module->exports.size(); ++i)
```
KarlSchimpf: Same here.

src/resolve-names.cc
```diff
@@ -318,16 +322,16 @@ static void visit_module(Context* ctx, Module* module) {
...
+    visit_export(ctx, module->exports[i]);
+  for (size_t i = 0; i < module->globals.size(); ++i)
```
KarlSchimpf: same here. and next two for loops.

src/resolve-names.cc
```diff
@@ -400,25 +402,24 @@ static void visit_command(Context* ctx, Command* command) {
...
-      visit_raw_module(ctx, &command->assert_uninstantiable.module);
+      visit_raw_module(ctx, command->assert_uninstantiable.module);
       break;
   }
 }
 
 static void visit_script(Context* ctx, Script* script) {
-  for (size_t i = 0; i < script->commands.size; ++i)
-    visit_command(ctx, &script->commands.data[i]);
+  for (size_t i = 0; i < script->commands.size(); ++i)
```
KarlSchimpf: Same here.

src/tools/wasm-link.cc
```diff
@@ -199,16 +221,16 @@ static void apply_relocation(Section* section, Reloc* r) {
...
     return;
 
   if (s_debug)
     writef(&s_log_stream, "apply_relocations: %s\n",
            get_section_name(section->section_code));
 
   /* Perform relocations in-place */
-  for (size_t i = 0; i < section->relocations.size; i++) {
-    Reloc* reloc = &section->relocations.data[i];
+  for (size_t i = 0; i < section->relocations.size(); i++) {
```
KarlSchimpf: Use for range loop?

src/tools/wasm-link.cc
```diff
@@ -244,15 +272,15 @@ static void write_slice(Stream* stream, StringSlice str, const char* desc) {
...
+                                const SectionPtrVector& sections) {
   /* Total section size includes the element count leb128 which is
    * always 1 in the current spec */
   uint32_t table_count = 1;
   uint32_t flags = WABT_BINARY_LIMITS_HAS_MAX_FLAG;
   uint32_t elem_count = 0;
 
-  for (size_t i = 0; i < sections->size; i++) {
-    Section* sec = sections->data[i];
+  for (size_t i = 0; i < sections.size(); i++) {
```
KarlSchimpf: same here.

src/tools/wasm-link.cc
```diff
@@ -268,19 +296,19 @@ static void write_table_section(Context* ctx,
 
 static void write_export_section(Context* ctx) {
   uint32_t total_exports = 0;
-  for (size_t i = 0; i < ctx->inputs.size; i++) {
-    LinkerInputBinary* binary = &ctx->inputs.data[i];
-    total_exports += binary->exports.size;
+  for (size_t i = 0; i < ctx->inputs.size(); i++) {
```
KarlSchimpf: Same here.

src/tools/wasm-link.cc
```diff
@@ -268,19 +296,19 @@ static void write_table_section(Context* ctx,
...
 
   Stream* stream = &ctx->stream;
   WRITE_UNKNOWN_SIZE(stream);
   write_u32_leb128(stream, total_exports, "export count");
 
-  for (size_t i = 0; i < ctx->inputs.size; i++) {
-    LinkerInputBinary* binary = &ctx->inputs.data[i];
-    for (size_t j = 0; j < binary->exports.size; j++) {
-      Export* export_ = &binary->exports.data[j];
+  for (size_t i = 0; i < ctx->inputs.size(); i++) {
```
KarlSchimpf: For range loops for nested loops?

src/tools/wasm-link.cc
```diff
@@ -299,13 +327,14 @@ static void write_export_section(Context* ctx) {
...
-static void write_elem_section(Context* ctx, const SectionPtrVector* sections) {
+static void write_elem_section(Context* ctx,
+                               const SectionPtrVector& sections) {
   Stream* stream = &ctx->stream;
   WRITE_UNKNOWN_SIZE(stream);
 
   uint32_t total_elem_count = 0;
-  for (size_t i = 0; i < sections->size; i++) {
-    Section* sec = sections->data[i];
+  for (size_t i = 0; i < sections.size(); i++) {
```
KarlSchimpf: Same here.

src/tools/wasm-link.cc
```diff
@@ -318,8 +347,8 @@ static void write_elem_section(Context* ctx, const SectionPtrVector* sections) {
 
   ctx->current_section_payload_offset = stream->offset;
 
-  for (size_t i = 0; i < sections->size; i++) {
-    Section* sec = sections->data[i];
+  for (size_t i = 0; i < sections.size(); i++) {
```
KarlSchimpf: Same here.

src/tools/wasm-link.cc
```diff
@@ -337,8 +366,8 @@ static void write_memory_section(Context* ctx,
   Limits limits;
   WABT_ZERO_MEMORY(limits);
   limits.has_max = true;
-  for (size_t i = 0; i < sections->size; i++) {
-    Section* sec = sections->data[i];
+  for (size_t i = 0; i < sections.size(); i++) {
```
KarlSchimpf: Same here.

src/tools/wasm-link.cc
```diff
@@ -367,48 +396,48 @@ static void write_global_import(Context* ctx, GlobalImport* import) {
 
 static void write_import_section(Context* ctx) {
   uint32_t num_imports = 0;
-  for (size_t i = 0; i < ctx->inputs.size; i++) {
-    LinkerInputBinary* binary = &ctx->inputs.data[i];
-    FunctionImportVector* imports = &binary->function_imports;
-    for (size_t j = 0; j < imports->size; j++) {
-      FunctionImport* import = &imports->data[j];
+  for (size_t i = 0; i < ctx->inputs.size(); i++) {
```
KarlSchimpf: Same here.

src/tools/wasm-link.cc
```diff
@@ -367,48 +396,48 @@ static void write_global_import(Context* ctx, GlobalImport* import) {
...
 
   WRITE_UNKNOWN_SIZE(&ctx->stream);
   write_u32_leb128(&ctx->stream, num_imports, "num imports");
 
-  for (size_t i = 0; i < ctx->inputs.size; i++) {
-    LinkerInputBinary* binary = &ctx->inputs.data[i];
-    FunctionImportVector* imports = &binary->function_imports;
-    for (size_t j = 0; j < imports->size; j++) {
-      FunctionImport* import = &imports->data[j];
+  for (size_t i = 0; i < ctx->inputs.size(); i++) {
```
KarlSchimpf: Same here?

src/tools/wasm-link.cc
```diff
@@ -367,48 +396,48 @@ static void write_global_import(Context* ctx, GlobalImport* import) {
...
+                                   const SectionPtrVector& sections,
                                    uint32_t total_count) {
   Stream* stream = &ctx->stream;
   WRITE_UNKNOWN_SIZE(stream);
 
   write_u32_leb128(stream, total_count, "function count");
 
-  for (size_t i = 0; i < sections->size; i++) {
-    Section* sec = sections->data[i];
+  for (size_t i = 0; i < sections.size(); i++) {
```
KarlSchimpf: Same here?

src/tools/wasm-link.cc
```diff
@@ -426,28 +455,28 @@ static void write_function_section(Context* ctx,
...
                                uint32_t total_count) {
   Stream* stream = &ctx->stream;
   WRITE_UNKNOWN_SIZE(stream);
 
   write_u32_leb128(stream, total_count, "data segment count");
-  for (size_t i = 0; i < sections->size; i++) {
-    Section* sec = sections->data[i];
-    for (size_t j = 0; j < sec->data_segments.size; j++) {
-      DataSegment* segment = &sec->data_segments.data[j];
+  for (size_t i = 0; i < sections.size(); i++) {
```
KarlSchimpf: Same here.

src/tools/wasm-link.cc
```diff
@@ -459,11 +488,11 @@ static void write_data_section(Context* ctx,
 static void write_names_section(Context* ctx) {
   uint32_t total_count = 0;
   size_t k;
-  for (size_t i = 0; i < ctx->inputs.size; i++) {
-    LinkerInputBinary* binary = &ctx->inputs.data[i];
-    for (size_t j = 0; j < binary->debug_names.size; j++) {
-      if (j < binary->function_imports.size) {
-        if (!binary->function_imports.data[j].active)
+  for (size_t i = 0; i < ctx->inputs.size(); i++) {
```
KarlSchimpf: Same here for outer loop?

src/tools/wasm-link.cc
```diff
@@ -500,13 +529,13 @@ static void write_names_section(Context* ctx) {
...
                                 BinarySection section_code,
-                                SectionPtrVector* sections) {
+                                const SectionPtrVector& sections) {
   uint32_t total_relocs = 0;
 
   /* First pass to know total reloc count */
-  for (size_t i = 0; i < sections->size; i++) {
-    Section* sec = sections->data[i];
-    total_relocs += sec->relocations.size;
+  for (size_t i = 0; i < sections.size(); i++) {
```
KarlSchimpf: Same here.

binji:
Pretty sure I got all the places where it was easy to switch to a range-based for. There are some others that use the index which probably could switch, but I figure it's better to do those later if we want.

PR #367, opened 2017-03-23
Windows parse hexfloat

Cellule:
Use parse_double_hex and parse_float_hex to parse float hexstring on Windows because some crt implementations of strtof and strtod doesn't support hexstrings.

This is related to #362.
After some more thought I decided to approach this issue differently.

First of all, I don't think we should try to keep supporting Visual Studio 2013. There are way too many things that needs to be changed to support it and I don't think it is worth it.

However, in chakra, we are using an older version of the crt that has that strtof/strtod bug. The main issue being not supporting hexstrings. Which this pull request fixes.

There is some precision bugs in the implementation which is not a big deal and hopefully we'll update the crt version soon and this problem will go away.

I hope that this change is not too invasive and can be merged in the mainline. It would avoid the need for a fork of wabt for our purposes

### DIFF
--- a/src/literal.cc
+++ b/src/literal.cc
@@ -283,37 +283,37 @@ static void parse_float_hex(const char* s,
   if (!seen_dot)
     significand_exponent += significand_shift;
 
-  assert(s < end && *s == 'p');
-  s++;
-
   if (significand == 0) {
     /* 0 or -0 */
     *out_bits = make_float(is_neg, F32_MIN_EXP, 0);
     return;
   }
 
-  assert(s < end);
   int exponent = 0;
   bool exponent_is_neg = false;
-  /* exponent is always positive, but significand_exponent is signed.
-   significand_exponent_add is negated if exponent will be negative, so it  can
-   be easily summed to see if the exponent is too large (see below) */
-  int significand_exponent_add = 0;
-  if (*s == '-') {
-    exponent_is_neg = true;
-    significand_exponent_add = -significand_exponent;
+  if (s < end) {
+    assert(*s == 'p');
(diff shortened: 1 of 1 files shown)


src/literal.cc
```diff
@@ -718,6 +725,12 @@ Result parse_double(LiteralType literal_type,
                     uint64_t* out_bits) {
   switch (literal_type) {
     case LiteralType::Int:
+#if COMPILER_IS_MSVC
+      if (string_starts_with(s, end, "0x"))
+      {
+        goto parseDoubleHex;
```
sbc100: using `goto` here seems little extreme.  Could you instead do something like this before the switch?

```
#if COMPILER_IS_MSVC
if (literal_type == LiteralType::Int && string_starts_with(s, end, "0x"))
  literal_type = LiteralType::Hexfloat
#endif
```
Also, this is probably worthy of comment explaining why.

Cellule: I think you are right, plus some compilers don't like unused labels

src/literal.cc
```diff
@@ -384,6 +384,13 @@ Result parse_float(LiteralType literal_type,
                    const char* s,
                    const char* end,
                    uint32_t* out_bits) {
+#if COMPILER_IS_MSVC
+  if (literal_type == LiteralType::Int && string_starts_with(s, end, "0x"))
+  {
```
sbc100: The style for this repo is to not have opening braces on a line by themselves.

(tip: You can use clang-format to format in accordance with the local style since there is a .clang-format file at the top level of the repo)

PR #368, opened 2017-03-24
Support arithmetic/canonical NaN; update testsuite

binji:
Positive NaN (+nan) is defined as being a NaN value where the sign bit
is clear, the exponent is all ones, and the tag has only the "quiet" bit
set. The quiet bit is the most-significant bit of the tag. For example,
for a 32-bit float, +nan is 0x7cf00000.

"Canonical NaN" is either +nan or -nan, where -nan is +nan with the sign
bit set.

"Arithmetic NaN" is defined as any other quiet NaN (i.e. the quiet bit
must be set, but any other bit can be either 0 or 1.)

This change doesn't update the interpreter because it is only a
loosening of the previous NaN behavior.


sunfishcode:
I took a quick look at the implementation here, and it looks correct to me.

PR #373, opened 2017-03-29
wasmlink: Fix function indexes in name section

sbc100:
The code was assuming that all functions were named
and using a simple increment rather than calculating
the actual function index in the new binary.


sbc100:
Gonna split this up

PR #374, opened 2017-03-29
wasmlink: Fix function indexes in name section

sbc100:
The code was assuming that all functions were named
and using a simple increment rather than calculating
the actual function index in the new binary.

### DIFF
--- a/src/tools/wasm-link.cc
+++ b/src/tools/wasm-link.cc
@@ -478,12 +478,12 @@ static void write_data_section(Context* ctx,
 
 static void write_names_section(Context* ctx) {
   uint32_t total_count = 0;
-  size_t k;
-  for (size_t i = 0; i < ctx->inputs.size(); i++) {
-    LinkerInputBinary* binary = ctx->inputs[i].get();
-    for (size_t j = 0; j < binary->debug_names.size(); j++) {
-      if (j < binary->function_imports.size()) {
-        if (!binary->function_imports[j].active)
+  for (const std::unique_ptr<LinkerInputBinary>& binary: ctx->inputs) {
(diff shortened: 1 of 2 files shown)


pipcet:
I'm confused by this PR; we have the following code in binary-reader-linker.cc:

```
    /* We don't currently support merging name sections unless they contain
     * a name for every function. */
    uint32_t total_funcs = binary->function_imports.size();
    for (const std::unique_ptr<Section>& section : binary->sections) {
      if (section->section_code == BinarySection::Function) {
        total_funcs += section->count;
        break;
      }
    }
    if (total_funcs != sec->count) {
      WABT_FATAL("name section count (%d) does not match function count (%d)\n",
                 sec->count, total_funcs);
    }
```

So how can this PR make a difference?

pipcet:
It seems `wasm2wast` actually writes an empty name for the unnamed function:

```
0000068: 046e 616d 6501 3105 000d 2469 6d70 6f72  .name.1...$impor
0000078: 745f 6675 6e63 3101 0d24 696d 706f 7274  t_func1..$import
0000088: 5f66 756e 6332 0206 246e 616d 6531 0306  _func2..$name1..
0000098: 246e 616d 6532 0400 0226 0500 0001 0002  $name2...&......
00000a8: 0100 0724 7061 7261 6d31 0301 0007 2470  ...$param1....$p
00000b8: 6172 616d 3204 0100 0724 7061 7261 6d32  aram2....$param2
```

Which is one bug hiding behind another one, I guess.

pipcet:
Test patch here: https://github.com/WebAssembly/wabt/compare/master...pipcet:fix-empty-function-name?expand=1

That breaks test/link/names.txt, and it's incomplete because it doesn't handle the locals section, so no PR yet.

PR #376, opened 2017-03-29
Use classes + virtual functions for BinaryReader

binji:
This adds a few new classes:

* BinaryReader: the abstract base class
* BinaryReaderNop: implements all of BinaryReader, but does nothing
* LoggingBinaryReader: logs calls through BinaryReader, and forwards to
  another BinaryReader

Typically this means we can remove the Context structs from these
implementations, since that data can just move into the BinaryReader
subclasses.

I also took the opportunity to rename the new member functions to
MixedCase instead of snake_case, since that's more common in C++.

### DIFF
--- a/src/binary-reader-objdump.cc
+++ b/src/binary-reader-objdump.cc
@@ -23,7 +23,7 @@
 
 #include <vector>
 
-#include "binary-reader.h"
+#include "binary-reader-nop.h"
 #include "literal.h"
 
 namespace wabt {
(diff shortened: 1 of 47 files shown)


pipcet:
Would it make sense to make `state` a member of the BinaryReader class rather than passing it by reference to all callbacks? Not a hundred percent clean to have a data member in the abstract base class, but it would save a lot of code...

binji:
@pipcet I had it that way at first, but it makes the forwarding in `LoggingBinaryReader` clunky, since you have to copy the values in before every callback. This seemed less error-prone, and would expand to other things too (like having a ChainBinaryReader or whatever). What do you think?

pipcet:
@binji I see. I must say I'd still prefer not to have that state argument, which appears to be unused 90% of the time. It's like having two `void *` arguments in a C callback...

I must confess I can't think of a perfectly clean way of doing this, my best attempt is:

When parsing starts, the BinaryReader's BeginModule method is called with a `BinaryReader::State* state` argument. The base class simply stores that argument and provides a `getInputOffset()` method to access the offset (that appears to be the only part that is actually used?).

That would be the least code, I think (LoggingReader can pass through the state pointer rather than having to copy data) and shouldn't reduce performance, while still allowing the state class to grow to include more state.

It's a bit C-like in using actual pointers though, so I fully understand if you prefer leaving things as in the PR; either way, I think, this PR is a huge improvement.

binji:
Yeah, that's not bad. I'll try it out and see.

binji:
Thanks for the suggestion, @pipcet. This is much nicer.

src/binary-reader-objdump.cc
```diff
@@ -32,77 +32,247 @@ namespace {
...
 }
 
-static Result begin_section(BinaryReaderContext* ctx,
-                            BinarySection section_code,
-                            uint32_t size) {
-  Context* context = static_cast<Context*>(ctx->user_data);
-  context->section_starts[static_cast<size_t>(section_code)] = ctx->offset;
+Result BinaryReaderObjdumpPrepass::OnFunctionName(uint32_t index,
+                                                  StringSlice name) {
+  if (options->mode == ObjdumpMode::Prepass) {
```
sbc100: Why is this condition needed?  Shouldn't it just be split into two different methods?  On a related note is seems strange for BinaryReaderObjdump to inherit from BinaryReaderObjdumpPrepass.  Could they both inherit from some common base class instead?

src/binary-reader-objdump.cc
```diff
@@ -32,77 +32,247 @@ namespace {
...
+Result BinaryReaderObjdumpPrepass::OnReloc(RelocType type,
+                                           uint32_t offset,
+                                           uint32_t index,
+                                           int32_t addend) {
+  uint32_t total_offset =
+      section_starts[static_cast<size_t>(reloc_section)] + offset;
+  PrintDetails("   - %-18s idx=%#-4x addend=%#-4x offset=%#x(file=%#x)\n",
+               get_reloc_type_name(type), index, addend, offset, total_offset);
+  if (options->mode == ObjdumpMode::Prepass &&
+      reloc_section == BinarySection::Code) {
```
sbc100: Ditto

src/wasm-link.h
```diff
@@ -83,12 +83,12 @@ struct Section {
...
   union {
     /* CUSTOM section data */
-    SectionDataCustom data_custom;
+    SectionDataCustom custom;
     /* DATA section data */
     std::vector<DataSegment>* data_segments;
     /* MEMORY section data */
     Limits memory_limits;
-  };
+  } data;
```
sbc100: Is this change related?

test/interp/basic-logging.txt
```diff
@@ -45,50 +45,28 @@
...
+  EndExportSection
+  BeginCodeSection(7)
+    OnFunctionBodyCount(1)
+    BeginFunctionBody(0)
+    OnLocalDeclCount(0)
+    OnI32ConstExpr(42 (0x2a))
+    OnReturnExpr
+    EndFunctionBody(0)
+  EndCodeSection
+EndModule
```
sbc100: Looks like the looking used to happen twice here?  Was that am existing bug?

src/wasm-link.h
binji: Kinda. I started getting a bug because this union wasn't initialized to zero. It seemed easiest to name it so I could do `WABT_ZERO_MEMORY(data)`

test/interp/basic-logging.txt
binji: Oh weird, I didn't even notice. Yeah, must have been an existing bug.

src/binary-reader-objdump.cc
binji: Done.

test/interp/basic-logging.txt
binji: Yeah, I forgot. the BinaryReaderInterpreter did two passes over the binary before since you need to first validate all the modifications to the data and elem segments, then only apply them if all are valid. This turned out to be a pain for the new system, so I ended up using two vectors to store the modifications and updating them later.

PR #377, opened 2017-03-29
Cleanup default test tools declartion in run-tests.py

sbc100:
There was a bunch of redundant splitting and joining
going on.


binji:
Thanks, lgtm

PR #379, opened 2017-03-30
Add -p option to run-tests.py to show the commands it runs

sbc100:
Without this it can be hard to dig down that find out
exactly what is being run

### DIFF
--- a/test/run-tests.py
+++ b/test/run-tests.py
@@ -704,8 +704,6 @@ def main(args):
       parser.error('--fail-fast only works with -j1')
     if options.stop_interactive:
       parser.error('--stop-interactive only works with -j1')
-    if options.print_cmd:
-      parser.error('--print-cmd only works with -j1')
 
   if options.patterns:
     pattern_re = '|'.join(
--- a/test/run-tests.py
+++ b/test/run-tests.py
@@ -540,13 +540,16 @@ def RunTest(info, options, variables, verbose_level=0):
   # Clone variables dict so it can be safely modified.
   variables = dict(variables)
 
+  cwd = REPO_ROOT_DIR
+  gen_input_path = info.CreateInputFile()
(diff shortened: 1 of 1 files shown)


test/run-tests.py
```diff
@@ -711,6 +717,8 @@ def main(args):
       parser.error('--fail-fast only works with -j1')
     if options.stop_interactive:
       parser.error('--stop-interactive only works with -j1')
+    if options.print_cmd:
```
binji: Why? Seems like it should work multi-threaded too.

test/run-tests.py
```diff
@@ -702,6 +705,9 @@ def main(args):
   parser.add_argument('--no-roundtrip',
                       help='don\'t run roundtrip.py on all tests',
                       action='store_false', default=True, dest='roundtrip')
+  parser.add_argument('-p', '--print-cmd',
```
binji: Maybe this should forward through to the subcommands instead of using the verbose flag for that.

test/run-tests.py
sbc100: It just one of those commands that doesn't really make sense in the multithreaded run since that output will be interleaved.   But i guess that could be considered a personal preference?

test/run-tests.py
sbc100: I didn't consider that but it would have made this change more invasive, since not all sub-commands support -p I think.

test/run-tests.py
binji: I guess it depends on how ugly the interleaving is.

test/run-tests.py
binji: Yeah, I guess so. Still, seems like it would be nicer to unify if possible.

sbc100:
I removed the limitation on -j1, and I don't pass it through to the child because that would cause the expected test output to fail

binji:
sgtm

PR #381, opened 2017-03-30
Allow names section to be sparse

sbc100:
Only emit names for function that actually have one.
Previously we were emitting empty strings for functions
without names.  The new version of the names section
allow for a sparse list.

Also, cleanup the funtion name handling in wasm-link.

### DIFF
--- a/src/binary-reader-linker.cc
+++ b/src/binary-reader-linker.cc
@@ -70,6 +70,8 @@ class BinaryReaderLinker : public BinaryReaderNop {
                                    const void* data,
                                    uint32_t size);
 
+  virtual Result BeginNamesSection(uint32_t size);
+
   virtual Result OnFunctionName(uint32_t function_index,
                                 StringSlice function_name);
 
@@ -88,6 +90,7 @@ class BinaryReaderLinker : public BinaryReaderNop {
 
   Section* reloc_section = nullptr;
   Section* current_section = nullptr;
+  uint32_t function_count = 0;
 };
 
 BinaryReaderLinker::BinaryReaderLinker(LinkerInputBinary* binary)
@@ -164,7 +167,7 @@ Result BinaryReaderLinker::OnImportGlobal(uint32_t import_index,
 }
 
 Result BinaryReaderLinker::OnFunctionCount(uint32_t count) {
-  binary->debug_names.resize(count + binary->function_imports.size());
+  function_count = count;
   return Result::Ok;
 }
 
@@ -259,6 +262,11 @@ Result BinaryReaderLinker::OnExport(uint32_t index,
   export_->name = name;
   export_->kind = kind;
   export_->index = item_index;
+  return Result::Ok;
+}
+
+Result BinaryReaderLinker::BeginNamesSection(uint32_t size) {
+  binary->debug_names.resize(function_count + binary->function_imports.size());
   return Result::Ok;
 }
 
--- a/src/binary-reader-linker.cc
+++ b/src/binary-reader-linker.cc
@@ -263,7 +263,6 @@ Result BinaryReaderLinker::OnExport(uint32_t index,
 }
 
 Result BinaryReaderLinker::OnFunctionName(uint32_t index, StringSlice name) {
-  assert(index < binary->debug_names.size());
   binary->debug_names[index] = string_slice_to_string(name);
   return Result::Ok;
 }
--- a/src/binary-reader-linker.cc
+++ b/src/binary-reader-linker.cc
@@ -34,8 +34,6 @@ class BinaryReaderLinker : public BinaryReaderNop {
 
   virtual Result BeginSection(BinarySection section_type, uint32_t size);
 
-  virtual Result BeginCustomSection(uint32_t size, StringSlice section_name);
-
   virtual Result OnImport(uint32_t index,
                           StringSlice module_name,
                           StringSlice field_name);
@@ -50,6 +48,8 @@ class BinaryReaderLinker : public BinaryReaderNop {
                                 uint32_t global_index,
                                 Type type,
                                 bool mutable_);
+
+  virtual Result OnFunctionCount(uint32_t count);
 
   virtual Result OnTable(uint32_t index,
                          Type elem_type,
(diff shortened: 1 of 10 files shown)


src/binary-writer.cc
```diff
@@ -880,17 +878,29 @@ static Result write_module(Context* ctx, const Module* module) {
...
-    write_u32_leb128(&ctx->stream, module->funcs.size(), "num functions");
-    for (size_t i = 0; i < module->funcs.size(); ++i) {
-      const Func* func = module->funcs[i];
-      write_u32_leb128(&ctx->stream, i, "function index");
-      wabt_snprintf(desc, sizeof(desc), "func name %" PRIzd, i);
-      write_str(&ctx->stream, func->name.start, func->name.length,
-                PrintChars::Yes, desc);
+
+    size_t named_functions = 0;
+    for (const auto func: module->funcs) {
```
binji: I've been avoiding using `auto` except for iterators and such. I think it makes it a bit clearer. Also, I believe the Chromium style guide suggests not using auto for pointers, though I can't quite remember why.

sbc100: Done

pipcet:
https://github.com/WebAssembly/wabt/pull/374 for context.

I don't think this change is correct as it stands, I'll comment in the code. In addition, it's not clear to me how we should handle empty names—the spec appears to allow them, which would require us using `std::optional<>` or rolling our own.

The other limitation is this patch handles sparse `"name".function` subsections, but not sparse `"name".locals` subsections.

src/binary-reader-linker.cc
```diff
@@ -163,6 +163,11 @@ Result BinaryReaderLinker::OnImportGlobal(uint32_t import_index,
   return Result::Ok;
 }
 
+Result BinaryReaderLinker::OnFunctionCount(uint32_t count) {
+  binary->debug_names.resize(count + binary->function_imports.size());
```
pipcet: The "count" is the count of function indices+names to follow, not the maximum function index. In addition, function imports aren't counted separately.

IOW, it's legal for count = 1 to be followed by index = 1000 and a name, which would blow up below.

src/binary-reader-linker.cc
```diff
@@ -314,12 +263,8 @@ Result BinaryReaderLinker::OnExport(uint32_t index,
 }
 
 Result BinaryReaderLinker::OnFunctionName(uint32_t index, StringSlice name) {
-  while (binary->debug_names.size() < index) {
-    binary->debug_names.emplace_back();
-  }
-  if (binary->debug_names.size() == index) {
-    binary->debug_names.push_back(string_slice_to_string(name));
-  }
+  assert(index < binary->debug_names.size());
```
pipcet: This assert would, I think, trigger in the index=1000 case described above.

src/binary-reader-linker.cc
sbc100: What I'm trying to do here is resize to the size of the function index space.

My understanding is that function index space goes from 0 to (local function count + imported function count).   This callback OnFunctionCount, gives us the number of local functions and function_imports records to number of imported functions.

src/binary-reader-linker.cc
sbc100: You are correct, we probably want to just ignore names for functions that are outside of the function index space.  Although no sane tools should generate such names.  I will fix.

sbc100: Actually it looks like the current doc says index should be within the function index space: https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#function-names.

So I think its OK to error out here, but perhaps I should change to and error rather than an assert?

src/binary-reader-linker.cc
pipcet: Oops! You're perfectly right, I was misreading OnFunctionCount as OnFunctionNamesCount. My apologies.

sbc100:
I removed to assert and added a test to make sure the binary-reader verified the function indexes.

pipcet:
Now I've found my thinko, this looks good to me, but the related code in wasm-link.cc, which now becomes relevant because of the dropped constraint, looks questionable:
```
  for (const std::unique_ptr<LinkerInputBinary>& binary: ctx->inputs) {
    for (size_t i = 0; i < binary->debug_names.size(); i++) {
      if (binary->debug_names[i].empty())
        continue;
      if (i < binary->function_imports.size()) {
        if (!binary->function_imports[i].active)
          continue;
      }
      write_u32_leb128(stream, i + binary->function_index_offset, "function index");
      write_string(stream, binary->debug_names[i], "function name");
    }
  }
```

Does that do the right thing for imports which are not active? My assumption and the previous code agreed that inactive imports wouldn't be emitted, so we should be writing `i + binary->function_index_offset - delta`, `delta` being a per-`binary` count of inactive imports discovered so far.

src/binary-reader-linker.cc
pipcet: I was mistaken in my initial comment. Error rather than assert sounds good to me though!

pipcet:
Hmm. I'm adding some tests (just trivial variations on test/link/names.txt), and seeing odd behavior and segfaults (for example, a module without a function section but with imports causes a segfault as OnFunctionCount is never called).

sbc100:
Good catch.  I didn't think of that.  Will add fix and add another test case.  Thanks for the thorough review!

pipcet:
https://github.com/WebAssembly/wabt/compare/master...pipcet:name-tests?expand=1 for two of the tests.

sbc100:
Ah, your names2.txt does seems to expose a bug in wasm-link.   However, it not caused by this change, and in fact this change exposes it as an error rather then a SEGFAULT.     So I think we should address that separately.  It looks like its generating a negative function index in the names section is clearly wrong.

pipcet:
Thanks for investigating, and fixing the segfault! Yes, the issue is in wasm-link.cc; there's what's maybe a fix at 
https://github.com/pipcet/wabt/tree/fix-names.

Addressing it separately makes sense to me, it's in wasm-link.cc which this PR doesn't touch.

(There's one test failure caused by the new error message.)

PR #383, opened 2017-03-31
Return non-zero on travis if tests fail

sbc100:
Oops, we were silently ignoring tests failures on travis.

Also turn on -v when running travis tests so it records the tests it runs

### DIFF
--- a/scripts/travis-test.sh
+++ b/scripts/travis-test.sh
@@ -22,14 +22,18 @@ RTN=0
 SCRIPT_DIR="$(cd "$(dirname "$0")"; pwd -P)"
 source "${SCRIPT_DIR}/travis-common.sh"
 
+cd ${ROOT_DIR}
+
 log_and_run() {
   echo $*
-  (exec $*) && true
-  [[ $? != 0 ]] && RTN=1
+  if ! $*; then
+    echo "travis-test.sh: sub-command failed: $*"
+    RTN=1;
+  fi
 }
 
 run_tests() {
-  (cd ${ROOT_DIR} && log_and_run test/run-tests.py -v --bindir ${BINDIR} $* --timeout=10)
+  log_and_run test/run-tests.py -v --bindir ${BINDIR} $* --timeout=10
   log_and_run ${BINDIR}/wabt-unittests
 }
 
@@ -55,8 +59,10 @@ for BUILD_TYPE in ${BUILD_TYPES_UPPER}; do
   else
     if set_run_test_args ${COMPILER} ${BUILD_TYPE}; then
       run_tests
+      break
     fi
   fi
 done
 
+echo "travis-test.sh done: $RTN"
 exit $RTN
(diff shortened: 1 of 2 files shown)


scripts/travis-test.sh
```diff
@@ -18,17 +18,19 @@
...
 set -o errexit
 
+RTN=0
 SCRIPT_DIR="$(cd "$(dirname "$0")"; pwd -P)"
 source "${SCRIPT_DIR}/travis-common.sh"
 
 log_and_run() {
   echo $*
-  exec $*
+  (exec $*) && true
```
pipcet: Probably a silly question, but what is the `&& true` supposed to do here? I'm no Travis expert, but in a Bourne shell that's just a nop, right?

binji: My understanding is that this runs the command in a subshell and captures the exit code, even though this script is running in error exit mode. That way we can keep running tests even if one of them fails.

pipcet: Thanks! I think the whole line is equivalent to just `$*`, but if it works with Travis, it's probably best not to fiddle with it :-)

The new code looks correct to me and appears to fix the bug, so this was mere curiosity.

test/binary/relocs.txt
```diff
@@ -17,4 +17,5 @@ section("reloc.name") {
 (;; STDOUT ;;;
 (module
   (type (;0;) (func (result i32))))
+XXXXX
```
pipcet: Er, we don't want to merge this bit, right?

sbc100: Exactly. I'm just using this to verify travis will fail on a failing test (which it doesn't right  now!)

pipcet: Thanks again. Good catch, by the way!

scripts/travis-test.sh
```diff
@@ -53,6 +59,10 @@ for BUILD_TYPE in ${BUILD_TYPES_UPPER}; do
   else
     if set_run_test_args ${COMPILER} ${BUILD_TYPE}; then
       run_tests
+      break
```
binji: why break?

sbc100: oops, reverted

scripts/travis-test.sh
```diff
@@ -18,17 +18,23 @@
...
+  if ! $*; then
+    echo "travis-test.sh: sub-command failed: $*"
+    RTN=1;
+  fi
 }
 
 run_tests() {
-  (cd ${ROOT_DIR} && log_and_run test/run-tests.py --bindir ${BINDIR} $* --timeout=10) && true
-  (log_and_run ${BINDIR}/wabt-unittests) && true
+  log_and_run test/run-tests.py -v --bindir ${BINDIR} $* --timeout=10
```
binji: Can you mention that you turned on verbosity in the commit message too?

PR #384, opened 2017-03-31
Fix function name indices for imported functions

pipcet:
When linking, a source binary's imported function indices and its implemented function indices can be non-contiguous; the previous code didn't catch that.

### DIFF
--- a/src/tools/wasm-link.cc
+++ b/src/tools/wasm-link.cc
@@ -504,14 +504,22 @@ static void write_names_section(Context* ctx) {
   WRITE_UNKNOWN_SIZE(stream);
   write_u32_leb128(stream, total_count, "element count");
   for (const std::unique_ptr<LinkerInputBinary>& binary: ctx->inputs) {
+    size_t delta = 0;
+    uint32_t off;
     for (size_t i = 0; i < binary->debug_names.size(); i++) {
       if (binary->debug_names[i].empty())
         continue;
       if (i < binary->function_imports.size()) {
-        if (!binary->function_imports[i].active)
+        if (!binary->function_imports[i].active) {
+          delta++;
           continue;
+        }
+        off = binary->imported_function_index_offset - delta;
+      } else {
+        off = binary->function_index_offset;
       }
-      write_u32_leb128(stream, i + binary->function_index_offset, "function index");
+
+      write_u32_leb128(stream, i + off, "function index");
       write_string(stream, binary->debug_names[i], "function name");
     }
   }
--- a/test/link/names2.txt
+++ b/test/link/names2.txt
@@ -1,0 +1,77 @@
+;;; TOOL: run-wasm-link
+;;; FLAGS: --debug-names -r
+(module
+  (import "__extern" "missing" (func $import_func2))
+  (import "__extern" "baz" (func $import_func1))
+  (export "foo" (func $name1))
+  (func $name1 (param $param1 i32)
+     i32.const 1
+     call 0)
+  (func $name2 (param $param2 i64)
+     i64.const 1
+     call 1)
+  (func (param $param2 i64))
+)
+(module
+  (export "baz" (func 0))
+  (func $name3 (param $param3 i32)
+     i32.const 2
+     call 0)
+)
+(;; STDOUT ;;;
+linked.wasm:	file format wasm 0x000001
+Sections:
+     Type start=0x0000000a end=0x0000001a (size=0x00000010) count: 4
+   Import start=0x00000020 end=0x00000034 (size=0x00000014) count: 1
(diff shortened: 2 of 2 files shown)


src/tools/wasm-link.cc
```diff
@@ -504,14 +504,22 @@ static void write_names_section(Context* ctx) {
...
+        if (!binary->function_imports[i].active) {
+          delta++;
           continue;
+        }
+        off = binary->imported_function_index_offset - delta;
+      } else {
+        off = binary->function_index_offset;
       }
-      write_u32_leb128(stream, i + binary->function_index_offset, "function index");
+
```
sbc100: Maybe we can use relocate_func_index here, which I think is designed to do just this.

test/link/names2.txt
```diff
@@ -0,0 +1,77 @@
+;;; TOOL: run-wasm-link
+;;; FLAGS: --debug-names -r
+(module
+  (import "__extern" "missing" (func $import_func2))
+  (import "__extern" "baz" (func $import_func1))
```
sbc100: Is there an modification to the existing names.txt test that can trigger this bug?  Or a smaller new test case?

src/tools/wasm-link.cc
pipcet: I don't think that deals with inactive imports? I must confess I don't fully understand the code, so it's probably best to use `relocate_func_index` and fix that (if necessary).

sbc100: Hmm, I think are right.  Imported function indexes can't simply be offset by constant.

I guess we should fix that.    If you want to land this in mean time I'd be ok with that (although perhaps we could update the exiting names.txt test rather than adding a new one?)

test/link/names2.txt
pipcet: Sure, updated. Unfortunately, using `relocate_func_index` results in deviating test output.

test/link/names.txt
```diff
@@ -39,45 +41,48 @@ Type:
...
- 000078: 0b                         | end
+00007e <$name1>:
+ 000080: 41 01                      | i32.const 0x1
+ 000082: 10 86 80 80 80 00          | call 0x6
+           000083: R_FUNC_INDEX_LEB   2
+ 000088: 0b                         | end
+000089 <$name2>:
+ 00008b: 42 01                      | i64.const 1
+ 00008d: 10 83 80 80 80 00          | call 0x3
+           00008e: R_FUNC_INDEX_LEB   3
```
pipcet: I think this line is wrong, and should be 0x2 (which it is with the possibly fixed `relocate_func_index` I'm trying), but I'd appreciate a second opinion. $name2 calls index 3 in the original binary, which is $import_func_2, which is func[2] in the linked binary. Right?

pipcet:
@sbc100 I think I fixed `relocate_func_index` (by caching, to avoid O(n^2) behavior--I still hope to run this on glibc soon :-) ). However, it changes the test output again, and I'd really like to make sure that the new output is correct and the old one wasn't. If you could have a look, that'd be great.

src/tools/wasm-link.cc
```diff
@@ -695,6 +704,16 @@ static void calculate_reloc_offsets(Context* ctx) {
     binary->imported_function_index_offset = total_function_imports;
     binary->imported_global_index_offset = total_global_imports;
     binary->memory_page_offset = memory_page_offset;
+
+    size_t delta = 0;
+    for (size_t i = 0; i < binary->function_imports.size(); i++) {
+      if (!binary->function_imports[i].active) {
+        delta++;
```
sbc100: else here since we don't want to store relocated_function_index for inactive imports?

sbc100:
I guess that fact that only the names test has a delta means that we could use some better test coverage elsewhere because this bug effects more than just names I think.

src/tools/wasm-link.cc
```diff
@@ -169,14 +169,20 @@ static uint32_t relocate_func_index(LinkerInputBinary* binary,
...
     FunctionImport* import = &binary->function_imports[function_index];
-    offset = binary->imported_function_index_offset;
+    uint32_t new_index = import->relocated_function_index;
     if (!import->active) {
       function_index = import->foreign_index;
       offset = import->foreign_binary->function_index_offset;
       if (s_debug)
         writef(&s_log_stream,
                "reloc for disabled import. new index = %d + %d\n",
                function_index, offset);
```
pipcet: I thought so at first, but actually we hit this code if we resolve a relocation internally. If a.wasm defines a.f and b.wasm imports a.f and calls it, the call relocation will end up in this code.

pipcet:
Cool. I can't merge, so feel free to when you're happy with the code, or let me know if anything further needs changing.

PR #385, opened 2017-04-01
Run tests on AppVeyor

binji:


binji:
This almost works, the only thing that's failing is conversion from converting 0x8000000000000401 to double on Win64. Seems that v8 ran into the same issue: https://codereview.chromium.org/1738623003.

binji:
OK, this seems to be working, PTAL

PR #388, opened 2017-04-04
[wasmlink] Relocate indexes in the gererated reloc section

sbc100:
We recently switched to embedding the index of the relocation
within the relocation entry itself but overlooked the fact
that this needs to be updated when writing out the linked
reloc section.

### DIFF
--- a/src/tools/wasm-link.cc
+++ b/src/tools/wasm-link.cc
@@ -536,10 +536,8 @@ static void write_reloc_section(Context* ctx,
   uint32_t total_relocs = 0;
 
   /* First pass to know total reloc count */
-  for (size_t i = 0; i < sections.size(); i++) {
-    Section* sec = sections[i];
+  for (Section* sec: sections)
     total_relocs += sec->relocations.size();
-  }
 
   if (!total_relocs)
     return;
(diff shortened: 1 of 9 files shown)


pipcet:
It looks like the old code used the unrelocated indices? That is certainly wrong. I've looked over the test output for a while, and the new output seems correct to me.

Is there anywhere else we're using relocated function indices that we might have missed?

sbc100:
There may be other places, yes.  I suggest we look into doing more execution tests on the resulting binaries, and fixing issues as we find time.  As an incremental improvement I think this change stands alone quite nicely though.

src/tools/wasm-link.cc
```diff
@@ -555,14 +553,20 @@ static void write_reloc_section(Context* ctx,
...
+      uint32_t new_offset = reloc.offset + sec->output_payload_offset;
       write_u32_leb128(&ctx->stream, new_offset, "reloc offset");
-      write_u32_leb128(&ctx->stream, relocs[j].index, "reloc index");
+      uint32_t index = reloc.index;
+      switch (reloc.type) {
+        case RelocType::FuncIndexLEB:
+          index = relocate_func_index(sec->binary, index);
+          break;
+        default:
+          break;
```
pipcet: Should this switch statement be expanded to cover, say, table indices, and possibly global indices? I think I'd prefer a fatal error as the `default`, and an explicit list of reloc types that don't need relocating before the `break`.

src/tools/wasm-link.cc
```diff
@@ -555,14 +553,20 @@ static void write_reloc_section(Context* ctx,
...
+  for (Section* sec: sections) {
+    for (const Reloc& reloc: sec->relocations) {
+      write_u32_leb128_enum(&ctx->stream, reloc.type, "reloc type");
+      uint32_t new_offset = reloc.offset + sec->output_payload_offset;
       write_u32_leb128(&ctx->stream, new_offset, "reloc offset");
-      write_u32_leb128(&ctx->stream, relocs[j].index, "reloc index");
+      uint32_t index = reloc.index;
+      switch (reloc.type) {
+        case RelocType::FuncIndexLEB:
+          index = relocate_func_index(sec->binary, index);
```
pipcet: I'd prefer `new_index = relocate_func_index(sec->binary, reloc.index)`, so we avoid the same variable referring to two different index spaces.

pipcet:
I agree! I'm not any kind of official reviewer, obviously, but I've taken the time to look at the debug output in detail (and added some extra debugging code at https://github.com/WebAssembly/wabt/compare/master...pipcet:extra-debugging?expand=1 which might or might not be useful for doing that), and it all looks good to me, as does the code except for the (very) minor nits.

src/tools/wasm-link.cc
sbc100: Good idea.  Done.

src/tools/wasm-link.cc
sbc100: Done.

sbc100:
I totally appreciate your feedback/reviews!   I added support for another reloc type and a fatal error on any other time (types we don't currently support/test).  Thanks for taking the time to look at this stuff in detail.

PR #391, opened 2017-04-04
Use 8 hex digits for wasm version rather than 6

pipcet:
Very definitely a cosmetic issue, but I think 0x00000001 or 0x1 are both preferrable to 0x000001.


binji:
I'm fine with it either way. @sbc100, opinion?

sbc100:
Not a big deal I guess, but perhaps just 0x1 is better.

binji:
Agreed, do you mind switching it to just 0x1, @pipcet?

pipcet:
Of course not, I was just being lazy and waiting for #395 to land first :-)

PR #392, opened 2017-04-05
Update testsuite to spec f1b89dfaf379060c7e35eb90b7daeb14d4ade3f7

binji:


binji:
Fails on Windows x64, seems to be the truncation+nearest tests w/ `assert_return_arithmetic_nan`:
```
(assert_return_arithmetic_nan (invoke "trunc" (f32.const -nan:0x200000)))
(assert_return_arithmetic_nan (invoke "trunc" (f32.const nan:0x200000)))
(assert_return_arithmetic_nan (invoke "nearest" (f32.const -nan:0x200000)))
(assert_return_arithmetic_nan (invoke "nearest" (f32.const nan:0x200000)))
(assert_return_arithmetic_nan (invoke "trunc" (f64.const -nan:0x4000000000000)))
(assert_return_arithmetic_nan (invoke "trunc" (f64.const nan:0x4000000000000)))
(assert_return_arithmetic_nan (invoke "nearest" (f64.const -nan:0x4000000000000)))
(assert_return_arithmetic_nan (invoke "nearest" (f64.const nan:0x4000000000000)))
```

binji:
Fixed; seems that win64 truncf/nearbyintf don't set the quiet bit.

src/interpreter.cc
```diff
@@ -668,11 +668,15 @@ DEFINE_BITCAST(bitcast_u64_to_f64, uint64_t, double)
...
-#define UNOP_FLOAT(type, func)                                 \
-  do {                                                         \
-    FLOAT_TYPE_##type value = BITCAST_TO_##type(POP_##type()); \
-    PUSH_##type(BITCAST_FROM_##type(func(value)));             \
-    break;                                                     \
+#define UNOP_FLOAT(type, func)                                   \
+  do {                                                           \
+    FLOAT_TYPE_##type value = BITCAST_TO_##type(POP_##type());   \
+    VALUE_TYPE_##type result = BITCAST_FROM_##type(func(value)); \
+    if (WABT_UNLIKELY(IS_NAN_##type(result))) {                  \
```
sbc100: If this is a win64 only thing is it worth an ifdef?  or a comment?

binji: There are a few other places where we explicitly set the quiet nan bit; since it's implementation defined what happens with NaNs AIUI, it's probably better to be explicit.

PR #393, opened 2017-04-05
[wasmdump] Don't print opcodes unless we are disassembling

sbc100:


binji:
lgtm, tests? :-)

binji:
Removing in favor of #395?

PR #395, opened 2017-04-05
Refactor wasmdump so that the disassembly pass has its own reader class

sbc100:
Its almost a pure refactor of the code in binary-reader-objdump.cc.  The one exception is the
change in the way the header printing is done.

### DIFF
--- a/src/binary-reader-objdump.cc
+++ b/src/binary-reader-objdump.cc
@@ -60,8 +60,7 @@ class BinaryReaderObjdumpBase : public BinaryReaderNop {
   bool print_details = false;
   BinarySection reloc_section = BinarySection::Invalid;
   uint32_t section_starts[kBinarySectionCount];
-  int section_found = false;
-  bool header_printed = false;
+  bool section_found = false;
   Stream* out_stream = nullptr;
 };
 
@@ -179,7 +178,6 @@ Result BinaryReaderObjdumpBase::BeginModule(uint32_t version) {
       }
 
       printf("%s:\tfile format wasm %#08x\n", basename, version);
-      header_printed = true;
       break;
     }
     case ObjdumpMode::RawData:
@@ -589,11 +587,9 @@ Result BinaryReaderObjdump::OnCount(uint32_t count) {
 }
 
 Result BinaryReaderObjdump::EndModule() {
-  if (options->section_name) {
-    if (!section_found) {
-      printf("Section not found: %s\n", options->section_name);
-      return Result::Error;
-    }
+  if (options->section_name && !section_found) {
+    printf("Section not found: %s\n", options->section_name);
+    return Result::Error;
   }
 
   return Result::Ok;
--- a/src/binary-reader-objdump.cc
+++ b/src/binary-reader-objdump.cc
@@ -231,9 +231,9 @@ Result BinaryReaderObjdumpPrepass::OnReloc(RelocType type,
   return Result::Ok;
 }
 
-class BinaryReaderObjdumpDisasseble : public BinaryReaderObjdumpBase {
+class BinaryReaderObjdumpDisassemble : public BinaryReaderObjdumpBase {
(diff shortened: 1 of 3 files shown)


src/binary-reader-objdump.cc
```diff
@@ -57,12 +60,18 @@ class BinaryReaderObjdumpBase : public BinaryReaderNop {
   bool print_details = false;
   BinarySection reloc_section = BinarySection::Invalid;
   uint32_t section_starts[kBinarySectionCount];
+  int section_found = false;
+  bool header_printed = false;
```
pipcet: Not strictly related to this change, but is `header_printed` actually used anywhere? It was a bit confusing to have both `header_printed` and `options->print_header`, so maybe we can get rid of both?

src/binary-reader-objdump.cc
```diff
@@ -57,12 +60,18 @@ class BinaryReaderObjdumpBase : public BinaryReaderNop {
   bool print_details = false;
   BinarySection reloc_section = BinarySection::Invalid;
   uint32_t section_starts[kBinarySectionCount];
+  int section_found = false;
```
pipcet: (also not related to this change): `bool`

src/binary-reader-objdump.cc
sbc100: Done

src/binary-reader-objdump.cc
sbc100: Done

src/binary-reader-objdump.cc
```diff
@@ -57,12 +60,17 @@ class BinaryReaderObjdumpBase : public BinaryReaderNop {
...
 };
 
 BinaryReaderObjdumpBase::BinaryReaderObjdumpBase(const uint8_t* data,
                                                  size_t size,
                                                  ObjdumpOptions* options)
-    : options(options), data(data), size(size) {
+    : options(options),
+      data(data),
+      size(size),
+      out_stream(init_stdout_stream()) {
```
pipcet: Probably not a major issue, but I'm a bit worried about calling init_stdout_stream() twice. I think it would be preferrable to pass in a `Stream *` to the BinaryReaderObjdumpDisassemble constructor.

src/binary-reader-objdump.cc
```diff
@@ -579,6 +465,136 @@ Result BinaryReaderObjdump::OnOpcodeBlockSig(uint32_t num_types,
...
+  virtual Result OnInitExprF64ConstExpr(uint32_t index, uint64_t value);
+  virtual Result OnInitExprGetGlobalExpr(uint32_t index, uint32_t global_index);
+  virtual Result OnInitExprI32ConstExpr(uint32_t index, uint32_t value);
+  virtual Result OnInitExprI64ConstExpr(uint32_t index, uint64_t value);
+
+ private:
+  Result OnCount(uint32_t count);
+  void LogOpcode(const uint8_t* data, size_t data_size, const char* fmt, ...);
+
+  Opcode current_opcode = Opcode::Unreachable;
```
binji: These members all seem to have moved elsewhere.

src/binary-reader-objdump.cc
sbc100: I moved this member to the derived class.  The class is still constructed more than once but that was already the case in the existing code.

sbc100: Both the old code and the new code call init_stdout_stream() in the constructor of BinaryReaderObjdump.  In the old code the BinaryReaderObjdump is constructed up to 3 times (once for each of non-prepass phases).    In the new code it is constructed up to 2 times, since the BinaryReaderObjdumpDisassabler is not separated out.

pipcet: Thanks for explaining! You're quite right, we can fix that later if it ever turns out to be an issue.

PR #397, opened 2017-04-08
Opcodes i64.clz, i64.ctz and i64.popcnt only have one argument

Wallbraker:


binji:
Good catch, thanks!

pipcet:
Is the code correct for `f32.nearest` and similar floating-point opcodes with a single argument? Right now it's:
```
  V(F32, F32, F32, 0, 0x8b, F32Abs, "f32.abs")                        \
  V(F32, F32, F32, 0, 0x8c, F32Neg, "f32.neg")                        \
  V(F32, F32, F32, 0, 0x8d, F32Ceil, "f32.ceil")                      \
  V(F32, F32, F32, 0, 0x8e, F32Floor, "f32.floor")                    \
  V(F32, F32, F32, 0, 0x8f, F32Trunc, "f32.trunc")                    \
  V(F32, F32, F32, 0, 0x90, F32Nearest, "f32.nearest")                \
  V(F32, F32, F32, 0, 0x91, F32Sqrt, "f32.sqrt")                      \
```

binji:
Yeah, it looks like those are incorrect too. It isn't causing any bugs currently because any code that uses this macro already knows to ignore that argument. But it's good to fix.

PR #400, opened 2017-04-15
Define the implicit func type before the function

binji:
When a function is defined without an explicit type, it will first
search the already defined (implicitly or otherwise) func types. If one
is found, it will use it. Otherwise it will define its own and use that.

The code previously would define this new type after the function, which
works but doesn't roundtrip properly because it ends up defining the
same func type twice.


binji:
Looks like the only failures here are from the appveyor build which are fixed on master, so landing.

PR #401, opened 2017-04-17
Fix uninitialized value in binary-writer-spec

binji:


binji:
Hm, I'm concerned about the failures on AppVeyor, but I'm guessing those were already there. I think this is a good change for now and I'll investigate the win64 failures stuff in the next PR.

PR #404, opened 2017-04-24
Make ASTWriter, BinaryWriter, BinaryWriterSpec C++ classes

binji:


src/ast-writer.cc
```diff
@@ -58,302 +58,338 @@ enum class NextChar {
...
+  assert(indent_ >= 0);
 }
 
-static void write_indent(Context* ctx) {
+void ASTWriter::WriteIndent() {
   static char s_indent[] =
       "                                                                       "
       "                                                                       ";
   static size_t s_indent_len = sizeof(s_indent) - 1;
-  size_t indent = ctx->indent;
```
KarlSchimpf: Why did you move from a local variable to the field? Doesn't this change the effect (you change the indent of the writer)?

binji: Hm, good point. Wonder why this wasn't breaking tests?

PR #406, opened 2017-04-25
Improve wasmdump output for relocations and data segments

sbc100:
For data segments, print the file offsets so they match
the file offsets shown when dumping relocations.

For relocations, only show the addend when one is present
and correctly display negative addends in the same way that
objdump does (e.g. symbol_foo-0x10 and symbol_foo+0x10)

### DIFF
--- a/src/binary-reader-objdump.cc
+++ b/src/binary-reader-objdump.cc
@@ -126,7 +126,7 @@ class BinaryReaderObjdumpPrepass : public BinaryReaderObjdumpBase {
   virtual Result OnReloc(RelocType type,
                          uint32_t offset,
                          uint32_t index,
-                         int32_t addend);
+                         uint32_t addend);
 };
 
 BinaryReaderObjdumpPrepass::BinaryReaderObjdumpPrepass(const uint8_t* data,
@@ -144,7 +144,7 @@ Result BinaryReaderObjdumpPrepass::OnFunctionName(uint32_t index,
 Result BinaryReaderObjdumpPrepass::OnReloc(RelocType type,
                                            uint32_t offset,
                                            uint32_t index,
-                                           int32_t addend) {
+                                           uint32_t addend) {
   BinaryReaderObjdumpBase::OnReloc(type, offset, index, addend);
   if (reloc_section == BinarySection::Code) {
     options->code_relocations.emplace_back(type, offset, index, addend);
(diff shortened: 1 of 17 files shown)


src/binary-reader-objdump.cc
```diff
@@ -824,11 +824,22 @@ Result BinaryReaderObjdump::OnRelocCount(uint32_t count,
...
       section_starts[static_cast<size_t>(reloc_section)] + offset;
-  PrintDetails("   - %-18s idx=%#-4x addend=%#-4x offset=%#x(file=%#x)\n",
-               get_reloc_type_name(type), index, addend, offset, total_offset);
+  PrintDetails("   - %-18s offset=%#08x(file=%#08x) index=%#08x",
+               get_reloc_type_name(type), offset, total_offset, index);
+  if (addend) {
+    int32_t signed_addend = static_cast<int32_t>(addend);
+    if (signed_addend < 0) {
+      PrintDetails("-");
+      signed_addend = -signed_addend;
```
binji: I think this is UB when signed_addend is -0x80000000. Probably not possible to generate this case, though.

test/link/table.txt
```diff
@@ -44,11 +44,11 @@ Elem:
 Custom:
  - name: "reloc.Elem"
   - section: Elem
-   - R_FUNC_INDEX_LEB   idx=0    addend=0    offset=0x6(file=0x38)
```
binji: no tests with a non-zero addend?

sbc100: Because we don't currently generate GLOBAL_ADDR relocations.  I haven't found a way to represent them in the wast format.  I guess we could use the binary builds to build them?

sbc100: Sorry I mean MEMORY_ADDR relocs.  They are the only type with an addend ( https://github.com/WebAssembly/tool-conventions/blob/master/Linking.md)

binji: Hm, I guess I'm just concerned that this will break otherwise. But if there's no easy way to do it, I guess it's OK. Do you mind creating an issue for it though?

sbc100: https://github.com/WebAssembly/wabt/issues/408

src/binary-reader-objdump.cc
sbc100: ha!  that would be quite an addend.  Funny that we were just chatting about that exact case the other day.

PR #411, opened 2017-04-26
wasmdump: support multiple input files

sbc100:
This mimics how objdump behaves. Also makes the code more
readable.

Looking into adding a test for this but looks tricky with the current framework.


sbc100:
Added a test

sbc100:
Does the test change look good to you?

binji:
I don't see a test...?

sbc100:
Oops, forgot to add the new test file.  Also i changed test/run-wasmdump.py so that it runs wasmdump in a single invocation rather than once per file so we get implicit coverage already.

binji:
Cool, looks good.

PR #413, opened 2017-05-02
Rename all occurences of AST

binji:
* ast.{cc,h} => ir.{cc,h}
* binary-reader-ast.{cc,h} => binary-reader-ir.{cc,h}
* ast-writer.{cc,h} => wat-writer.{cc,h}

Everything else changes ast => wast.

When deciding between wast vs. wat, the wat format should only allow for
a single module, without assertions.


binji:
This is a big change, but there's not really anything interesting going on aside from the name choices.

PR #416, opened 2017-05-05
Add fuzzing helper scripts, fix fuzzing bugs

binji:
The fuzzing dictionary was old, so I updated it. Also, I've added some
simple fuzzing shell scripts that make it a bit easier to run.

Bug fixes:

* Validate function index in local name section before calling callback.
* Fix invalid assert in parse_{float,double}_infinity, which assumed
  that "infinity" is required (it could be "inf").
* Bail out of resolve local names if there is no current function (e.g.
  if attempting to resolve names in an elem offset expression).
* Catch bad_alloc in wast2wasm and wasm2wast. Without this, afl-fuzz
  will see allocation failure as a crash.


binji:
Yes, we don't disable exceptions (though maybe we should), so allocation failure throws bad_alloc. And yes, the fuzzer causes allocation failures because it runs with a very small ulimit memory size. If you don't catch the allocation failure it just terminates which looks like a crash.

binji:
Hm, looks like `-fno-exceptions` will just turn a throw into an abort anyway, which won't really improve matters.

sbc100:
I think we should build with -fno-exceptions like chromium (and v8?).  How do other projects that use this flag deal with this issue?

binji:
I did a quick search to see if anybody does this, but didn't find anything. Looks like the nicest way would be to conditionally compile in the use of exceptions for afl-fuzz builds.

sbc100:
If there is no simpler way then I guess thats fine.  In any case if you are going to add try/catch around the body of the main functions I think you should split out the try body to avoid extra nesting and make the try catch statements close together visually.

binji:
Added `WITH_EXCEPTIONS` option, disabled by default.

binji:
Any further comments on this one?

binji:
Landing, assuming there are no further issues.

PR #418, opened 2017-05-07
Add highlights to readme code-blocks

Alhadis:
Nothing groundbreaking going on here. This PR simply improves the readability of the readme's code samples by colouring comments and output lines differently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ This will fetch the testsuite and gtest repos, which are needed for some tests.
 You'll need [CMake](https://cmake.org). If you just run `make`, it will run
 CMake for you, and put the result in `out/clang/Debug/` by default:
 
-> Note: If you are on OSX, you will need to use CMake version 3.2 or higher
+> Note: If you are on macOS, you will need to use CMake version 3.2 or higher
 
 ```console
 $ make
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ target; instead they aim for full fidelity and compliance with the spec (e.g.
 
 Clone as normal, but don't forget to get the submodules as well:
 
-```
+```console
 $ git clone --recursive https://github.com/WebAssembly/wabt
 $ cd wabt
 ```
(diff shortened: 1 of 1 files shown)


binji:
Looks pretty good. What's the difference between `sh` and `console`? Have you joined the [community group](https://www.w3.org/community/webassembly/)?

Alhadis:
The [language package](https://github.com/atom/language-shellscript) which GItHub uses to provide shell highlighting has two different forms:

- [`sh`](https://github.com/atom/language-shellscript/blob/master/grammars/shell-unix-bash.cson): For highlighting shell-script source code
- [`console`](https://github.com/atom/language-shellscript/blob/master/grammars/shell-session.cson): For highlighting code samples in readmes (such as this one)

The latter will colour differently every line which doesn't start with a prompt symbol. It also won't highlight comments (since some shells use `#` as a prompt string).

Hope that makes sense.

binji:
Thanks, that makes sense. Well, this looks good to me. I'm happy to merge, as long as you join the community group. Thanks!

README.md
```diff
@@ -36,9 +36,9 @@ This will fetch the testsuite and gtest repos, which are needed for some tests.
 You'll need [CMake](https://cmake.org). If you just run `make`, it will run
 CMake for you, and put the result in `out/clang/Debug/` by default:
 
-> Note: If you are on OSX, you will need to use CMake version 3.2 or higher
+> Note: If you are on macOS, you will need to use CMake version 3.2 or higher
```
sbc100: Is this preferable to OSX?   What is the rationale?

binji: Looks like that's what they call it now: https://en.wikipedia.org/wiki/Macintosh_operating_systems + https://www.apple.com/macos/sierra/

Alhadis: Yes, it was rebranded `macOS` last year. Many repositories have since updated their docs to use the new capitalisation.

Alhadis:
@binji Joined.

binji:
Thanks!

PR #421, opened 2017-05-10
Improve error messages in binary reader

sbc100:


src/binary-reader.cc
```diff
@@ -1213,26 +1213,26 @@ static void read_export_section(Context* ctx, uint32_t section_size) {
...
-                       "invalid export external kind");
+                       "invalid export external kind: %#x", external_kind);
 
     uint32_t item_index;
     in_u32_leb128(ctx, &item_index, "export item index");
     switch (static_cast<ExternalKind>(external_kind)) {
       case ExternalKind::Func:
         RAISE_ERROR_UNLESS(item_index < num_total_funcs(ctx),
-                           "invalid export func index: %d", item_index);
+                           "invalid export func index: %#x", item_index);
```
binji: Why make these hex values?

sbc100: Fair enough.  Done.

PR #422, opened 2017-05-10
[wasmdump] Don't assume names section is in order

sbc100:


pipcet:
My understanding is the spec mandates that the name section is in order. We should probably just throw an error?

sbc100:
I don't see anything in the spec about saying that the indexs have to be in order.  Perhaps I missed something?  I'm not sure if @lukewagner intended it perhaps?

binji:
Ah, @pipcet is right: https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#name-map

> sequence of naming sorted by index

pipcet:
Sorry for not providing the reference; yes, I meant the bit @binji quoted.

sbc100:
Ah, I didn't notice this.  Thanks!

PR #423, opened 2017-05-10
Regenerate expected test output

sbc100:
The expected test output was out of sync because whitespace
at the start and end of stdout doesn't cause test failures
(because stdout/stderr are stripped before comparison).


binji:
How was this generated? Can you fix `run-tests` to do this?

sbc100:
I just ran run-tests -r to generate.    I thought about fixing run-tests a while ago but it seemed like a unlikely occurrence with a trivial fix (regen all).  In any case the fix can be in a followup?

sbc100:
#423 .. in case we want this fixed.

PR #426, opened 2017-05-10
Ensure entries in name section appear in order

sbc100:


src/binary-reader.cc
```diff
@@ -930,11 +930,16 @@ static void read_names_section(Context* ctx, uint32_t section_size) {
         uint32_t num_names;
         in_u32_leb128(ctx, &num_names, "name count");
         CALLBACK(OnFunctionNamesCount, num_names);
+        uint32_t last_function_index = 0;
         for (uint32_t j = 0; j < num_names; ++j) {
           uint32_t function_index;
           StringSlice function_name;
 
           in_u32_leb128(ctx, &function_index, "function index");
+          RAISE_ERROR_UNLESS(
```
binji: I don't think this is right, since you could just reuse function_index 0 multiple times.

sbc100: Hmm.. well technically the spec says "sorted by index" which a repeating sequence of zero indexes would be :)    This was enough to make sure my lld port was going the right thing.

binji: Now that we have `Index` you can just initialize the last value to `kInvalidIndex` instead of 0.

sbc100: OK Done.  We should probably clarify the spec too.

sbc100:
Added a bunch more tests

sbc100:
@binji could you take another look since I added a bunch more test cases and rebased

PR #428, opened 2017-05-11
Use codemirror for wast2wasm JS demo

binji:
This also includes a simple codemirror syntax highlighter for the wast
format, using the `simple-mode.js` codemirror plugin.

I've also moved the examples out from `demo.js` to `examples.js`.

The codemirror.css and codemirror.js files are marked as binary so they
don't show up in diffs; we shouldn't be patching them anyway.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -11,3 +11,5 @@
 *.y text
 demo/primer.css binary
 demo/libwabt.js binary
+demo/third_party/codemirror/codemirror.css binary
+demo/third_party/codemirror/codemirror.js binary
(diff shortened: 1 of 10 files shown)


.gitattributes
```diff
@@ -11,3 +11,5 @@
 *.y text
 demo/primer.css binary
 demo/libwabt.js binary
+demo/third_party/codemirror/codemirror.css binary
+demo/third_party/codemirror/codemirror.js binary
```
sbc100: I'm not sure that not wanting to see the diffs is a good reason for marking something as binary.   Binary should be for binaries.

binji: But I read it on the internet! :-)
https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes

sbc100: I don't have strong opinions here. Up to you.

PR #429, opened 2017-05-11
Implement UTF-8 validation, update testsuite

binji:
This is enough to pass the testuite, but doesn't yet validate utf8
strings in the text format.


binji:
Looks like there are additional tests [here](https://github.com/WebAssembly/spec/pull/470) to handle the text format utf8 stuff. So this should be OK to land without for now.

Issue #322, opened 2017-02-27
Use Index/Address/Offset types instead of size_t/int, etc.


binji:
Binaryen set a good precedent here, it would be better to be more careful about the types that we use for manipulating wasm binaries.

PR #433, opened 2017-05-12
Use Index/Address/Offset instead of uint32_t

binji:
An `Index` is an index into one of the WebAssembly index spaces. It also
is used for counts for these spaces, as well as parameter counts and
result counts.

An `Address` is an index into linear memory, or the size of a data
region in linear memory.

An `Offset` is an offset into the host's file or memory buffer.

This fixes issue #322.

### DIFF
--- a/src/wat-writer.cc
+++ b/src/wat-writer.cc
@@ -85,7 +85,7 @@ class WatWriter {
   void WriteStringSlice(const StringSlice* str, NextChar next_char);
   bool WriteStringSliceOpt(const StringSlice* str, NextChar next_char);
   void WriteStringSliceOrIndex(const StringSlice* str,
-                               uint32_t index,
+                               Index index,
                                NextChar next_char);
   void WriteQuotedData(const void* data, size_t length);
   void WriteQuotedStringSlice(const StringSlice* str, NextChar next_char);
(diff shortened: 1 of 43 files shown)


sbc100:
Can you explain why you used Index for things like param counts?   This doesn't really seem like a index space in the same why... but either way this change seem like an improvement.

src/wat-writer.cc
```diff
@@ -122,14 +122,14 @@ class WatWriter {
   Result result_ = Result::Ok;
   int indent_ = 0;
   NextChar next_char_ = NextChar::None;
-  int depth_ = 0;
+  Index depth_ = 0;
```
sbc100: Why is depth_ an index?

binji: Depth is an index into the label index space, see: http://webassembly.github.io/spec/syntax/modules.html#indices

binji:
> Can you explain why you used Index for things like param counts?

Well, the spec uses `localidx` for accessing via `get_local` etc., so it seemed reasonable. I was considering using a different typedef for counts vs. indexes (for example using `Address` for linear memory sizes seems a bit weird), but I don't know that it would actually make things clearer.

PR #438, opened 2017-05-18
Add support for type relocations

sbc100:
This relocation type is used for the signature type
immediate of the call_indirect instruction.


sbc100:
@KarlSchimpf can you take a look since binji is OOO?

PR #440, opened 2017-05-18
Allow any module name

winksaville:
Currently a module name must be `__extern` with this change any module
name is allowed.

### DIFF
--- a/src/binary-reader-linker.cc
+++ b/src/binary-reader-linker.cc
@@ -130,18 +130,6 @@ Result BinaryReaderLinker::OnReloc(RelocType type,
 Result BinaryReaderLinker::OnImport(Index index,
                                     StringSlice module_name,
                                     StringSlice field_name) {
-#if 0
-  // Is this module been seen in the past
-  std::vector<FunctionImport>& imports = binary->function_imports;
-  for (size_t j = 0; j < imports.size(); j++) {
-    FunctionImport* import = &imports[j];
-    if (string_slices_are_equal(&module_name, &import->module_name)) {
-      // Found
-      return Result::Ok;
-    }
-  }
-  // A new module should anything be done?
-#endif
   return Result::Ok;
 }
 
--- a/src/binary-reader-linker.cc
+++ b/src/binary-reader-linker.cc
@@ -34,9 +34,6 @@ class BinaryReaderLinker : public BinaryReaderNop {
 
   Result BeginSection(BinarySection section_type, Offset size) override;
 
-  Result OnImport(Index index,
-                  StringSlice module_name,
-                  StringSlice field_name) override;
   Result OnImportFunc(Index import_index,
                       StringSlice module_name,
                       StringSlice field_name,
@@ -124,12 +121,6 @@ Result BinaryReaderLinker::OnReloc(RelocType type,
 
   reloc_section->relocations.emplace_back(type, offset, index, addend);
 
-  return Result::Ok;
-}
-
-Result BinaryReaderLinker::OnImport(Index index,
-                                    StringSlice module_name,
-                                    StringSlice field_name) {
   return Result::Ok;
 }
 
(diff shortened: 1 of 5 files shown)


src/binary-reader-linker.cc
```diff
@@ -130,20 +130,29 @@ Result BinaryReaderLinker::OnReloc(RelocType type,
...
+  std::vector<FunctionImport>& imports = binary->function_imports;
+  for (size_t j = 0; j < imports.size(); j++) {
+    FunctionImport* import = &imports[j];
+    if (string_slices_are_equal(&module_name, &import->module_name)) {
+      // Found
+      return Result::Ok;
+    }
   }
+  // A new module should anything be done?
+#endif
```
sbc100: Can you remove this dead code?

src/binary-reader-linker.cc
```diff
@@ -130,20 +130,29 @@ Result BinaryReaderLinker::OnReloc(RelocType type,
...
   return Result::Ok;
 }
 
 Result BinaryReaderLinker::OnImportFunc(Index import_index,
                                         StringSlice module_name,
                                         StringSlice field_name,
-                                        Index global_index,
-                                        Index sig_index) {
+                                        uint32_t global_index,
+                                        uint32_t sig_index) {
```
sbc100: Did you mean to make this change?

src/binary-reader-linker.cc
winksaville: Happy to remove, is there anything anyone knows we might want to do?

src/binary-reader-linker.cc
```diff
@@ -130,10 +130,6 @@ Result BinaryReaderLinker::OnReloc(RelocType type,
 Result BinaryReaderLinker::OnImport(Index index,
                                     StringSlice module_name,
```
sbc100: Can you remove this method completely now?

winksaville:
There are now a couple tests, is this adequate?

Question: In an `import` statement you can provide a `module_name` but I don't see where you can actually name a module? I would have expected something like below where I give the name `anything` to the second module which is the `module_name` used when importing `baz`:
```
(module
  (import "__extern" "missing0" (func $import_func0))
  (import "_extern1" "missing1" (func))
  (import "anything" "baz" (func $import_func1))
  (import "extern2" "missing2" (func $import_func2))
  (export "foo" (func $name1))
  (func $name1 (param $param1 i32)
     i32.const 1
     call 2)
  (func $name2 (param $param2 i64)
     i64.const 1
     call 3)
  (func (param $param2 i64))
)
(module "anything"
  (export "baz" (func 0))
  (func $name3 (param $param3 i32)
     i32.const 2
     call 0)
)
```

winksaville:
I'm a member of the [w3.org/community/webassembly](https://www.w3.org/community/webassembly/participants) search for `saville`, if that's what you mean.

src/binary-reader-linker.cc
winksaville: Done

winksaville:
I suggesting waiting until the CI checks have completed before committing, just to be safe :)

winksaville:
@sbc100, commit at will.

sbc100:
In answer to your question about module names, currently for the linker prototype and the linking conventions in https://github.com/WebAssembly/tool-conventions/blob/master/Linking.md we don't take advantage of wasm's two level namespace.  For s2wasm (and the lld port I'm working on) we simply put all symbols inside the 'env' module.  In the future we could start using the two level namespace, in particular I can see this being useful for dynamic linking.

winksaville:
@sbc100, thanks for the info on module linking, but how do you name a module?

sbc100:
wasm modules don't have names.   Or maybe it makes more sense to say that we haven't defined a convention for that the module field of the imports means. At least not for static linking.    We currently use a flat namespace, as most C/C++ developers expect.

winksaville:
I find that odd, especially since import statements have a `module name`. Historically I come from an embedded world (although that was awhile ago) and have used things like linker scripts to control where various sections are placed. Obviously its early days for wasm, but I'm quite excited about wasm in more than just web systems and see wider possibilities.

Anyway, can you suggest the best place I might ask about `module name` support? Or maybe you suggest "forget it for now" ;)

sbc100:
You might want to open a bug in the design repo?   However, I imagine that that answer will be that how modules are linked together is outside of the code spec and belongs in the tooling conventions.

winksaville:
I'll do that and we'll see what happens, txs.

PR #441, opened 2017-05-18
[wasmdump] Display function names at call sites in disassembly

sbc100:
Also print Index types as decimal rather than hex
since this seems to be the convention elsewhere.


src/binary-reader-objdump.cc
```diff
@@ -296,6 +297,16 @@ Result BinaryReaderObjdumpDisassemble::OnOpcodeBare() {
   return Result::Ok;
 }
 
+Result BinaryReaderObjdumpDisassemble::OnOpcodeIndex(Index value) {
```
KarlSchimpf: Minor question. Since the calling context should know if the index corresponds to a call, why not have two different functions "OnOpcodeIndex()" and "OnOpcodeCallIndex()"?

In this view, you wouldn't need to check the current_opcode, and OnOpcodeIndex() would not need to be conditionalized. However, OnOpcodeCallIndex still would, since there may not be a function name.

sbc100: I was trying to avoid having too many callbacks here, but maybe you are right.. let me see.

sbc100: So the idea is that this disassembler can be a little dumb and doesn't need to implment one callback for each opcode, it can just implement OnOpcode and then OnOpcode<ArgType>, so it seems string to have specific  OnOpcode<ArgType>, since we already have OnCallExpr which basically does that for the call opcode specifically.   But I agree that this change is moving towards caring about the specfic opcode...which is counter to that goal.

sbc100: I think I will rename these OnOpcode* callback in a follow up to make this more clear

src/binary-reader.cc
```diff
@@ -604,15 +604,15 @@ static void read_function_body(Context* ctx, Offset end_offset) {
         break;
 
       case Opcode::I32Const: {
-        uint32_t value = 0;
+        uint32_t value;
```
binji: hm, surprised this isn't giving warnings for potentially uninitialized values.

src/binary-reader-objdump.cc
binji: Yeah, I think this is OK for now.

PR #446, opened 2017-05-22
Fix disassembly of 64-bit values in wasmdump

### LINKED ISSUE #445: i64.const too big are not printed correctly by wasmdump
Given a cpp file

```c_cpp
#include <stdint.h>

uint64_t fct1() { return 1llu << 63; }
uint64_t fct2() { return (1llu << 63) + 255; }
```
the .s generated by clang contains the correct value

```
fct1: i64.const	$push0=, -9223372036854775808
fct2: i64.const	$push0=, -9223372036854775553
```
the .wast is ok too
```
 (func $_Z4fct6v (result i64)
  (i64.const -9223372036854775808)
 )
 (func $_Z4fct7v (result i64)
  (i64.const -9223372036854775553)
 )
```

.wasm is generated by binaryen, wasmdump print them as
```
004e7f func[1679]:
 004e81: 42 80 80 80 80 80 80 80 80 80 | i64.const 0
 004e8c: 0b                         | end
004e8d func[1680]:
 004e8f: 42 ff 81 80 80 80 80 80 80 80 | i64.const 255
 004e9a: 0b                         | end
```
output should look like
```
004e7f func[1679]:
 004e81: 42 80 80 80 80 80 80 80 80 80 7f|  i64.const -9223372036854775808
 004e8c: 0b                         |  end
004e8d func[1680]:
 004e8f: 42 ff 81 80 80 80 80 80 80 80 7f|  i64.const -9223372036854775553
 004e9a: 0b                         |  end
```

wasmdump correctly eat the right number of byte.

Another problem but unrelated to wasmdump, in the last byte, only bit 0 is used and I expected all other bits set to 0, I don't know if leb128 encoding specify than unused bits of last byte when reaching max length are garbage, must be set to 1 or must be set to 0. Any thought about that?

binji:
Fixes #445.


binji:
@phil-el good point, though I'd actually prefer to have `OnOpcodeUint32` print as decimal instead of hex too. It looks like this callback is only used in `i32.const`, `current_memory`, `grow_memory`. All of which should probably print as "%u" instead of "%#x", I think. @sbc100, wdyt?

sbc100:
sgtm

Issue #448, opened 2017-05-22
wasm2wast should add $ to debug names


sbc100:
IIRC the $ that preceeds names in the wast format is not actually part of the name and should be stripped on intput and generated on output.

Right now I run wasm2wast on an llvm-generated object the wast file is invalid because the names are not preceded by $.

binji:
OK, I think this is reasonable. I think we should confirm this behavior with others though.

PR #449, opened 2017-05-22
Don't include leading $ in wasm binary name section

sbc100:
Attempt at fixing #448 .. still need to figure out how to strip the $ on parse.


binji:
You could strip it in the parser, but then the names wouldn't be as nice in errors, I think. Probably the simplest place to do it is in `binary-writer.cc`, just add a new function `write_name` instead of `write_str` that strips the initial $, I think.

binji:
Also, for the same reason, it might be nicer to prepend the "$" in `binary-reader-ir.cc` instead of `wat-writer.cc`.

sbc100:
OK.. ptal

PR #450, opened 2017-05-23
Use + with $(MAKE) to use jobserver

binji:
Without this, Make complains:

    make[1]: warning: jobserver unavailable: using -j1.  Add '+' to
    parent make rule.


sbc100:
What does this actually go?  (other than stop the warning being printed)

binji:
Oh, without it you can't build in parallel. `make -j10` just uses `-j1`.

PR #451, opened 2017-05-23
Clean up BinaryReader; rename to BinaryReaderDelegate

binji:
This removes the use of setjmp/longjmp as well.

### DIFF
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -40,10 +40,12 @@
       return Result::Error; \
   } while (0)
 
-#define ERROR_UNLESS(expr, ...)    \
-  do {                             \
-    if (!(expr))                   \
-      return OnError(__VA_ARGS__); \
+#define ERROR_UNLESS(expr, ...) \
+  do {                          \
+    if (!(expr)) {              \
+      PrintError(__VA_ARGS__);  \
+      return Result::Error;     \
+    }                           \
   } while (0)
 
 #define CALLBACK0(member)                                   \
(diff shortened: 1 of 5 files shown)


src/binary-reader.cc
```diff
@@ -206,70 +140,207 @@ size_t read_i32_leb128(const uint8_t* p,
...
+      state_(static_cast<const uint8_t*>(data), size),
+      logging_delegate_(options->log_stream, delegate),
+      delegate_(options->log_stream ? &logging_delegate_ : delegate),
+      options_(options),
+      last_known_section_(BinarySection::Invalid) {
+  delegate->OnSetState(&state_);
+}
+
+Result WABT_PRINTF_FORMAT(2, 3) BinaryReader::OnError(const char* format, ...) {
+  WABT_SNPRINTF_ALLOCA(buffer, length, format);
```
sbc100: This is really just nit/suggestion.

Could this method be better named?  OnError looks like a delegate method name (and indeed it is one).  All of the "return OnError()" read a little funny to me too.  There doesn't seem to be any point in the return value here either since its always Result::Error.   Would it be clearer if the calling code pattern was more like?:

```
ShowError(....)   // retuns void
return Result::Error;
```

ShowError?  or RaiseError?  Or HandleError?

binji: Yeah, in other places I've done
```
PrintError(...)
return Result::Error;
```

I thought this new way was a little nicer since it is a one-liner, but it's less clear. I'll change it.

binji:
The callgraph didn't change, setjmp/longjmp was used to throw exceptions, but now I just bubble the errors up the stack.

binji:
OK, renamed `OnError` -> `PrintError`. Also switched some `if (...) { PrintError(...); return` to `ERROR_UNLESS(...)`. Take a look at https://github.com/WebAssembly/wabt/pull/451/commits/9442c76916701f72f0630f5533562dde59090620.

binji:
Sorry!
/me races to land first... :-P

PR #453, opened 2017-05-23
[wasm-link] Add some helper methods to LinkerInputBinary

sbc100:
Make things a little more C++-y and some helper methods
to make the code more clear.

### DIFF
--- a/src/tools/wasm-link.cc
+++ b/src/tools/wasm-link.cc
@@ -168,9 +168,8 @@ bool LinkerInputBinary::IsValidFunctionIndex(Index index) {
 }
 
 Index LinkerInputBinary::RelocateFuncIndex(Index function_index) {
-  assert(IsValidFunctionIndex(function_index));
   Index offset;
-  if (function_index >= function_imports.size()) {
+  if (!IsFunctionImport(function_index)) {
     /* locally declared function call */
     offset = function_index_offset;
     if (s_debug)
--- a/src/tools/wasm-link.cc
+++ b/src/tools/wasm-link.cc
@@ -154,17 +154,30 @@ LinkerInputBinary::~LinkerInputBinary() {
   delete[] data;
 }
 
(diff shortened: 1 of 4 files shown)


src/tools/wasm-link.cc
```diff
@@ -154,17 +154,30 @@ LinkerInputBinary::~LinkerInputBinary() {
...
+
+bool LinkerInputBinary::IsValidFunctionIndex(Index index) {
+  return index < function_imports.size() + function_count;
+}
+
+Index LinkerInputBinary::RelocateFuncIndex(Index function_index) {
+  assert(IsValidFunctionIndex(function_index));
   Index offset;
-  if (function_index >= binary->function_imports.size()) {
+  if (function_index >= function_imports.size()) {
```
binji: could use `!IsFunctionImport(...)` and remove the assert (since it's already in `IsFunctionImport`)

src/wasm-link.h
```diff
@@ -98,7 +98,8 @@ struct Section {
 
 typedef std::vector<Section*> SectionPtrVector;
 
-struct LinkerInputBinary {
+class LinkerInputBinary {
+public:
```
binji: nit: indent 1 space

src/wasm-link.h
```diff
@@ -124,8 +125,17 @@ struct LinkerInputBinary {
   Index memory_page_offset;
 
   Index table_elem_count;
+  Index function_count;
 
   std::vector<std::string> debug_names;
+
+  Index RelocateFuncIndex(Index findex);
```
binji: nit: put member functions before member data.

src/wasm-link.h
```diff
@@ -124,8 +125,17 @@ struct LinkerInputBinary {
   Index memory_page_offset;
 
   Index table_elem_count;
+  Index function_count;
```
binji: might be nice to make the member data private at some point, but this is fine for now.

PR #454, opened 2017-05-23
Move interpreter stuff into its own namespace

binji:
The only names that conflict are `wabt::Opcode` vs.
`interpreter::Opcode` and `wabt::Result` vs. `interpreter::Result`.
Other names would conflict too (e.g. `FuncSignature`) but those aren't
ever included in the interpreter-related code.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -102,8 +102,8 @@ enum class Result {
   Error,
 };
 
-#define WABT_SUCCEEDED(x) ((x) == Result::Ok)
-#define WABT_FAILED(x) ((x) == Result::Error)
+#define WABT_SUCCEEDED(x) ((x) == ::wabt::Result::Ok)
+#define WABT_FAILED(x) ((x) == ::wabt::Result::Error)
 
 inline std::string WABT_PRINTF_FORMAT(1, 2)
     string_printf(const char* format, ...) {
--- a/src/interpreter.h
+++ b/src/interpreter.h
@@ -29,9 +29,9 @@
 
 namespace wabt {
 
-// TODO(binji): move all the Interpreter stuff into its own namespace.
-
 class Stream;
+
+namespace interpreter {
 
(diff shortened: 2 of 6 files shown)


src/common.h
```diff
@@ -102,8 +102,8 @@ enum class Result {
   Error,
 };
 
-#define WABT_SUCCEEDED(x) ((x) == Result::Ok)
-#define WABT_FAILED(x) ((x) == Result::Error)
+#define WABT_SUCCEEDED(x) ((x) == ::wabt::Result::Ok)
+#define WABT_FAILED(x) ((x) == ::wabt::Result::Error)
```
sbc100: Do you need the leading :: here?  I've not seen this pattern used much.  Why here?

src/interpreter.h
```diff
@@ -92,73 +92,71 @@ static const IstreamOffset kInvalidIstreamOffset = ~0;
 #define WABT_TABLE_ENTRY_DROP_OFFSET sizeof(uint32_t)
 #define WABT_TABLE_ENTRY_KEEP_OFFSET (sizeof(IstreamOffset) + sizeof(uint32_t))
 
-enum class InterpreterOpcode {
+enum class Opcode {
```
sbc100: Won't this make using simple tools like grep to find the definition of OpCode a little harder?  I guess it fine..

src/common.h
binji: It's probably not needed. It just makes sure that you're accessing the global namespace, in case `wabt::Result::Ok` was defined in some other namespace.

src/interpreter.h
binji: Yes, that is a concern. But I think it's a win when you balance it against having simpler names.

src/common.h
sbc100: Really?  wabt::Result::Ok existing in two places is surprising.   But I'm prepared to believe it :)

PR #457, opened 2017-05-25
Only use single line test output if attached to a tty

sbc100:
This means we get verbose test output when (for example)
running within ninja.

### DIFF
--- a/test/run-tests.py
+++ b/test/run-tests.py
@@ -751,16 +751,14 @@ def main(args):
     print('\nInterrupted testing\n')
 
   status.Done()
-  ret = 0
 
+  ret = 0
   if status.failed:
     sys.stderr.write('**** FAILED %s\n' % ('*' * (80 - 14)))
     for info in status.failed_tests:
       sys.stderr.write('- %s\n    %s\n' % (info.GetName(),
                                            ' '.join(info.last_cmd)))
     ret = 1
-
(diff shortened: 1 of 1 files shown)


test/run-tests.py
```diff
@@ -752,18 +750,16 @@ def main(args):
...
   ret = 0
-
   if status.failed:
     sys.stderr.write('**** FAILED %s\n' % ('*' * (80 - 14)))
     for info in status.failed_tests:
       sys.stderr.write('- %s\n    %s\n' % (info.GetName(),
                                            ' '.join(info.last_cmd)))
     ret = 1
 
-  status.Print()
```
binji: Ah, I remember why I added this. After this PR, the output is inconsistent when you run it:
```
$ test/run-tests.py parse
[+503|-0|%100] (1.43s) test/parse/string-hex.txt (roundtrip fold-exprs)
$ test/run-tests.py parse
[+529|-0|%100] (1.37s) test/parse/string-hex.txt (roundtrip fold-exprs)
```

sbc100: I can't repro that..  I get 528 each time.

binji: Hm, well I'm running this on a machine with `multiprocessing.cpu_count() => 32`, maybe that makes a difference?

PR #460, opened 2017-05-26
Add support for folding expressions in wat-writer

binji:


test/roundtrip/fold-fac.txt
```diff
@@ -0,0 +1,60 @@
...
+    (set_local $i
+      (get_local $n))
+    (set_local $res
+      (i64.const 1))
+    (block $B0
+      (loop $L1
+        (if $I2
+          (i64.eq
+            (get_local $i)
+            (i64.const 0))
```
sbc100: Why not put pack these onto a single line?

binji:
> I thought we were moving away from this syntax?

Nope, it's going to be official.

test/roundtrip/fold-fac.txt
binji: That would be a nice improvement, but it would take some additional logic, I think. What we really want is something clever like clang-format to determine whether the rest of the operands fit nicely.

PR #463, opened 2017-05-30
Update relocation offsets when fixing up section size

### LINKED ISSUE #447: Function relocations not executing with wasm-link
In a simple linking test I've got something similar to the following linking scenario:

```
(module
  (import "__extern" "foo" (func $import0 (param i32) (result i32)))
  (import "__extern" "bar" (func $import1 (param i32) (result i32)))
  (func $local_func (param i32)
     get_local 0
     call $local_func
     call $import0
     call $import1
     call 3)
  (func (param i32)
    get_local 0)
)
(module
  (import "__extern" "baz" (func $import0 (param f64)))
  (export "foo" (func $local_func))
  (func $local_func (param i64)
     f64.const 1
     call $import0
     i64.const 10
     call $local_func)
)
```

The `call 3` in the first module should be updated to become a `call 2` in the linked output. I'm seeing the custom relocation data being created to represent this relocation in the binary, but when running in the browser it doesn't seem to pick this up and instead gives a stack error.

Is this because browser support for relocations is pending? Or is there something that can be done here to get these relocations to work in the browser?

sbc100:
Fixes #447

### DIFF
--- a/src/binary-writer.cc
+++ b/src/binary-writer.cc
@@ -199,9 +199,9 @@ class BinaryWriter {
  private:
   void WriteHeader(const char* name, int index);
   Offset WriteU32Leb128Space(Offset leb_size_guess, const char* desc);
-  void WriteFixupU32Leb128Size(Offset offset,
-                               Offset leb_size_guess,
-                               const char* desc);
+  Offset WriteFixupU32Leb128Size(Offset offset,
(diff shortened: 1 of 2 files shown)


guybedford:
Thanks so much for the quick fix here, hope it wasn't too much of a distraction.

PR #466, opened 2017-06-02
Cleanup Lexer

binji:
* Add class LexerSource: it encapsulates reading data from a source,
  either buffer or file.
* Add class LexerSourceLineFinder: it uses a LexerSource to read a line
  from the source for displaying in errors. It lazily caches line
  numbers -> file offsets, so no work is done unless an error occurs.
* Make WastLexer a class instead of a struct.


src/lexer-source.h
```diff
@@ -0,0 +1,120 @@
...
+  Result ReadRange(OffsetRange, std::vector<char>* out_data) override;
+
+ private:
+  const void* data_;
+  Offset size_;
+  Offset read_offset_;
+};
+
+class LexerSourceLineFinder {
+ public:
```
sbc100: Make this its own file and header?

src/lexer-source.h
```diff
@@ -0,0 +1,120 @@
...
+namespace wabt {
+
+template <typename T>
+struct Range {
+  Range() : start(0), end(0) {}
+  Range(T start, T end) : start(start), end(end) {}
+  T start;
+  T end;
+
+  T size() const { return end - start; }
```
binji: Yeah, that's what I figured. And also since size() is all over STL. But I'm OK either way, honestly.

src/lexer-source.h
binji: Done.

PR #471, opened 2017-06-02
Make ExprVisitor into C++ class

binji:
Add ExprVisitor::Delegate for callbacks, and ExprVisitor::DelegateNop
when some functions aren't implemented (similar to BinaryReader).


src/apply-names.cc
```diff
@@ -32,9 +33,26 @@ namespace wabt {
 
 namespace {
 
-struct Context {
+struct Context : ExprVisitor::DelegateNop {
```
sbc100: My gut tells me we should be using class here since there are many (virtual) methods, but maybe my gut is wrong.

src/generate-names.cc
```diff
@@ -33,8 +34,12 @@ namespace wabt {
 
 namespace {
 
-struct Context {
-  Context() : module(nullptr), label_count(0) { WABT_ZERO_MEMORY(visitor); }
+struct Context : public ExprVisitor::DelegateNop {
+  Context() : module(nullptr), visitor(this), label_count(0) {}
```
sbc100: Ditto, struct inheriting from classes seems a string.

src/apply-names.cc
binji: Agreed, just thought it might be nicer to do that in another PR

PR #475, opened 2017-06-04
Make TypeChecker into C++ class

binji:


binji:
Assuming this is uncontroversial, landing.

PR #476, opened 2017-06-04
Make Validator into C++ class

binji:


binji:
Assuming this cleanup is uncontroversial. Landing.

PR #479, opened 2017-06-06
[objdump] Split out state from ObjdumpOptions

sbc100:
Also, use C++ constructor inheritance to avoid some
boilerplate constructors.

### DIFF
--- a/src/binary-reader-objdump.cc
+++ b/src/binary-reader-objdump.cc
@@ -34,7 +34,8 @@ class BinaryReaderObjdumpBase : public BinaryReaderNop {
  public:
   BinaryReaderObjdumpBase(const uint8_t* data,
                           size_t size,
-                          ObjdumpOptions* options);
+                          ObjdumpOptions* options,
+                          ObjdumpState* state);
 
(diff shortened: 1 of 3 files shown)


src/binary-reader-objdump.cc
```diff
@@ -46,9 +47,10 @@ class BinaryReaderObjdumpBase : public BinaryReaderNop {
  protected:
   const char* GetFunctionName(Index index);
 
-  ObjdumpOptions* options = nullptr;
-  const uint8_t* data = nullptr;
-  size_t size = 0;
+  ObjdumpOptions* options;
+  ObjdumpState* objdump_state_;
```
binji: seems weird to have `objdump_state_` but the rest without trailing underscore.

binji: Also, could just call this `state_`

sbc100: There is already state member from the reader interface (admittedly without an _, but have state and state_ would be strange).

sbc100: Removed the trailing _ for now

PR #480, opened 2017-06-07
Store all test intermediate files in output directory

sbc100:
Previously we would you temp directories that we would
clean up after the test run.  This way the intermediates
are always available for inspection after a run.

### DIFF
--- a/test/gen-spec-js.py
+++ b/test/gen-spec-js.py
@@ -512,22 +512,24 @@ def main(args):
   # modules is a list of pairs: [(module_command, [assert_command, ...]), ...]
   modules = CollectInvalidModuleCommands(all_commands)
 
-  with utils.TempDirectory(options.temp_dir, 'gen-spec-js-') as temp_dir:
-    extender = ModuleExtender(wast2wasm, wasm2wast, temp_dir)
-    for module_command, assert_commands in modules:
-      if assert_commands:
-        wasm_path = os.path.join(json_dir, module_command['filename'])
-        new_module_filename = extender.Extend(wasm_path, assert_commands)
-        module_command['filename'] = os.path.relpath(new_module_filename,
-                                                     json_dir)
+  temp_dir = options.temp_dir
+  assert(temp_dir)
 
-    output = StringIO()
-    if options.prefix:
-      with open(options.prefix) as prefix_file:
-        output.write(prefix_file.read())
-        output.write('\n')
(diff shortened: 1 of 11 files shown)


test/gen-spec-js.py
```diff
@@ -512,22 +512,24 @@ def main(args):
...
-
-    output = StringIO()
-    if options.prefix:
-      with open(options.prefix) as prefix_file:
-        output.write(prefix_file.read())
-        output.write('\n')
-
-    JSWriter(json_dir, all_commands, output).Write()
+  temp_dir = options.temp_dir
+  assert(temp_dir)
```
binji: I run these commands from outside the test runner occasionally, so I'd like something nicer than an assert if I forget to provide a temporary directory. Can you use `parser.error` or something instead?

test/run-tests.py
```diff
@@ -534,7 +547,13 @@ def RunTest(info, options, variables, verbose_level=0):
   cwd = REPO_ROOT_DIR
   gen_input_path = info.CreateInputFile()
   rel_gen_input_path = os.path.relpath(gen_input_path, cwd)
-  variables['out_dir'] = os.path.dirname(rel_gen_input_path)
+
+  out_dir = os.path.splitext(rel_gen_input_path)[0] + "_out"
```
binji: So this would convert a test named `test/parse/expr/br.txt` to an output directory named `out/test/parse/expr/br_out`, right?

Can we just use the name without "_out" instead?

test/gen-spec-js.py
binji: And honestly, the purpose of this change (keeping the output) still works, even if you don't remove the TempDirectory stuff, right? Why remove it?

test/run-tests.py
sbc100: I just wanted to be sure that nothing was still using /tmp.   With this change the test will never delete their intermediate files.

test/gen-spec-js.py
sbc100: I just wanted to be sure that nothing was still using /tmp. With this change the test will never delete their intermediate files.

test/run-tests.py
sbc100: Done.

sbc100:
I think as followup change we could move some of the arg parsing and setup into a shared location so all these scripts behave the same.

sbc100:
OK, I reverted the controversial parts.

PR #481, opened 2017-06-07
[objdump] Be more explicit about type and element indexes

sbc100:


sbc100:
Yes, I was concerted about the the elem indexes being interpreted as being absolute table indexes.  However,  i think this is presenting more information, and TBH all of the objects I deal with have segments with init_expr = 0.

PR #484, opened 2017-06-08
Update testsuite; more lexer/parser changes

binji:
* Add support for quoted modules: `(module quote "...")`
* Binary modules must be annotated: `(module binary "...")`
* Multiple result blocks are no longer a parser error:
  `(func (result i32) (result i32) ...)`
* Function types can specify unused bind variables:
  `(type (func (param $foo)))`
* Rename `RawModule` -> `ScriptModule`. This encapsulates a module that
  may not be parsed yet, whether binary or "quoted".
* Validate load/store offsets and alignment in the parser, not in the
  validator. The spec tests assume that you can catch these errors with
  `assert_malformed`.
* Parse wast files in `wasm-interp` when checking malformed/invalid/etc.
  modules. This allows us to run all assertions at the same time, which
  is nice. `wasm-interp` should probably be renamed, though.
* Two tests in `type.wast` fail because they use:
  `(assert_invalid (module quote "..."))`. I'd prefer that we don't
  support this, since it's unnecessary, and additional work. I'll fix in
  a follow-up CL if we decide this is worth keeping.


binji:
Hm, looks slash vs. backslash problems on windows.

PR #488, opened 2017-06-09
Use C++ class for HostImportDelegate

binji:


src/tools/wasm-interp.cc
```diff
@@ -401,107 +401,103 @@ static interpreter::Result default_host_callback(const HostFunc* func,
...
+};
 
 static void init_environment(Environment* env) {
   HostModule* host_module =
       env->AppendHostModule(string_slice_from_cstr("spectest"));
-  host_module->import_delegate.import_func = spectest_import_func;
-  host_module->import_delegate.import_table = spectest_import_table;
-  host_module->import_delegate.import_memory = spectest_import_memory;
-  host_module->import_delegate.import_global = spectest_import_global;
+  host_module->import_delegate.reset(new SpectestHostImportDelegate());
```
sbc100: I've become used to see make_unique() used for this.  llvm implements it locally since its in C++14 only I think: http://llvm.org/docs/doxygen/html/STLExtras_8h_source.html

binji: Yeah, there's a few places it would be nice to have `make_unique`. We should add our own implementation as well (just copy the one from binaryen.)

PR #490, opened 2017-06-10
Add syntax for try blocks to the wast parser.

KarlSchimpf:
Adds the syntax for the proposed try blocks to the wast parser.

Currently only wast2wasm knows about this syntax (using the CL argument "--except"),

Added two tests (in test-exceptions) that are incomplete, but do parse.

Also modified the wast parser so that debugging can be turned on (see wast2wasm --debug-parser flag).


src/ir.h
```diff
@@ -135,6 +141,9 @@ struct Expr {
   static Expr* CreateBrTable(VarVector* targets, Var default_target);
   static Expr* CreateCall(Var);
   static Expr* CreateCallIndirect(Var);
+  static Expr* CreateCatch(Var);
+  static Expr* CreateCatchAll();
+  static Expr* CreateCatchBlock(Expr* catch_, Expr* first);
```
binji: Maybe nicer to just have Catch and CatchAll each be a block? Is there a benefit to having a generic CatchBlock expr?

src/wast-parser.y
```diff
@@ -623,6 +708,29 @@ if_ :
     }
 ;
 
+rethrow_check : RETHROW {
+      if (!parser->flags->allow_exceptions) {
```
binji: I think I'd prefer having a function that does this check, rather than having it as part of the grammar. Does that break the error-handling?

test-exceptions/try-text.wast
```diff
@@ -0,0 +1,13 @@
+;; out/clang/Debug/wast2wasm --except test-exceptions/try-text.wast
```
binji: You should be able to just use `;; FLAGS: --except` here

src/tools/wast2wasm.cc
```diff
@@ -90,8 +93,12 @@ static Option s_options[] = {
     {FLAG_VERBOSE, 'v', "verbose", nullptr, NOPE,
      "use multiple times for more info"},
     {FLAG_HELP, 'h', "help", nullptr, NOPE, "print this help message"},
+    {FLAG_DEBUG_PARSER, 0, "debug-parser", nullptr, NOPE,
+     "Turn on debugging the parser of wast files"},
     {FLAG_DUMP_MODULE, 'd', "dump-module", nullptr, NOPE,
      "print a hexdump of the module to stdout"},
+    {FLAG_EXCEPTIONS, 0, "except", nullptr, NOPE,
```
KarlSchimpf: I agree I didn't know the best way to introduce this. Changing to "future-exceptions".

src/wast-parser.h
```diff
@@ -24,7 +24,13 @@ namespace wabt {
 struct Script;
 class SourceErrorHandler;
 
-Result parse_wast(WastLexer* lexer, Script** out_script, SourceErrorHandler*);
+struct WastParseFlags {
```
KarlSchimpf: done

src/wast-parser.y
```diff
@@ -560,6 +578,41 @@ block :
     }
 ;
 
+
+catch_all_check : CATCH_ALL {
+      if (!parser->flags->allow_exceptions) {
```
KarlSchimpf: Good suggestion.

src/wast-parser.y
```diff
@@ -560,6 +578,41 @@ block :
...
+        wast_parser_error(&@1, lexer, parser, "Catch blocks not allowed");
+      }
+      @$ = @1;
```
KarlSchimpf: Removed.

src/wast-parser.y
```diff
@@ -584,7 +637,39 @@ expr1 :
...
+  | try_check labeling_opt block RPAR catch_list {
+      $3->label = $2;
+      Expr* try_ = Expr::CreateTry($3, $5.first);
+      $$ = join_exprs1(&@1, try_);
+    }
+  ;
+
+catch_list :
+    catch_ {
+      $$ = $1;
```
KarlSchimpf: omitted.

src/wast-parser.y
KarlSchimpf: Added CHECK_ALLOW_EXCEPTIONS.

test-exceptions/try-text.wast
KarlSchimpf: Actually, it just represents the manual command to run, since I don't know how to use the scripting language.

src/ir.h
KarlSchimpf: First, they are not a block (in the same way that an "else" of the "if" isn't a block).

Changing to have catch and catch_all be a { var , expr* } pair.

KarlSchimpf: There shouldn't be a catch block at all. The problem was I was trying to find the right syntax (without parse conflicts) that I lost sight of how it should be structured. The simple (clean) solution appears to have the body of the "try" s-exp be a "block", and have the catches (including instruction list) appear after the block.

test-exceptions/try-text.wast
KarlSchimpf: Ok, figured it out. Made test official.

KarlSchimpf:
Fixed issues, ready for review.

src/wast-parser.y
```diff
@@ -1462,6 +1540,19 @@ ExprList join_exprs2(Location* loc, ExprList* expr1, Expr* expr2) {
   return result;
 }
 
+ExprList join_expr_lists(Location* loc, ExprList* expr1, ExprList* expr2) {
```
binji: loc is unused?

test/exceptions/bad-try.txt
```diff
@@ -0,0 +1,23 @@
+;;; EXE: %(wast2wasm)s
```
binji: You can just use `;;; TOOL: wast2wasm` (see the top of run-tests.py for the list of valid tools)

Or you can just omit, since wast2wasm is the default. Though that's historical (since wast2wasm used to be the only tool) so maybe it's better to be explicit. :-)

src/wast-parser.y
binji: Hm, this looks like it could just use `append_expr_list` twice. Though honestly the whole thing should be rewritten to C++-style, so this is mostly just nitpicking.

KarlSchimpf: Removed loc.

KarlSchimpf: Changed to use append_expr_list.

test/exceptions/bad-try.txt
KarlSchimpf: Changed to use TOOL

PR #491, opened 2017-06-11
Add wasm2wast functionality

binji:
* Add demo index landing page
* Move old demo index -> wast2wasm/index.html
* Add demo for wasm2wast
* ~~Fix wast2wasm example (`i64` -> `(result i64)`)~~
* ~~Fix bug in BinaryReaderIR w/ debug names:
  The debug name don't have a '$', and add it in when converting to IR.
  This worked properly for func->name, but wasn't updated for the func
  bindings.~~
* Fix bug when setting default values in JS:
  `foo || true` will never be false, need to check for undefined
  specifically.
* Fix bug in `allocateBuffer`; the offset in TypedArray.set is the
  offset of the source buffer not the destination. You have to create a
  new view to change the destination offset.
* Add new functionality to the emscripten build:
  - Read binary file into `WasmModule`
  - Converting from `WasmScript`/`WasmModule` to text or binary
* Wrap `libwabt.js` module in IIFE so as to not pollute the global
  namespace with emscripten junk


binji:
@sbc100 mind taking a look? It's kind of a big change, so I could look at breaking it up if that helps.

src/binary-error-handler.h
```diff
@@ -55,6 +57,18 @@ class BinaryErrorHandlerFile : public BinaryErrorHandler {
   PrintHeader print_header_;
 };
 
+class BinaryErrorHandlerBuffer : public BinaryErrorHandler {
+ public:
+  explicit BinaryErrorHandlerBuffer();
```
sbc100: Does explicit have any meaning without any args?

test/roundtrip/inline-export-func-name.txt
```diff
@@ -0,0 +1,12 @@
+;;; TOOL: run-roundtrip
+;;; FLAGS: --stdout --inline-export --debug-names
```
sbc100: Yeah this should probably be a seperate change.

src/binary-error-handler.h
binji: Nope, that's a mistake. :-)

test/roundtrip/inline-export-func-name.txt
binji: Refactored this into a couple changes (#498, #499)

PR #494, opened 2017-06-12
Add exception declarations and examples.

KarlSchimpf:
Extends the wast parser to recognize exception declarations, including local, imported, and exported.

Also fixed labeling issues with try blocks and rethrows. Fixes catch_all section that doesn't get an arguments.

Adds examples of these new extensions. Also updates the examples to include throws and rethrows.


src/wast-parser.y
```diff
@@ -650,9 +652,10 @@ expr1 :
       assert(if_->type == ExprType::If);
       if_->if_.true_->label = $2;
     }
-  | try_check LPAR BLOCK labeling_opt block RPAR catch_list {
-      $5->label = $4;
-      Expr* try_ = Expr::CreateTry($5, $7.first);
+  | try_check labeling_opt LPAR BLOCK labeling_opt block RPAR catch_list {
```
KarlSchimpf: Note: I don't know if this is what we want, in terms of labels. The try label is for the rethrow instruction, while the block label is for other control.

This may be wrong, and we want to use the same label for both cases. If you agree that a single label should be used (like the WAT case), then where should it be located?

binji: Oh, I guess I missed this before. It seems to me that try is the block, and you shouldn't need both, right?
So it should be `try_check labeling_opt block catch_list` instead. Or does that not work?

src/wast-parser.y
```diff
@@ -743,6 +746,23 @@ const_expr :
     instr_list
 ;
 
+/* Exceptions */
+except :
```
binji: except vs exception is a bit confusing here, though I'm not sure I have an idea for something better. :-) Thoughts?

src/validator.cc
```diff
@@ -771,8 +775,12 @@ void Validator::CheckImport(const Location* loc, const Import* import) {
   }
 }
 
-void Validator::CheckExport(const Export* export_) {
+void Validator::CheckExport(const Location* loc, const Export* export_) {
   switch (export_->kind) {
+    case ExternalKind::Except:
+      // TODO(karlschimpf) Define.
+      PrintError(loc, "except except: don't know how to validate");
```
sbc100: "except export"?

src/wat-writer.cc
```diff
@@ -987,14 +987,20 @@ void WatWriter::WriteImport(const Import* import) {
...
   }
   WriteCloseNewline();
 }
 
 void WatWriter::WriteExport(const Export* export_) {
   if (options_->inline_export)
     return;
-  static const char* s_kind_names[] = {"func", "table", "memory", "global"};
+  static const char* s_kind_names[] = {"func", "table", "memory", "global",
+                                       "except"};
```
sbc100: Hmm.. this could use get_kind_name() instead?

src/validator.cc
KarlSchimpf: Fixed.

src/wast-parser.y
KarlSchimpf: Yes. Its the "(" shift/reduce conflict problem with the catches. Hence, I made an explicit explicit block for the code in the try block, but that introduced issues concerning the label.

In the parenthesis form, I also would like some type of node clustering the "try block body", similar to the "then" and the "else" in an "if". Hence, I chose a "block" node without a label. Only the try node gets a label.

src/wast-parser.y
KarlSchimpf: I agree that this is not clear. Renamed "except" to "exception_field" to clarify it is a (module) "field".

src/wat-writer.cc
KarlSchimpf: Good point. Fixing.

src/wast-parser.y
```diff
@@ -665,9 +665,9 @@ expr1 :
       assert(if_->type == ExprType::If);
       if_->if_.true_->label = $2;
     }
-  | try_check LPAR BLOCK labeling_opt block RPAR catch_list {
-      $5->label = $4;
+  | try_check labeling_opt LPAR BLOCK block RPAR catch_list {
```
binji: This still seems a bit strange to me. If we use the flat format there is an implicit block, but the folded format syntax requires a block here? Just parsing it, it looks as though there would be an additional block expression in the generated output.

KarlSchimpf: It is an implicit block (it doesn't build a block Expr). The block is just acting like a "then" for the "try".

Alternatively, I could make the block ALWAYS explicit. The problem is that I am trying to find a way to parse this to match the semantics of the proposal.

KarlSchimpf: Ok, was able to rewrite so no explicit (block ....) had to be added around the try body.

src/wast-parser.y
```diff
@@ -665,13 +666,41 @@ expr1 :
       assert(if_->type == ExprType::If);
       if_->if_.true_->label = $2;
     }
-  | try_check labeling_opt LPAR BLOCK block RPAR catch_list {
-      Expr* try_ = Expr::CreateTry($5, $7.first);
-      try_->try_block.label = $2;
-      $$ = join_exprs1(&@1, try_);
+  | try_check labeling_opt try_ {
```
KarlSchimpf: Here is the new s-expression form for a try block.

src/wast-parser.y
```diff
@@ -665,13 +666,41 @@ expr1 :
...
+    block_sig try_ {
+      $$ = $2;
+      Block* block = $$->try_block.block;
+      $$->try_block.block->sig.insert(block->sig.end(), $1->begin(), $1->end());
+      delete $1;
+    }
+  | try_inst_list
+  ;
+
+try_inst_list :
```
binji: nit: try_instr_list

src/wast-parser.y
```diff
@@ -665,13 +666,41 @@ expr1 :
...
+      block->label = $2;
+      $$ = join_exprs1(&@1, $3);
     }
   ;
 
+try_ :
+    block_sig try_ {
+      $$ = $2;
+      Block* block = $$->try_block.block;
+      $$->try_block.block->sig.insert(block->sig.end(), $1->begin(), $1->end());
```
binji: nit: Could use block here instead of `$$->try_block.block`

src/wast-parser.y
```diff
@@ -665,13 +666,41 @@ expr1 :
...
+      delete $1;
+    }
+  | try_inst_list
+  ;
+
+try_inst_list :
+      instr catch_list {
```
binji: This means `try_instr_list` cannot be empty, correct? Does that match the desired semantics?

KarlSchimpf: I believe so. Does it make sense to have an empty try block since there can't be a throw if empty?

binji: No, I suppose not. Though that seems to be an inconsistency with the flat format. But maybe we can just add a TODO and figure it out later.

src/wast-parser.y
KarlSchimpf: Simplified.

src/wast-parser.y
KarlSchimpf: Fixing.

src/wast-parser.y
KarlSchimpf: Good point. Modifying grammar to allow empty blocks.

PR #497, opened 2017-06-13
Make OptionParser into C++ class

binji:
Added a new feature where you can specify how many arguments are
allowed. It is an error to provide too many or too few.

Also add "help" tests for all of the other tools.

### DIFF
--- a/src/option-parser.cc
+++ b/src/option-parser.cc
@@ -28,19 +28,88 @@
 
 namespace wabt {
 
-static int option_match(const char* s,
-                        const char* full,
+OptionParser::Option::Option(char short_name,
+                             const std::string& long_name,
+                             const std::string& metavar,
+                             HasArgument has_argument,
+                             const std::string& help,
+                             const Callback& callback)
+    : short_name(short_name),
+      long_name(long_name),
+      metavar(metavar),
+      has_argument(has_argument),
+      help(help),
+      callback(callback) {}
+
+OptionParser::Argument::Argument(const std::string& name,
+                                 ArgumentCount count,
+                                 const Callback& callback)
+    : name(name), count(count), callback(callback) {}
+
+OptionParser::OptionParser(const char* program_name, const char* description)
+    : program_name_(program_name),
+      description_(description),
+      on_error_([this](const std::string& message) { DefaultError(message); }) {
+}
+
+void OptionParser::AddOption(const Option& option) {
+  options_.emplace_back(option);
+}
+
+void OptionParser::AddArgument(const std::string& name,
+                               ArgumentCount count,
+                               const Callback& callback) {
+  arguments_.emplace_back(name, count, callback);
+}
+
+void OptionParser::AddOption(char short_name,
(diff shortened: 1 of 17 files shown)


src/option-parser.cc
```diff
@@ -86,97 +172,115 @@ void parse_options(OptionParser* parser, int argc, char** argv) {
         }
 
         if (best_count > 1) {
-          error(parser, "ambiguous option \"%s\"", arg);
+          Errorf("ambiguous option \"%s\"", arg);
           continue;
         } else if (best_count == 0) {
-          error(parser, "unknown option \"%s\"", arg);
+          Errorf("unknown option \"%s\"", arg);
```
sbc100: Just single quotes here to avoid escaping?  This also happens to match how gnu tools format their option errors:

```ls --dsflgkj
ls: unrecognized option '--dsflgkj'
Try 'ls --help' for more information.
```

src/option-parser.h
```diff
@@ -17,47 +17,83 @@
...
 
 namespace wabt {
 
-enum class HasArgument {
-  No = 0,
-  Yes = 1,
-};
+class OptionParser {
+ public:
+  enum class HasArgument { No, Yes };
```
sbc100: If you keep the explicit = 0 and = 1 can you continue to write "if (has_argument)"?  Seem a bit clunky to have to explicity compare to something called 'Yes' for variables like has_argument which are clearly boolean already.

src/option-parser.h
```diff
@@ -17,47 +17,83 @@
...
+           const std::string& long_name,
+           const std::string& metavar,
+           HasArgument has_argument,
+           const std::string& help,
+           const Callback&);
+
+    char short_name;
+    std::string long_name;
+    std::string metavar;
+    HasArgument has_argument;
```
sbc100: Or maybe this member can just be a bool?

src/tools/wasm-link.cc
```diff
@@ -68,53 +53,26 @@ struct Context {
...
-      break;
-
-    case FLAG_RELOCATABLE:
-      s_relocatable = true;
-      break;
-
-    case FLAG_HELP:
-      print_help(parser, PROGRAM_NAME);
-      exit(0);
-      break;
```
sbc100: Call me old fashioned by I actually prefer this old syntax.. i find it less noisy.   Don't let that block this change though.

binji: You're right, it is a bit noisy. I was a bit concerned about that, so I tried to add helper functions to make the common cases nice. One additional thought I had was to have an explicit BoolOption/StringOption so you could just do:
```
parser.AddBoolOption('f', "fold-exprs", "...", &s_fold_exprs_value);
```

But I figure that can be layered on top of this change, which is more general.

src/option-parser.cc
binji: Done.

src/option-parser.h
binji: Done.

PR #500, opened 2017-06-15
Rewrite wabt::Opcode as C++ class

binji:
The class stores just one enum, so should be optimized to be passed as a
register. Since it is an object, the get_opcode_* accessors can now be
members:

    get_opcode_name(opcode) => opcode.GetName()

One side-effect of this change is that the opcode enumeration does not
match its binary encoding. This is a potential source of errors, but
will allows us to use a contiguous set of values for the enumeration, so
Opcode::GetInfo is an O(1) lookup. It will also be beneficial when
adding operators with prefixed encodings.


binji:
I was looking into what it would take to add the saturating float->int operators, and realized they can't be added without changing the way we handle operator encodings.

binji:
@KarlSchimpf ping

PR #503, opened 2017-06-16
Share float parse code literal.cc with templates

binji:
This will make it easier to update the parser to disallow out-of-range
hexfloat values (added recently to the spec parser).

### DIFF
--- a/src/literal.cc
+++ b/src/literal.cc
@@ -22,56 +22,14 @@
 #include <cstdlib>
 #include <cstring>
 
-#define HEX_DIGIT_BITS 4
-
-/* The PLUS_ONE values are used because normal IEEE floats have an implicit
- * leading one, so they have an additional bit of precision. */
-
-#define F32_SIGN_SHIFT 31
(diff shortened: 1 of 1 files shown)


binji:
BTW: I ran the hexfloat tests locally after this change. Unfortunately, these tests don't run on the bots because they're too slow.

src/literal.cc
```diff
@@ -22,56 +22,101 @@
...
+template <typename T>
+struct FloatTraitsBase {};
+
+// The "PlusOne" values are used because normal IEEE floats have an implicit
+// leading one, so they have an additional bit of precision.
+
+template <>
+struct FloatTraitsBase<float> {
+  typedef uint32_t Uint;
+  static constexpr int kBits = sizeof(Uint) * 8;
```
sbc100: 4?

sbc100: IUC ... makes sense

binji:
No missing preposition, but I could word it better.

PR #506, opened 2017-06-17
Modify lexer to look past parenthesis when needed for conflicts.

KarlSchimpf:
One of the problems with the wast parser is that the s-expression rules are difficult to write because one needs to look past the parenthesis (to the following opcode) to resolve conflicts.

Because of this, one needs to be very careful, and write right-recursive rules, to avoid conflicts in the generated LR parser. This not only complicates grammar updates, but forces the parse stack to be unbounded.

This PR focuses on fixing this problem, by allowing the generation of tokens of the form LPAR_NAME for tokens "LPAR "NAME". These tokens can be used in the grammar to disambiguate rules containing

   LPAR NAME

with

   LPAR_NAME LPAR NAME

to remove shift/reduce conflicts in the grammar.

This new lexing capability is used to simplify  the s-expression form of catch blocks.


src/circ_array.h
```diff
@@ -0,0 +1,115 @@
...
+
+namespace wabt {
+
+// TODO(karlschimpf) Complete the API
+// TODO(karlschimpf) Apply constructors/destructors on base type T
+//                   as collection size changes (if not POD).
+// TODO(karlschimpf) Deal with "front" overflow after adding max<ize_t>
+//                   elements.
+template<class T, const size_t kCapacity>
+class circ_array {
```
binji: I'd prefer CircularArray, personally. (Or maybe RingArray?)

src/circ_array.h
```diff
@@ -0,0 +1,115 @@
+/*
```
binji: Convention is to use hyphens for file names.

src/wast-lexer.cc
```diff
@@ -76,25 +126,35 @@
...
 namespace wabt {
 
+struct WastLexer::LexToken {
+  Location loc_;
+  int value_;
+  Token lval_;
+};
+  
+struct WastLexer::Lookahead {
+  circ_array<LexToken, 3> tokens_;
```
binji: Make this a power-of-two, so the modulo can turn into a bitwise and?

src/wast-lexer.cc
```diff
@@ -129,10 +193,15 @@ std::unique_ptr<WastLexer> WastLexer::CreateBufferLexer(const char* filename,
   return std::unique_ptr<WastLexer>(new WastLexer(std::move(source), filename));
 }
 
+bool WastLexer::lookahead_is_lpar() {
```
binji: nit: `IsLookaheadLpar`. I've been trying to use CamelCase for new functions (gonna switch it all over at some point...)

src/wast-lexer.h
```diff
@@ -60,10 +61,18 @@ class WastLexer {
   char* buffer_;
   size_t buffer_size_;
   char* marker_;
-  char* token_;
+  char* next_pos_;
   char* cursor_;
   char* limit_;
 
+  // The following defines a lookahead queue for inserting EXPECTS_PAREN_NAME
```
binji: This comment seems strange: what is EXPECTS_PAREN_NAME?

src/wast-lexer.cc
```diff
@@ -32,27 +32,77 @@
...
+#define ERROR(...)                                  \
+  YY_USER_ACTION(loc);                              \
   wast_parser_error(loc, this, parser, __VA_ARGS__)
 
 #define BEGIN(c) cond = (c)
 #define FILL(n)                                \
   do {                                         \
     if (WABT_FAILED(Fill(loc, parser, (n)))) { \
-      RETURN(EOF);                             \
+      int value = NAME_TO_VALUE(EOF);          \
```
binji: doesn't this do the same thing as RETURN(EOF)?

src/wast-lexer.cc
```diff
@@ -32,27 +32,77 @@
...
-    loc->last_column = COLUMN(cursor_); \
-  }
+#define PUSH_LOOKAHEAD_TOKEN              \
+  do {                                    \
+    WastLexer::LexToken tok;              \
+    lookahead_->tokens_.push_back(tok);   \
+    token_ = &lookahead_->tokens_.back(); \
+  } while (0)
+
+#define YY_USER_ACTION(loc)                \
```
binji: There's nothing magical about YY_USER_ACTION name anymore, probably better to give it a clearer name like SET_LOCATION or FILL_LOCATION or something.

src/wast-lexer.cc
```diff
@@ -32,27 +32,77 @@
...
 #define INITIAL_LEXER_BUFFER_SIZE (64 * 1024)
 
-#define YY_USER_ACTION                  \
-  {                                     \
-    loc->filename = filename_;          \
-    loc->line = line_;                  \
-    loc->first_column = COLUMN(token_); \
-    loc->last_column = COLUMN(cursor_); \
-  }
+#define PUSH_LOOKAHEAD_TOKEN              \
```
binji: As long as you're cleaning this stuff up... :-)

I wonder if some of these macros can be member functions of WastLexer instead? Or maybe even primarily functions with a small macro wrapper for name conversion. Wdyt?

src/wast-lexer.cc
```diff
@@ -107,12 +167,16 @@ WastLexer::WastLexer(std::unique_ptr<LexerSource> source, const char* filename)
       buffer_(nullptr),
       buffer_size_(0),
       marker_(nullptr),
-      token_(nullptr),
+      next_pos_(nullptr),
       cursor_(nullptr),
-      limit_(nullptr) {}
+      limit_(nullptr),
+      token_(nullptr) {
+  lookahead_ = new WastLexer::Lookahead();
```
binji: nit: May as well just move this into the initializer above.

src/wast-lexer.cc
KarlSchimpf: Unfortunately, a macro is "tokenized" first, and "#" concatenates tokens. Hence, when nested, it converts the token sequence to "RETURN "(EOF" ")". It took me a long time to realize that this was the main reason I was having trouble with many of the previous macros

binji: Ah, of course. Because `EOF` is already a macro that expands to `(-1)`. Maybe best to just change the name of the EOF token then.

binji:
Thanks, looks much nicer!

PR #512, opened 2017-06-19
Allow duplicate names in the name section

sbc100:
While this was previously allowed, it would result
in an invalid wast file since mutliple functions
would end up with the same name.

When generating labels we now add a unique numeric
suffix (.1, .2 etc) a la llvm-link.

We could add magic behaviour to strip these out
again when producing the name sectio in wast2wasm
which would allow for cleaner roundtrip, but might
cuase confusion/issues for those who might want
to legitametly name thier functions with such
suffixes.

### DIFF
--- a/src/binary-reader-ir.cc
+++ b/src/binary-reader-ir.cc
@@ -824,12 +824,10 @@ Result BinaryReaderIR::OnFunctionName(Index index, StringSlice name) {
 
   Func* func = module->funcs[index];
   std::string dollar_name = std::string("$") + string_slice_to_string(name);
-  if (module->func_bindings.count(dollar_name) != 0) {
-    std::string orig_name = dollar_name;
-    int counter = 1;
-    do {
-      dollar_name = orig_name + "." + std::to_string(counter++);
-    } while (module->func_bindings.count(dollar_name) != 0);
+  int counter = 1;
+  std::string orig_name = dollar_name;
+  while (module->func_bindings.count(dollar_name) != 0) {
+    dollar_name = orig_name + "." + std::to_string(counter++);
   }
   func->name = dup_string_slice(string_to_string_slice(dollar_name));
   module->func_bindings.emplace(dollar_name, Binding(index));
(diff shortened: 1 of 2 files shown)


src/binary-reader-ir.cc
```diff
@@ -824,6 +824,13 @@ Result BinaryReaderIR::OnFunctionName(Index index, StringSlice name) {
 
   Func* func = module->funcs[index];
   std::string dollar_name = std::string("$") + string_slice_to_string(name);
+  if (module->func_bindings.count(dollar_name) != 0) {
```
binji: Can't you turn this into a while loop?

src/binary-reader-ir.cc
```diff
@@ -824,6 +824,13 @@ Result BinaryReaderIR::OnFunctionName(Index index, StringSlice name) {
...
+    std::string orig_name = dollar_name;
+    int counter = 1;
+    do {
+      dollar_name = orig_name + "." + std::to_string(counter++);
+    } while (module->func_bindings.count(dollar_name) != 0);
+  }
   func->name = dup_string_slice(string_to_string_slice(dollar_name));
   module->func_bindings.emplace(dollar_name, Binding(index));
   return Result::Ok;
```
binji: What about duplicated param/local names?

sbc100: That is not possible.  Local names need to be unique.

src/binary-reader-ir.cc
sbc100: Done

src/binary-reader-ir.cc
binji: In C object files or in the wasm format? I don't see why locals/params should be different from function names for the wasm format.

sbc100: We are allowing duplicate names since there is a real world screnario where this is possible (multiple compilations units have the same local function names).  I don't know of any place where allowing params with the same name would be useful/desirable.

binji: Fair, but if that's the case we should specify that in the names section.

sbc100: You mean in Design.md explicitly say that duplicate local names are not allowed?  Yes I guess so.  Also we should/could assert there in the code too.

PR #514, opened 2017-06-20
Fix the validator to be able to validate exception handling constructs.

KarlSchimpf:
Fixes the validator to be able to validate exception handling constructs.

Also updates:

    the expr-visitor to be able to handle exception constructs.
    the name resolver to be able to handle exception constructs.
    the type checker to be able to handle exception constructs.

Also updates wast-desugar to be able to turn on debugging of the wast parser.


src/resolve-names.cc
```diff
@@ -209,6 +225,7 @@ Result NameResolver::EndBlockExpr(Expr* expr) {
   return Result::Ok;
 }
 
+
```
KarlSchimpf: Remove this blank line.

src/common.cc
```diff
@@ -31,6 +31,8 @@
 
 namespace wabt {
 
+CommonClOptionsStruct CommonClOptions;
```
sbc100: Does CL hear mean "command line", in which case I think CLI is more common.

src/common.h
```diff
@@ -97,6 +97,13 @@ static const Address kInvalidAddress = ~0;
 static const Index kInvalidIndex = ~0;
 static const Offset kInvalidOffset = ~0;
 
+/* Common CL options. */
+struct CommonClOptionsStruct {
+  bool allow_exceptions = false;
+};
+
+extern CommonClOptionsStruct CommonClOptions;
```
sbc100: Using a global here seems a little wrong....  is there some better way?

src/validator.cc
```diff
@@ -959,6 +1023,42 @@ Result Validator::CheckGet(const Action* action, Type* out_type) {
...
+    *out_except = current_module_->excepts[index];
+  return Result::Ok;
+}
+
+void Validator::CheckExcept(const Location* loc, const Exception* except) {
+  for (Type ty : except->sig) {
+    switch (ty) {
+      default:
+        PrintError(loc, "Invalid exception type: %s", get_type_name(ty));
+        break;
```
sbc100: I'm used to seeing default that the end of the switch.  Is this a style used elsewhere in wabt?

src/validator.cc
```diff
@@ -142,10 +152,12 @@ class Validator {
   Index current_memory_index_ = 0;
   Index current_global_index_ = 0;
   Index num_imported_globals_ = 0;
+  Index current_except_index = 0;
```
KarlSchimpf: Fixed. Also added _ after try_contexts.

src/common.cc
KarlSchimpf: Removed, no longer needed.

src/common.h
KarlSchimpf: Removed. No longer needed.

KarlSchimpf: Removed, so no longer an issue.

src/expr-visitor.cc
```diff
@@ -61,14 +61,24 @@ Result ExprVisitor::VisitExpr(Expr* expr) {
       break;
 
     case ExprType::Catch:
-      // TODO(karlschimpf): Define
-      WABT_FATAL("Catch: don't know how to visit\n");
-      return Result::Error;
+      if (!CommonClOptions.allow_exceptions) {
```
KarlSchimpf: I changed the philosophy to assume that the reader/parser checks if these constructs are allowed, and removed the checks in the code.

I also reorganized the model for catches, and they are no longer an Expr.

The only two conditions that the validator now needs to check are:
1) Are all rethrow instructions in a catch block.
2) The catch all (if it appears) must be the last catch.

I also wrapped catch and catch_all into a single construct (with an invalid index representing a catch all).

src/resolve-names.cc
```diff
@@ -281,6 +298,67 @@ Result NameResolver::OnTeeLocalExpr(Expr* expr) {
...
+}
+
+Result NameResolver::EndTryBlockExpr(Expr* expr) {
+  PopLabel();
+  return Result::Ok;
+}
+
+Result NameResolver::BeginCatchExpr(Expr* expr) {
+  if (try_blocks.empty()) {
+    PrintError(&expr->loc, "Catch not in try block");
```
KarlSchimpf: The problem was I wasn't assuming that you could count of valid tree structure (just node construction). I realize now that I should have assumed the reader/parser are the only ones that need to check this.

src/type-checker.cc
```diff
@@ -344,6 +344,24 @@ Result TypeChecker::OnCompare(Opcode opcode) {
...
+  return Result::Ok;
+}
+
+Result TypeChecker::OnCatchBlock(const TypeVector* sig) {
+  Result result = Result::Ok;
+  Label* label;
+  CHECK_RESULT(TopLabel(&label));
+  COMBINE_RESULT(result, CheckLabelType(label, LabelType::Try));
+  COMBINE_RESULT(result, PopAndCheckSignature(label->sig, "try block"));
+  COMBINE_RESULT(result, CheckTypeStackEnd("if true branch"));
```
KarlSchimpf: done

src/common.h
KarlSchimpf: REmoved, no longer needed.

src/validator.cc
KarlSchimpf: Default moved to the end.

KarlSchimpf:
This PR is ready to reviewed again.

Note: The bots are failing due to "memory" leakage of 4 bytes when there is an error on the "throw" or the "rethrow" Expr. I haven't figured out the cause yet.

src/ir.h
```diff
@@ -130,6 +128,22 @@ struct Block {
   Expr* first;
 };
 
+struct Catch {
+  WABT_DISALLOW_COPY_AND_ASSIGN(Catch);
+  Catch() = delete;
```
binji: Doesn't hurt, but isn't this implicitly deleted because of the constructors below?

src/ir.h
```diff
@@ -130,6 +128,22 @@ struct Block {
...
+  explicit Catch(Expr* first);
+  Catch(Var var, Expr* first);
+  ~Catch();
+  Location loc;
+  Var var;
+  struct Expr* first;
+  bool IsCatchAll() const {
```
binji: What's the benefit of distinguishing catch vs. catch_all here rather than in the try block?

src/resolve-names.cc
```diff
@@ -84,6 +90,9 @@ class NameResolver : public ExprVisitor::DelegateNop {
   Func* current_func_ = nullptr;
   ExprVisitor visitor_;
   std::vector<Label*> labels_;
+#if 0
```
binji: Unused?

src/validator.cc
```diff
@@ -569,14 +570,42 @@ void Validator::CheckExpr(const Expr* expr) {
...
+      if (WABT_SUCCEEDED(CheckExceptVar(&expr->throw_.var, &except))) {
+        typechecker_.OnThrow(&except->sig);
+      }
       break;
 
-    case ExprType::TryBlock:
-      // TODO(karlschimpf) Define.
-      PrintError(&expr->loc, "TryBlock: don't know how to validate");
+    case ExprType::TryBlock: {
+      TryContext context;
```
binji: Ah, this is much clearer, I think. Thanks!

I think we'll eventually want to move more of this logic into TypeChecker, but this looks good for now.

test/exceptions/rethrow-to-wrong-block.txt
```diff
@@ -0,0 +1,20 @@
...
+        (i32.const 7)
+        (catch_all
+          (rethrow $b)
+        )
+      )
+    )
+  )
+) 
+(;; STDERR ;;;
+out/test/exceptions/rethrow-to-wrong-block.txt:10:12: invalid rethrow depth: 1 (max 2)
```
binji: Hm, this error is a bit strange. OK for now, but we'll probably want to use the maximum rethrow depth instead of the maximum label depth.

binji:
It looks like the memory leak comes from the `name` field of `Exception`. It needs to be destroyed manually (still working on switching over to using `string_view` and `std::string`...)

KarlSchimpf:
The problem is the name field, but it is because struct VAR, which has VERY SUBTLE constructor/assignment versions (i.e. var(Var&&) and var(const Var&), operator=(Var&&) and operator=(const Var&).

In particular, the (Var&&) version doesn't create a copy but the (const Var&) does.

I can't tell when/why each version is used (by reading the code), so I'm getting unexpected results in one context.

binji:
Yes, unfortunately `Var` needs to have a move constructor/assignment operator so it can properly handle the union members. It would be much nicer with `std::variant`...

But I don't see what this has to do with the memory leak.

KarlSchimpf:
​​
​It happens when I call Expr::CreateThrow() and Expr::CreateRethrow() in
wast-parser.y

I haven't yet figured out whether it is due to the parameter being passed
in, or the assignment in the body of the form:

   expr->(re)throw_.var = var;

One of these two contexts is creating a duplicate string, that doesn't
happen with similar code such as Expr::CreateBr().



   ​

Karl Schimpf

binji:
> It happens when I call Expr::CreateThrow() and Expr::CreateRethrow() in
wast-parser.y

I don't think that's correct -- I pulled your change -- it seems that you just need to add:
```
Exception::~Exception() { destroy_string_slice(&name); }
```
That fixed the leak when I tried it.

KarlSchimpf:
binji - Thanks for finding the memory leak!

PR #515, opened 2017-06-21
Update the testsuite; fix literal parsing issues

binji:
* Return error when floats overflow (used to return inf)
* Significantly simplfy the significand parser in FloatParser::ParseHex
* Add some new parse tests to hexfloat.cc; the previous tests only
  checked valid/canonical strings

### DIFF
--- a/src/literal.cc
+++ b/src/literal.cc
@@ -27,11 +27,11 @@ namespace wabt {
 namespace {
 
 int Clz(uint32_t value) {
-  return value != 0 ? wabt_clz_u32(value) : 32;
+  return value == 0 ? 32 : wabt_clz_u32(value);
 }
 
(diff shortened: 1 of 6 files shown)


binji:
@sbc100 ping

src/literal.cc
```diff
@@ -27,11 +27,11 @@ namespace wabt {
 namespace {
 
 int Clz(uint32_t value) {
-  return wabt_clz_u32(value);
+  return value != 0 ? wabt_clz_u32(value) : 32;
```
sbc100: Do you think its a little easer to read the other way around: `value == 0 ? 32 : wabt_clz_u32(value);`

binji: Done.

PR #516, opened 2017-06-21
Add validation to wasm2wast

binji:
* Modify Location so it can reference a binary file offset
* Set node locations in BinaryReaderIR
* Modify SourceErrorHandler to print text or binary locations

### DIFF
--- a/src/binary-reader-ir.cc
+++ b/src/binary-reader-ir.cc
@@ -51,7 +51,9 @@ LabelNode::LabelNode(LabelType label_type, Expr** first)
 
 class BinaryReaderIR : public BinaryReaderNop {
  public:
-  BinaryReaderIR(Module* out_module, BinaryErrorHandler* error_handler);
+  BinaryReaderIR(Module* out_module,
+                 const char* filename,
+                 BinaryErrorHandler* error_handler);
 
   bool OnError(const char* message) override;
 
@@ -191,6 +193,7 @@ class BinaryReaderIR : public BinaryReaderNop {
 
  private:
   bool HandleError(Offset offset, const char* message);
+  Location GetLocation() const;
   void PrintError(const char* format, ...);
   void PushLabel(LabelType label_type, Expr** first);
   Result PopLabel();
@@ -204,11 +207,20 @@ class BinaryReaderIR : public BinaryReaderNop {
   Func* current_func = nullptr;
   std::vector<LabelNode> label_stack;
   Expr** current_init_expr = nullptr;
+  const char* filename_;
 };
 
 BinaryReaderIR::BinaryReaderIR(Module* out_module,
-                                 BinaryErrorHandler* error_handler)
-    : error_handler(error_handler), module(out_module) {}
+                               const char* filename,
+                               BinaryErrorHandler* error_handler)
+    : error_handler(error_handler), module(out_module), filename_(filename) {}
+
+Location BinaryReaderIR::GetLocation() const {
+  Location loc;
+  loc.filename = filename_;
+  loc.offset = state->offset;
+  return loc;
+}
 
 void WABT_PRINTF_FORMAT(2, 3) BinaryReaderIR::PrintError(const char* format,
                                                           ...) {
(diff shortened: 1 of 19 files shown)


binji:
Sorry, do you mean I should clarify the PR description?

KarlSchimpf:
Sorry, it appears that GitHub lost my code comments!  Will try to add them again.

src/binary-reader-ir.cc
```diff
@@ -843,28 +863,38 @@ Result BinaryReaderIR::OnLocalNameLocalCount(Index index, Index count) {
 }
 
 Result BinaryReaderIR::OnInitExprF32ConstExpr(Index index, uint32_t value) {
-  *current_init_expr = Expr::CreateConst(Const(Const::F32(), value));
+  Expr* expr = Expr::CreateConst(Const(Const::F32(), value, GetLocation()));
+  expr->loc = GetLocation();
```
KarlSchimpf: Why pass in location to CreateConst and then assign after?

src/binary-reader-ir.cc
```diff
@@ -843,28 +863,38 @@ Result BinaryReaderIR::OnLocalNameLocalCount(Index index, Index count) {
...
+  *current_init_expr = expr;
   return Result::Ok;
 }
 
 Result BinaryReaderIR::OnInitExprF64ConstExpr(Index index, uint64_t value) {
-  *current_init_expr = Expr::CreateConst(Const(Const::F64(), value));
+  Expr* expr = Expr::CreateConst(Const(Const::F64(), value, GetLocation()));
+  expr->loc = GetLocation();
```
KarlSchimpf: Why pass in location to CreateConst() and then assign after.?

src/binary-reader-ir.cc
```diff
@@ -843,28 +863,38 @@ Result BinaryReaderIR::OnLocalNameLocalCount(Index index, Index count) {
...
+  *current_init_expr = expr;
   return Result::Ok;
 }
 
 Result BinaryReaderIR::OnInitExprGetGlobalExpr(Index index,
                                                Index global_index) {
-  *current_init_expr = Expr::CreateGetGlobal(Var(global_index));
+  Expr* expr = Expr::CreateGetGlobal(Var(global_index, GetLocation()));
+  expr->loc = GetLocation();
```
KarlSchimpf: Why pass location to CreateGetGlobal() and then assign after?

src/binary-reader-ir.cc
```diff
@@ -843,28 +863,38 @@ Result BinaryReaderIR::OnLocalNameLocalCount(Index index, Index count) {
...
+  *current_init_expr = expr;
   return Result::Ok;
 }
 
 Result BinaryReaderIR::OnInitExprI32ConstExpr(Index index, uint32_t value) {
-  *current_init_expr = Expr::CreateConst(Const(Const::I32(), value));
+  Expr* expr = Expr::CreateConst(Const(Const::I32(), value, GetLocation()));
+  expr->loc = GetLocation();
```
KarlSchimpf: Same here.

src/binary-reader-ir.cc
```diff
@@ -843,28 +863,38 @@ Result BinaryReaderIR::OnLocalNameLocalCount(Index index, Index count) {
...
+  *current_init_expr = expr;
   return Result::Ok;
 }
 
 Result BinaryReaderIR::OnInitExprI64ConstExpr(Index index, uint64_t value) {
-  *current_init_expr = Expr::CreateConst(Const(Const::I64(), value));
+  Expr* expr = Expr::CreateConst(Const(Const::I64(), value, GetLocation()));
+  expr->loc = GetLocation();
```
KarlSchimpf: Same here.

src/binary-reader-ir.cc
binji: This one assigns the Expr::loc, the other one assigns the Const::loc. In the text format these can be different. But honestly, it would be nicer to pass the location in to `Expr::CreateConst` directly, I just can't do it with a default parameter until I add a `Location` default constructor. This is currently hard to do because `Location` is used in places where it assumes it can have a trivial constructor. I could fix that here, but I figured it would be better to do in another PR.

PR #517, opened 2017-06-22
Add color output in SourceErrorHandler

binji:


binji:
Good idea, I was looking at [chalk](https://github.com/chalk/supports-color/blob/master/index.js) to see how they check for a color terminal. Seems pretty complex, so I wanted to punt until we have users who can test them. They seem to have added a `FORCE_COLOR` environment variable as a top-level override, so maybe we should just follow that.

sbc100:
yuk:
```if ('TRAVIS' in env || env.CI === 'Travis') {
			return 1;
```

yes an override env var sounds good.  Happy to be in a followup CL

binji:
Looks like `ENABLE_VIRTUAL_TERMINAL_PROCESSING` is only supported in windows 10, so we should fall back to `SetConsoleTextAttribute` like binaryen does.

binji:
OK, for now it's win10 only, and color output is disabled on Windows otherwise. ptal

binji:
Yeah, I'll add a test.

binji:
Also added ability to set environment variables in test, PTAL

PR #518, opened 2017-06-22
Update the binary writer to handle try/catch/throw/rethrow instructions.

KarlSchimpf:
Defines how to write the contents of exception handling instructions.

Note: It doesn't yet handle exception declarations.


binji:
Ah, I was gonna suggest you write a test -- but it looks like the tests you have are now producing output! (that's what's failing on the bots)

PR #520, opened 2017-06-23
Improve display of data segments

sbc100:
Use the virtual address (init offset) when display
the data rather that the file offset.

Show the segment size in the header.


src/binary-reader-objdump.cc
```diff
@@ -839,17 +944,56 @@ Result BinaryReaderObjdump::OnDataSegmentCount(Index count) {
...
+  switch (data_init_expr_.type) {
+    case InitExprType::I32:
+      voffset = data_init_expr_.value.i32;
+      break;
+    case InitExprType::I64:
+      voffset = data_init_expr_.value.i64;
+      break;
+    case InitExprType::F32:
+    case InitExprType::F64:
+      fprintf(stderr, "Floating point value not valid for data segement offset");
```
binji: sp: segment

PR #521, opened 2017-06-23
Consistent comments when ending anonymous namespace

sbc100:
This matches the wording prefered by llvm.

### DIFF
--- a/src/wat-writer.cc
+++ b/src/wat-writer.cc
@@ -1122,7 +1122,7 @@ void WatWriter::WriteInlineExport(const Export* export_) {
   }
 }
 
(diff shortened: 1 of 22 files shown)


src/wat-writer.cc
```diff
@@ -1122,7 +1122,7 @@ void WatWriter::WriteInlineExport(const Export* export_) {
   }
 }
 
-}  // namespace
+}  // end namespace anonymous
```
binji: end anonymous namespace

sbc100: I put that one in there as a test for the reviewer.

PR #525, opened 2017-06-23
Add type check to data segment offset

sbc100:
This is a conservative check that we can do in the binary
reader itself.  More extensive checking is still done in the
interpreter (i.e. vefiying the type of the global).

### DIFF
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -176,7 +176,7 @@ class BinaryReader {
   Index NumTotalMemories();
   Index NumTotalGlobals();
 
-  Result ReadInitExpr(Index index) WABT_WARN_UNUSED;
+  Result ReadInitExpr(Index index, bool require_i32 = false) WABT_WARN_UNUSED;
   Result ReadTable(Type* out_elem_type,
                    Limits* out_elem_limits) WABT_WARN_UNUSED;
   Result ReadMemory(Limits* out_page_limits) WABT_WARN_UNUSED;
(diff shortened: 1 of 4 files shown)


src/binary-reader.cc
```diff
@@ -1523,7 +1530,7 @@ Result BinaryReader::ReadDataSection(Offset section_size) {
     CHECK_RESULT(ReadIndex(&memory_index, "data segment memory index"));
     CALLBACK(BeginDataSegment, i, memory_index);
     CALLBACK(BeginDataSegmentInitExpr, i);
-    CHECK_RESULT(ReadInitExpr(i));
+    CHECK_RESULT(ReadInitExpr(i, true/*require_i32*/));
```
binji: I assume this is true for elem_segments too?

src/binary-reader.cc
```diff
@@ -176,7 +176,7 @@ class BinaryReader {
   Index NumTotalMemories();
   Index NumTotalGlobals();
 
-  Result ReadInitExpr(Index index) WABT_WARN_UNUSED;
+  Result ReadInitExpr(Index index, bool require_i32 = false) WABT_WARN_UNUSED;
```
binji: Might be nice to have a passthrough function called `ReadI32InitExpr` that sets this to `true`, especially if we add the same thing for elem segments too.

sbc100: Done.  PTAL

PR #526, opened 2017-06-23
Fix binary writer to handle exception declarations.

KarlSchimpf:
Generate exception section, imported exceptions, and exported exceptions. At this point, we can successfully convert wast files with exceptions to their corresponding wasm files.

Also updates exceptions to reflect the changes.

### DIFF
--- a/src/binary-writer.cc
+++ b/src/binary-writer.cc
@@ -224,6 +224,7 @@ class BinaryWriter {
   void WriteTable(const Table* table);
   void WriteMemory(const Memory* memory);
   void WriteGlobalHeader(const Global* global);
+  void WriteExceptType(const TypeVector* except_types);
   void WriteRelocSection(const RelocSection* reloc_section);
 
   Stream stream_;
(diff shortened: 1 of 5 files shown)


src/binary-writer.cc
```diff
@@ -1002,8 +1009,12 @@ Result BinaryWriter::WriteModule(const Module* module) {
   }
 
   if (module->excepts.size()) {
-    // TODO(karlschimpf) Define.
-    WABT_FATAL("write exception section not implemented");
+    BeginCustomSection("exception", LEB_SECTION_SIZE_GUESS);
+    write_u32_leb128(&stream_, module->excepts.size(), "exception count");
```
sbc100: Is exception not going to be a buildin section?  (sorry I should really just read the spec..)

KarlSchimpf: The spec at the moment does not introduce a new builtin section, and there were people who fought back against making it a builtin section.

However, since we are still a long way away from getting the proposal approved, it may change to a builtin section in the future.

PR #527, opened 2017-06-23
Use add add_custom_command rather than add_custom_target

sbc100:
add_custom_target means that build is alwasy dirty and will
always execute the copy commands.  add_custom_command means
the copy will only be trigged if the binaries get rebuilt.

This means the ninja build will report nothing to do if
it is re-run without changing any inputs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -283,10 +283,9 @@ if (NOT EMSCRIPTEN)
     list(APPEND WABT_EXECUTABLES ${name})
     set(WABT_EXECUTABLES ${WABT_EXECUTABLES} PARENT_SCOPE)
 
-    add_custom_target(${name}-copy-to-bin ALL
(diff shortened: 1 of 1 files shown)


sbc100:
The downside is that the contents of `bin` is not always what you expect.. because if you do ninja -C dir1.. then ninja -C dir2.. the second ninja command might not do anything and the binaries from dir1 might still be in `bin`...

But its a little annoying to have a ninja build that always build stuff.

binji:
Yeah, that's why I did it this way. Because sometimes I switch between a release build and a debug build, but I always want bin/ to be correct.

binji:
I guess I would be OK with this if there were some way to force copy the files to bin. Maybe in the Makefile? That way I could just run that target instead and always get the behavior I want.

binji:
@sbc100 any thoughts on this or should we close?

PR #529, opened 2017-06-24
Make wast-desugar work for exception handling.

KarlSchimpf:
Updates generate/apply names, as well as the wat writer, allowing wast-desugar to work.

Also fixes the following nits:

* Fixes NameApplier::UseNameForVar() to not leak a string slice (i.e. don't override name if
  it exists).

* Removes unnecessary "Module" arguments in class NameApplier (it is already a field).

### DIFF
--- a/src/wat-writer.cc
+++ b/src/wat-writer.cc
@@ -1165,6 +1165,7 @@ void WatWriter::BuildExportMaps() {
       }
 
       case ExternalKind::Except:
+        // TODO(karlschimpf): Build for inline exceptions.
         break;
     }
   }
--- a/src/wat-writer.cc
+++ b/src/wat-writer.cc
@@ -133,6 +133,8 @@ class WatWriter {
   void WriteFunc(const Module* module, const Func* func);
   void WriteBeginGlobal(const Global* global);
   void WriteGlobal(const Global* global);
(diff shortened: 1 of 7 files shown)


src/wat-writer.cc
```diff
@@ -1122,7 +1165,6 @@ void WatWriter::BuildExportMaps() {
       }
 
       case ExternalKind::Except:
-        WABT_FATAL("BuildExportMaps(except) not implemented");
```
binji: This will be needed when you add inline exports for exceptions, I think. May be good to add a TODO

test/desugar/try.txt
```diff
@@ -0,0 +1,29 @@
+;;; TOOL: wast-desugar
```
binji: What is this test demonstrating? I guess just that you can roundtrip the try format? Might be good to add some tests that show that folding works too.

KarlSchimpf: I mainly did it to show that the wat-writer is working. I didn't even notice it had a folding version. I will add this as a separate PR.

src/wat-writer.cc
KarlSchimpf: Done.

PR #530, opened 2017-06-24
Fix ununitailzed value_ in LexToken

winksaville:
This fixes an uninitialized warning/error when compiling with c++ 7.1.1.

### DIFF
--- a/src/wast-lexer.cc
+++ b/src/wast-lexer.cc
@@ -98,6 +98,8 @@ struct WastLexer::LexToken {
   Location loc_;
   int value_;
   Token lval_;
+
+  LexToken(void): value_(0) {}
 };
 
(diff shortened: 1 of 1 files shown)


winksaville:
@KarlSchimpf can you pull it, I can't.

KarlSchimpf:
Trying now.

KarlSchimpf:
Committed.

sbc100:
What is c++ 7.1.1 ?  Is that a gcc version ?

Would using "= 0" in the declaration have the same effect?  I wonder if we should prefer that in general (especially when the constructor body is empty)?

binji:
Yeah, I'd like to move toward using those more often, since I find them to be clearer generally.

PR #531, opened 2017-06-25
Fix the wat writer to be able to fold exception handling constructs.

KarlSchimpf:
Implements expression folding for exception handling constructs in the wat writer.

Also adds the following for debugging:

1) Names to  the enum ExprType.

2) Adds (compile-time) conditional debugging trace code to wat-writer.cc, making it easier to see how the IR is being translated.

### DIFF
--- a/src/ir.cc
+++ b/src/ir.cc
@@ -21,7 +21,7 @@
 
 namespace {
 
-const char* ExprTypeName[size_t(wabt::ExprType::Unreachable) + 1] = {
+const char* ExprTypeName[] = {
   "Binary",
   "Block",
   "Br",
@@ -59,8 +59,7 @@ const char* ExprTypeName[size_t(wabt::ExprType::Unreachable) + 1] = {
 namespace wabt {
 
 const char* GetExprTypeName(ExprType type) {
-  static_assert(size_t(ExprType::Unreachable) + 1
-                == WABT_ARRAY_SIZE(ExprTypeName),
+  static_assert(WABT_ENUM_COUNT(ExprType) == WABT_ARRAY_SIZE(ExprTypeName),
                 "Malformed ExprTypeName array");
   return ExprTypeName[size_t(type)];
 }
--- a/src/ir.cc
+++ b/src/ir.cc
@@ -176,26 +176,21 @@ void MakeTypeBindingReverseMapping(
   }
 }
 
-ModuleField* Module::AppendField() {
-  ModuleField* result = new ModuleField();
+void Module::AppendField(ModuleField* field) {
   if (!first_field)
-    first_field = result;
+    first_field = field;
   else if (last_field)
-    last_field->next = result;
-  last_field = result;
-  return result;
+    last_field->next = field;
+  last_field = field;
 }
 
 FuncType* Module::AppendImplicitFuncType(const Location& loc,
                                          const FuncSignature& sig) {
-  ModuleField* field = AppendField();
-  field->loc = loc;
-  field->type = ModuleFieldType::FuncType;
-  field->func_type = new FuncType();
-  field->func_type->sig = sig;
-
-  func_types.push_back(field->func_type);
-  return field->func_type;
+  FuncType* func_type = new FuncType();
+  func_type->sig = sig;
+  func_types.push_back(func_type);
+  AppendField(new FuncTypeModuleField(func_type, loc));
+  return func_type;
 }
 
 void DestroyExprList(Expr* first) {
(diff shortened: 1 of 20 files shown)


src/wat-writer.cc
```diff
@@ -464,6 +490,8 @@ void WatWriter::WriteConst(const Const* const_) {
 }
 
 void WatWriter::WriteExpr(const Expr* expr) {
+  if (s_trace)
```
binji: I'm not a huge fan of keeping debug tracing in, since I find it to clutter up the code. I'd be OK with it though if we could minimize a bit... something like this maybe?
```
TRACE_ENTER("(%s)", GetExprTypeName(*expr))
TRACE_EXIT(...)
```

I think you could write the macro like:
```
#define TRACE_ENTER(fmt, ...) \
  if (s_trace) fprintf(stderr, "-> %s" fmt "\n", __func__, __VA_ARGS__)
```

src/wat-writer.cc
```diff
@@ -807,9 +871,46 @@ void WatWriter::PushExpr(const Expr* expr,
...
+    fprintf(stderr, "-> PushExpr(%s, %" PRIindex ", %" PRIindex")\n",
+            GetExprTypeName(*expr), operand_count, result_count);
+  }
+}
+
+void WatWriter::PushCatch(const Catch* catch_) {
+  if (s_trace)
+    fprintf(stderr, "-> PushCatch()\n");
+  ExprTree tree(catch_);
+  expr_tree_stack_.push_back(std::move(tree));
```
binji: this could just be `emplace_back(catch_)` I think.

src/wat-writer.cc
```diff
@@ -858,11 +972,20 @@ void WatWriter::FlushExprTree(const ExprTree& expr_tree) {
...
     }
   }
+  if (s_trace)
+    fprintf(stderr, "<- FlushExprTree(%s)\n", GetExprTypeName(*expr_tree.expr));
 }
 
 void WatWriter::FlushExprTreeVector(const std::vector<ExprTree>& expr_trees) {
+  if (s_trace)
+    fprintf(stderr, "-> FlushExprTreeVector(%zu)\n:", expr_trees.size());
   for (auto expr_tree : expr_trees)
```
binji: nit: use braces for blocks that have more than one line.

src/wat-writer.cc
```diff
@@ -807,9 +871,46 @@ void WatWriter::PushExpr(const Expr* expr,
...
+  if (s_trace)
+    fprintf(stderr, "-> PushCatch()\n");
+  ExprTree tree(catch_);
+  expr_tree_stack_.push_back(std::move(tree));
+  FlushExprTreeStack();
+  if (s_trace)
+    fprintf(stderr, "<- PushCatch()\n");
+}
+
+void WatWriter::FlushCatchTree(const ExprTree& expr_tree) {
```
binji: I don't understand why catches need to be pushed on the stack. It seems like all of this code could be inlined directly into `FlushExprTree` below since you always immediately flush the stack after you push a new catch on.

src/wat-writer.cc
```diff
@@ -807,9 +871,46 @@ void WatWriter::PushExpr(const Expr* expr,
...
+  if (s_trace)
+    fprintf(stderr, "-> FlushCatchTree(%s)\n", expr_tree.describe().c_str());
+  assert(expr_tree.IsCatch());
+  WritePuts("(", NextChar::None);
+  if (expr_tree.catch_->IsCatchAll()) {
+    WritePutsNewline(Opcode::CatchAll_Opcode.GetName());
+  } else {
+    WritePutsSpace(Opcode::Catch_Opcode.GetName());
+    WriteVar(&expr_tree.catch_->var, NextChar::Space);
```
binji: Probably should write a newline after this, since it will be followed by an expr list.

src/wat-writer.cc
KarlSchimpf: Now you know why the tracing. It was not very obvious to me (until I had some tracing), what was even going on. The code of this PR is quite different than initially written, and only got simplified once I saw the picture of what the code was doing. Will simplify.

binji: That's fair... :-) Probably could use a comment or two explaining what's going on too. I'll see about adding something like that... the basic idea is relatively straightforward (I think!)

src/ir.cc
```diff
@@ -19,8 +19,56 @@
 #include <cassert>
 #include <cstddef>
 
+namespace {
+
+const char* ExprTypeName[size_t(wabt::ExprType::Unreachable) + 1] = {
```
KarlSchimpf: Done.

src/ir.h
```diff
@@ -119,9 +119,11 @@ enum class ExprType {
   Throw,
   TryBlock,
   Unary,
-  Unreachable,
+  Unreachable,  // Unreachable must appear last
```
KarlSchimpf: Done

src/wat-writer.cc
KarlSchimpf: Ok, I've simplified the tracing so that it is minimized.

I created files tracing.{h,cc} that implement a simple (scoped) class that generates enter/exit messages automatically. I then added two simple macros WABT_TRACE(method_name) and WABT_TRACE_ARGS(method_name, format, ...) to insert an instance of this (scoped) tracing class.

src/wat-writer.cc
KarlSchimpf: I was just following what PushExpr() did. Changing all occurrences.

src/wat-writer.cc
KarlSchimpf: Done.

src/wat-writer.cc
KarlSchimpf: Done.

binji:
I'm finding this a bit difficult to review because it has my merged changes included (the cast stuff). Do you think you could rebase instead?

KarlSchimpf:
Could not rebase in same branch. So I created a new PR #533 to continue this PR.

binji:
Oh, OK. Sorry about that!

PR #532, opened 2017-06-25
Use type hierarchy for ModuleField instead of union

binji:
Also added LLVM-style casting free functions:

- isa<T>: returns true if the object is of type T
- cast<T>: converts object to T* or asserts
- dyn_cast<T>: converts object to T* or returns nullptr

### DIFF
--- a/src/binary-reader-ir.cc
+++ b/src/binary-reader-ir.cc
@@ -25,6 +25,7 @@
 
 #include "binary-error-handler.h"
 #include "binary-reader-nop.h"
+#include "cast.h"
 #include "common.h"
 #include "ir.h"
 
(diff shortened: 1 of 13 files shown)


src/binary-reader-ir.cc
```diff
@@ -435,12 +422,10 @@ Result BinaryReaderIR::OnMemoryCount(Index count) {
...
 
 Result BinaryReaderIR::OnMemory(Index index, const Limits* page_limits) {
-  ModuleField* field = module->AppendField();
-  field->loc = GetLocation();
-  field->type = ModuleFieldType::Memory;
-  field->memory = new Memory();
-  field->memory->page_limits = *page_limits;
-  module->memories.push_back(field->memory);
+  auto memory = new Memory();
+  memory->page_limits = *page_limits;
```
binji: Totally agree, at some point I'm picturing that a lot of this code will just be one-liners constructing everything. Kinda want to just fix it all now, but I'm trying to control my refactoring urges :-}

PR #534, opened 2017-06-26
Explicitly initialize LexToken.value_ with 0

winksaville:
As suggested by Sam Clegg (@sbc100) this is a simpler solution to
my first solution, PR #530.

### DIFF
--- a/src/wast-lexer.cc
+++ b/src/wast-lexer.cc
@@ -96,10 +96,8 @@ namespace wabt {
 
 struct WastLexer::LexToken {
   Location loc_;
-  int value_;
+  int value_ = 0;
   Token lval_;
-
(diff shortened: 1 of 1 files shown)


src/wast-lexer.cc
```diff
@@ -96,10 +96,8 @@ namespace wabt {
 
 struct WastLexer::LexToken {
   Location loc_;
```
binji: Actually, this needs to be initialized too -- the current default constructor does not initialize the values. But that can be fixed in another PR.

winksaville: PR #535 is an attempt to add the default contrustor, may not be correct but it is passing tests.

PR #535, opened 2017-06-26
Add default constructor for Location and Const

winksaville:
This fixes some TODO's although may not be exactly right but is passing
the tests for me.

I've initialized all union fields to 0 and I assume the compiler will
remove any unnecessary initialization.

### DIFF
--- a/src/binding-hash.h
+++ b/src/binding-hash.h
@@ -29,9 +29,7 @@ namespace wabt {
 struct Var;
 
 struct Binding {
-  explicit Binding(Index index) : index(index) {
-    WABT_ZERO_MEMORY(loc);
-  }
+  explicit Binding(Index index) : index(index) {}
   Binding(const Location& loc, Index index) : loc(loc), index(index) {}
 
   Location loc;
--- a/src/common.h
+++ b/src/common.h
@@ -140,14 +140,12 @@ struct StringSlice {
 };
 
 struct Location {
-  // TODO(binji): The type is not stored directly on the location because we
-  // don't currently have a default constructor. We eventually should add a
-  // default constructor, but it will be tricky since this struct is used in
-  // classes that are trivially constructible.
   enum class Type {
     Text,
     Binary,
   };
+
+  Location() : filename(NULL), line(0), first_column(0), last_column(0) {}
 
   const char* filename;
   union {
@@ -163,10 +161,6 @@ struct Location {
     };
   };
 };
-
-// TODO(binji): This should be a default constructor, but can't currently; see
-// comment above.
-Location EmptyLocation();
 
 /* matches binary format, do not change */
 enum class Type {
--- a/src/ir.h
+++ b/src/ir.h
@@ -65,18 +65,11 @@ struct Const {
   struct F32 {};
   struct F64 {};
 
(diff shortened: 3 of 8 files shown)


src/common.h
```diff
@@ -162,10 +158,16 @@ struct Location {
       size_t offset;
     };
   };
+
+
+  Location() {
+    this->line = 0;
```
binji: initialize `filename` too. Also style is to use `:line(0), first_column(0),` etc

src/common.h
```diff
@@ -162,10 +158,16 @@ struct Location {
...
+    this->first_column = 0;
+    this->last_column = 0;
+    this->offset = 0;
+  }
 };
 
-// TODO(binji): This should be a default constructor, but can't currently; see
-// comment above.
 Location EmptyLocation();
```
binji: Probably worth just removing this and replacing callers with `Location()`

src/binding-hash.h
```diff
@@ -30,7 +30,7 @@ struct Var;
 
 struct Binding {
   explicit Binding(Index index) : index(index) {
-    WABT_ZERO_MEMORY(loc);
+    //WABT_ZERO_MEMORY(loc);
```
binji: Remove.

src/ir.h
```diff
@@ -66,13 +66,17 @@ struct Const {
   struct F64 {};
 
   // Keep the default constructor trivial so it can be used as a union member.
-  Const() = default;
+  Const() {
```
binji: I'd prefer picking something valid, like just forwarding this to `Const(I32(), 0)`

src/ir.h
```diff
@@ -66,13 +66,17 @@ struct Const {
...
+    this->u32 = 0;
+    this->u64 = 0;
+    this->f32_bits = 0;
+    this->f64_bits = 0;
+  }
   Const(I32, uint32_t);
   Const(I64, uint64_t);
   Const(F32, uint32_t);
   Const(F64, uint64_t);
-  // TODO(binji): Make these default arguments on the constructors above when
```
binji: If you remove, can you remove the constructors and make the above into default args?

binji:
Bot failure is because you updated `wast-parser.y`, so you need to also run `make update-bison` too (to update the prebuilt parser)

src/ir.h
winksaville: done

src/ir.h
winksaville: done

src/binding-hash.h
winksaville: done

src/common.h
winksaville: done

src/common.h
winksaville: Turns out I can initialize one union or the other, but not both if I use initializers so I'm doing the larger union. Another option is to have no constructor and "inline" the initialization of filename with:
```
const char* filename = NULL;
```
That works and passes all tests. Maybe you can ask a C++ Guru if that also guarantees zero initialization of the union. From [the first answer here](https://stackoverflow.com/questions/4203010/how-to-initialize-member-struct-in-initializer-list-of-c-class) I think it might.

PR #536, opened 2017-06-27
Support for CommonJS and AMD loaders in libwabt.js

dcodeIO:
It is currently necessary to postprocess libwabt.js to support different kinds of module loaders, especially CommonJS-style `require` and AMD-style `define`. This PR adds CommonJS and AMD module loader support directly to wabt.pre/post.js and exposes `wabt` as a global var only if neither is available.

Unfortunately, I haven't yet figured out where exactly libwabt.js is built with emscripten, so the new generated file is not included. In case you decide to merge the PR, I would appreciate if libwabt.js could be updated as well.

### DIFF
--- a/src/wabt.post.js
+++ b/src/wabt.post.js
@@ -350,10 +350,17 @@ WasmModule.prototype.destroy = function() {
   Module._wabt_destroy_module(this.module_addr);
 };
 
-return {
+var exports = {
   ready: Promise.resolve(),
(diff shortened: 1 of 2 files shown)


binji:
`libwabt.js` is currently built by me running `make emscripten-release` and copying the result from the output directory into the demo directory. ;-)

It would be much better to have this automated, but for now I can just update this manually.

PR #538, opened 2017-06-27
Use type hierarchy for Command instead of union

binji:
Also remove CommandType::AssertInvalidNonBinary. It was only used for
the case when assert_invalid could not generate a binary file because
names could not be resolved. This error is now handled by
assert_malformed with a quoted binary.

### DIFF
--- a/src/ir.h
+++ b/src/ir.h
@@ -707,9 +707,6 @@ enum class CommandType {
   Register,
   AssertMalformed,
   AssertInvalid,
-  /* This is a module that is invalid, but cannot be written as a binary module
-   * (e.g. it has unresolvable names.) */
-  AssertInvalidNonBinary,
   AssertUnlinkable,
   AssertUninstantiable,
   AssertReturn,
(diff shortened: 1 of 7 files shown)


src/validator.cc
```diff
@@ -1080,35 +1080,35 @@ void Validator::CheckAssertReturnNanCommand(const Action* action) {
 void Validator::CheckCommand(const Command* command) {
```
KarlSchimpf: Note: If you made CheckCommand() a (virtual) method of Command, this switch can be removed.

src/wast-parser.y
```diff
@@ -1531,14 +1500,14 @@ script :
 
       int last_module_index = -1;
       for (size_t i = 0; i < $$->commands.size(); ++i) {
-        Command& command = *$$->commands[i].get();
+        Command* command = $$->commands[i].get();
```
KarlSchimpf: Somehow, this is screaming for an abstraction on the command object. However, I wouldn't worry about changing that on this PR.

src/validator.cc
binji: True. I like keeping the validation logic in validator.cc, though. If we went this route, I think I'd prefer adding visitors for these to prevent having to cast all the time.

src/wast-parser.y
binji: Haha, definitely! It's pretty silly that every caller has to manually manage the module this way. It would be a nice cleanup for someone to do. Or likely I'll get around to it in a few PRs. :-)

src/ir.h
```diff
@@ -723,28 +720,111 @@ enum class CommandType {
...
 static const int kCommandTypeCount = WABT_ENUM_COUNT(CommandType);
 
-struct Command {
+class Command {
+ public:
   WABT_DISALLOW_COPY_AND_ASSIGN(Command);
-  Command();
-  ~Command();
+  Command() = delete;
+  virtual ~Command() {}
```
sbc100: = default?

binji: Oh, good point.

PR #540, opened 2017-06-28
Remove WABT_ZERO_MEMORY WABT_FAILED and WABT_SUCCEEDED macros.

KarlSchimpf:
Changed these macros to inline functions.

This change was prompted by the difficulty of using WABT_ZERO_MEMORY to clear the contents of this in a constructor.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -33,10 +33,6 @@
 
 #define WABT_FATAL(...) fprintf(stderr, __VA_ARGS__), exit(1)
 #define WABT_ARRAY_SIZE(a) (sizeof(a) / sizeof(a[0]))
-#define WABT_ZERO_MEMORY(var)                                          \
-  WABT_STATIC_ASSERT(                                                  \
-      std::is_pod<std::remove_reference<decltype(var)>::type>::value); \
-  memset(static_cast<void*>(&(var)), 0, sizeof(var))
 
 #define WABT_USE(x) static_cast<void>(x)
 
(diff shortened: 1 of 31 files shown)


src/common.h
```diff
@@ -103,8 +99,13 @@ enum class Result {
   Error,
 };
 
-#define WABT_SUCCEEDED(x) ((x) == ::wabt::Result::Ok)
-#define WABT_FAILED(x) ((x) == ::wabt::Result::Error)
+template<typename T>
+void ZeroMemory(T& v) {
+  memset(&v, 0, sizeof(v));
```
binji: I think we should keep the static assert -- it's caught a number of bugs while I've refactored.

src/common.h
```diff
@@ -103,8 +99,13 @@ enum class Result {
...
+}
+
+inline bool Succeeded(Result result) { return result == Result::Ok; }
```
binji: Honestly, if we're refactoring this, I'd rather make it so that we could just `Result` directly as a bool.

binji: That said, this is nicer for now.

src/interpreter.cc
```diff
@@ -94,9 +94,9 @@ FuncSignature::FuncSignature(Index param_count,
       result_types(result_types, result_types + result_count) {}
 
 Import::Import() : kind(ExternalKind::Func) {
-  WABT_ZERO_MEMORY(module_name);
-  WABT_ZERO_MEMORY(field_name);
-  WABT_ZERO_MEMORY(func.sig_index);
+  ZeroMemory(module_name);
```
winksaville: Shouldn't we have StringSlice (an maybe other struct's) initialize itself, zeroing them manually doesn't seem very robust.

binji: These will eventually be `std::string` generally, and be passed via `string_view`. But more refactoring needs to be done before that's possible -- again, due requiring trivial constructors in a few places.

src/common.h
KarlSchimpf: I'll leave it as is for now. but will change latter to a bool. I just find the needing to test with macros, and that Result::Ok is zero, too easy to make a mistake.

src/common.h
KarlSchimpf: Adding back in.

PR #544, opened 2017-06-30
Use intrusive_list for Expr

binji:
Nothing much special here, but it is becoming increasingly annoying that
the parser Token cannot have value types (so everything must be moved
then deleted).

### DIFF
--- a/src/binary-reader-ir.cc
+++ b/src/binary-reader-ir.cc
@@ -40,15 +40,14 @@ namespace wabt {
 namespace {
 
 struct LabelNode {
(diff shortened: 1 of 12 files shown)


binji:
> I agree that having to allocate/delete in the parser is annoying, but (at the moment) don't see an obvious solution.

Yeah, it is an annoying limitation of the bison parsing model (at least the C skeleton). At this point there's enough clunkiness, I started thinking about switching to a recursive descent parser. But don't worry, there are more important things to do first. :-)

PR #545, opened 2017-06-30
Extend binary reader to handle exception constructs.

KarlSchimpf:
Extends the binary reader to handle exception constructs.

Also updates wasm2wast to be able to read binary files with exceptions in them.

Also updates run-roundtrip.py to understand new command line arguments for exceptions.


src/binary-reader-ir.cc
```diff
@@ -660,6 +698,10 @@ Result BinaryReaderIR::OnNopExpr() {
   return AppendExpr(expr);
 }
 
+Result BinaryReaderIR::OnRethrowExpr(Index depth) {
+  return AppendExpr(new RethrowExpr(Var(depth)));
```
binji: You can construct `Var(depth, GetLocation())`. I noticed that I forgot to do this for other `Var` constructors in this file too.

src/binary-reader.cc
```diff
@@ -989,10 +1013,53 @@ Result BinaryReader::ReadFunctionBody(Offset end_offset) {
...
+        Index index;
+        CHECK_RESULT(ReadIndex(&index, "exception index"));
+        CALLBACK(OnCatchExpr, index);
+        CALLBACK(OnOpcodeIndex, index);
+        break;
+      }
+
+      case Opcode::CatchAll: {
+        ERROR_UNLESS_FUTURE_EXCEPTIONS_OPCODE(opcode);
+        CALLBACK(OnCatchAllExpr);
```
binji: Should call `OnOpcodeBare` callback too.

src/tools/wasm2wast.cc
```diff
@@ -72,8 +72,15 @@ static void parse_options(int argc, char** argv) {
         s_outfile = argument;
         ConvertBackslashToSlash(&s_outfile);
       });
+  // TODO(karlschimpf) Figure out why fold-exprs is on by default.
```
binji: Weird -- do you have a repro case for this?

src/tools/wasm2wast.cc
```diff
@@ -72,8 +72,15 @@ static void parse_options(int argc, char** argv) {
...
   parser.AddOption('f', "fold-exprs", "Write folded expressions where possible",
                    []() { s_write_wat_options.fold_exprs = true; });
+  parser.AddOption('n', "no-fold-exprs", "Don't fold expressions",
```
binji: I'd rather not add this for now, since it's the default. At least not with a short letter.

test/roundtrip/try.txt
```diff
@@ -0,0 +1,30 @@
+;;; TOOL: run-roundtrip
```
binji: It would be good to add some dump tests too, to ensure that these are being decoded properly by wasm-objdump.

src/binary-reader-ir.cc
```diff
@@ -270,6 +289,13 @@ Result BinaryReaderIR::AppendExpr(Expr* expr) {
   return Result::Ok;
 }
 
+Exception* BinaryReaderIR::CreateException(TypeVector& sig) {
```
binji: Just add a Exception constructor?

KarlSchimpf: Done.

src/binary-reader-ir.cc
KarlSchimpf: Done, and fixed other cases I could find.

src/binary-reader-ir.cc
```diff
@@ -687,11 +729,62 @@ Result BinaryReaderIR::OnStoreExpr(Opcode opcode,
...
   auto expr = new TeeLocalExpr(Var(local_index, GetLocation()));
   return AppendExpr(expr);
 }
 
+Result BinaryReaderIR::OnTryExpr(Index num_types, Type* sig_types) {
+  auto expr = new TryExpr();
+  expr->block = new Block();
+  expr->block->sig.assign(sig_types, sig_types + num_types);
+  if (Failed(AppendExpr(expr))) {
+    delete expr;
```
KarlSchimpf: Done, and fixed other calls to AppendExpr when it fails.

src/binary-reader-ir.cc
```diff
@@ -687,11 +729,62 @@ Result BinaryReaderIR::OnStoreExpr(Opcode opcode,
...
+    return Result::Error;
+  }
+  PushLabel(LabelType::Try, &expr->block->exprs);
+  LabelNode* label;
+  if (Failed(TopLabel(&label))) {
+    PrintError("Internal error installing try block");
```
KarlSchimpf: I agree, if PushLabel() is written correctly, this should succeed. Changing to assert.

src/binary-reader-ir.cc
```diff
@@ -687,11 +729,62 @@ Result BinaryReaderIR::OnStoreExpr(Opcode opcode,
...
+  }
+  if (label != nullptr)
+    PrintError("catch not inside try block");
+  delete catch_;
+  return Result::Error;
+}
+
+Result BinaryReaderIR::OnCatchExpr(Index except_index) {
+  ExprList empty;
+  return AppendCatch(new Catch(Var(except_index), std::move(empty)));
```
KarlSchimpf: Added more constructors to class Catch.

src/binary-reader-ir.cc
```diff
@@ -687,11 +729,62 @@ Result BinaryReaderIR::OnStoreExpr(Opcode opcode,
...
+  return Result::Error;
+}
+
+Result BinaryReaderIR::OnCatchAllExpr() {
+  ExprList empty;
+  return AppendCatch(new Catch(std::move(empty)));
```
KarlSchimpf: Done.

src/binary-reader.cc
KarlSchimpf: Done.

src/tools/wasm2wast.cc
KarlSchimpf: Hmm, I don't know. The problem no longer happens. I guess I must of accidentally broke this, and then fixed the breakage after changing things.

Removing the no-fold-exprs flag.

src/tools/wasm2wast.cc
KarlSchimpf: Removed. Problem no longer happens.

test/roundtrip/try.txt
KarlSchimpf: Will add tests once wasm-objdump is made aware of exceptions (i.e. in a separate PR).

PR #547, opened 2017-07-03
Adds the concept of exceptions to wasm-objdump

KarlSchimpf:
Modifies the wasm objdump code to understand exceptions.

Adds command-line argument --future-exceptions to wasm-objdump and run-objdump.py.

Modifies class BinaryReaderNop so that its default behavior can be defined by a flag that defines if
exception handling should be accepted.

### DIFF
--- a/src/binary-reader-objdump.cc
+++ b/src/binary-reader-objdump.cc
@@ -70,6 +70,7 @@ BinaryReaderObjdumpBase::BinaryReaderObjdumpBase(const uint8_t* data,
       data(data),
       size(size) {
   ZeroMemory(section_starts);
+  BinaryReaderNop::allow_future_exceptions = options->allow_future_exceptions;
 }
 
 Result BinaryReaderObjdumpBase::BeginSection(BinarySection section_code,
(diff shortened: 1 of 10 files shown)


src/binary-reader-objdump.cc
```diff
@@ -752,6 +768,24 @@ Result BinaryReaderObjdump::OnImportGlobal(Index import_index,
   return Result::Ok;
 }
 
+Result BinaryReaderObjdump::OnImportException(Index import_index,
+                                              StringSlice module_name,
+                                              StringSlice field_name,
+                                              Index except_index,
+                                              TypeVector& sig) {
+  PrintDetails(" - except[%" PRIindex "] (", except_index);
```
binji: There aren't any tests that display this output.

src/binary-reader-objdump.cc
```diff
@@ -1042,6 +1076,25 @@ Result BinaryReaderObjdump::OnSymbolInfo(StringSlice name,
...
 
+Result BinaryReaderObjdump::OnExceptionCount(Index count) {
+  return OnCount(count);
+}
+
+Result BinaryReaderObjdump::OnExceptionType(
+    Index index, TypeVector& sig) {
+  if (!ShouldPrintDetails())
+    return Result::Ok;
+  printf(" - except[%" PRIindex "] (", index);
```
binji: Or this.

PR #549, opened 2017-07-04
Add inline exports to wat-writer.

KarlSchimpf:
Noticed that inline exports was not handled for exceptions. This PR adds the missing code, and a test to verify it works.


KarlSchimpf:
Closing. My  development state is bad.

PR #551, opened 2017-07-04
Add more test examples for wasm-objdump.

KarlSchimpf:
Added more test examples for wasm-objdump, and requested by @binji.


test/run-objdump.py
```diff
@@ -81,6 +83,7 @@ def main(args):
       find_exe.GetWasmdumpExecutable(options.bindir),
       error_cmdline=options.error_cmdline)
   wasm_objdump.AppendOptionalArgs({
+      '--details': options.details,
```
sbc100: I think this is the same as '-x' which is already set by --dump-verbose.

KarlSchimpf: Ok. Removed --details flag from run-objdump.py and used --dump-verbose instead.

PR #553, opened 2017-07-06
Remove BinaryErrorHandler, rename SourceErrorHandler

binji:
Since SourceErrorHandler can support binary locations, it doesn't make
much sense to have two different error handlers. There is now only an
ErrorHandler base class, and the first argument specifies whether the
error handler is expecting text or binary locations.


src/error-handler.h
```diff
@@ -24,14 +24,14 @@
...
 
-  virtual ~SourceErrorHandler() {}
+  virtual ~ErrorHandler() {}
 
   // Returns true if the error was handled.
-  virtual bool OnError(const Location*,
+  virtual bool OnError(const Location&,
                        const std::string& error,
                        const std::string& source_line,
                        size_t source_line_column_offset) = 0;
```
KarlSchimpf: Why not introduce an inline version without source line and column, for use with the binary case. I.e. something like:

bool OnError(Offset offset, const std::string& error) {
  return OnError(Location(offset), error, std::string(), 0);
}

That way, the binary use doesn't need to explicitly convert/fill in arguments.

src/binary-reader-interpreter.cc
```diff
@@ -322,7 +322,7 @@ Label* BinaryReaderInterpreter::TopLabel() {
 }
 
 bool BinaryReaderInterpreter::HandleError(Offset offset, const char* message) {
-  return error_handler->OnError(offset, message);
+  return error_handler->OnError(Location(offset), message, std::string(), 0);
```
KarlSchimpf: Use binary version of OnError (see error-handler.h comments).

src/binary-reader-ir.cc
```diff
@@ -289,7 +289,7 @@ Result BinaryReaderIR::AppendExpr(Expr* expr) {
 }
 
 bool BinaryReaderIR::HandleError(Offset offset, const char* message) {
-  return error_handler->OnError(offset, message);
+  return error_handler->OnError(Location(offset), message, std::string(), 0);
```
KarlSchimpf: Again, use binary version of OnError.

src/error-handler.h
binji: Good idea.

src/common.h
```diff
@@ -147,7 +147,8 @@ struct Location {
     Binary,
   };
 
-  Location() : filename(NULL), line(0), first_column(0), last_column(0) {}
+  Location() : filename(nullptr), line(0), first_column(0), last_column(0) {}
+  explicit Location(size_t offset) : filename(nullptr), offset(offset) {}
 
   const char* filename;
```
sbc100: Add "= nullptr" here to avoid setting in both contructors above?

test/dump/try-details.txt
```diff
@@ -14,16 +14,21 @@
...
 Type:
  - type[0] () -> i32
 Function:
  - func[0] sig=0
 Custom:
  - name: "exception"
  - except[0] (i32)
+
 Code Disassembly:
+
```
sbc100: Strange.  I wonder how these newlines were originally missing?   Separate PR maybe?

binji: Yeah, I think they accidentally got included when I rebuilt all the tests. It is that same issue you found before, where empty lines are not caught by the diff.

src/common.h
binji: Done.

PR #554, opened 2017-07-06
Use intrusive list for ModuleFields

binji:


src/validator.cc
```diff
@@ -753,8 +753,8 @@ void Validator::CheckTable(const Location* loc, const Table* table) {
 }
 
 void Validator::CheckElemSegments(const Module* module) {
-  for (ModuleField* field = module->first_field; field; field = field->next) {
-    if (auto elem_segment_field = dyn_cast<ElemSegmentModuleField>(field)) {
+  for (const ModuleField& field : module->fields) {
```
KarlSchimpf: Much more readable!

binji:
I borrowed the name from boost: http://www.boost.org/doc/libs/1_64_0/doc/html/intrusive.html

KarlSchimpf:
Its intrusive, because the list structure is part of the object being built, rather than the list being separate data structures. For example, std::list  and std::deque aren't intrusive because the list is defined separately from the elements.

PR #555, opened 2017-07-06
Add support data-size and alignment linking metadata

sbc100:


src/tools/wasm-link.cc
```diff
@@ -758,8 +785,11 @@ static void write_binary(Context* ctx) {
   write_names_section(ctx);
 
   /* Generate a new set of reloction sections */
-  for (size_t i = FIRST_KNOWN_SECTION; i < kBinarySectionCount; i++) {
-    write_reloc_section(ctx, static_cast<BinarySection>(i), sections[i]);
+  if (s_relocatable) {
+    write_linking_section(ctx, 0, 0);
```
binji: Seems weird to have an alignment of 0. Should this be 1?

PR #556, opened 2017-07-06
Use std::string instead of StringSlice in Var

binji:
This change keeps the union, but hides it behind accessors. The
implementation is a bit complex since std::string requires constructors
and destructors to be called. As a result we have to use placement new
and explicit destructors.

### DIFF
--- a/src/binding-hash.h
+++ b/src/binding-hash.h
@@ -49,14 +49,13 @@ class BindingHash : public std::unordered_multimap<std::string, Binding> {
 
   Index FindIndex(const Var&) const;
 
-  Index FindIndex(const string_view& name) const {
-    auto iter = find(name.to_string());
+  Index FindIndex(const std::string& name) const {
+    auto iter = find(name);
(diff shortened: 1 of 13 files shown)


src/binding-hash.h
```diff
@@ -48,11 +49,14 @@ class BindingHash : public std::unordered_multimap<std::string, Binding> {
...
+    return iter != end() ? iter->second.index : kInvalidIndex;
+  }
+
   Index FindIndex(const StringSlice& name) const {
     auto iter = find(string_slice_to_string(name));
-    if (iter != end())
-      return iter->second.index;
-    return kInvalidIndex;
+    return iter != end() ? iter->second.index : kInvalidIndex;
   }
```
KarlSchimpf: Maybe consider adding a method

Index FindIndex(std::string name) const {
  auto iter  = find(name);
  return iter != end() ? iter->second.index : kInvalidIndex;
}

and then simplify the bodies of the other two FindIndex() methods to call this with the extracted string?

PR #559, opened 2017-07-07
Remove uses of StringSlice from IR

binji:
All member variables are now std::string, and most parameters are
string_view. This allows us to remove many of the destructors.


KarlSchimpf:
LGTM + 1 nit.

binji:
@KarlSchimpf where's the nit? :-)

PR #560, opened 2017-07-07
Clean up BinaryWriterSpec::GetModuleFilename

binji:
This previously had an upper limit to the extension length.

I've also taken this opportunity to modify the Writer and Stream APIs to
take `string_view` instead of `const char*` for filenames.

### DIFF
--- a/src/binary-writer-spec.cc
+++ b/src/binary-writer-spec.cc
@@ -67,7 +67,7 @@ class BinaryWriterSpec {
   Result WriteScript(Script* script);
 
  private:
-  char* GetModuleFilename(const char* extension);
+  std::string GetModuleFilename(const char* extension);
   void WriteString(const char* s);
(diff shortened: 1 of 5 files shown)


src/binary-writer-spec.cc
```diff
@@ -113,14 +114,13 @@ BinaryWriterSpec::BinaryWriterSpec(const char* source_filename,
...
-                    module_filename_noext_.c_str(), num_modules_, extension);
-  ConvertBackslashToSlash(str, length);
-  return str;
+std::string BinaryWriterSpec::GetModuleFilename(const char* extension) {
+  std::string result = module_filename_noext_;
+  result += '.';
+  result += std::to_string(num_modules_);
+  result += extension;
+  ConvertBackslashToSlash(&result);
+  return result;
```
KarlSchimpf: Much cleaner!

PR #564, opened 2017-07-10
Use templates instead of macros in interpreter

binji:
There's still more to be done, but this is a decent cleanup.

### DIFF
--- a/src/interpreter.cc
+++ b/src/interpreter.cc
@@ -20,6 +20,8 @@
 #include <cassert>
 #include <cinttypes>
 #include <cmath>
+#include <limits>
+#include <type_traits>
 #include <vector>
 
 #include "stream.h"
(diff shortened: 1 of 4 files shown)


binji:
Works, but I'm not particularly happy with this solution. For future reference, the problem is that on MSVC ia32, it will use SSE2 for calculating float math, but all calling conventions pass floats/doubles via x87 registers instead of xmm registers unless you use the `__vectorcall` calling convention. This causes incorrect rounding in some cases.

binji:
I think a better solution is to always keep the float/double as an integer value at function boundaries. So Add/Sub/etc. should bitcast on entry and exit.

binji:
OK, I think this is better. PTAL

binji:
Yeah, I kept trying to find a way to make it simpler, but this was as good as I could do. If you have any ideas, I'd love to hear them!

PR #567, opened 2017-07-12
Use std::vector for DataSegment data

binji:
* Also switch ScriptModule data to std::vector
* In wast-parser.y:
  - Change DupTextList to write to std::vector
  - quoted_text uses CopyStringContents instead of DupTextList

### DIFF
--- a/src/ir.cc
+++ b/src/ir.cc
@@ -334,11 +334,6 @@ Expr::Expr(ExprType type) : type(type) {}
 Table::Table() {
   ZeroMemory(elem_limits);
 }
-DataSegment::DataSegment() : data(nullptr), size(0) {}
-
-DataSegment::~DataSegment() {
-  delete[] data;
-}
 
 Memory::Memory() {
   ZeroMemory(page_limits);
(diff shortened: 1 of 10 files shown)


src/ir.cc
```diff
@@ -379,32 +374,31 @@ ScriptModule::ScriptModule(Type type) : type(type) {
...
   typedef std::string std_string;
+
   switch (type) {
     case ScriptModule::Type::Text:
       delete text;
       break;
     case ScriptModule::Type::Binary:
       binary.name.~std_string();
-      delete [] binary.data;
+      binary.data.~vector();
```
sbc100: I'm not used to seeing this syntax.  Is it needed because we are dealing with union?

binji: Yeah, you have to manually call the destructor when deleting union members. Pretty annoying, but the alternative is to use variant (C++17 only) or use virtual functions.

KarlSchimpf: Yes. It is the placement call to the destructor, to match the placement call to the constructor in the previous method. Placement construction/destruction is necessary for unions.

binji: Updated to use `Construct` and `Destruct`.

PR #573, opened 2017-07-14
Add saturating float truncation operators

binji:


binji:
I still need to add flags (`--future-trunc-sat`?) but otherwise everything else is implemented.

src/interpreter-opcode.def
```diff
@@ -27,12 +27,12 @@
...
+ *          tr  t1    t2   m  prefix code  Name text
  * ============================================================  */
 
 #include "opcode.def"
-WABT_OPCODE(___, ___, ___, 0, 0xc0, Alloca, "alloca")
-WABT_OPCODE(___, ___, ___, 0, 0xc1, BrUnless, "br_unless")
-WABT_OPCODE(___, ___, ___, 0, 0xc2, CallHost, "call_host")
-WABT_OPCODE(___, ___, ___, 0, 0xc3, Data, "data")
-WABT_OPCODE(___, ___, ___, 0, 0xc4, DropKeep, "drop_keep")
+WABT_OPCODE(___, ___, ___, 0, 0,     0xc0, Alloca, "alloca")
```
KarlSchimpf: Out of curiosity, why aren't these in opcode.def?

binji: These are special opcodes that are only used for the interpreter. It definitely complicates things, so I've been considering switching the interpreter to execute directly from wasm instead of recompiling it first. Not sure it's worth the effort, though.

PR #574, opened 2017-07-14
Make the exception section a known section.

KarlSchimpf:
Change the (experimental) exception section to a known section with value 13.

It also accepts a name section as a known section with value 12.

### DIFF
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -1291,9 +1291,6 @@ Result BinaryReader::ReadCustomSection(Offset section_size) {
     CHECK_RESULT(ReadRelocSection(section_size));
   } else if (section_name == WABT_BINARY_SECTION_LINKING) {
     CHECK_RESULT(ReadLinkingSection(section_size));
-  } else if (options_->allow_future_exceptions &&
-             section_name == WABT_BINARY_SECTION_EXCEPTION) {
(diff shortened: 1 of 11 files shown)


src/binary.h
```diff
@@ -40,7 +40,9 @@
   V(Start, start, 8)                   \
   V(Elem, elem, 9)                     \
   V(Code, code, 10)                    \
-  V(Data, data, 11)
+  V(Data, data, 11)                    \
+  V(Names, name, 12)                   \
```
binji: When did the name section become a known section?

src/binary-reader.cc
```diff
@@ -1687,8 +1690,13 @@ Result BinaryReader::ReadSections() {
     ERROR_UNLESS(state_.offset == read_end_,
                  "unfinished section (expected end: 0x%" PRIzx ")", read_end_);
 
-    if (section != BinarySection::Custom)
-      last_known_section_ = section;
+    switch (section) {
+      case BinarySection::Custom:
+      case BinarySection::Exception:
```
binji: Why not set it here? Otherwise you could put an exception section out of order.

binji:
Closing, since the exception proposal seems to be moving in a different direction.

eholk:
I think we're still likely to have an exception section, but we might as well just do another PR once we know for sure.

PR #576, opened 2017-07-17
Remove StringSlice

binji:
* Added StringTerminal only used by the lexer and parser, since we need an
  object that has a non-trivial constructor/destructor (so it can be
  included in the Token union).
* Name all terminal Token members with `t_` prefix
* Rename Literal -> LiteralTerminal, and only use for the lexer/parser.
* Move LiteralType to literal.h
* Use std::vector<std::string> for TextList and rename to TextVector
* Renamed CopyStringContents -> UnescapeString, and make it a template
  so it can write to a std::string and a std::vector.
* Rename DupTextList -> UnescapeTextVector, and simplify it
  considerably.

### DIFF
--- a/src/wast-parser.y
+++ b/src/wast-parser.y
@@ -145,7 +145,7 @@ class BinaryErrorHandlerModule : public ErrorHandler {
 };
 
 template <typename OutputIter>
-void UnescapeString(string_view text, OutputIter dest) {
+void RemoveEscapes(string_view text, OutputIter dest) {
   // Remove surrounding quotes; if any. This may be empty if the string was
   // invalid (e.g. if it contained a bad escape sequence).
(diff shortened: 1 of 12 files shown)


src/wast-parser.y
```diff
@@ -149,6 +144,65 @@ class BinaryErrorHandlerModule : public ErrorHandler {
   WastParser* parser_;
 };
 
+template <typename OutputIter>
+void UnescapeString(string_view text, OutputIter dest) {
```
KarlSchimpf: Perhaps better named 'RemoveEscapes' (the string can be dropped since it is implied by the first argument).

binji: Done. I also renamed UnescapeTextVector to RemoveEscapes too.

PR #579, opened 2017-07-19
Rename snake_case to MixedCase.

binji:
There are no functional changes.

### DIFF
--- a/src/binary-reader-logging.cc
+++ b/src/binary-reader-logging.cc
@@ -34,7 +34,7 @@ namespace wabt {
 
 namespace {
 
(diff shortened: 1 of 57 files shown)


src/binary-reader-logging.cc
```diff
@@ -34,7 +34,7 @@ namespace wabt {
 
 namespace {
 
-void sprint_limits(char* dst, size_t size, const Limits* limits) {
+void SprintLimits(char* dst, size_t size, const Limits* limits) {
```
jgravelle-google: Should this be `Sprint` limits or `SPrint` limits?

binji: Hm, good point.

PR #580, opened 2017-07-22
Refactor wasm-opcodecnt.

binji:
Simplify some of the code using map instead of vector. Also added the
ability to count immediates of any opcode.

### DIFF
--- a/src/binary-reader-opcnt.cc
+++ b/src/binary-reader-opcnt.cc
@@ -49,6 +49,10 @@ OpcodeInfo::OpcodeInfo(Opcode opcode, Kind kind, T* data, size_t count, T extra)
 
 template <typename T>
 std::pair<const T*, size_t> OpcodeInfo::GetDataArray() const {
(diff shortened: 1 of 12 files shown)


src/binary-reader-opcnt.cc
```diff
@@ -24,100 +24,240 @@
...
+  Result OnBrTableExpr(Index num_targets,
+                       Index* target_depths,
+                       Index default_target_depth) override;
+  Result OnEndExpr() override;
+  Result OnEndFunc() override;
 
  private:
-  OpcntData* opcnt_data;
+  template <typename... Args>
+  void Emplace(Args&&... args);
```
KarlSchimpf: Why not make this have a return type of Result, and return Ok, so that OpcodeXXX routines can just return the result of Emplace, rather than a separate return statement returning Ok?

binji: Done.

PR #582, opened 2017-07-25
Update testsuite

binji:
* Implicit function types are now appended to the module, not directly
  after the function/import that uses them.
* Float and int literals can have an underscore separating digits.

### DIFF
--- a/src/literal.cc
+++ b/src/literal.cc
@@ -160,10 +160,11 @@ Result FloatParser<T>::ParseFloat(const char* s,
   // WebAssembly floats can contain underscores, but strto* can't parse those,
   // so remove them first.
   assert(s <= end);
-  char buffer[end - s + 1];
+  const size_t kBufferSize = end - s + 1;  // +1 for \0.
+  char* buffer = static_cast<char*>(alloca(kBufferSize));
   auto buffer_end =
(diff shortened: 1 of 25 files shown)


src/literal.cc
```diff
@@ -260,7 +271,9 @@ Result FloatParser<T>::ParseHex(const char* s,
   int significand_exponent = 0;  // Exponent adjustment due to dot placement.
   for (; s < end; ++s) {
     uint32_t digit;
-    if (*s == '.') {
+    if (*s == '_') {
```
KarlSchimpf: Is this still possible? Didn't the code above remove underscores?

src/literal.cc
```diff
@@ -544,6 +562,8 @@ Result ParseUint64(const char* s, const char* end, uint64_t* out) {
     }
   } else {
     for (; s < end; ++s) {
+      if (*s == '_')
```
KarlSchimpf: Same here.

src/literal.cc
binji: Yeah, the code above only removes underscores for the `strtof`/`strtod` path. I guess I could remove them from other passes too, but that's the only case where it was necessary.

PR #591, opened 2017-08-11
Rewrite parser as recursive descent

binji:
* Remove Bison dependency
* Remove pre-generated parser files
* Rename build config from no-re2c-bison to no-re2c
* Add a simple make_unique implementation
* Move handling of module bindings into ir.cc
* Simplify lexer
  - Remove lookahead, the parser handles this now
  - Unify Token/LexerToken, it only contains terminal values now
  - Refactor setting token type and value into one function (e.g.
    LITERAL, RETURN => RETURN_LITERAL)
* New Parser
  - Uses two tokens of lookahead (use Peek(), PeekAfter())
  - Consume() consumes one token of any kind
  - Match(t) consumes the current token if it matches
  - PeekMatch(t) returns true iff the token matches, but doesn't consume
  - Basic error synchronization; plenty of room for improvement here


test/parse/bad-toplevel.txt
```diff
@@ -1,13 +1,7 @@
 ;;; ERROR: 1
 (foo bar)
 (;; STDERR ;;;
-out/test/parse/bad-toplevel.txt:2:2: error: unexpected token "foo"
+out/test/parse/bad-toplevel.txt:2:1: error: unexpected token "(", expected a module field or a command.
```
jayphelps: I know this is a first pass conversion (and it's awesome) but this one jumped out at me and confusing. Perhaps when an invalid field is discovered it consumes the Lpar [before erroring](https://github.com/WebAssembly/wabt/pull/591/files#diff-3f63016a605cad7c823d8c04253ed961R682) so the invalid token is given?

binji: Good call; this is handled in a few other places but I think I must have missed it here. :+1:

src/wast-lexer.cc
```diff
@@ -22,48 +22,22 @@
...
-#define FILL(n)                                \
-  do {                                         \
-    if (Failed(Fill(loc, parser, (n)))) {      \
-      int value = NAME_TO_VALUE(EOF);          \
-      SetToken(value);                         \
-      return PopLookaheadToken(lval, loc);     \
-    }                                          \
+#define FILL(n)              \
+  do {                       \
+    if (Failed(Fill((n)))) { \
```
KarlSchimpf: What if there were lookahead tokens? In such a case you would return eof too early.

src/wast-lexer.cc
```diff
@@ -86,23 +60,163 @@
...
 
 namespace wabt {
 
-struct WastLexer::LexToken {
-  Location loc_;
-  int value_ = 0;
-  Token lval_;
-};
+const char* GetTokenTypeName(TokenType token_type) {
+  const char* s_names[] = {
```
KarlSchimpf: Shouldn't this be static (i.e. initialized only once)?

src/wast-parser.cc
```diff
@@ -0,0 +1,2013 @@
...
+bool WastParser::Match(TokenType type) {
+  if (PeekMatch(type)) {
+    Consume();
+    return true;
+  }
+  return false;
+}
+
+bool WastParser::MatchLpar(TokenType type) {
+  if (PeekMatchLpar(type)) {
```
KarlSchimpf: Why not a call to check if PeekAfter() == type?

src/wast-parser.cc
```diff
@@ -0,0 +1,2013 @@
...
+    out_type_list->push_back(Consume().type);
+
+  CHECK_RESULT(ErrorIfLpar({"i32", "i64", "f32", "f64"}));
+  return Result::Ok;
+}
+
+Result WastParser::ParseQuotedText(std::string* text) {
+  WABT_TRACE(ParseQuotedText);
+  if (!PeekMatch(TokenType::Text))
+    return ErrorExpected({"a quoted string"}, "\"foo\"");
```
KarlSchimpf: ?foo?

src/wast-parser.cc
```diff
@@ -0,0 +1,2013 @@
...
+  if (tokens_.empty())
+    tokens_.push_back(lexer_->GetToken(this));
+  return tokens_.front();
+}
+
+Location WastParser::GetLocation() {
+  return GetToken().loc;
+}
+
+TokenType WastParser::Peek() {
```
KarlSchimpf: Why doesn't this check if there available tokens?

I assume this is because you are assuming that multiple lookahead only happens in very restricted contexts (such as "(" "token"). If you are making this assumption, make it more clear.

Another choice would be to replace Peek() and PeekAfter() with:

TokenType WasmParser::Peek(size_t n = 0) {
  while (tokens_.size() <= n)
    tokens_.push_back(lexer_->GetToken(this);
  return tokens_.at(n).,token_type;
}

This solution would allow more general peek behavior.

src/wast-parser.cc
```diff
@@ -0,0 +1,2013 @@
...
+#define WABT_TRACING 0
+#include "tracing.h"
+
+#define CHECK_RESULT(expr)  \
+  do {                      \
+    if (Failed(expr))       \
+      return Result::Error; \
+  } while (0)
+
+#define EXPECT(token_type) CHECK_RESULT(Expect(TokenType::token_type))
```
KarlSchimpf: Why not make this a function and let the compiler inline it?

KarlSchimpf: Similarly, create a "check()" function

void check(Token_type ty) { CHECK_RESULT(expr); }

KarlSchimpf: and

void ExpectCheck(TokenType ty) { CHECK_RESULT(Expect(ty)); }

src/wast-lexer.cc
KarlSchimpf: Never mind. I realize now that lookahead is handled in the parser.

src/wast-parser.cc
binji: I made this a macro mostly because of the magic automatic control flow (since `CHECK_RESULT` will return on failure)

src/wast-lexer.cc
binji: I moved the lookahead tokens to the parser, so the lexer can always just return the immediate value. Seemed more natural to me this way, since only the parser actually cares about being able to look ahead.

src/wast-parser.cc
binji: I thought about this, but since we only need two tokens of lookahead, I thought it would be better to make that clear by not making it general. Guess I should add some more documentation about all this.

src/wast-parser.cc
binji: That's what `PeekMatchLpar` does, but it also checks that `Peek()` is `TokenType::Lpar`

src/wast-parser.cc
binji: This is just for example text. Maybe not needed, but I thought it made the errors a little nicer:
```
error: unexpected token "(", expected a quoted string (e.g. "foo").
```

src/wast-parser.cc
KarlSchimpf: Good point. I didn't read it carefully enough. current form is acceptable.

src/wast-parser.cc
KarlSchimpf: Verified. Ok.

src/wast-parser.cc
KarlSchimpf: While I agree  that you only need two tokens of look-ahead, the current implementation was making some larger assumptions, such as you are always assuming that at least one token of look-ahead is being maintained (my solution removes this restriction).

I prefer a solution that doesn't have implicit assumptions because they are ALWAYS harder to maintain. In most contexts, an extension will come along that violates it, and then it is much harder to convince yourself what places need to be fixed. I understand that the Peek(0) (or simply Peek()) that I suggest does add an additional test (to see if a token needs to be retrieved), but I don't see that as a major performance cost.

I also like only having one "Peek" method, but that is my feeling.

binji: OK, I'll give it a shot :-)

binji: Done, PTAL

PR #595, opened 2017-08-26
Clean up various structs in ir.h

binji:
* Nicer static constructors for Const (e.g. `Const::F32()`)
* Protect `type` member when used as discriminant in type hierarchy 
  (e.g. `Expr`, `ModuleField`, etc.)
* Remove pointer member fields where possible. Where not, use `std::unique_ptr` instead.
* Remove nearly all uses of unions (`Var` and `Const` still use them)

### DIFF
--- a/src/validator.cc
+++ b/src/validator.cc
@@ -780,30 +780,38 @@ void Validator::CheckDataSegments(const Module* module) {
 }
 
 void Validator::CheckImport(const Location* loc, const Import* import) {
-  switch (import->kind) {
+  switch (import->kind()) {
     case ExternalKind::Except:
       ++current_except_index_;
-      CheckExcept(loc, import->except);
+      CheckExcept(loc, &cast<ExceptionImport>(import)->except);
       break;
-    case ExternalKind::Func:
-      if (import->func->decl.has_func_type)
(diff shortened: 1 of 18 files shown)


binji:
This looks like a big change, but most of it is just switching `->` to `.` and similar. Gonna land now, but feel free to comment on confusing/weird/incorrect bits, and I'll address in future PRs.

src/validator.cc
```diff
@@ -779,42 +768,50 @@ void Validator::CheckMemory(const Location* loc, const Memory* memory) {
 void Validator::CheckDataSegments(const Module* module) {
   for (const ModuleField& field : module->fields) {
     if (auto data_segment_field = dyn_cast<DataSegmentModuleField>(&field)) {
-      DataSegment* data_segment = data_segment_field->data_segment;
+      auto&& data_segment = data_segment_field->data_segment;
```
sbc100: I thought the/our convention was to use auto only when the type is obvious somewhere else in the statement?  (But maybe I'm just mis-remembering).

binji: Yeah, probably a good idea. I'll undo these.

binji: See #598.

PR #596, opened 2017-08-28
Removed duplicate initialization of EMSCRIPTEN_DIR var in Makefile

dcodeIO:
Should now allow overriding the variable with a custom emscripten location.

For reference, EMSCRIPTEN_DIR is already conditionally set [here](https://github.com/WebAssembly/wabt/blob/master/Makefile#L26).


binji:
Oops, I didn't notice this, thanks for catching! This isn't the right fix though -- those directory names are meant to be used to generate the subdirectories under `out/`.

dcodeIO:
I see, thanks!

PR #599, opened 2017-08-30
Use AppendField in binary-reader-ir.cc

binji:
This matches the behavior in wast-parser.cc, and is clearer.

### DIFF
--- a/src/binary-reader-ir.cc
+++ b/src/binary-reader-ir.cc
@@ -303,8 +303,7 @@ Result BinaryReaderIR::OnType(Index index,
   FuncType& func_type = field->func_type;
   func_type.sig.param_types.assign(param_types, param_types + param_count);
   func_type.sig.result_types.assign(result_types, result_types + result_count);
-  module->func_types.push_back(&func_type);
-  module->fields.push_back(field.release());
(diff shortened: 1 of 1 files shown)


src/binary-reader-ir.cc
```diff
@@ -818,8 +786,7 @@ Result BinaryReaderIR::BeginDataSegment(Index index, Index memory_index) {
   auto field = MakeUnique<DataSegmentModuleField>(GetLocation());
   DataSegment& data_segment = field->data_segment;
   data_segment.memory_var = Var(memory_index, GetLocation());
-  module->data_segments.push_back(&data_segment);
-  module->fields.push_back(field.release());
+  module->AppendField(field.release());
```
sbc100: What are some of these .release() and others std::move()?  (sorry basic C++ question).

binji: `.release()` is used when the caller wants a raw pointer and `std::move()` is used when the caller wants a `std::unique_ptr`. It probably should be a `unique_ptr` in all cases, actually, since they're always taking ownership.

binji: PR to do this: #602.

PR #603, opened 2017-08-31
Update relocation names to match tool convention doc

binji:
See https://github.com/WebAssembly/tool-conventions/blob/master/Linking.md#relocation-sections


binji:
Still waiting on mac travis builder. I don't think there will be an issues. If there is, I'll fix in a subsequent PR.

PR #618, opened 2017-09-08
Move Interpreter::Opcodes into Opcode

binji:
Having the interpreter use its own opcodes was convenient because it
adds a few additional codes (e.g. alloca, br_unless, etc).

As soon as we add the atomic instructions, this count goes above 1-byte,
so we need some mechanism for storing those codes. We could just
increase the opcode size to uint16_t, but it's simpler to just use the
same encoding as the WebAssembly binary format.

The one drawback is that mapping from the encoding to Opcode is a
log2(n) operation currently. This doesn't affect the speed of running
tests at all because it is too insignificant, but it's worth mentioning.


sbc100:
Why not define the enum to have the same values as the encoding?

binji:
Because the encoding uses a prefix byte followed by LEB, so it can't really map 1-1. Maybe it would be fine to just do `(prefix << 8) | opcode` though.

PR #619, opened 2017-09-08
Add virtual destructor to Stream

binji:
Seems to be causing a warning on some builds; I didn't see it on Travis,
though.


binji:
Landing since this will likely fix the waterfall failures.

PR #624, opened 2017-09-13
Add an interface to enable features

Cellule:
Make it possible for a consumer other than a console application to set features.

### DIFF
--- a/src/feature.h
+++ b/src/feature.h
@@ -28,12 +28,8 @@ class Features {
   void AddOptions(OptionParser*);
 
 #define WABT_FEATURE(variable, flag, help) \
-  bool variable##_enabled() const { return variable##_enabled_; }
-#include "src/feature.def"
-#undef WABT_FEATURE
-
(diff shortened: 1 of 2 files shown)


src/feature.h
```diff
@@ -32,6 +32,11 @@ class Features {
 #include "src/feature.def"
 #undef WABT_FEATURE
 
+#define WABT_FEATURE(variable, flag, help) \
+  void variable##_enable() { variable##_enabled_ = true; }
```
binji: This would turn into `threads_enable()` which sounds a bit odd. Maybe `enable_threads()` instead?

src/feature.h
```diff
@@ -32,6 +32,11 @@ class Features {
 #include "src/feature.def"
 #undef WABT_FEATURE
 
+#define WABT_FEATURE(variable, flag, help) \
```
binji: Do you mind combining with the `WABT_FEATURE` define above?

PR #629, opened 2017-09-15
Correctly report the value of bad opcodes

sbc100:
Previously we were always reporting 0.

Partial fix for #627

### DIFF
--- a/src/opcode.cc
+++ b/src/opcode.cc
@@ -37,9 +37,6 @@ Opcode::Info Opcode::infos_[] = {
 #undef WABT_OPCODE
 
 // static
-Opcode::Info Opcode::invalid_info_ = {
-    "<invalid>", Type::Void, Type::Void, Type::Void, 0, 0, 0, 0,
-};
 
 // static
 Opcode Opcode::FromCode(uint32_t code) {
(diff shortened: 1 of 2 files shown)


test/binary/bad-opcode.txt
```diff
@@ -0,0 +1,23 @@
...
+section(CODE) {
+  count[1]
+  func {
+    locals[0]
+    invalid_op[0xff]
+  }
+}
+(;; STDERR ;;;
+Error running "wasm-objdump":
+*ERROR*: @0x0000001a: unexpected opcode: 195 (0xc3)
```
binji: Should be `0xff`, right?

src/opcode.cc
```diff
@@ -62,7 +59,9 @@ Opcode Opcode::FromCode(uint8_t prefix, uint32_t code) {
 }
 
 Opcode::Info Opcode::GetInfo() const {
-  return enum_ < Invalid ? infos_[enum_] : invalid_info_;
+  if (enum_ >= Invalid)
```
binji: I think we need to do something more clever in `Opcode::FromCode` above instead of returning `Opcode(Invalid)`. Maybe we can use a negative value for `enum_` to mean an invalid opcode with code `-enum_`.

binji:
Closing since https://github.com/WebAssembly/wabt/pull/630 landed.

PR #630, opened 2017-09-16
Correctly report the value of bad opcodes

binji:
Previously the opcode would always return 0.

Since `Opcode` only stores an enumeration, we represent invalid opcodes
as a negative number. This way `Opcode::GetInfo` can return the invalid
prefix and opcode number.

For now, these values are both 1-byte, but the current encoding can
be expanded to 2**24 opcodes.

This is a partial fix for #627.

### DIFF
--- a/src/opcode.cc
+++ b/src/opcode.cc
@@ -37,11 +37,6 @@ Opcode::Info Opcode::infos_[] = {
 #undef WABT_OPCODE
 
 // static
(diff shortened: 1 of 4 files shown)


binji:
Forked this from your PR @sbc100, hope you don't mind.

binji:
I added a comment to help explain a bit.

PR #636, opened 2017-09-22
[wasm-objdump] Continue reading sections on error

binji:
wasm-objdump wants to always try to read as much of the binary as
possible. This change adds a new `BinaryReader` option called
`stop_on_first_error` which can be set to `false` to enable this
behavior.

Currently this will bail on any error in a section, but will try to read
the next section. We may expand the error-handling behavior to try and
recover inside of a section as well, but it's probably best to keep it
simple for now.

Because `wasm-objdump` does multiple passes over the binary, the errors
would normally be displayed multiple times. We also add a `OnError`
override which will suppress the error message for all passes other than
the first.

I took this opportunity to clean up a `binary-reader.cc` a little too:

* Remove IN_SIZE macro and use a template instead
* Change C comments to C++ comments

### DIFF
--- a/src/binary-reader-objdump.cc
+++ b/src/binary-reader-objdump.cc
@@ -37,6 +37,8 @@ class BinaryReaderObjdumpBase : public BinaryReaderNop {
                           ObjdumpOptions* options,
                           ObjdumpState* state);
 
+  bool OnError(const char* message) override;
+
   Result BeginModule(uint32_t version) override;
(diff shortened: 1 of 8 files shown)


binji:
@sbc100 ping!

sbc100:
Would it be easy to split the cleanup (comments + templates) into its own CL?

binji:
@sbc100 Yeah, probably should. I'm always trying to sneak stuff in... :-)

PR #640, opened 2017-09-27
Some cleanup in binary-reader.cc

binji:
* Remove IN_SIZE macro and use a template instead
* Change C comments to C++ comments


binji:
Split out from https://github.com/WebAssembly/wabt/pull/636.

PR #647, opened 2017-10-06
Fix some fuzzer-found regressions

binji:
* Assertion/hang when parsing an invalid value type list
* Assertion parsing an invalid module field
* Assertion when two start sections are specified, and one of them has a
  named function

### DIFF
--- a/test/regress/regress-14.txt
+++ b/test/regress/regress-14.txt
@@ -4,7 +4,7 @@
     (return (i32.const 42))))
 
 (;; STDERR ;;;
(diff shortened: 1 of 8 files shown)


test/regress/regress-14.txt
```diff
@@ -0,0 +1,10 @@
+;;; ERROR: 1
+(module
+  (func (result(i32)
+    (return (i32.const 42))))
+
+(;; STDERR ;;;
+out/test/regress/regress-14.txt:3:17: error: unexpected token "i32", expected i32, i64, f32 or f64.
```
binji: Hm, this is a bad error.

PR #655, opened 2017-10-14
Update demos to add JS eval to wat2wasm demo

krisselden:
Some minor redesign of demos.

image


binji:
Very cool! Thanks for doing this. The only thing I was wondering was about whether it makes sense to use template strings and `let`, but I guess they've been around for a long time now so there's not much worry.

binji:
By the way, can you join the [community group](https://www.w3.org/community/webassembly/)? It's required to contribute to WebAssembly GH projects. Sorry I forgot to mention this before.

krisselden:
As far as I know, there is no browser that supports WebAssembly that does not also support the features I used.

I created an account and filled out the form requesting to join as a collaborator.

binji:
> As far as I know, there is no browser that supports WebAssembly that does not also support the features I used.

Right, but the library was compiled w/ asm.js so didn't require WebAssembly before. In any case, I think it should be fine.

> I created an account and filled out the form requesting to join as a collaborator.

Thanks!

Issue #543, opened 2017-06-29
Reorganize tools


binji:
We should change the tools to the following (adding on to #354):

- [x] `wat2wasm`: converts .wat text files only to .wasm binary files, and optionally validates
- [x] `wast2json`: converts .wast spec test files to .json and .wasm files
- ~~`wat-validate`: standalone .wat file validation, no conversion~~
- [x] `wasm2wat`: converts .wasm files to .wat text files, and optionally validates (rename of `wasm2wast`)
- [x] `wasm-validate`: standalone .wasm file validation, no conversion
- [x] `spectest-interp`: (better name?) run the spec tests in interpreter (current `wasm-interp` functionality)
- [x] `wasm-interp`: only run a single .wasm file, later adding passing arguments (see #198)

All of the functionality is available currently, this is just splitting it out. Thoughts?

sbc100:
I think maybe we can live with just one validate tool.  I would suggest you drop wat-validate.   If you want to validate a wat you can try to compile it with wat2wasm (which should validate by default), just like you would with a C++ source file.

binji:
Sounds good.

PR #664, opened 2017-10-31
Add wasm-validate tool

binji:
It's currently identical to wasm2wat, without the part to write out a
.wat file.


binji:
This is the final part of issue #543.

PR #673, opened 2017-11-21
Add various math operations to c-writer

kripken:


kripken:
Heh it looks like this conflicts with code that just landed which does the same stuff... so this can probably be closed.

binji:
Ah, thanks! Yeah, I started trying to see if I can pass the spec tests, so I'm implementing these even though no big wasm files I've found use them.

src/c-writer.cc
```diff
@@ -1654,28 +1672,43 @@ void CWriter::Write(const BinaryExpr& expr) {
...
+      WritePrefixBinaryExpr(expr.opcode, "I64_ROTR");
       break;
 
     case Opcode::F32Min:
+      WritePrefixBinaryExpr(expr.opcode, "fminf");
+      break;
+
     case Opcode::F64Min:
-      UNIMPLEMENTED(expr.opcode.GetName());
+      WritePrefixBinaryExpr(expr.opcode, "fmin");
```
binji: Didn't notice this! Looks like it has subtly different behavior w/ NaNs though:

```
fmin(3.0, NAN) => 3.0
(f32.min (f32.const 3.0) (f32.const nan)) => nan
```

From looking at http://en.cppreference.com/w/cpp/numeric/math/fmin.

kripken:
The binaryen fuzzer emits them, that's how I noticed ;)

binji:
(deleted mostly because origin/maths and origin/master both have the same prefix so my tab completion gets stuck :-})

PR #677, opened 2017-11-27
   Wabt simd v128.const instruction initial PR:

lizhengxing:
This is an initial PR for Wabt simd support to get some early feedback.
   This PR only implement the v128.const instruction.

   It will construct a simd 128-bits const.
   The expected wat format is:
   v128.const i32 0x00000000 0x11111111 0x22222222 0x33333333

   For simplify, this PR only implement the wat2wasm and wasm2wat functions.

   The following PRs will implement the full functions in Wabt.
   such as: Interp, objdump, logging, etc....

### DIFF
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -83,6 +83,7 @@ class BinaryReader {
   Result ReadU32(uint32_t* out_value, const char* desc) WABT_WARN_UNUSED;
   Result ReadF32(uint32_t* out_value, const char* desc) WABT_WARN_UNUSED;
   Result ReadF64(uint64_t* out_value, const char* desc) WABT_WARN_UNUSED;
+  Result ReadV128(v128* out_value, const char* desc) WABT_WARN_UNUSED;
   Result ReadU32Leb128(uint32_t* out_value, const char* desc) WABT_WARN_UNUSED;
   Result ReadS32Leb128(uint32_t* out_value, const char* desc) WABT_WARN_UNUSED;
   Result ReadS64Leb128(uint64_t* out_value, const char* desc) WABT_WARN_UNUSED;
@@ -234,6 +235,10 @@ Result BinaryReader::ReadF64(uint64_t* out_value, const char* desc) {
   return ReadT(out_value, "double", desc);
 }
 
+Result BinaryReader::ReadV128(v128* out_value, const char* desc) {
+  return ReadT(out_value, "v128", desc);
+}
+
 Result BinaryReader::ReadU32Leb128(uint32_t* out_value, const char* desc) {
   const uint8_t* p = state_.data + state_.offset;
   const uint8_t* end = state_.data + read_end_;
@@ -327,6 +332,7 @@ static bool is_concrete_type(Type type) {
     case Type::I64:
     case Type::F32:
     case Type::F64:
+    case Type::V128:
       return true;
 
     default:
@@ -388,6 +394,15 @@ Result BinaryReader::ReadInitExpr(Index index, bool require_i32) {
       uint64_t value_bits = 0;
       CHECK_RESULT(ReadF64(&value_bits, "init_expr f64.const value"));
       CALLBACK(OnInitExprF64ConstExpr, index, value_bits);
+      break;
+    }
+
+    case Opcode::V128Const: {
+      ERROR_UNLESS_OPCODE_ENABLED(opcode);
+      v128 value_bits;
+      ZeroMemory(value_bits);
+      CHECK_RESULT(ReadV128(&value_bits, "init_expr v128.const value"));
+      CALLBACK(OnInitExprV128ConstExpr, index, value_bits);
       break;
     }
 
@@ -626,6 +641,16 @@ Result BinaryReader::ReadFunctionBody(Offset end_offset) {
         CHECK_RESULT(ReadF64(&value_bits, "f64.const value"));
         CALLBACK(OnF64ConstExpr, value_bits);
         CALLBACK(OnOpcodeF64, value_bits);
+        break;
+      }
+
+      case Opcode::V128Const: {
+        ERROR_UNLESS_OPCODE_ENABLED(opcode);
+        v128 value_bits;
+        ZeroMemory(value_bits);
+        CHECK_RESULT(ReadV128(&value_bits, "v128.const value"));
+        CALLBACK(OnV128ConstExpr, value_bits);
+        CALLBACK(OnOpcodeV128, value_bits);
         break;
       }
 
--- a/src/common.h
+++ b/src/common.h
@@ -90,6 +90,10 @@
 #define PRIaddress "u"
 #define PRIoffset PRIzx
 
+struct v128 {
+  uint32_t v[4];
+};
+
 namespace wabt {
 
 typedef uint32_t Index;    // An index into one of the many index spaces.
@@ -189,6 +193,7 @@ enum class Type {
   I64 = -0x02,
   F32 = -0x03,
   F64 = -0x04,
+  V128 = -0x05,
   Anyfunc = -0x10,
   Func = -0x20,
   Void = -0x40,
@@ -297,6 +302,8 @@ static WABT_INLINE const char* GetTypeName(Type type) {
       return "f32";
     case Type::F64:
       return "f64";
+    case Type::V128:
+      return "v128";
     case Type::Anyfunc:
       return "anyfunc";
     case Type::Func:
--- a/src/wast-parser.cc
+++ b/src/wast-parser.cc
@@ -586,7 +586,7 @@ bool WastParser::ParseVarListOpt(VarVector* out_var_list) {
 Result WastParser::ParseValueType(Type* out_type) {
   WABT_TRACE(ParseValueType);
   if (!PeekMatch(TokenType::ValueType)) {
-    return ErrorExpected({"i32", "i64", "f32", "f64"});
+    return ErrorExpected({"i32", "i64", "f32", "f64", "v128"});
   }
 
   *out_type = Consume().type();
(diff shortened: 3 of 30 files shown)


lizhengxing:
@binji Review please!
@arunetm fyi

src/wast-parser.cc
```diff
@@ -1479,6 +1480,14 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
       break;
     }
 
+    // Simd
+    case TokenType::SimdConst: {
```
binji: I think it'd be better to return a TokenType::Const but with a different opcode, like the other const instructions.

src/wast-parser.cc
```diff
@@ -1488,6 +1497,96 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
   return Result::Ok;
 }
 
+// Simd
+// The expected V128 const lists is:
+// i32 0xXXXXXXXX 0xXXXXXXXX 0xXXXXXXXX 0xXXXXXXXX
+Result WastParser::ParseSimdConstList(ConstVector* consts) {
```
binji: Probably better if this had a different name; currently "ParseConstList" name is used to parse a list of constant values for assertions, like `(i32.const 1) (f32.const 2)` etc.

This only seems to be used by `ParseSimdConst` below, maybe it can just be inlined? Or do you think you'll need it elsewhere?

src/wast-parser.cc
```diff
@@ -1488,6 +1497,96 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
...
+          Error(const_.loc, "invalid literal \"%s\"", literal.text.c_str());
+        }
+
+        consts->push_back(const_);
+      }
+      break;
+    }
+
+    default:
+      assert(!"ParseSimdConstListcalled with invalid input const type");
```
binji: This should be an `Error`, not an assert.

src/wast-parser.cc
```diff
@@ -1488,6 +1497,96 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
...
+  switch (opcode) {
+    case Opcode::V128Const: {
+      const_->type = Type::V128;
+      // 1. check the const type and range.
+      int nSize = simdconsts.size();
+      assert(nSize == 4);
+      // 2. Copy it into const_.v128_bits.
+      union i128bits{
+        int32_t i32Array[4];
+        uint8_t i8Array[16];
```
binji: Why is the union needed? I think you can use the i32array directly in memcpy below.

src/wast-parser.cc
```diff
@@ -1488,6 +1497,96 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
...
+
+  switch (opcode) {
+    case Opcode::V128Const: {
+      const_->type = Type::V128;
+      // 1. check the const type and range.
+      int nSize = simdconsts.size();
+      assert(nSize == 4);
+      // 2. Copy it into const_.v128_bits.
+      union i128bits{
+        int32_t i32Array[4];
```
binji: Should be uint32_t, I think

src/wast-parser.cc
```diff
@@ -1488,6 +1497,96 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
...
+
+    default:
+      assert(!"ParseSimdConstListcalled with invalid input const type");
+      return Result::Error;
+  }
+
+  return Result::Ok;
+}
+
+Result WastParser::ParseSimdConst(Const* const_) {
```
binji: I think it would be better if this was called from `ParseConst` so we could share places where that is used currently.

src/wat-writer.cc
```diff
@@ -452,6 +452,19 @@ void WatWriter::WriteConst(const Const& const_) {
       break;
     }
 
+    // Simd
+    case Type::V128: {
+      int32_t i32Array[4];
+      memcpy(&i32Array, &(const_.v128_bits), sizeof(v128));
+      WritePutsSpace(Opcode::V128Const_Opcode.GetName());
+      Writef("i32 0x%08x", i32Array[0]);
```
binji: You may as well roll this into one `Writef` call, I think

src/common.h
```diff
@@ -90,6 +90,9 @@
 #define PRIaddress "u"
 #define PRIoffset PRIzx
 
+// Simd.
+typedef __m128 v128;
```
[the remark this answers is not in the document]
lizhengxing: I'm not sure what's "define this internally" means.  Move "typedef __m128 v128" to other header file?

BTW:
How should we deal with the platform that doesn't support 128-bit integer?

src/wast-parser.cc
lizhengxing: Yes.  I do wanted to implement v128.const like this way at beginning.

But parse the simd const lists is slightly different to Int32/int64/F32/F64 const.  I'm worried that the changes will mess up the original code.  That's why I use a dedicated token for simd const.

Anyway, I'm re-implementing it and try to keep the code as clean as current.

src/wast-parser.cc
lizhengxing: Inline it will cause ParseSimdConst Function bigger.  Let me try it.

lizhengxing:
@binji 
Thanks for your review and nice comments. please check my questions about the comments. thanks again.

@arunetm FYI

src/common.h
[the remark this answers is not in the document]
binji: Sorry, I meant define something in wabt directly, like:
```
struct v128 {
  uint32_t v[0];
};
```

We don't really need to use the native type, since most of the time we'll just be loading and storing the values anyway. The interpreter may want to use a real v128 type, but even there it might be nice to do it with scalar values instead.

src/wast-parser.cc
binji: Yeah, I think you'll have to restructure `ParseConst` a bit, but that's OK. Don't worry too much about messing up the code. :-)

lizhengxing:
@binji 
Sorry, Some other issues blocked me submit the updated PR later.
Please review it if you are available, thanks!

The main changes are:
1. updated the PR according to upstream's comments.
2. added some tests.

@arunetm FYI

src/common.h
```diff
@@ -90,6 +90,10 @@
 #define PRIaddress "u"
 #define PRIoffset PRIzx
 
+typedef struct v128_data {
```
binji: This is C++, so you can just do `struct v128 { ... };`

src/wat-writer.cc
```diff
@@ -485,6 +485,15 @@ void WatWriter::WriteConst(const Const& const_) {
       break;
     }
 
+    case Type::V128: {
+      int32_t i32Array[4];
+      memcpy(&i32Array, &(const_.v128_bits), sizeof(v128));
```
binji: Remove and just use `const_.v128_bits.v[0]`, etc. below.

src/wast-parser.cc
```diff
@@ -1499,28 +1499,85 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
...
+        if (Failed(result)) {
+          Error(loc, "invalid literal \"%s\"", literal.text.c_str());
+          return Result::Error;
+        }
+      }
+      break;
+    }
+
+    default:
+      Error(const_->loc, "ParseSimdConst called with invalid input const type, Type::I32 expected");
```
binji: Since this will be displayed to the user, it's better to avoid internal function names. Maybe just "expected i32 at start of simd constant". What do you think?

src/wast-parser.cc
```diff
@@ -1545,13 +1602,23 @@ Result WastParser::ParseConst(Const* const_) {
       result = ParseDouble(literal.type, s, end, &const_->f64_bits);
       break;
 
+    case Opcode::V128Const:
+      const_->type = Type::V128;
+      // Parse V128 Simd Const (16 bytes).
+      result = ParseSimdConst(const_, in_type, sizeof(v128));
```
binji: Since this is always going to be `size(v128)` for now, probably best to just hard code it in `ParseSimdConst`.

src/wast-parser.cc
```diff
@@ -1499,28 +1499,85 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
...
+// The current expected V128 const lists is:
+// i32 0xXXXXXXXX 0xXXXXXXXX 0xXXXXXXXX 0xXXXXXXXX
+Result WastParser::ParseSimdConst(Const* const_, Type in_type, int32_t nSimdConstBytes) {
+  WABT_TRACE(ParseSimdConst);
+
+  // Parse the Simd Consts according to input data type.
+  switch (in_type) {
+    case Type::I32: {
+      const_->loc = GetLocation();
+      int nCount = nSimdConstBytes/sizeof(uint32_t);
```
binji: We can hardcode this for now, since we know that `nSimdConstBytes` is always `sizeof(v128)`.

binji: nit: `count` instead of `nCount`

src/binary-reader.cc
```diff
@@ -391,6 +397,14 @@ Result BinaryReader::ReadInitExpr(Index index, bool require_i32) {
       break;
     }
 
+    case Opcode::V128Const: {
+      v128 value_bits;
```
binji: Since SIMD is a future feature, we want to only allow it if the feature flag is enabled. See other examples using the `ERROR_UNLESS_OPCODE_ENABLED` macro in this file.

src/binary-reader.cc
```diff
@@ -629,6 +643,15 @@ Result BinaryReader::ReadFunctionBody(Offset end_offset) {
         break;
       }
 
+      case Opcode::V128Const: {
+        v128 value_bits;
```
binji: Same here.

src/wast-parser.cc
```diff
@@ -1545,13 +1602,23 @@ Result WastParser::ParseConst(Const* const_) {
       result = ParseDouble(literal.type, s, end, &const_->f64_bits);
       break;
 
+    case Opcode::V128Const:
+      const_->type = Type::V128;
```
binji: We only want to allow parsing this if the SIMD feature is enabled, see other examples in this file using `ErrorUnlessOpcodeEnabled`.

src/binary-reader.cc
lizhengxing: Agree.

src/binary-reader.cc
lizhengxing: Agree.

src/common.h
lizhengxing: Agree.

src/wast-parser.cc
lizhengxing: The reason why I didn't use hardcode here is: 
I want to this function can be used for future 256 or 512 simd type.

lizhengxing: Agree.

src/wast-parser.cc
lizhengxing: Agree.

src/wast-parser.cc
lizhengxing: Agree.

src/wast-parser.cc
lizhengxing: The reason here is for future possible 256, 512 simd type too.

src/wat-writer.cc
lizhengxing: Agree.

lizhengxing:
@binji 
Thanks for your review. I will update the PR. thanks!

@arunetm FYI!

lizhengxing:
@binji 
I updated the PR according to your suggestion except for the hardcode for simd value size.

BTW:
1. Running test/run-tests.py --enable-simd -v --bindir out/clang/Debug/ will meet 
"run-tests.py: error: unrecognized arguments: --enable-simd"

Would you please take a look for it?

2. Some tests may be mixed simd and non-simd functions, How should I control the --enable-simd test option when I submit a PR.

Thanks!

@arunetm FYI!

binji:
You can specify the flags in the test file itself, so in `test/parse/expr/bad-const-v128-type-i32-expected.txt`, you can write:
```
;;; ERROR: 1
;;; FLAGS: --enable-simd
(module (func v128.const ...
```

lizhengxing:
@binji 
Thanks! 
How about test/parse/expr/const.txt?  extract the v128.const to another dedicated test file or use "FLAGS: --enable-simd" in test/parse/expr/const.txt?  

@arunetm FYI

binji:
> How about test/parse/expr/const.txt? extract the v128.const to another dedicated test file or use "FLAGS: --enable-simd" in test/parse/expr/const.txt?

It's OK to just add the `--enable-simd` flag to that file.

lizhengxing:
@binji 
Thanks!

I updated the test files, please review it again once you are available .

@arunetm FYI!

PR #678, opened 2017-11-28
Fix ubsan build and remove workaround in favor of ubsan.blacklist

sbc100:


sbc100:
Do you know how to check for clang version in cmake so we can warn about doing ubsan with clang < 3.8?

PR #682, opened 2017-12-01
wat2wasm: Always write output to a file

sbc100:
This involves deriving the output filename from the input
filename when none is otherwise specified.

Also split out filename handling utilities and add unittests
for them.

### DIFF
--- a/src/filenames.h
+++ b/src/filenames.h
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2016 WebAssembly Community Group participants
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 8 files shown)


sbc100:
Happy to split out the filename utiiity function into its own change if you prefer?

src/filenames.h
```diff
@@ -0,0 +1,48 @@
...
+// "foo" => ""
+//  "/foo/bar/foo.wasm" => ".wasm"
+string_view GetExtension(string_view filename);
+
+// Strip extension, e.g.:
+//
+// "foo", => "foo"
+// "foo.json" => "foo"
+//  "foo.wasm" => "foo"
+// "foo.bar" => "foo.bar"
```
binji: this line is wrong now

src/test-filenames.cc
```diff
@@ -0,0 +1,58 @@
...
+  ASSERT_EQ(len, sv.size());
+  for (size_t i = 0; i < len; ++i) {
+    ASSERT_EQ(s[i], sv[i]);
+  }
+}
+
+}  // end anonymous namespace
+
+TEST(filenames, GetBasename) {
+  assert_string_view_eq("foo.txt", GetBasename("/bar/dir/foo.txt"));
```
binji: empty string for all these tests?

src/filenames.h
```diff
@@ -0,0 +1,48 @@
...
+string_view StripExtension(string_view s);
+
+// Strip everything up to and including the last slash, e.g.:
+//
+// "/foo/bar/baz", => "baz"
+// "/usr/local/include/stdio.h", => "stdio.h"
+// "foo.bar", => "foo.bar"
```
binji: Example that shows this works for `\` too?

sbc100:
Done

PR #684, opened 2017-12-01
Make more use of string_view over 'const char *'

sbc100:


src/wast-lexer.h
```diff
@@ -59,7 +59,7 @@ class WastLexer {
 
   std::unique_ptr<LexerSource> source_;
   LexerSourceLineFinder line_finder_;
-  const char* filename_;
+  string_view filename_;
```
binji: maybe this one should be a `std::string` instead.

Issue #685, opened 2017-12-03
Strange symbols don't roundtrip through wasm2wat and wat2wasm


jtgeibel:
Over in https://github.com/alexcrichton/wasm-gc/pull/8 I'm experimenting with demangling rust symbols in wasm output.  This results in symbol names such as `<alloc::vec::Vec<T>>::extend_from_slice` and `<&'a T as core::fmt::Debug>::fmt`.

While `wasm2wat` produces very nice looking output, I get a segmentation fault when feeding this output back into `wat2wasm`.  I'm guessing that the space characters are causing issues when parsing the `.wat` file but other characters could be causing problems as well.

[example.wasm.gz](https://github.com/WebAssembly/wabt/files/1524991/wasm.gc.wasm.gz)

binji:
Ah, I see! Yeah, this seems to be a limitation of the text format, since any utf-8 sequence is allowed for a name in the binary format. For now, I guess `wasm2wat` should probably just produce a replacement character.

I wasn't able to repro the segfault though, here's what I tried:

```
$ wasm2wat wasm.gc.wasm -o wasm.gc.wat
$ wat2wasm wasm.gc.wat
wasm.gc.wat:82:15: error: unexpected token T, expected ).
  (func $<&'a T as core::fmt::Display>::fmt::h0bee0bd948d5b839 (type 2) (para...
              ^
wasm.gc.wat:82:17: error: unexpected token as.
  (func $<&'a T as core::fmt::Display>::fmt::h0bee0bd948d5b839 (type 2) (para...
                ^^
wasm.gc.wat:82:20: error: unexpected token core::fmt::Display>::fmt::h0bee0bd948d5b839.
  (func $<&'a T as core::fmt::Display>::fmt::h0bee0bd948d5b839 (type 2) (para...
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
wasm.gc.wat:82:70: error: unexpected token 2, expected (.
...::fmt::Display>::fmt::h0bee0bd948d5b839 (type 2) (param i32 i32) (result i32)
                                                 ^
```

PR #687, opened 2017-12-04
Replace invalid characters with `_` in wat-writer

binji:
We should add a way to express all names that are valid in the binary
format in the text format. Until then, it's better to be able to
produce a valid `wat` file.

See issue #685.

### DIFF
--- a/src/wat-writer.cc
+++ b/src/wat-writer.cc
@@ -56,6 +56,20 @@ static const uint8_t s_is_char_escaped[] = {
     1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
     1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1};
 
+// This table matches the characters allowed by wast-lexer.cc for `symbol`.
+// The disallowed printable characters are: "(),;[]{} and <space>.
+static const uint8_t s_valid_name_chars[256] = {
+    //         0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
+    /* 0x00 */ 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    /* 0x10 */ 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    /* 0x20 */ 0, 1, 0, 1, 1, 1, 1, 1, 0, 0, 1, 1, 0, 1, 1, 1,
+    /* 0x30 */ 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 1, 1, 1, 1,
+    /* 0x40 */ 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+    /* 0x50 */ 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 1, 0, 1, 1,
+    /* 0x60 */ 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+    /* 0x70 */ 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 1, 0, 1, 0,
+};
+
 enum class NextChar {
   None,
   Space,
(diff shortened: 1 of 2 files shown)


src/wat-writer.cc
```diff
@@ -56,6 +56,20 @@ static const uint8_t s_is_char_escaped[] = {
     1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
     1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1};
 
+// This table matches the characters allowed by wast-lexer.cc for `symbol`.
+// The disallowed printable characters are: "(),;[]{} and <space>.
```
jtgeibel: @binji thanks a lot for investigating!  I was hoping a few additional characters would be allowed in the lexer, but oh well.

@alexcrichton and @shepmaster for your reference the above characters will be replaced by`_` in `wasm2wat`.  Out of these characters I know `(),[] and <space>` are used in rust unmangled symbol names.  Below are a few examples I was worried about but now that I see them it doesn't seem too bad.

```
<(i32, i32) as wasm::Wasm>::wasm::hfc09b9ea108eba04
<_i32__i32__as_wasm::Wasm>::wasm::hfc09b9ea108eba04

<&'a T as core::fmt::Debug>::fmt::h273375ad95a1eb7b
<&'a_T_as_core::fmt::Debug>::fmt::h273375ad95a1eb7b

<core::fmt::Write::write_fmt::Adapter<'a, T> as core::fmt::Write>::write_char::h25e7e7a8edf51d28
<core::fmt::Write::write_fmt::Adapter<'a__T>_as_core::fmt::Write>::write_char::h25e7e7a8edf51d28

<std::error::<impl core::convert::From<alloc::string::String> for alloc::boxed::Box<std::error::Error + core::marker::Sync + core::marker::Send + 'static>>::from::StringError as std::error::Error>::description::hdf1479d893d74cf2
<std::error::<impl_core::convert::From<alloc::string::String>_for_alloc::boxed::Box<std::error::Error_+_core::marker::Sync_+_core::marker::Send_+_'static>>::from::StringError as std::error::Error>::description::hdf1479d893d74cf2
```

jtgeibel:
I've been thinking more about long term solutions and have a few ideas to float.  I've tried to combine existing constructs defined in the spec so that existing code can be leveraged.

At call sites, if `has_invalid_chars` is true then we could fall back on the function index and provide a comment.  The comment would be encoded as a string, mainly to handle the case where a symbol contains a `U+0A` which would end the line comment.  This would also align with how string literals end up encoded in the data section.  It would look like this:

```
(call $traditional_symbol
  (get_local $p0)
  (get_local $p1)))
(call 35 ;; "core::fmt::num::<impl core::fmt::Display for usize>::fmt::h2012ce6f955e8c94"
  (get_local $p0)
  (get_local $p1)))
```

At the function declaration, to use the comment trick we would need to move the type, param, and result fields to the next line.  It would look like this:

```
(func 35 ;; "core::fmt::num::<impl core::fmt::Display for usize>::fmt::h2012ce6f955e8c94"
  (type $t3) (param $p0 i32) (param $p1 i32) (result i32)
  (local $l0 i32) (local $l1 i32) (local $l2 i32) (local $l3 i32)
  ...
```

Alternatively, it may be possible to add a `(name "...")` construct to the spec.

```
(func 35 (name "core::fmt::num::<impl core::fmt::Display for usize>::fmt::h2012ce6f955e8c94") (type $t3) (param $p0 i32) (param $p1 i32) (result i32)
  (local $l0 i32) (local $l1 i32) (local $l2 i32) (local $l3 i32)
  ...
```

I think the main advantage of this option is that it would make it easier when parsing a `.wat` file than trying to detect a magic comment.  I guess this brings me to a question.  In general is it expected that `wasm2wat | wat2wasm` produces byte identical output and if so does that apply to the custom name section?  If that isn't a goal, then we would just accept that strange symbols may not survive the round-trip.  If this is a goal, then we would also have to consider how this intersects with disambiguation of identical symbols and named local variables.

binji:
> I guess this brings me to a question. In general is it expected that wasm2wat | wat2wasm produces byte identical output and if so does that apply to the custom name section? 

Personally I think it should, in general. Since you can express short varints with a longer length they won't necessarily by byte identical, but they'll mean the same thing.

The broader topic of round-tripping custom sections came up here as well: https://github.com/WebAssembly/design/issues/1153. Using the annotation syntax there, it might look like:

```
(func $myFunc (@name "myFunkyNameFunc()[];;;")
  ...
)
```

binji:
Gonna land this for now since it is better than producing a bad `.wat` file.

PR #691, opened 2017-12-09
[cleanup] Always use braces with if

binji:


sbc100:
Hmm.. llvm has the opposite policy and I've grown the like it for being concise.

binji:
There are things I like about it too, but it's pretty rare to see it in a style guide these days. To the extent that new languages don't even allow it! I figure we should try to be on the right side of history here. :-)

PR #695, opened 2017-12-18
Python3 - just testing the water

sbc100:


binji:
I've been trying to keep things 2/3 compatible. AFAIK it's mostly there, since I do a bunch of testing on my arch laptop (which defaults to python3)

sbc100:
I know .. but I've been wanting to experiment with type annotations.      And I was kinda thinking, why do we need python2 compat here?

binji:
Ah, I see.

CMakeLists.txt
```diff
@@ -373,7 +373,7 @@ if (NOT EMSCRIPTEN)
   endif ()
 
   # test running
-  find_package(PythonInterp 2.7 REQUIRED)
+  find_package(PythonInterp 3.4 REQUIRED)
```
rathann: A better change would be:
``` diff
-    find_package(PythonInterp 2.7 REQUIRED)
+    find_package(Python COMPONENTS Interpreter 3.4 REQUIRED)
```
or even

``` diff
-    find_package(PythonInterp 2.7 REQUIRED)
+    find_package(Python3 COMPONENTS Interpreter 3.4 REQUIRED)
```

See https://cmake.org/cmake/help/latest/module/FindPython.html and https://cmake.org/cmake/help/latest/module/FindPython3.html .

binji:
This is pretty old, and we have python3 testing on CI now. Let's close this.

sbc100:
But this change was to drop support for python2.. which allows us to cleanup some stuff.

CMakeLists.txt
sbc100: I think I tried that and ran into some issues.  Can do leave that for a followup?

sbc100:
Reopened as https://github.com/WebAssembly/wabt/pull/1321

PR #698, opened 2017-12-18
Add SIMD suppor in run-interp.py/run-objdump.py/run-roundtrip.py

lizhengxing:
Currently run-interp.py doesn't recongize --enable-simd flag in test file.

This PR add SIMD suppor in run-interp.py/run-objdump.py/run-roundtrip.py.


lizhengxing:
@binji 
Currently run-interp.py doesn't recongize --enable-simd flag in test file. I added the --enable-simd flag support in run-interp.py/run-objdump.py/run-roundtrip.py files.

@arunetm FYI!

PR #699, opened 2017-12-19
Add support for init function in linking metadata

sbc100:
Add support for init function in linking metadata
    
Also remove support for DataAlignment which was removed
from the "spec" (Linking.md) and only output DataSize
if its non-zero.

### DIFF
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -1320,6 +1320,7 @@ Result BinaryReader::ReadLinkingSection(Offset section_size) {
           CHECK_RESULT(ReadU32Leb128(&func, "function index"));
           CALLBACK(OnInitFunction, priority, func);
         }
+        break;
       default:
         // Unknown subsection, skip it.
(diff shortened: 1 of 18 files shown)


src/binary-reader.cc
```diff
@@ -1318,7 +1310,16 @@ Result BinaryReader::ReadLinkingSection(Offset section_size) {
...
+      case LinkingEntryType::InitFunctions:
+        CHECK_RESULT(ReadU32Leb128(&count, "info count"));
+        CALLBACK(OnInitFunctionCount, count);
+        while (count--) {
+          uint32_t priority;
+          uint32_t func;
+          CHECK_RESULT(ReadU32Leb128(&priority, "priority"));
+          CHECK_RESULT(ReadU32Leb128(&func, "function index"));
+          CALLBACK(OnInitFunction, priority, func);
+        }
```
binji: break;

PR #702, opened 2017-12-19
Wabt simd v128.const instruction PR (Part 2):

lizhengxing:
This is a sequential PR of Wabt simd v128.const instruction initial PR (096f7ab84749b5725a5780671111434c34cc17ea)

The PR implement wasm interp, objdump and validate functions for v128.const.

### DIFF
--- a/src/binary-reader-interp.cc
+++ b/src/binary-reader-interp.cc
@@ -241,6 +241,7 @@ class BinaryReaderInterp : public BinaryReaderNop {
   wabt::Result EmitI8(uint8_t value);
   wabt::Result EmitI32(uint32_t value);
   wabt::Result EmitI64(uint64_t value);
+  wabt::Result EmitV128(v128 value);
   wabt::Result EmitI32At(IstreamOffset offset, uint32_t value);
   wabt::Result EmitDropKeep(uint32_t drop, uint8_t keep);
   wabt::Result AppendFixup(IstreamOffsetVectorVector* fixups_vector,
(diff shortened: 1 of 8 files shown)


lizhengxing:
@binji 
This is part 2 of v128.const instruction PR, please review it, thanks!

@arunetm FYI!

src/interp.cc
```diff
@@ -71,6 +71,10 @@ std::string TypedValueToString(const TypedValue& tv) {
       return StringPrintf("f64:%f", value);
     }
 
+    case Type::V128:
+      return StringPrintf("v128:0x%08x 0x%08x 0x%08x 0x%08x", tv.value.v128_bits.v[0],\
```
binji: nit: remove `\`

src/binary-reader-interp.cc
```diff
@@ -425,6 +426,10 @@ wabt::Result BinaryReaderInterp::EmitI64(uint64_t value) {
   return EmitData(&value, sizeof(value));
 }
 
+wabt::Result BinaryReaderInterp::EmitI128(v128 value) {
```
binji: `EmitV128` is better here, I think

src/interp.cc

src/binary-reader-interp.cc
lizhengxing: Done.

lizhengxing:
@binji 
Updated the PR, please check it, thanks!

@arunetm FYI!

PR #704, opened 2017-12-20
SIMD: i8x16.splat instruction implementation.

lizhengxing:


lizhengxing:
@binji 
This PR implements the simd i8x16.splat instruction. Thanks!

@arunetm FYI!

src/interp.cc
```diff
@@ -1214,6 +1214,25 @@ ValueTypeRep<T> Xchg(ValueTypeRep<T> lhs_rep, ValueTypeRep<T> rhs_rep) {
   return rhs_rep;
 }
 
+// i(8,16,32,64) f(32,64) X (2,4,8,16) splat ==> v128
+template <typename T, typename V>
+ValueTypeRep<T> SimdSplat(V LaneData) {
+  union SimdBits{
```
binji: I think this typecast through a union is undefined behavior. I'd prefer we use `Bitcast` here, which uses memcpy under the hood, which I believe is OK. So something like:
```
V lanes[sizeof(T) / sizeof(V)];
...
T data = Bitcast<T>(lanes);
```

For this to work properly though, `Bitcast` will have to be modified to take its argument by rvalue reference: `Dst Bitcast(Src&& value)`

src/interp.cc
```diff
@@ -2226,6 +2245,12 @@ Result Thread::Run(int num_instructions) {
         break;
       }
 
+      case Opcode::I8X16Splat: {
+        uint8_t dLane = static_cast<uint8_t>(Pop<uint32_t>());
```
binji: nit: style for local variables is `snake_case`, no hungarian notation. Same for above

src/interp.cc
lizhengxing: Ok, I will update the PR.

BTW: 
I still don't understand why you say "this typecast through a union is undefined behavior", Would you please explain it with more details if possible?

Thanks!

src/interp.cc
lizhengxing: Got it.

src/interp.cc

src/interp.cc

lizhengxing:
@binji 
The PR was updated according to your suggestion. PTAL!

@arunetm FYI!

src/interp.cc
```diff
@@ -1214,6 +1214,23 @@ ValueTypeRep<T> Xchg(ValueTypeRep<T> lhs_rep, ValueTypeRep<T> rhs_rep) {
   return rhs_rep;
 }
 
+// i(8,16,32,64) f(32,64) X (2,4,8,16) splat ==> v128
+template <typename T, typename V>
+ValueTypeRep<T> SimdSplat(V Lane_Data) {
```
binji: should be all lowercase, like `lane_data`, but we can clean that up in a future PR.

lizhengxing: @binji 
Got it, thanks!

PR #705, opened 2017-12-26
SIMD: i16x8.splat i32x4.splat i64x2.splat f32x4.splat f64x2.splat ins…

lizhengxing:
…truction implementation.

### DIFF
--- a/src/interp.cc
+++ b/src/interp.cc
@@ -1216,19 +1216,19 @@ ValueTypeRep<T> Xchg(ValueTypeRep<T> lhs_rep, ValueTypeRep<T> rhs_rep) {
 
 // i(8,16,32,64) f(32,64) X (2,4,8,16) splat ==> v128
 template <typename T, typename V>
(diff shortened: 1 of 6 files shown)


lizhengxing:
@binji 
This PR implements the simd i16x8 / i32x4 / i64x2 / f32x4 / f64x2.splat instructions. PTAL, Thanks!

@arunetm FYI!

src/interp.cc
```diff
@@ -2249,6 +2249,36 @@ Result Thread::Run(int num_instructions) {
         break;
       }
 
+      case Opcode::I16X8Splat: {
+        uint16_t dLane = static_cast<uint16_t>(Pop<uint32_t>());
```
binji: nit: `lane`

binji: nit: `static_cast` is not needed since `Pop<T>` returns type `T` -- here and below

lizhengxing:
@binji 
Update the PR according to feedbacks, PTAL, thanks!

@arunetm FYI!

PR #710, opened 2018-01-08
Add wasm2c tool

binji:
Add `wasm2c`, a new tool that reads a `.wasm` file and generates a C
source file and its accompanying header file. The C output currently
only supports gcc/clang compilers, since it uses builtins for some
functionality.

The resulting C code is not standalone; there are runtime functions that
must be provided, as well as pointers to all imports.

The C runtime symbols that must be provided are as follows:

* `void wasm_rt_trap(wasm_rt_trap_t code)`:
  Called when the WebAssembly code traps. This function must not return.
* `u32 wasm_rt_register_func_type(u32 param_count, u32 result_count, ...)`:
  Register a function type with the given signature. This function must
  check whether this signature has already been registered and return
  the original index.
* `void wasm_rt_allocate_memory(wasm_rt_memory_t*, u32 initial, u32 max)`:
  Allocate the memory buffer for the given memory object, given the
  number of pages. The memory must be zeroed before returning.
* `u32 wasm_rt_grow_memory(wasm_rt_memory_t*, u32 delta)`:
  Grow memory by the given number of pages. If allocation fails, or the
  new pages size is larger than the maximum, return -1. Otherwise return
  the previous number of pages. The newly allocated memory must be
  zeroed.
* `void wasm_rt_allocate_table(wasm_rt_table_t*, u32 initial, u32 max)`:
  Allocate the buffer for the given table object. The buffer must be
  zeroed before returning.
* `u32 wasm_rt_call_stack_depth`:
  A symbol that tracks the current call stack depth. If this value
  exceeds `WASM_RT_MAX_CALL_STACK_DEPTH` then a trap occurs. This value
  defaults to 500, but can redefined.

An example implementation can be found in `spec-wasm2c-prefix.c`.

All functionality from the WebAssembly MVP is supported, and the
generated code passes all of the core spec tests. There is a new test
tool called `run-spec-wasm2c.py` which runs the following:

* `wast2json` to convert the spec test to json and wasm files
* `wasm2c` to convert the wasm to C source and headers
* a C compiler (default `cc`) to compile and link all C source files,
  including a C test runner (`spec-wasm2c-prefix.c`)
* Finally, the resulting executable to produce output

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -1,0 +1,2437 @@
+/*
+ * Copyright 2017 WebAssembly Community Group participants
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "src/c-writer.h"
+
+#include <cctype>
+#include <cinttypes>
+#include <map>
+#include <set>
+
+#include "src/cast.h"
+#include "src/common.h"
+#include "src/ir.h"
+#include "src/literal.h"
+#include "src/stream.h"
+#include "src/string-view.h"
+
+#define INDENT_SIZE 2
+
+#define UNIMPLEMENTED(x) printf("unimplemented: %s\n", (x)), abort()
+
+namespace wabt {
+
+namespace {
+
+struct Label {
+  Label(LabelType label_type,
+        const std::string& name,
+        const BlockSignature& sig,
+        size_t type_stack_size,
+        bool used = false)
+      : label_type(label_type),
+        name(name),
+        sig(sig),
+        type_stack_size(type_stack_size),
+        used(used) {}
+
+  bool HasValue() const {
+    return label_type != LabelType::Loop && !sig.empty();
+  }
+
+  LabelType label_type;
+  const std::string& name;
+  const BlockSignature& sig;
+  size_t type_stack_size;
+  bool used = false;
+};
+
+template <int>
+struct Name {
+  explicit Name(const std::string& name) : name(name) {}
+  const std::string& name;
+};
+
+typedef Name<0> LocalName;
+typedef Name<1> GlobalName;
+typedef Name<2> ExternalPtr;
+typedef Name<3> ExternalRef;
+
+struct GotoLabel {
+  explicit GotoLabel(const Var& var) : var(var) {}
+  const Var& var;
+};
+
+struct LabelDecl {
+  explicit LabelDecl(const std::string& name) : name(name) {}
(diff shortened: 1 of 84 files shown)


binji:
Travis failure looks like bot flake.

pepyakin:
For those who wondering how output is looks like I hacked a demo:

https://pepyakin.github.io/wabt/wasm2c/index.html

it is rough copy of wat2wasm atm since I don't have a much time, but i think it could serve as a demo

source: https://github.com/pepyakin/wabt

src/c-writer.cc
```diff
@@ -0,0 +1,2437 @@
...
+        const Func& func = *module_->GetFunc(var);
+        Index num_params = func.GetNumParams();
+        Index num_results = func.GetNumResults();
+        assert(type_stack_.size() >= num_params);
+        if (num_results > 0) {
+          assert(num_results == 1);
+          Write(StackVar(num_params - 1, func.GetResultType(0)), " = ");
+        }
+
+        Write(GlobalVar(var), "(");
```
pepyakin: it appears that there is a problem with this line, because the following code

```WebAssembly
(module
  (func $fac (param f64) (result f64)
    get_local 0
    f64.const 1
    f64.lt
    if (result f64)
      f64.const 1
    else
      get_local 0
      get_local 0
      f64.const 1
      f64.sub
      call $fac
      f64.mul
    end)
  (export "fac" (func $fac)))
```

is compiled as

```c
static f64 fac(f64 _) {
  FUNC_PROLOGUE;
  u32 i0;
  f64 d0, d1, d2;
  d0 = _;
  d1 = 1;
  i0 = d0 < d1;
  if (i0) {
    d0 = 1;
  } else {
    d0 = _;
    d1 = _;
    d2 = 1;
    d1 -= d2;
    d1 = (d1);
    d0 *= d1;
  }
  FUNC_EPILOGUE;
  return d0;
}
```

src/c-writer.cc
```diff
@@ -0,0 +1,2437 @@
...
+        const Var& var = cast<GetGlobalExpr>(&expr)->var;
+        PushType(module_->GetGlobal(var)->type);
+        Write(StackVar(0), " = ", GlobalVar(var), ";", Newline());
+        break;
+      }
+
+      case ExprType::GetLocal: {
+        const Var& var = cast<GetLocalExpr>(&expr)->var;
+        PushType(func_->GetLocalType(var));
+        Write(StackVar(0), " = ", var, ";", Newline());
```
pepyakin: That there is something going on with var naming

```
(module
  (func $addTwo (param i32) (result i32)
    (local i32)
    get_local 1
  )
)
```

gets compiled as

```c
static u32 addTwo(u32 _) {
  u32 __0 = 0;
  FUNC_PROLOGUE;
  u32 i0;
  i0 = _;
  FUNC_EPILOGUE;
  return i0;
}
```

it should return `__0` (0) however it returns `_` (arg passed in the func).

binji:
@pepyakin Yeah, that's pretty strange. I'll take a look.

binji:
@pepyakin Ah, so the problem is that the `WriteC` function expects that the module has been validated, but also that all vars are named. You can do this in JS with `module.generateNames` and `module.applyNames`. The kind of annoying thing is that `module.validate` expects that all vars are indexes. So you end up having to do this:

```
...
module.resolveNames();
module.validate();
module.generateNames();
module.applyNames();
var cOutput = module.toC();
```

PR #712, opened 2018-01-10
Use templates for generating wasm2c source

binji:
This uses a very simple template syntax, without any advanced features:

```
%%top
...
%%bottom
...
```

This template will generate output with two C strings:

```
const char SECTION_NAME(top)[] = ...
...
const char SECTION_NAME(bottom)[] = ...
...
```

To update the generated files, run `make update-wasm2c`.


binji:
Looks like the only failure is the ubsan bot failure that I recently fixed.

PR #714, opened 2018-01-11
SIMD i16x8.add / i32x4.add / i64x2.add implementation.

lizhengxing:


lizhengxing:
@binji 
simd i16x8.add / i32x4.add / i64x2.add implementation, PTAL! 
Thanks!

@arunetm FYI!

PR #717, opened 2018-01-17
Check that name section is after known sections

binji:


test/binary/relocs.txt
```diff
@@ -2,9 +2,7 @@
 magic
 version
 section(TYPE) { count[1] function params[0] results[1] i32 }
-section("name") {
-  func_count[0]
-}
+section("name") {}
```
sbc100: Why this?

binji: This test was invalid (since the name section changed), but the `name_section_ok` check was hiding the bug.

sbc100: So the section is still present, but without any content at all?   I would have thought that was illegal.   Why is the old version wrong and the new version not?

binji: The new format is zero or more subsections, where each subsection has a type, a length, and a collection of names. The old format was a function count, followed by its name, and local count, and a collection of local names.

I could just remove the name section though, maybe that would be clearer.

sbc100: Oh I see so this is name section with zero subsections... yeah, maybe just remove it if its not serving any purpose?

binji:
Travis failure seems to be unrelated (i686 bot couldn't download gcc?)

PR #720, opened 2018-01-18
SIMD sub implementation.

lizhengxing:


lizhengxing:
@binji 
Simd sub implementation. PTAL, thanks!

@arunetm FYI!

PR #723, opened 2018-01-18
SIMD i8x16.neg / i16x8.neg / i32x4.neg / i64x2.neg implementation.

lizhengxing:


lizhengxing:
@binji 
SIMD neg instructions implementation. PTAL, thanks!

@arunetm FYI!

src/interp.cc
```diff
@@ -1047,6 +1070,12 @@ ValueTypeRep<R> IntEqz(ValueTypeRep<T> v_rep) {
   return ToRep(v_rep == 0);
 }
 
+// i{32,64}.neg
```
binji: nit: this isn't a real instruction, so it's better to just remove this comment.

src/opcode.def
```diff
@@ -254,6 +254,10 @@ WABT_OPCODE(V128, V128, V128, ___, 0, 0xfd, 0x1f,  I64X2Sub, "i64x2.sub")
 WABT_OPCODE(V128, V128, V128, ___, 0, 0xfd, 0x20,  I8X16Mul, "i8x16.mul")
 WABT_OPCODE(V128, V128, V128, ___, 0, 0xfd, 0x21,  I16X8Mul, "i16x8.mul")
 WABT_OPCODE(V128, V128, V128, ___, 0, 0xfd, 0x22,  I32X4Mul, "i32x4.mul")
+WABT_OPCODE(V128, V128, ___, ___, 0, 0xfd, 0x23,  I8X16Neg, "i8x16.neg")
```
binji: nit: align columns please!

src/interp.cc

src/opcode.def
lizhengxing: Done.

lizhengxing:
@binji 
updated the PR for windows platform. PTAL, thanks!

@arunetm FYI!

PR #725, opened 2018-01-19
Allow multiple commands per test in test runner

binji:
This CL allows each test to specify more than one command, each of which
will be run, in order, as long as the previous commands all return
errorcode `0`.

For example:
```
;;; EXE: %(command1)s
;;; FLAGS: --foo=1 --stuff
;;; EXE: %(command2)s
;;; FLAGS: --other-things
```

The goal here is to remove many of the test wrapper scripts (e.g.
`test/run-interp`, `test/run-objdump`, etc.) by having their
functionality represented by multiple commands instead, then wrapped
into a `TOOL` definition.

### DIFF
--- a/test/dump/invalid-data-segment-no-memory.txt
+++ b/test/dump/invalid-data-segment-no-memory.txt
@@ -1,4 +1,3 @@
-;;; TOOL: wat2wasm
 ;;; FLAGS: -v --no-check
 (module
   (data (i32.const 0) "hello"))
--- a/test/dump/invalid-data-segment-no-memory.txt
+++ b/test/dump/invalid-data-segment-no-memory.txt
@@ -1,3 +1,4 @@
+;;; TOOL: wat2wasm
 ;;; FLAGS: -v --no-check
 (module
   (data (i32.const 0) "hello"))
--- a/test/run-tests.py
+++ b/test/run-tests.py
@@ -316,7 +316,7 @@ class TestInfo(object):
     if self.tool not in ROUNDTRIP_TOOLS:
       raise NoRoundtripError()
 
-    if not (len(self.cmds) == 1 and len(result.cmds) == 1):
+    if len(self.cmds) != 1:
       raise NoRoundtripError()
 
     result = TestInfo()
--- a/test/run-tests.py
+++ b/test/run-tests.py
@@ -199,62 +199,13 @@ class Cell(object):
     return self.value[0]
 
 
-def RunCommandListWithTimeout(command, cwd, timeout, console_out=False,
-                              env=None):
-  process = None
-  is_timeout = Cell(False)
-
-  def KillProcess(timeout=True):
-    if process:
-      try:
-        if IS_WINDOWS:
-          # http://stackoverflow.com/a/10830753: deleting child processes in
-          # Windows
-          subprocess.call(['taskkill', '/F', '/T', '/PID', str(process.pid)])
-        else:
-          os.killpg(os.getpgid(process.pid), 15)
-      except OSError:
-        pass
-    is_timeout.Set(timeout)
-
-  try:
-    start_time = time.time()
-    kwargs = {}
-    if not IS_WINDOWS:
-      kwargs['preexec_fn'] = os.setsid
-
-    # http://stackoverflow.com/a/10012262: subprocess with a timeout
-    # http://stackoverflow.com/a/22582602: kill subprocess and children
-    process = subprocess.Popen(command, cwd=cwd, env=env,
-                               stdout=None if console_out else subprocess.PIPE,
-                               stderr=None if console_out else subprocess.PIPE,
-                               universal_newlines=True, **kwargs)
-    timer = threading.Timer(timeout, KillProcess)
-    try:
-      timer.start()
-      stdout, stderr = process.communicate()
-    finally:
-      returncode = process.returncode
-      process = None
-      timer.cancel()
-    if is_timeout.Get():
-      raise Error('TIMEOUT\nSTDOUT:\n%s\nSTDERR:\n%s\n' % (stdout, stderr))
-    duration = time.time() - start_time
-  except OSError as e:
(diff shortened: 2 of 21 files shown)


test/two-commands.txt
```diff
@@ -0,0 +1,11 @@
+;;; TOOL: wat2wasm
+;;; FLAGS: -v
```
sbc100: llvm's lit uses "RUN" for this so maybe we should mirror that?  Could/should be separate change I guess.

sbc100: Eventually it might make sense to have flags be part of RUN too, no need for a separate line is there?

test/dump/invalid-data-segment-no-memory.txt
```diff
@@ -1,3 +1,4 @@
+;;; TOOL: wat2wasm
```
sbc100: So part of this change to remove the default tool?  Perhaps that could be a separate CL since it seems somewhat orthogonal to supporting multiple tools.

test/run-tests.py
```diff
@@ -265,11 +327,16 @@ def CreateRoundtripInfo(self, fold_exprs):
     result.input_ = self.input_
     result.expected_stdout = ''
     result.expected_stderr = ''
+
     # TODO(binji): It's kind of cheesy to keep the enable flag based on prefix.
     # Maybe it would be nicer to add a new directive ENABLE instead.
-    result.flags = [flag for flag in self.flags if flag.startswith('--enable')]
+    old_cmd = self.cmds[0]
+    new_cmd = result.cmds[0]
+    new_cmd.flags = [flag for flag in old_cmd.flags
```
sbc100: i think its ok to use single char variable names in list comprehensions, especially if it means not having to wrap them.

test/run-tests.py
```diff
@@ -413,32 +478,9 @@ def Parse(self, filename):
 
     # If the test didn't specify a executable (either via EXE or indirectly
     # via TOOL, then use the default tool)
-    if not self.exe:
+    if not self.cmds:
```
sbc100: So there is still a default?

test/run-tests.py
```diff
@@ -189,7 +199,8 @@ def Get(self):
     return self.value[0]
 
 
-def RunCommandWithTimeout(command, cwd, timeout, console_out=False, env=None):
+def RunCommandListWithTimeout(command, cwd, timeout, console_out=False,
```
sbc100: Why rename this?  It still takes a single command doesn't it?

test/run-tests.py
```diff
@@ -237,6 +248,52 @@ def KillProcess(timeout=True):
   return stdout, stderr, returncode, duration
 
 
+def AsFlagsList(value):
```
sbc100: I prefer `args` to `flags` in general.  Maybe "SplitArgs"?

test/run-tests.py
```diff
@@ -597,15 +641,30 @@ def RunTest(info, options, variables, verbose_level=0):
...
+  total_duration = 0
 
-  try:
-    return RunCommandWithTimeout(cmd, cwd, timeout, verbose_level > 0, env)
-  except (Error, KeyboardInterrupt) as e:
-    return e
+  for cmd in info.cmds:
+    cmd_list = cmd.AsList(variables, options.arg, verbose_level)
+    if options.print_cmd:
+      print(' '.join(cmd_list))
```
sbc100: How about cmd.Print()  or `print cmd` then implement Command.__str__()?

test/run-tests.py
```diff
@@ -597,15 +641,30 @@ def RunTest(info, options, variables, verbose_level=0):
...
+  total_stderr = ''
+  returncode = 0
+  total_duration = 0
 
-  try:
-    return RunCommandWithTimeout(cmd, cwd, timeout, verbose_level > 0, env)
-  except (Error, KeyboardInterrupt) as e:
-    return e
+  for cmd in info.cmds:
+    cmd_list = cmd.AsList(variables, options.arg, verbose_level)
```
sbc100: I don't really like the name `AsList` but I can't think of a better one right now.  Something like ".Get()" or ".Format()"?  I don't really like those either :)

test/run-tests.py
```diff
@@ -597,15 +641,30 @@ def RunTest(info, options, variables, verbose_level=0):
...
+      stdout, stderr, returncode, duration = (
+          RunCommandListWithTimeout(cmd_list, cwd, timeout,
+                                    verbose_level > 0, env))
+      total_stdout += stdout
+      total_stderr += stderr
+      total_duration += duration
+
+      if returncode != 0:
+        break
+    except (Error, KeyboardInterrupt) as e:
```
sbc100: Move the accept up so it only surrounds the RunCommandListWithTimeout call?

test/dump/invalid-data-segment-no-memory.txt
binji: No, it still has a default tool, but it doesn't work with just `FLAGS`. I guess it would be better to just set it as the default if the `FLAGS` directive is found.

test/run-tests.py
binji: Yeah, good point. I wanted to differentiate between passing a `Command` object and a list of strings (`CommandList`) but maybe it's not worth it.

test/run-tests.py
binji: Me too, but I wanted to do the flags -> args rename in another CL

test/run-tests.py
binji: Done.

test/run-tests.py
binji: Done.

test/two-commands.txt
binji: Yeah, I want to do that too (move FLAGS into RUN)

binji:
OK, this should be a bit cleaner, PTAL!

PR #726, opened 2018-01-19
Fix for SIMD i8x16.neg / i16x8.neg / i32x4.neg / i64x2.neg implementation. (#723)

lizhengxing:
SIMD neg are unary instructions and at wrong place in Disassemble() in interp.cc, This
Fix move them to the right place.


lizhengxing:
@binji 
Fix one bug in SIMD neg PR, PTAL, thanks!

@arunetm FYI!

binji:
Thanks!

PR #727, opened 2018-01-19
SIMD saturating add implementation.

lizhengxing:


lizhengxing:
@binji 
SIMD saturating add implementation. PTAL, thanks!

@arunetm FYI!

src/interp.cc
```diff
@@ -912,6 +912,21 @@ ValueTypeRep<T> Add(ValueTypeRep<T> lhs_rep, ValueTypeRep<T> rhs_rep) {
   return ToRep(FromRep<T>(lhs_rep) + FromRep<T>(rhs_rep));
 }
 
+template <typename T, typename R, R rmax, R rmin>
```
binji: Rather than passing `rmax` and `rmin` as template params, you can use `std::numeric_limit<R>::min()` and `std::numeric_limit<R>::max()`.

src/interp.cc
```diff
@@ -912,6 +912,21 @@ ValueTypeRep<T> Add(ValueTypeRep<T> lhs_rep, ValueTypeRep<T> rhs_rep) {
...
+ValueTypeRep<T> AddSaturate(ValueTypeRep<T> lhs_rep, ValueTypeRep<T> rhs_rep) {
+  T max = static_cast<T>(rmax);
```
binji: Are these `static_cast`s necessary? It seems like they'll implicitly cast since `T` is always a larger size type than `R`.

src/interp.cc
lizhengxing: Great! It makes code cleaner.

BTW:
The saturating sub is almost the same as saturating add. I want to use a BOOL template param to determine the saturating operation is add or sub, saturating sub and saturating add will share most of the code in this way. Is it feasible?

Thanks!

src/interp.cc
lizhengxing: Yes, you are right. 

I just want to make the signed and unsigned number promotion in code more explicitly.

Thanks!

src/interp.cc
binji: A bool template param is a little odd, perhaps. Maybe you can use a `Clamp<T, R>` function instead to share most of the code?

lizhengxing: Thanks!

It's not easy to use a Clamp<T, R> function to share code for saturating operation based on current code.

I will use the separated function to implement saturating add and sub.

lizhengxing:
@binji 
PTAL, thanks!

@arunetm FYI!

PR #729, opened 2018-01-21
[test] `RUN` command has optional args

binji:
Also fix bug when printing the last command that failed.

### DIFF
--- a/test/run-tests.py
+++ b/test/run-tests.py
@@ -208,10 +208,14 @@ def SplitArgs(value):
 class CommandTemplate(object):
 
   def __init__(self, exe):
-    self.exe = exe
-    self.args = []
+    self.args = self._FixPythonExecutable(SplitArgs(exe))
     self.verbose_args = []
-    self.last_cmd = None
+
+  def _FixPythonExecutable(self, args):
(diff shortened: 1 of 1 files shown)


test/run-tests.py
```diff
@@ -208,10 +208,14 @@ def SplitArgs(value):
...
-    self.args = []
+    self.args = self._FixPythonExecutable(SplitArgs(exe))
     self.verbose_args = []
-    self.last_cmd = None
+
+  def _FixPythonExecutable(self, args):
+    exe, rest = args[0], args[1:]
+    if os.path.splitext(exe)[1] == '.py':
+      return [sys.executable, os.path.join(REPO_ROOT_DIR, exe)] + rest
+    return args
```
sbc100: This method doesn't use `self` so maybe make it a separate function like SplitArgs?

binji:
> Maybe include an example that uses this new option?

The next PR has one (see https://github.com/WebAssembly/wabt/tree/tool-multi-cmd).

> (Also. its sunday!)

I did most of this on Friday, just wanted to get the PR up so I could build on Travis! :-)

PR #734, opened 2018-01-25
Creating emscripten-exported C bindings for wast handling functions.

icefoxen:
The goal is to make it possible to run `wast2json` from an external library binding to the C functions, so we can make a Rust binding for them: https://github.com/pepyakin/wabt-rs/issues/1

A little WIP; I need to bind `WriteBinarySpecScript` and maybe one or two other utility things but I wanted to get this out there to see if this is something you would consider accepting.


binji:
Thanks, this change looks good. We'll probably want to make a real C API at some point, but for now this is the best place to do it.

binji:
BTW, have you joined the [WebAssembly CG](https://www.w3.org/community/webassembly/)? For now, it's required to contribute to WebAssembly projects (though this topic is coming up for discussion at [tomorrow's CG meeting](https://github.com/WebAssembly/meetings/blob/master/2018/CG-01-26.md#agenda-items).)

icefoxen:
That's kinda goody, but ok, I'll sign up.

icefoxen:
Ok, I think this is done and ready to be merged, but frankly I am not sure I did it right because it's 1 am and I don't think the unit test suite even builds this code.  I can't find docs on how to build with emscripten so for the moment this is what we get.

Also, joined the WebAssembly CG.

pepyakin:
That is just awesome, thanks @icefoxen !

src/emscripten-helpers.cc
```diff
@@ -104,6 +123,43 @@ wabt::Result::Enum wabt_validate_module(
...
 
+wabt::Result::Enum wabt_validate_script(
+    wabt::WastLexer* lexer,
+    wabt::Script* script,
+    wabt::ErrorHandlerBuffer* error_handler) {
+  wabt::ValidateOptions options;
+  return ValidateScript(lexer, script, error_handler, &options);
+}
+
+WabtWriteBinaryResult* wabt_write_binary_spec_script(wabt::Script* script,
```
binji: Should this be `WabtWriteModuleResult`?

icefoxen:
Ok, it builds now.  Thank you!

pepyakin:
I'm wondering, is it feasible to export spec tests directly in C API, i.e. without roundtriping to wasm and JSON files?

icefoxen:
I don't know, that's a good question...

binji:
Yeah, currently it writes all files to the filesystem, but it would be better if it wrote them to a bunch of `Stream`s like `WriteBinaryModule` does.

pepyakin:
Or maybe we can go further! What if we expose `Script` in C API?

https://github.com/WebAssembly/wabt/blob/1f735e8f5125f84abba2a91d4f44bc8ee30544a8/src/tools/spectest-interp.cc#L100-L104

Or in case `wabt-rs` maybe we can employ `rust-bindgen` to generate bindings directly to C++?

binji:
Ah, I see what you mean. Yes, we should add a C API for this. I think it would expose this `Script` though:

https://github.com/WebAssembly/wabt/blob/master/src/ir.h#L947-L957

PR #744, opened 2018-02-01
Use Streams in the WriteBinarySpecScript API

binji:
`WriteBinarySpecScript` writes a JSON file and its associated module
files (either `.wasm` or `.wat` files). The previous API wrote directly
to the filesystem, but it is more useful to allow the caller to specify
where to write the output instead.

The function now provides a callback `WriteBinarySpecStreamFactory`,
which takes a filename as parameter and returns a `Stream` to write to.

There is also a convenience function that will write the results to a
vector of filename/MemoryStream pairs.

### DIFF
--- a/src/binary-writer-spec.h
+++ b/src/binary-writer-spec.h
@@ -17,21 +17,43 @@
 #ifndef WABT_BINARY_WRITER_SPEC_H_
 #define WABT_BINARY_WRITER_SPEC_H_
 
+#include <functional>
+#include <utility>
+#include <vector>
+
 #include "src/binary-writer.h"
 #include "src/common.h"
 #include "src/ir.h"
 
 namespace wabt {
 
-struct WriteBinarySpecOptions {
-  Stream* log_stream = nullptr;
-  const char* json_filename = nullptr;
-  WriteBinaryOptions write_binary_options;
+struct FilenameMemoryStreamPair {
+  FilenameMemoryStreamPair(string_view filename,
+                           std::unique_ptr<MemoryStream> stream)
+      : filename(filename), stream(std::move(stream)) {}
+  std::string filename;
+  std::unique_ptr<MemoryStream> stream;
 };
 
-Result WriteBinarySpecScript(struct Script*,
-                             const char* source_filename,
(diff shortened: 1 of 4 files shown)


pepyakin:
Hello @sbc100 ! Could you please take a look at this?

src/binary-writer-spec.h
```diff
@@ -17,21 +17,43 @@
...
-                             const WriteBinarySpecOptions*);
+typedef std::function<Stream*(string_view filename)>
+    WriteBinarySpecStreamFactory;
+
+Result WriteBinarySpecScript(Stream* json_stream,
+                             WriteBinarySpecStreamFactory module_stream_factory,
+                             Script*,
+                             string_view source_filename,
+                             string_view module_filename_noext,
+                             const WriteBinaryOptions*);
```
sbc100: What are some args not named?  Is there some kind of rule here?

src/tools/wast2json.cc
```diff
@@ -115,12 +116,23 @@ int ProgramMain(int argc, char** argv) {
...
+          StripExtension(s_outfile ? s_outfile : s_infile).to_string();
+      result = WriteBinarySpecScript(
+          &json_stream, script.get(), s_infile, module_filename_noext,
+          &s_write_binary_options, &module_streams, s_log_stream.get());
+
+      if (s_outfile) {
+        json_stream.WriteToFile(s_outfile);
+      }
+
+      std::for_each(module_streams.begin(), module_streams.end(),
```
sbc100: Is this preferable to using a normal for loop?    Seems harder to visually parse and more verbose.

sbc100:
`WriteBinarySpecStreamFactory` .. what is this Java project?

src/binary-writer-spec.h
binji: I often omit names that are redundant, kind of like Blink style. So you don't have `Script* script` or `const WriteBinaryOptions* options`. But it's not really anything formalized.

binji:
> WriteBinarySpecStreamFactory .. what is this Java a project?

Ha! I didn't like it much either, do you have a better name?

PR #747, opened 2018-02-06
SIMD saturating sub implementation.

lizhengxing:


lizhengxing:
@binji 
This is simd saturating sub implementation, PTAL! 
Thanks!

@arunetm FYI!

test/interp/simd-binary.txt
```diff
@@ -130,6 +130,26 @@
     v128.const i32 0x00ffffff 0x0400ffff 0x00000003 0x00000004
     v128.const i32 0x00020001 0xfe000002 0x00000003 0x00000004
     i16x8.add_saturate_u)
+
+  ;; i8x16 saturating sub (signed and unsigned)
+  (func (export "i8x16_sub_saturate_signed_0") (result v128)
+    v128.const i32 0x00000001 0x0000007f 0x00000003 0x00000080
+    v128.const i32 0x00000001 0x00000002 0x00000003 0x000000ff
+    i8x16.sub_saturate_s)
```
binji: Can you two sums that demonstrate saturation here, underflow and overflow?

test/interp/simd-binary.txt
```diff
@@ -130,6 +130,26 @@
...
+  (func (export "i8x16_sub_saturate_unsigned_0") (result v128)
+    v128.const i32 0x00ff0001 0x04000002 0x00000003 0x00000004
+    v128.const i32 0x00020001 0xfe000002 0x00000003 0x00000004
```
binji: `0x4 - 0xfe` should be `0`, right? I think this is happening because the arithmetic is done with `uint32_t`, which will wrap to `0xffffff06` before being clamped to `0xff`.

test/interp/simd-binary.txt
```diff
@@ -130,6 +130,26 @@
...
+    i8x16.sub_saturate_u)
+
+  ;; i16x8 saturating sub (signed and unsigned)
+  (func (export "i16x8_sub_saturate_signed_0") (result v128)
+    v128.const i32 0x00000001 0x00007fff 0x00000003 0x00008000
```
binji: Same here, please add examples that saturate.

test/interp/simd-binary.txt
lizhengxing: @binji 
Done.

test/interp/simd-binary.txt
lizhengxing: @binji 
Yes, you are right. Should always keep result value as signed value for Saturating Sub operation.

test/interp/simd-binary.txt
lizhengxing: @binji 
Done.

PR #758, opened 2018-02-14
SIMD Bitwise logic instructions implementation.

lizhengxing:
Including:
v128.and, v128.or, v128.xor, v128.not


lizhengxing:
@binji 
Simd Bitwise logic instructions implementation. PTAL!
Thanks!

@arunetm FYI!

PR #759, opened 2018-02-14
SIMD v128.bitselect instruction implementation.

lizhengxing:


lizhengxing:
@binji 
V128.bitselect implementation. PTAL!
Thanks!

@arunetm FYI!

src/binary-reader-interp.cc
```diff
@@ -184,7 +184,7 @@ class BinaryReaderInterp : public BinaryReaderNop {
   wabt::Result OnLoopExpr(Index num_types, Type* sig_types) override;
   wabt::Result OnNopExpr() override;
   wabt::Result OnReturnExpr() override;
-  wabt::Result OnSelectExpr() override;
+  wabt::Result OnSelectExpr(Opcode opcode) override;
```
binji: `bitselect` is different enough from `select` to have its own `ExprType`, I think. It actually seems to be a straightforward tertiary operator. If there are other SIMD instructions that have 3 operands, I'd suggest `Tertiary`, otherwise I think just calling it `BitSelect` is fine.

src/opcode.def
```diff
@@ -282,6 +282,7 @@ WABT_OPCODE(V128, V128, V128, ___, 0, 0xfd, 0x3b,  V128And, "v128.and")
 WABT_OPCODE(V128, V128, V128, ___, 0, 0xfd, 0x3c,  V128Or,  "v128.or")
 WABT_OPCODE(V128, V128, V128, ___, 0, 0xfd, 0x3d,  V128Xor, "v128.xor")
 WABT_OPCODE(V128, V128, ___,  ___, 0, 0xfd, 0x3e,  V128Not, "v128.not")
+WABT_OPCODE(V128, ___,  ___,  ___, 0, 0xfd, 0x3f,  V128BitSelect, "v128.bitselect")
```
binji: This can be specified with 4 `V128`s as well, I believe.

src/binary-reader-interp.cc
lizhengxing: Done.

src/opcode.def
lizhengxing: Done.

lizhengxing:
@binji 
re-implement the simd v128 bitselect instruction, PTAL!
Thanks!

@arunetm FYI!

Issue #760, opened 2018-02-15
Small module makes wabt to do an enourmous allocation


pepyakin:
Fuzzer come up with a such module (base64 form)

```
AGFzbQEAAAAAKAoAAABhcycAAABtAQAAACYBAAAAAC8AAAAAYXNtAQAAAABhc20BAAcHB/////EHBwcHAAA=
```

Running it makes my test machine to blow up:

```
$ echo "AGFzbQEAAAAAKAoAAABhcycAAABtAQAAACYBAAAAAC8AAAAAYXNtAQAAAABhc20BAAcHB/////EHBwcHAAA=" | base64 -d > ~/test.wasm

$ wasm2wat ~/test.wasm
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
Aborted
```

it has 2GB of RAM + 4GB of swap.

I understand that this is quite degenerate case, so I'm wondering is it in scope for this project to handle such fails gracefully?

rossberg:
Reference interpreter says:

    test.wasm:0x34: decoding error: length out of bounds

which may hint at the problem.

binji:
It looks like the export section has 2118123519 entries. When the binary reader reads this, it reserves that many entries in `module_->exports`. On my machine at least, that succeeds. But I suppose that's just Linux never failing on allocation.

Calling `reserve` is just an optimization, and removing it (or catching `std::bad_alloc`) would fix this bug at least.

wabt also has the cmake flag `WITH_EXCEPTIONS` (which is off by default) which will catch `std::bad_alloc` at the top-level and print an error message instead. This isn't really very graceful, though.

For now, I'll add try/catch blocks around the reserve calls when `WITH_EXCEPTIONS` is enabled.

pepyakin:
btw, my test machine is on linux

pepyakin:
Here is a few other failures:

https://gist.github.com/pepyakin/f25917936923647c5a7a2fdde69fc41d

these are so large that mmap fails

binji:
Hm, none of these are failing on my machine either. That said, I found the need to add `WITH_EXCEPTIONS` when I was fuzzing too, so perhaps that should be default, or you should try enabling that when running your fuzzer.

rossberg:
So what the interpreter does, for any length value it reads, is to first check whether that is within bounds of the binary's total size. If it isn't, you can stop right there (like in this case). If it is, then by construction, allocation should be of manageable size.

[...]

PR #761, opened 2018-02-15
Add try/catch around calls to reserve

binji:
Only when exceptions are enabled via the CMake `WITH_EXCEPTIONS` option.

See issue #760.

### DIFF
--- a/src/binary-reader-ir.cc
+++ b/src/binary-reader-ir.cc
@@ -302,7 +302,9 @@ bool BinaryReaderIR::OnError(const char* message) {
 }
 
 Result BinaryReaderIR::OnTypeCount(Index count) {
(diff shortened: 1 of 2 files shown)


sbc100:
I didn't know we supported exception.. why?

binji:
We mostly don't support exceptions, but provide it as an option because when an exception would be thrown (e.g. `std::bad_alloc`) the process is terminated as a crash. So if you're fuzzing, this looks like an error and hides other "real" errors.

This PR doesn't really improve things much, but it since `reserve` is just an optimization, if it does throw `std::bad_alloc` we can just catch and ignore it, which will mean the memory allocation will either fail later or won't fail at all since the reservation wasn't actually needed.

Issue #752, opened 2018-02-08
segfault when pointing `wasm2wat` to a module with a really big function


pepyakin:
Here is a binary

https://github.com/pepyakin/wasmi/files/1708839/livesplit_core.zip

afc3f3a8

```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=2, address=0x7fff5f3fff30)
  * frame #0: 0x0000000100035679 wasm2wat`wabt::TypeChecker::PushLabel(this=<unavailable>, label_type=<unavailable>, sig=<unavailable>) at type-checker.cc:75
    frame #1: 0x00000001000391e4 wasm2wat`wabt::TypeChecker::OnBlock(this=0x00007fff5fbfef30, sig=0x000000010076a288 size=1) at type-checker.cc:325
    frame #2: 0x000000010004216a wasm2wat`wabt::(anonymous namespace)::Validator::CheckExpr(this=0x00007fff5fbfeee0, expr=0x000000010076a230) at validator.cc:494
    frame #3: 0x00000001000419f1 wasm2wat`wabt::(anonymous namespace)::Validator::CheckExprList(this=0x00007fff5fbfeee0, loc=0x000000010076a1b8, exprs=0x000000010076a210) at validator.cc:417
    frame #4: 0x00000001000421a7 wasm2wat`wabt::(anonymous namespace)::Validator::CheckExpr(this=0x00007fff5fbfeee0, expr=0x000000010076a1a0) at validator.cc:495

< 10268 frames earlier... >

    frame #10272: 0x00000001000421a7 wasm2wat`wabt::(anonymous namespace)::Validator::CheckExpr(this=0x00007fff5fbfeee0, expr=0x00000001006b1a40) at validator.cc:495
    frame #10273: 0x00000001000419f1 wasm2wat`wabt::(anonymous namespace)::Validator::CheckExprList(this=0x00007fff5fbfeee0, loc=0x00000001006b19c8, exprs=0x00000001006b1a20) at validator.cc:417
    frame #10274: 0x0000000100042b19 wasm2wat`wabt::(anonymous namespace)::Validator::CheckExpr(this=0x00007fff5fbfeee0, expr=0x00000001006b19b0) at validator.cc:599
[...]

PR #765, opened 2018-02-21
Rewrite ExprVisitor to be non-recursive

binji:
This will help with wasm modules that have very deeply nested blocks.

See issue #752.

### DIFF
--- a/src/expr-visitor.cc
+++ b/src/expr-visitor.cc
@@ -24,305 +24,83 @@ namespace wabt {
 ExprVisitor::ExprVisitor(Delegate* delegate) : delegate_(delegate) {}
 
 Result ExprVisitor::VisitExpr(Expr* root_expr) {
-  enum class State {
-    Default,
-    Block,
-    IfTrue,
-    IfFalse,
-    Loop,
-    Try,
-    TryCatch,
-  };
+  state_stack_.clear();
+  expr_stack_.clear();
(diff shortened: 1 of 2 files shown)


src/expr-visitor.cc
```diff
@@ -23,179 +23,333 @@ namespace wabt {
...
+    expr_iter_stack.emplace_back(expr_list.begin());
+    catch_index_stack.emplace_back(catch_index);
+  };
+
+  auto pop_catch = [&]() {
+    state_stack.pop_back();
+    expr_stack.pop_back();
+    expr_iter_stack.pop_back();
+    catch_index_stack.pop_back();
+  };
```
sbc100: These looks like they could/should be methods on object contains the 5 stacks.   Why not use a class/struct with methods?

src/expr-visitor.cc
```diff
@@ -23,179 +23,333 @@ namespace wabt {
...
+  while (!state_stack.empty()) {
+    State state = state_stack.back();
+    auto* expr = expr_stack.back();
+
+    switch (state) {
+      case State::Default:
+        pop_default();
+
+        switch (expr->type()) {
+          case ExprType::AtomicLoad:
```
sbc100: this is quite a lot of nesting.  Is it worth creating a method for the handling of a single State?

binji:
Done, PTAL

PR #766, opened 2018-02-21
SIMD Comparison Equality instructions implementation.

lizhengxing:
Including:
(i8x16/i16x8/i32x4/f32x4/f64x2).eq


lizhengxing:
@binji 
Simd EQ implementation, PTAL!
Thanks!

@arunetm FYI!

PR #767, opened 2018-02-21
SIMD Comparison Non-Equality instructions implementation.

lizhengxing:
Including:
(i8x16/i16x8/i32x4/f32x4/f64x2).ne


lizhengxing:
@binji 
Simd Non-Equality implementation, PTAL!
Thanks!

@arunetm FYI!

PR #768, opened 2018-02-21
Use ExprVisitor in WatWriter

binji:
Since ExprVisitor is non-recursive, this will allow modules with deeply
nested blocks to be written.

See issue #752.

### DIFF
--- a/src/wat-writer.cc
+++ b/src/wat-writer.cc
@@ -143,9 +143,6 @@ class WatWriter {
                        const Block& block,
                        const char* text);
   void WriteEndBlock();
-  void WriteBlock(LabelType label_type,
-                  const Block& block,
-                  const char* start_text);
   void WriteConst(const Const& const_);
   void WriteExpr(const Expr* expr);
   template <typename T>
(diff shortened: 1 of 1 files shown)


src/wat-writer.cc
```diff
@@ -517,217 +520,324 @@ void WatWriter::WriteLoadStoreExpr(const Expr* expr) {
...
+  visitor.VisitExpr(const_cast<Expr*>(expr));
 }
 
 void WatWriter::WriteExprList(const ExprList& exprs) {
   WABT_TRACE(WriteExprList);
-  for (const Expr& expr : exprs)
-    WriteExpr(&expr);
+  ExprVisitorDelegate delegate(this);
+  ExprVisitor visitor(&delegate);
+  visitor.VisitExprList(const_cast<ExprList&>(exprs));
```
sbc100: Why const_cast here?

If the visitiing really modifies the list then perhaps the arg to this function should not be const either?

binji: It doesn't modify the list, but the visitor doesn't have a const overload of `VisitExprList`. I probably should just add one.

PR #769, opened 2018-02-22
Add support for new symbol table format

sbc100:
This should not land until the upstream support in llvm also lands


src/binary-reader-objdump.cc
```diff
@@ -157,19 +167,47 @@ class BinaryReaderObjdumpPrepass : public BinaryReaderObjdumpBase {
  public:
   using BinaryReaderObjdumpBase::BinaryReaderObjdumpBase;
 
-  Result OnFunctionName(Index function_index,
-                        string_view function_name) override;
+  Result OnFunctionName(Index index,
```
binji: nit: I generally haven't been defining functions like this inline in the class body. Any reason to prefer it this way?

src/binary-reader-objdump.cc
```diff
@@ -1127,28 +1176,76 @@ Result BinaryReaderObjdump::OnStackGlobal(Index stack_global) {
...
+                                     Index index,
+                                     string_view name,
+                                     uint32_t flags,
+                                     Index segment,
+                                     uint32_t offset,
+                                     uint32_t size) {
+  std::string sym_name = name.to_string();
+  switch (type) {
+    case SymbolType::Function:
+      if (sym_name.empty())
```
binji: nit: braces around all ifs

src/binary-reader-objdump.cc
```diff
@@ -1127,28 +1176,76 @@ Result BinaryReaderObjdump::OnStackGlobal(Index stack_global) {
...
+      break;
+    case SymbolVisibility::Default:
+      vis_name = "default";
+      break;
+  }
+
+  PrintDetails("   - sym[%d] <" PRIstringview ">", symbol_index,
+               WABT_PRINTF_STRING_VIEW_ARG(sym_name));
+  switch (type) {
+  case SymbolType::Data:
```
binji: nit: case is indented 2 spaces

src/binary-reader.cc
```diff
@@ -1348,15 +1348,37 @@ Result BinaryReader::ReadLinkingSection(Offset section_size) {
...
+            case SymbolType::Data:
+              CHECK_RESULT(ReadStr(&name, "symbol name"));
+              if ((flags & WABT_SYMBOL_FLAG_UNDEFINED) == 0) {
+                CHECK_RESULT(ReadU32Leb128(&segment, "segment"));
+                CHECK_RESULT(ReadU32Leb128(&offset, "offset"));
+                CHECK_RESULT(ReadU32Leb128(&size, "size"));
+              }
+              break;
+          }
+          CALLBACK(OnSymbol, i, static_cast<SymbolType>(kind), index, name,
```
binji: Seems a little strange to send down `segment`, `offset` and `size` even though those are only valid for the `Data` symbol type. I guess this is simpler than separating those out into their own callback, though.

sbc100: Yeah I was thinking seperate callbacks too.   Do you think I should make three different one instead of this singe one, or in addition?

src/binary-reader-objdump.cc
sbc100: My thinking was because (a) these are all internal to the cpp file (b) they are both small forwarding functions (c) it makes the purpose of this class and these methods clear simply by ready the class def.

Whats wrong with inlining short methods like these?

src/binary-reader.cc
binji: In the other callbacks like this, I have one generic one with the common data, then have other callbacks with type-specific data.

sbc100:
refactored, PTAL

src/binary-reader-objdump.cc
```diff
@@ -1208,22 +1187,49 @@ Result BinaryReaderObjdump::OnSymbol(Index symbol_index,
...
+  return Result::Ok;
+}
+
+Result BinaryReaderObjdump::OnFunctionSymbol(Index index,
+                                             uint32_t flags,
+                                             string_view name,
+                                             Index func_index) {
+  std::string sym_name = name.to_string();
+  if (sym_name.empty()) {
+      sym_name = GetFunctionName(func_index);
```
binji: nit: 2 space indent

src/binary-reader-objdump.cc
```diff
@@ -172,16 +172,12 @@ class BinaryReaderObjdumpPrepass : public BinaryReaderObjdumpBase {
...
-                  Index segment,
-                  uint32_t offset,
-                  uint32_t size) override {
-    if (type == SymbolType::Function && !name.empty())
-      SetFunctionName(index, name);
+  Result OnFunctionSymbol(Index index,
+                          uint32_t flags,
+                          string_view name,
+                          Index func_index) override {
+    if (!name.empty())
```
binji: nit: braces

src/binary-reader-logging.cc
```diff
@@ -449,16 +449,38 @@ Result BinaryReaderLogging::OnReloc(RelocType type,
...
                                      SymbolType type,
-                                     Index index,
-                                     string_view name,
-                                     uint32_t flags,
-                                     Index segment,
-                                     uint32_t offset,
-                                     uint32_t size) {
-  LOGF("(OnSymbol name: " PRIstringview ", flags: 0x%x)\n",
+                                     uint32_t flags) {
+  LOGF("(OnSymbol type: %s flags: 0x%x)\n", "", flags);
```
binji: type is "" ... unfinished?

test/binary/linking-section.txt
```diff
@@ -31,17 +31,19 @@ section("linking") {
...
     index[0]
-    str("foo")
+    str("func_sym")
 
-    type[0]
+    type[2]
     flags[2]
     index[0]
-    str("bar")
+    str("global_syum")
```
binji: global_sym

PR #770, opened 2018-02-22
SIMD Comparison Less than instructions implementation.

lizhengxing:
Including:
i8x16.lt_s, i8x16_lt_u
i16x8.lt_s, i16x8_lt_u
i32x4.lt_s, i32x4_lt_u
f32x4.lt
f64x2.lt


lizhengxing:
@binji 
SIMD less than implementation, PTAL!
Thanks!

@aruntem FYI!

PR #773, opened 2018-02-28
WIP on support for level1 exception spec

binji:
Implemented:

* Parsing `try`, `if_except`, `throw`, `rethrow`
* Validation
* Binary and text output

Still missing:

* `except_ref` for locals
* Interpreter implementation


test/dump/except.txt
```diff
@@ -0,0 +1,37 @@
...
+0000018: 02                                        ; exception type count
+0000019: 7d                                        ; f32
+000001a: 7c                                        ; f64
+0000009: 11                                        ; FIXUP section size
+
+except.wasm:	file format wasm 0x1
+
+Section Details:
+
+Custom:
```
eholk: Does this mean the exception section is being treated as a custom section?

binji: It looks like that's what the proposal currently has: https://github.com/WebAssembly/exception-handling/blob/master/proposals/Level-1.md#exception-section

eholk: Ah, okay. I think the exception section should be a Wasm section like say the types sections. I thought the idea behind custom sections was that they could be ignored or dropped without changing the semantics of the module, which isn't the case for exceptions.

We should probably clarify this in the exceptions proposal.

binji: Yeah, there was discussion about this previously. Can't remember where, though. I seem to recall an issue being that this would be the first section number that is out of order. But I don't think that's a big deal.

eholk: It likes like the discussion was at https://github.com/WebAssembly/exception-handling/issues/25 and https://github.com/WebAssembly/exception-handling/pull/27.

I think the idea was that future sections would have out of order IDs, but that's by design. Having a fixed order was more important than having them in numerical order.

binji:
OK, gonna land this. @KarlSchimpf, feel free to add comments if you have any, I'll fix in later PRs.

PR #779, opened 2018-02-28
SIMD (f32x4/f64x2).neg implementation.

lizhengxing:


lizhengxing:
@binji 
SIMD (f32x4/f64x2).neg implementation, PTAL!
Thanks!

@arunetm FYI!

PR #782, opened 2018-02-28
Match value type enum to spec values and read/write as uint8_t

sbc100:
These seems to match what is in the spec where these values
are specified as single type terminals.

See: https://webassembly.github.io/spec/core/binary/types.html


sbc100:
The spec and BinaryFormat.md also seem to disagree somewhat on this.  The BinaryFormat.md seems to use varint7 .. where as (I think) the spec specified a raw byte.  Am I wrong?  Should we update BinaryFormat.md?

binji:
BinaryEncoding.md hasn't been updated in a while, I wouldn't rely on it for anything anymore.

PR #783, opened 2018-03-01
SIMD (f32x4/f64x2).abs implementation.

lizhengxing:


lizhengxing:
@binji 
SIMD (f32x4/f64x2).abs implementation. PTAL!
Thanks!

@arunetm FYI!

PR #784, opened 2018-03-02
SIMD Floating-point min and max implementation.

lizhengxing:
Including:
f32x4.min, f64x2.min, f32x4.max, f64x2.max


lizhengxing:
@binji 
Simd f32x4.min, f64x2.min, f32x4.max, f64x2.max implementation. PTAL!
Thanks!

@arunetm FYI!

PR #786, opened 2018-03-03
SIMD Floating-point add and sub implementation.

lizhengxing:
Including:
f32x4.add, f64x2.add, f32x4.sub, f64x2.sub


lizhengxing:
@binji 
Simd float add/sub implementation PR, PTAL!
Thanks!

@arunetm FYI!

PR #787, opened 2018-03-05
SIMD Floating-point div and mul implementation.

lizhengxing:
Including:
f32x4.div, f64x2.div, f32x4.mul, f64x2.mul


lizhengxing:
@binji 
SIMD div/mul instructions, PTAL!
Thanks!

@arunetm FYI!

PR #788, opened 2018-03-05
Use early return in PrintStackIfFailed

sbc100:
Also add some missing braces.


src/type-checker.cc
```diff
@@ -242,48 +244,49 @@ static std::string TypesToString(const TypeVector& types,
 void TypeChecker::PrintStackIfFailed(Result result,
                                      const char* desc,
                                      const TypeVector& expected) {
-  if (Failed(result)) {
-    size_t limit = 0;
-    Label* label;
-    if (Succeeded(TopLabel(&label))) {
-      limit = label->type_stack_limit;
-    }
+  if (Succeeded(result))
```
binji: nit: braces here too

Thanks for fixing the other spots, btw!

PR #794, opened 2018-03-06
Generate symbol table when creating relocations in binary-writer.cc

sbc100:


src/binary-reader-objdump.cc
```diff
@@ -134,24 +135,29 @@ const char* BinaryReaderObjdumpBase::GetGlobalName(Index index) const {
...
+  ObjdumpSymbol& sym = objdump_state_->symtab[symbol_index];
+  switch (sym.kind) {
+    case SymbolType::Function:
+      return GetFunctionName(sym.index);
+    case SymbolType::Data:
+      return sym.name;
+    case SymbolType::Global:
+      return GetGlobalName(sym.index);
+  }
+}
```
binji: gcc gives an error here: `error: control reaches end of non-void function`

src/binary-reader-objdump.cc
```diff
@@ -1241,7 +1292,9 @@ Result BinaryReaderObjdump::OnGlobalSymbol(Index index,
                                            Index global_index) {
   std::string sym_name = name.to_string();
   if (sym_name.empty()) {
-    sym_name = GetGlobalName(global_index);
+    if (const char* N = GetGlobalName(global_index)) {
```
binji: nit: gotta be something better than `N` here :-)

src/binary-reader-objdump.h
```diff
@@ -50,13 +50,20 @@ struct ObjdumpOptions {
   const char* section_name;
 };
 
+struct ObjdumpSymbol {
+  wabt::SymbolType kind;
+  string_view name;
```
binji: Maybe `std::string` here? Who owns the string_view memory?

src/binary-writer.cc
```diff
@@ -289,6 +298,30 @@ Index BinaryWriter::GetExceptVarDepth(const Var* var) {
...
+      break;
+    case RelocType::GlobalIndexLEB:
+      type = SymbolType::Global;
+      name = module_->globals[index]->name;
+      break;
+    default:
+      WABT_UNREACHABLE;
+  }
+  auto iter = symtab_.find(name);
+  if (iter != symtab_.end())
```
binji: nit: braces

src/binary-writer.cc
```diff
@@ -289,6 +298,30 @@ Index BinaryWriter::GetExceptVarDepth(const Var* var) {
...
+    return iter->second;
+
+  Index sym_index = Index(symbols_.size());
+  symtab_[name] = sym_index;
+  symbols_.push_back(Symbol{sym_index, type, index});
```
binji: `emplace_back` instead?

src/ir.h
```diff
@@ -133,7 +133,6 @@ struct FuncSignature {
 };
 
 struct FuncType {
-  FuncType() = default;
```
binji: hm, not sure how I feel about this change. I don't think these should always have names.

sbc100: But all the callsites to provide a name.   No point in having non-used constructor is there?

src/binary-writer.cc
sbc100: I don't think that works with POD structs

src/binary-reader-objdump.h
sbc100: good point.  I was assuming it was the binary file, but I guess the strings that are provided in the reader callbacks don't outlive the reader itself.

src/ir.h
binji: Guess so.

PR #795, opened 2018-03-06
SIMD Integer to floating point conversion implementation.

lizhengxing:
Including:
f32x4.convert_s/i32x4
f32x4.convert_u/i32x4
f64x2.convert_s/i64x2
f64x2.convert_u/i64x2


lizhengxing:
@binji 
SIMD float convert instructions, PTAL!
Thanks!

@arunetm FYI!

src/interp.cc
```diff
@@ -2903,6 +2909,22 @@ Result Thread::Run(int num_instructions) {
...
+      case Opcode::F32X4ConvertUI32X4:
+        CHECK_TRAP(SimdUnop<v128, uint32_t>(SimdConvert<float, uint32_t>));
+        break;
+
+      case Opcode::F64X2ConvertSI64X2:
+        CHECK_TRAP(SimdUnop<v128, int64_t>(SimdConvert<double, int64_t>));
+        break;
+
+      case Opcode::F64X2ConvertUI64X2:
+        CHECK_TRAP(SimdUnop<v128, uint64_t>(SimdConvert<double, uint64_t>));
```
binji: This case uses `wabt_convert_uint64_to_double` to handle a floating point rounding issue on windows. Can you do the same here?

lizhengxing:
@binji 
Updated the PR, PTAL!
Thanks!

@arunetm FYI!

src/interp.cc
```diff
@@ -2903,6 +2909,26 @@ Result Thread::Run(int num_instructions) {
...
+      case Opcode::F32X4ConvertUI32X4:
+        CHECK_TRAP(SimdUnop<v128, uint32_t>(SimdConvert<float, uint32_t>));
+        break;
+
+      case Opcode::F64X2ConvertSI64X2:
+        CHECK_TRAP(SimdUnop<v128, int64_t>(SimdConvert<double, int64_t>));
+        break;
+
+      case Opcode::F64X2ConvertUI64X2:
+        #if COMPILER_IS_MSVC && defined(_M_X64)
```
binji: Hmm, I'd prefer to keep the #if in `config.h.in`. I forgot that `wabt_convert_*` is not a function otherwise. Can you instead convert the macro in `config.h.in` to an inline function and always use that here? Sorry for the extra effort!

lizhengxing: Sure, I will do that.

binji: Hm, looking at the appveyor build, it looks like it breaks on Windows. Not quite sure why though.

lizhengxing:
@binji 
updated the PR, PTAL!
Thanks!

@arunetm FYI!

PR #796, opened 2018-03-07
objdump: If the export name for a function if one exists

sbc100:


sbc100:
Not sure what to do about the display of non-ascii, but it seems like that problem already exists.

binji:
Nice, lgtm!

PR #797, opened 2018-03-07
SIMD Floating point to integer with saturation implementation.

lizhengxing:
Including:
i32x4.trunc_s/f32x4:sat
i32x4.trunc_u/f32x4:sat
i64x2.trunc_s/f64x2:sat
i64x2.trunc_u/f64x2:sat


lizhengxing:
@binji 
SIMD float trunc with saturation instructions, PTAL!
Thanks!

@arunetm FYI!

PR #800, opened 2018-03-09
Improve output for object of symbol table

sbc100:


src/binary-reader-objdump.cc
```diff
@@ -1289,13 +1292,14 @@ Result BinaryReaderObjdump::OnFunctionSymbol(Index index,
                                              uint32_t flags,
                                              string_view name,
                                              Index func_index) {
+  char type_char = flags & WABT_SYMBOL_FLAG_UNDEFINED ? 'F' : 'U';
```
binji: Shouldn't this be the other way around? `? 'U' : 'F'`

PR #801, opened 2018-03-09
SIMD v128.load/v128.store implementation.

lizhengxing:


lizhengxing:
@binji 
SIMD v128 load/store implementation, PTAL!
Thanks!

@arunetm FYI!

PR #802, opened 2018-03-09
Simd i8x16.extract_lane_s instruction implementation.

lizhengxing:
Including:
1. All necessary code for SIMD lanes accessing.
2. i8x16.extract_lane_s implementation.


lizhengxing:
@binji 
SIMD i8x16.extract_lane_s implementation, PTAL!

SIMD Lane accessing instructions implementation is a little bit complex. It might take you much time to review it. Thanks!

@arunetm FYI!

src/interp.cc
```diff
@@ -1358,6 +1360,24 @@ ValueTypeRep<T> SimdSplat(V lane_data) {
   return ToRep(Bitcast<T>(simd_data));
 }
 
+// Simd instructions with Lane operand.
+// a0: input v128 value. a1: lane index. 
+// typename T: lane data type.
+template <typename R, typename V0, typename V1, typename T>
```
binji: Is there a need to have differing lane index types? I think this can always be int (instead of using V1).

src/interp.cc
```diff
@@ -1358,6 +1360,24 @@ ValueTypeRep<T> SimdSplat(V lane_data) {
...
+ValueTypeRep<R> SimdExtractLane(V0 a0, V1 a1) {
```
binji: Better name for `a1`? Perhaps `lane` or `index`.

src/binary-reader-interp.cc
```diff
@@ -195,6 +195,7 @@ class BinaryReaderInterp : public BinaryReaderNop {
   wabt::Result OnTernaryExpr(wabt::Opcode opcode) override;
   wabt::Result OnUnreachableExpr() override;
   wabt::Result EndFunctionBody(Index index) override;
+  wabt::Result OnSimd1LaneImmExpr(wabt::Opcode opcode, Type type, uint32_t value) override;
```
binji: Why is this called `Simd1LaneImm`? I suppose it will be used for extract and replace lane, is that why?

src/binary-reader-logging.cc
```diff
@@ -353,6 +353,11 @@ Result BinaryReaderLogging::OnTryExpr(Index num_types, Type* sig_types) {
   return reader_->OnTryExpr(num_types, sig_types);
 }
 
+Result  BinaryReaderLogging::OnSimd1LaneImmExpr(Opcode opcode, Type type, uint32_t value) {
+  LOGF("OnSimdLaneExpr (lane: %d)\n", value);
```
binji: Match function name here: `OnSimd1LaneImmExpr`

src/interp.cc
```diff
@@ -2414,6 +2434,13 @@ Result Thread::Run(int num_instructions) {
         break;
       }
 
+      case Opcode::I8X16ExtractLaneS: {
+        v128 dLane = static_cast<v128>(Pop<v128>());
```
binji: Better name than `dLane` here? Other examples in this file use `value`.

src/ir.h
```diff
@@ -266,6 +267,18 @@ typedef OpcodeExpr<ExprType::Convert> ConvertExpr;
 typedef OpcodeExpr<ExprType::Unary> UnaryExpr;
 typedef OpcodeExpr<ExprType::Ternary> TernaryExpr;
 
+template <ExprType TypeEnum, typename T>
```
binji: No need for template param T here, the lane index will always be < 16. Using `int` instead is sufficient.

src/wast-parser.cc
```diff
@@ -1504,6 +1516,32 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
...
+                                           Location loc,
+                                           Const* const_) {
+  WABT_TRACE(ParseSimdConstWithCheck);
+
+  const_->type = type;
+  CHECK_RESULT(ParseSimdConst(const_, Type::I32, sizeof(V)));
+
+  switch (opcode) {
+    case Opcode::I8X16ExtractLaneS: {
+      if(const_->v128_bits.v[0] >= 16)
```
binji: Better to do this check in the validator, so it can be shared when reading text or binary.

src/wast-parser.cc
```diff
@@ -1495,6 +1496,17 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
       break;
     }
 
+    case TokenType::Simd1LaneImm: {
+      Token token = Consume();
+      ErrorUnlessOpcodeEnabled(token);
+      Const const_;
+      CHECK_RESULT((ParseSimdConstWithCheck<uint32_t>(token.opcode(), Type::I32,
```
binji: Why parse as SimdConst? This should just use `ParseNat` to parse a natural number.

src/interp.cc
lizhengxing: Agree.

src/interp.cc
lizhengxing: Agree.

src/binary-reader-interp.cc
lizhengxing: Yes, extract and replace lane only have 1 lane immediate parameter. But v8x16.shuffle have 16 lane Immediate parameters.

src/binary-reader-logging.cc
lizhengxing: Sorry for the omission. I will fix it, Thanks!

src/interp.cc
lizhengxing: Agree.

src/ir.h
lizhengxing: Consider the future v8x16.shuffle implementation.

src/wast-parser.cc
lizhengxing: Ok, I will move the check into validator.

Is it ok to omit such type of error in parsing stage?

src/wast-parser.cc
lizhengxing: Still Consider the future v8x16.shuffle implementation.

ParseNat can only parse 1 natural number. But ParseSimdConst can parse several natural numbers according to the input byte numbers.

src/wast-parser.cc
binji: Yes, I think so. Any module parsed via text will have to be validated before it can be used in most cases anyway.

src/ir.h
binji: I see, though I'd say that `v8x16.shuffle` is different enough to warrant a diferrent Expr type.

src/wast-parser.cc
binji: OK, this seems a bit overly-generic to me, but I'm OK with this direction for now. We can always refactor later if necessary.

lizhengxing: Done.
use ParsrNat for simd extract/replace lane as v8x16.shuffle will be another Expr type.

src/wast-parser.cc
lizhengxing: Done.

src/ir.h
lizhengxing: Done.
I will use a different Expr type for v8x16.shuffle.

src/interp.cc

src/binary-reader-logging.cc
lizhengxing: Done.

src/interp.cc

src/interp.cc

lizhengxing:
@binji 
Updated the PR according to your suggestions, PTAL!
Thanks!

@arunetm FYI!

PR #803, opened 2018-03-10
Add some documentation for wasm2c

binji:
Also remove `WASM_RT_DEFINE_EXTERNAL`, as it doesn't seem to be used.

### DIFF
--- a/wasm2c.md
+++ b/wasm2c.md
@@ -1,0 +1,332 @@
+# wasm2c: Convert wasm files to C source and header
+
+`wasm2c` takes a WebAssembly module and produces an equivalent C source and
+header. Some examples:
+
+```sh
+# parse binary file test.wasm and write test.c and test.h
+$ wasm2c test.wasm -o test.c
+
+# parse test.wasm, write test.c and test.h, but ignore the debug names, if any
+$ wasm2c test.wasm --no-debug-names -o test.c
+```
+
+## Tutorial: .wat -> .wasm -> .c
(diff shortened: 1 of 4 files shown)


binji:
WIP currently, but I thought I'd add a PR early so folks could take a look.

wasm2c.md
```diff
@@ -0,0 +1,332 @@
...
+Next is the `wasm_rt_type_t` enum, which is used for specifying function
+signatures. The four WebAssembly value types are included:
+
+```c
+typedef enum {
+  WASM_RT_I32,
+  WASM_RT_I64,
+  WASM_RT_F32,
+  WASM_RT_F64,
+} wasm_rt_type_t;
```
aardappel: Not obvious why you'd want these at runtime (and quite a bit of the other data in this section as well). I'd imagine you'd simply generate a cast at the call site affert you `assert` the two function indices are the the same.

binji: You only need them if you're linking multiple modules. Since two modules can both define the same function signature with different indexes, the embedder needs to determine whether they're the same.

Perhaps there should be an option to say: I only need one module, don't give me this other stuff.

binji:
Still working on this, but found it was nicer to factor out the `wasm_rt*` stuff first: https://github.com/WebAssembly/wabt/pull/813

binji:
Coming back to this now. Going to land what I have and add more in another PR.

Issue #807, opened 2018-03-14
wasm-objdump should have a way to show locals declared in a function


eholk:
It'd be helpful in debugging generated Wasm code. It doesn't have to be wasm-objdump, but that seems like the most logical tool.

binji:
Good point, it shouldn't be too hard to add.

PR #808, opened 2018-03-15
Print locals in objdump disassembly

binji:
Fixes issue #807.


binji:
Hm, I guess it would be nicer to show the actual bytes that generate that info too.

binji:
Discussed offline, decided it would be better to show bytes, and disassembly only. PTAL!

eholk:
Thanks for getting this done so quickly! I applied the patch locally and it helped me debug my issue.

PR #811, opened 2018-03-16
SIMD v8x16.shuffle implementation.

lizhengxing:


lizhengxing:
@binji 
SIMD v8x16.shuffle instruction, PTAL!
Thanks!

@arunetm FYI!

PR #813, opened 2018-03-16
Separate out wasm-rt.h and wasm-rt-impl.{c,h}

binji:
This makes it easier to use outside of running wasm2c spec tests.


binji:
This is mostly just reorganizing code, so I'm gonna land for now. Feel free to add comments if you notice any issues.

PR #829, opened 2018-04-20
Enable memory growth when compiling libwabt.js

dcodeIO:
This should fix memory issues as described in https://github.com/wasdk/WebAssemblyStudio/pull/202#issuecomment-383209090.

Options are taken from [Binaryen's build-js.sh](https://github.com/WebAssembly/binaryen/blob/fd3b3e54bd97abbf8269b33d937ad2f44ba4bb60/build-js.sh#L41) script. One downside in enabling memory growth is that the code will emit as `"almost asm";` instead of `"use asm";`, but it seems that can't be avoided if the goal is to support any size of WASM binaries.


kripken:
Looks like good options, but note that `-s ELIMINATE_DUPLICATE_FUNCTIONS=1` will make the build slower (just in asm.js, not in wasm), and `--llvm-lto 1` should be tested carefully (LLVM LTO is not on by default anywhere AFAIK and had bugs in the past).

dcodeIO:
> --llvm-lto 1 should be tested carefully (LLVM LTO is not on by default anywhere AFAIK and had bugs in the past)

Thanks for the insights! I've tested the generated binary using the new options against [this simple test suite](https://github.com/AssemblyScript/wabt.js/blob/master/tests/index.js) and it didn't trigger any errors.

PR #830, opened 2018-04-23
Refer to target section by index in relocation section

sbc100:
See https://github.com/WebAssembly/tool-conventions/issues/52

### DIFF
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -1419,16 +1419,11 @@ Result BinaryReader::ReadNameSection(Offset section_size) {
 
 Result BinaryReader::ReadRelocSection(Offset section_size) {
   CALLBACK(BeginRelocSection, section_size);
-  uint32_t section;
-  CHECK_RESULT(ReadU32Leb128(&section, "section"));
-  string_view section_name;
-  if (static_cast<BinarySection>(section) == BinarySection::Custom) {
-    CHECK_RESULT(ReadStr(&section_name, "section name"));
-  }
+  uint32_t section_index;
+  CHECK_RESULT(ReadU32Leb128(&section_index, "section index"));
   Index num_relocs;
   CHECK_RESULT(ReadCount(&num_relocs, "relocation count"));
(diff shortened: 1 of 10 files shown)


SanderSpies:
Not sure if it should be in this PR, but to get it to work I needed to add the linking version.

binji:
lgtm

src/binary-reader.cc
```diff
@@ -1454,6 +1449,9 @@ Result BinaryReader::ReadRelocSection(Offset section_size) {
 
 Result BinaryReader::ReadLinkingSection(Offset section_size) {
   CALLBACK(BeginLinkingSection, section_size);
+  uint32_t version;
+  CHECK_RESULT(ReadU32Leb128(&version, "version"));
```
alexcrichton: FWIW I think this breaks using tools like `wasm2wat` with code generated by rustc today. Rustc is still on LLVM 6 (and the LLD at the time) which I don't think encodes the version here, so this is causing `wasm2wat` to fail on any rustc-generated binary (which currently retain a small `linking` section).

Would it be possible to ignore errors here or otherwise not fail decoding if this fails? According to the [binary encoding spec](https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#name-section):

> If a WebAssembly implementation interprets the payload of any custom section during module validation or compilation, errors in that payload must not invalidate the module.

binji: Ah, thanks for mentioning this. We've discussed in the past that it would be best to have an option to require or ignore errors in custom sections, and this is a good example of why that's needed.

I'll see about adding that today.

alexcrichton: Awesome, thanks!

I'm also hoping we can figure out a strategy for ourselves to upgrade LLVM in the meantime :)

binji: This landed here: https://github.com/WebAssembly/wabt/commit/9890d45ff4e04a89ba53137d5f569c5038df8d27

You can use the `--ignore-custom-section-errors` flag in `wasm2wat` or `wasm-validate` for this.

alexcrichton: Awesome, thanks!

PR #831, opened 2018-04-30
Read and write module names in the names section

binji:


binji:
Yep: https://webassembly.github.io/spec/core/appendix/custom.html#binary-modulenamesec

lukewagner:
Thanks, works great!  At first I thought it didn't work because the first thing I tried was `wasm-objdump -x`.  But `wasm2wat` shows the module name great, so I think maybe just `wasm-objdump` could print the module name?

binji:
> But wasm2wat shows the module name great, so I think maybe just wasm-objdump could print the module name?

Added in https://github.com/WebAssembly/wabt/pull/833.

PR #832, opened 2018-05-03
Fix one bug for converting v128.shuffle from wasm format to wat format.

lizhengxing:
When converting v128.shuffle from wasm format to wat format, the '$' char
must be removed as the const value in v128.shuffle instruction expects
a valid simd v128 const format,
i.e. 0xXXX 0xXXX 0xXXX 0xXXX or Int32 Int32 Int32 Int32

otherwise it will encounter errors when converting
v128.shuffle from wat format back to wasm format again.

### DIFF
--- a/src/wat-writer.cc
+++ b/src/wat-writer.cc
@@ -869,7 +869,7 @@ Result WatWriter::ExprVisitorDelegate::OnSimdLaneOpExpr(SimdLaneOpExpr* expr) {
 Result WatWriter::ExprVisitorDelegate::OnSimdShuffleOpExpr(
     SimdShuffleOpExpr* expr) {
   writer_->WritePutsSpace(expr->opcode.GetName());
-  writer_->Writef(" $0x%08x %08x %08x %08x", (expr->val.v[0]), (expr->val.v[1]),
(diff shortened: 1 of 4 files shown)


lizhengxing:
@binji 
One bug fix, PTAL, thanks!

@arunetm FYI!

binji:
Hm, interesting. I wonder why this wasn't caught with the roundtrip tests. I suppose those only run with parse tests. Can you add a parse test for this? Perhaps add a new file in `test/parse/expr/simd.txt`. Thanks!

lizhengxing:
@binji 
Sure, But maybe a little bit late as I'm busying in other issue. Thanks!

binji:
No problem, there's no rush!

lizhengxing:
@binji 
Updated the PR, PTAL, thanks!

@arunetm FYI!

lizhengxing:
@binji 
I tested this PR on local machine and passed all tests, But it still met a check error on wabt server.
Woud you please give some hints of this check error as I'm still busying in other work?

Thanks!

binji:
> I tested this PR on local machine and passed all tests, But it still met a check error on wabt server.
Woud you please give some hints of this check error as I'm still busying in other work?

Sorry, there was an emscripten regression that caused this. I've fixed the bug upstream but the fix hasn't been rolled out yet.

lizhengxing:
@binji 
Got it, thank you very much!

PR #834, opened 2018-05-08
Display module name in wasm-objdump

binji:
See comment here: https://github.com/WebAssembly/wabt/pull/831#issuecomment-387491592


binji:
Maybe we should display the module name somewhere else more prominent than the name section, but I didn't really see an obvious place to do that.

sbc100:
Yes, how about one the first line which currently:

`module-name.wasm:	file format wasm 0x1`

Could be:

`module-name.wasm[my_module]:	file format wasm 0x1`

PR #839, opened 2018-05-14
wasm-objdump: Clearly distinguish undefined symbols in symbol table

sbc100:


binji:
Can you describe this a bit more in the CL description?

Issue #838, opened 2018-05-14
0000099: error: unable to read u32 leb128: stack pointer index


MarkSwanson:
I'm compiling Rust code to wasm, and I'm getting the leb128 issue with wasm-validate.

Rust code:

#[no_mangle]
pub extern "C" fn test(n: i32) -> i32 {
    n
}

I'm using the latest wabt master branch as of today, and a Rust May 9, 2018 nightly.
$ cargo build --release --target wasm32-unknown-unknown
$ wasm-validate target/wasm32-unknown-unknown/release/test.wasm

Has LLVM perhaps been updated to use a different format? Maybe wabt is expecting an older format?

binji:
I chatted with @sbc100 about this, and we're wondering if you're using an older version of wabt. The newest versions should be expecting a linking metadata version which is likely provided by the generated wasm file.

sbc100:
I can take a look at this one.

sbc100:
Looks like then new version field is not being added to the linking metadata.  It was recently added so its not surprising.   Also, the issue was being mis-reported due to a bug: https://github.com/WebAssembly/wabt/pull/841.  Oops.

After that change land the error message is more useful:
```
/bin/wasm-objdump ~/Downloads/test.wasm -h

test.wasm:	file format wasm 0x1
*ERROR*: @0x00000097: invalid linking metadata version: 3
```

MarkSwanson:
Thanks for looking into this.
Do you think the correct thing to do next would be for me (or someone) to log a bug with llvm / LLD?

binji:
@sbc100 wrote the wasm implementation of LLD so I imagine he will fix it.

I did think of another thing to change in wabt though; currently `wasm-validate` is failing to parse the binary because it hits the `linking` custom section and it doesn't match. But validation never relies on a custom section, they always should be ignored. You can pass `--ignore-custom-section-errors` to have it not do this, but it should probably be the default (at least for `wasm-validate` -- for `wasm2wat`, I'm not sure).

sbc100:
I suspect that the compiler that generated that object file just needs to be updated to a version that writes the newer metadata format.  Do you know how that object was generated?  By which llvm version exactly?

I don't think `--ignore-custom-section-errors` should be the default.  If you have old/invalid metadata you probably want to catch it early.  Thats the point of wasm-validate right?

MarkSwanson:
I generated the object like this:
cargo build --release wasm32-unknown-unknown
[...]

PR #841, opened 2018-05-14
Fix broken version check on metadata section

sbc100:


src/binary-reader.cc
```diff
@@ -1459,7 +1459,7 @@ Result BinaryReader::ReadLinkingSection(Offset section_size) {
   CALLBACK(BeginLinkingSection, section_size);
   uint32_t version;
   CHECK_RESULT(ReadU32Leb128(&version, "version"));
-  ERROR_UNLESS(version = 1, "invalid linking metadata version");
+  ERROR_UNLESS(version == 1, "invalid linking metadata version: %u", version);
```
binji: brutal!

shepmaster:
Howdy! This exposes an error when using WASM generated by Rust:

```
error: invalid linking metadata version: 3
```

Is there a flag I should be using for this?

shepmaster:
Perhaps related to the discussion in #838

sbc100:
Yes, you may be able to use `--ignore-custom-section-errors`.  The problem should fix itself once Rust pulls  in a new llvm version.

shepmaster:
Hmm, the flag doesn't seem to take effect; perhaps I'm doing something wrong?

```
$ wasm2wat --help | grep ignore-custom
      --ignore-custom-section-errors          Ignore errors in custom sections

$ wasm2wat --fold-exprs --generate-names --inline-exports --ignore-custom-section-errors -o /tmp/out.wat target/wasm32-unknown-unknown/debug/playground.wasm.gc
002f892: error: invalid linking metadata version: 3
```

shepmaster:
Moved discussion to #842.

PR #846, opened 2018-05-22
Add support for new relocation types used in DWARF sections

sbc100:


src/binary-reader-objdump.cc
```diff
@@ -175,6 +185,19 @@ class BinaryReaderObjdumpPrepass : public BinaryReaderObjdumpBase {
  public:
   using BinaryReaderObjdumpBase::BinaryReaderObjdumpBase;
 
+  Result BeginSection(BinarySection section_code, Offset size) override {
+    BinaryReaderObjdumpBase::BeginSection(section_code, size);
+    if (section_code != BinarySection::Custom)
```
binji: braces around if

PR #851, opened 2018-06-02
[objdump] Include elem segment offset when printing table elements

sbc100:


src/binary-reader-objdump.cc
```diff
@@ -1131,41 +1146,62 @@ void BinaryReaderObjdump::PrintInitExpr(const InitExpr& expr) {
   }
 }
 
-Result BinaryReaderObjdump::OnInitExprF32ConstExpr(Index index,
-                                                   uint32_t value) {
-  InitExpr expr;
-  expr.type = InitExprType::F32;
-  expr.value.f32 = value;
+static Result InitExprToConstOffset(const InitExpr& expr,
+                                    uint32_t& out_offset) {
```
binji: nit: use * for out param

Issue #853, opened 2018-06-04
Emscripten travis failure


binji:
Not quite sure what's failing here, though maybe I'm doing something weird w/ `--pre-js`:

See [this travis failure](https://travis-ci.org/WebAssembly/wabt/jobs/387896293) for example.

```
Assertion failed: could not find the assigment to "asmLibraryArg". perhaps 
--pre-js or --post-js code moved it out of the global scope? (things like that 
should be done after emcc runs, as they do not need to be run through the 
optimizer which is the special thing about --pre-js/--post-js code)
```

@kripken, any clues?

binji:
Ah, thanks, I see. The error message makes more sense now. I'll give that a shot!

PR #855, opened 2018-06-04
Fix emscripten build

binji:
It's not correct to use `--pre-js` and `--post-js` to wrap the module in
a function instance; instead we're supposed to use `-s MODULARIZE=1`.

This still keeps the build as (almost) asm.js, as switching to wasm is a
bit more work (we need to preload the wasm binary module).

This fixes issue #853.


kripken:
lgtm. And sorry for the confusion with those emcc APIs, we only recently realized that bad things could happen if the optimizer was not aware of the scoping issue, and so started to issue an error for people adding an enclosing scope themselves.

binji:
No problem, I just didn't really know why it broke. The error probably should have clued me in, but it didn't really click until you explained it. :-)

PR #860, opened 2018-06-06
Use named offset type and clean up PushRep doc

ahorn:
Slightly OCD perhaps, fixes minor inconsistencies in the interpreter header.


binji:
lgtm, thanks!

PR #862, opened 2018-06-15
Fix integer overflow w/ too many locals

binji:
The maximum number of locals in a function is 2**32-1.

### DIFF
--- a/test/binary/bad-too-many-locals.txt
+++ b/test/binary/bad-too-many-locals.txt
@@ -1,0 +1,35 @@
+;;; TOOL: run-objdump-gen-wasm
+;;; ARGS: -x
+;;; ERROR: 1
+magic
+version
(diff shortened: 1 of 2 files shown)


test/binary/bad-too-many-locals.txt
```diff
@@ -0,0 +1,35 @@
...
+magic
+version
+section(TYPE) { count[1] function params[0] results[0] }
+section(FUNCTION) { count[1] type[0] }
+section(CODE) {
+  count[1]
+  func {
+    locals [
+      decl_count[2]
+      leb_u32(0xffffffff) i32
```
pepyakin: this is a sick tool! To create my test case, I had to take out my hex editor!

binji: Thanks! It has been pretty useful for generating invalid tests like these.

PR #865, opened 2018-06-18
Relax the check for alloca

### LINKED ISSUE #864: error no alloca
Build fails on OpenBSD (and other BSDs) because of the check for alloca.

rrrapha:
Resolves #864


binji:
lgtm, thanks!

PR #873, opened 2018-07-14
Use the wasm-objdump option -h for --help only, not for --headers

### LINKED ISSUE #872: The option -h of wasm-objdump is ambiguous
<pre>
options:
  -h, --headers                Print headers
  -j, --section=SECTION        Select just one section
  -s, --full-contents          Print raw section contents
  -d, --disassemble            Disassemble function bodies
      --debug                  Print extra debug information
  -x, --details                Show section details
  -r, --reloc                  Show relocations inline with disassembly
  -h, --help                   Print this help message
</pre>

rrrapha:
Resolves #872

### DIFF
--- a/src/tools/wasm-objdump.cc
+++ b/src/tools/wasm-objdump.cc
@@ -42,7 +42,7 @@ static std::unique_ptr<FileStream> s_log_stream;
 static void ParseOptions(int argc, char** argv) {
   OptionParser parser("wasm-objdump", s_description);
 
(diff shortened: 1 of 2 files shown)


binji:
Was there something wrong with this PR? I was going to land it.

rrrapha:
Sorry for the confusion.. you can of course merge this if you want. ;)
I'm not convinced this is the best solution though (see my latest comment here: #872).

PR #886, opened 2018-08-04
handle traps in host functions

poemm:
With these changes, when my host function returns `interp::Result::TrapHostTrapped`, execution traps as desired.

By the way, great work on wabt! Useful tools!


binji:
Thanks for the fix!

I took a look at https://github.com/poemm/pywebassembly too. Pretty cool stuff! I like that it follows the spec very closely; I started working on something similar in JS but stalled a bit. :-)

binji:
BTW, I added a test for this here: https://github.com/WebAssembly/wabt/pull/887

PR #890, opened 2018-08-14
[interp] Add rot13 interpreter example

binji:
This demonstrates sharing memory between a WebAssembly module and the
embedder.


Adlai-Holler:
Wow this is extremely useful as a demonstration of a minimal buffer mutation API! Thanks @binji I'll be borrowing this 😀

PR #893, opened 2018-08-14
Update Gtest to current revision

### LINKED ISSUE #892: third_party/gtest fails to build with Visual Studio
Current version of Google test is using non-standard namespace that is not supported in Visual Studio 2017: google/googletest#1111

Would it be possible to update gtest to newer version? Its current head seems to work without changes to WABT code.

penzn:
Closes #892


penzn:
I am going to rollback the revision a little bit. There is a [commit][c] in gtest that introduces "missing initializer" warning, which is unfortunate. With this revision it should be good with both Visual Studion and GCC/Clang.

Waiting for the builds.

[c]: https://github.com/google/googletest/commit/87a4cdd

penzn:
Alternatively, it can be solved by silencing the warning. I can do it if that works for you better.

binji:
Yeah, we can probably just silence the warning in the future. But I think it's fine to just use the older version for now, as long as it fixes the bug.

PR #896, opened 2018-08-16
add minimal manpages (mdoc-formatted)

### LINKED ISSUE #417: Manual pages
Hello,

After installing `wabt` through Homebrew, I noticed none of the executables had a corresponding manpage.

If assistance is needed, I'd be happy to write them for you. =) I thought I'd ask first instead of submit a PR up-front, as not all maintainers are up to maintaining manpages in addition to Markdown-based docs. Particularly when the subject is very build-centric in nature.

Let me know if there's anything I can do.

rrrapha:
Resolves #417

### DIFF
--- a/man/wasm-interp.1
+++ b/man/wasm-interp.1
@@ -1,0 +1,62 @@
+.Dd $Mdocdate$
+.Dt WABT 1
+.Os
+.Sh NAME
+.Nm wasm-interp
+.Nd decode and run a WebAssembly binary file
+.Sh SYNOPSIS
+.Nm wasm-interp
+.Op options
+.Ar filename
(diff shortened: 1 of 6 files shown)


rrrapha:
The information is mostly taken from README.md and extracted from the usage output. Only the tools mentioned in README.md are covered.
The manual pages are not hooked up to the build system yet.

binji:
Thanks! I don't know the syntax at all, but just a quick glance looks good to me.

sbc100:
Shouldn't they be added to the `make install` target too?

binji:
Yeah, but I think it's fine to do that in another PR.

PR #900, opened 2018-08-23
Reset Executor’s Thread at start of RunFunction

spoonincode:
At the end of Executor::RunFunction(), the Thread is Reset() so that
the Executor can be reused for subsequent RunFunction()s. However in
some cases the execution won’t make it that far in RunFunction() —
for example, if a HostModule’s callback throws an exception. Move the
call to the Thread’s Reset() to the start of RunFunction() so that
the Thread’s state is always as expected when Executor starts a function.


binji:
Thanks, lgtm!

PR #909, opened 2018-09-10
Fix wasm2c on windows (via MSVC)

binji:


src/wasm2c.c.tmpl
```diff
@@ -63,12 +76,88 @@ DEFINE_STORE(i64_store8, u8, u64);
...
+#if _MSC_VER
+#if _M_IX86
+static inline unsigned long LowDword(unsigned __int64 value) {
+  return (unsigned long)value;
+}
+
+static inline unsigned long HighDword(unsigned __int64 value) {
+  unsigned long high;
+  memcpy(&high, (unsigned char*)&value + sizeof(high), sizeof(high));
+  return high;
```
jaykrell: Why not `return value >> 32;`?

src/wasm2c.c.tmpl
```diff
@@ -133,15 +222,49 @@ DEFINE_STORE(i64_store32, u32, u64);
 #define I32_TRUNC_U_F64(x) TRUNC_U(u32, f64, UINT32_MAX, >,  x)
 #define I64_TRUNC_U_F64(x) TRUNC_U(u64, f64, UINT64_MAX, >=, x)
 
-#define DEFINE_REINTERPRET(name, t1, t2)  \
-  static inline t2 name(t1 x) {           \
-    t2 result;                            \
-    memcpy(&result, &x, sizeof(result));  \
-    return result;                        \
+#define DEFINE_REINTERPRET(name, t1, t2)   \
```
jaykrell: Some of these can be more efficient.
See:
https://github.com/microsoft/ChakraCore/pull/5890/files#diff-51806c1cfd597ef0db6e366e700bc6c7R199
https://github.com/microsoft/ChakraCore/pull/5890/files#diff-51806c1cfd597ef0db6e366e700bc6c7R210

src/wasm2c.c.tmpl
binji: True, I copied this from config.h.in, but I'm not sure why I wrote it this way. :-)

src/wasm2c.c.tmpl
binji: Good point, though I'd rather get this working and landed before optimizing. It still isn't quite there, unfortunately. It's pretty hard for me to test, since I don't normally run windows.

keithw:
Done in #1843.

PR #910, opened 2018-09-10
Optimize interpreter and `Opcode::FromCode`

binji:
`Opcode::FromCode` calculated the opcode given a prefix/code pair by
using lower_bound over the list of all `OpcodeInfo`s. This was happening
for every instruction, which is incredibly slow.

Since the interpreter's format is internal only, we can use any encoding
we want, so it's simpler and faster to use the `Opcode::Enum` directly
without calling `Opcode::FromCode`.

`Opcode::FromCode` is also used when reading a binary file, so it should
be optimized anyway. Instead of using the `infos_` table, which is
indexed by the opcode's `enum_` value, we create a new
statically-defined table that maps from prefix-code pair to its enum
value.

Unfortunately, this can't be done easily in C++ because it does not
currently support designated array initializers, so this table is
created in a C file instead, `opcode-code-table.c`.

### DIFF
--- a/src/opcode-code-table.c
+++ b/src/opcode-code-table.c
@@ -32,10 +32,10 @@ typedef enum WabtOpcodeEnum {
 _Static_assert(Invalid <= 65536, "Too many opcodes");
 
 /* The array index calculated below must match the one in Opcode::FromCode. */
-uint32_t WabtOpcodeCodeTable[] = {
+uint32_t WabtOpcodeCodeTable[65536] = {
 #define WABT_OPCODE(rtype, type1, type2, type3, mem_size, prefix, code, Name, \
                     text)                                                     \
-  [prefix * 256 + code] = Name,
+  [(prefix << 8) + code] = Name,
 #include "opcode.def"
 #undef WABT_OPCODE
 };
--- a/src/opcode-code-table.c
+++ b/src/opcode-code-table.c
@@ -14,9 +14,7 @@
  * limitations under the License.
  */
 
-/* This structure is defined in C because C++ doesn't (yet) allow you to use
- * designated array initializers, i.e. [10] = {foo}.
- */
+#include "src/opcode-code-table.h"
 
 #include <stdint.h>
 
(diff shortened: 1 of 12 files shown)


src/opcode-code-table.c
```diff
@@ -0,0 +1,41 @@
...
+typedef enum WabtOpcodeEnum {
+#define WABT_OPCODE(rtype, type1, type2, type3, mem_size, prefix, code, Name, \
+                    text)                                                     \
+  Name,
+#include "opcode.def"
+#undef WABT_OPCODE
+  Invalid,
+} WabtOpcodeEnum;
+
+_Static_assert(Invalid <= 65536, "Too many opcodes");
```
jgravelle-google: Maybe `static const size_t kNumOpcodes = 65536;` in opcode.h
Especially because the extern declaration there needs to match

src/opcode-code-table.c
```diff
@@ -0,0 +1,41 @@
...
+
+/* The array index calculated below must match the one in Opcode::FromCode. */
+uint32_t WabtOpcodeCodeTable[65536] = {
+#define WABT_OPCODE(rtype, type1, type2, type3, mem_size, prefix, code, Name, \
+                    text)                                                     \
+  [(prefix << 8) + code] = Name,
```
jgravelle-google: This file has to be compiled as C for this to work, right?
Not gonna lie, this is really neat. Meaning both "nifty" and "clean"

src/opcode-code-table.c
binji: It's a bit tricky since I can't include opcode.h here (since it's C++). So I'd have to make a C header and include it there. I was trying to avoid adding a new file, but there's no reason to do so, really.

src/opcode-code-table.c
binji: Yep, it's a C99-only feature (until C++20, I think). It works, but I dunno, I thought about doing this before and thought it was kind of gross. :-) But the alternatives are not great either, like building the table at runtime, generating it via script, or writing it out manually.

jgravelle-google: Maybe a `constexpr` static initializer? Should be doable even with C++11 constexprs. Not sure if it's less-gross though

aardappel: Hmm, a table of tables|null would be more space efficient.

PR #916, opened 2018-09-26
Update Wasm simd instruction opcodes

lizhengxing:
Update the wasm simd instruction opcodes as they're changed in latest spec.


lizhengxing:
@binji 
Updated the simd op according to the latest wasm simd spec, PTAL! 
Thanks!

@aretmr FYI!

PR #918, opened 2018-10-01
Tailcall

binji:
This doesn't do any of the real work yet, it just adds the
ReturnCall/ReturnCallIndirect Expr and Opcode types, and the
"--enable-tail-call" flag.

Still TODO:

* Parse the opcodes in binary-reader.cc
* Validate the opcodes in validator.cc and type-checker.cc
* Implement the opcodes in interp.cc
* Write standard wabt tests, and enable the spec proposal tests too


binji:
cc @frankmccabe

Issue #928, opened 2018-10-12
Implement further segment validation checks for bulk memory operations


alexcrichton:
First added in https://github.com/WebAssembly/wabt/pull/927 the bulk memory operations aren't fully validated yet. Right now they don't verify that the immediate indexes passed to `{table,memory}.{init,drop}` are valid segments for the table/memory sections. For validation to be complete we'll need to validate that the index is indeed in bounds ~~and it's also a passive segment~~ (not needed any more!).

conrad-watt:
We decided recently not to validate whether the index is a passive segment ahead of time (https://github.com/WebAssembly/bulk-memory-operations/issues/27, https://github.com/WebAssembly/bulk-memory-operations/pull/31).

alexcrichton:
Oh sorry, my bad! I've updated the description

binji:
I think these are all fixed now.

PR #930, opened 2018-10-13
Add more bulk memory tests; use vars for segments  

binji:
* Allow for names in data/elem segments:
```
(data $foo (i32.const 0) "abcdef")
```

* These names can be referenced by the memory/table instructions that
operate on segments:
```
memory.drop $foo
```

* Fix running wasm-objdump with bulk-memory instructions

* Check for valid data/elem segments in validation

* Check that bulk-memory is enabled when parsing text

Partial fix for #928; still needs interpreter checks too.


binji:
cc @alexcrichton

alexcrichton:
Thanks for adding these! This all looks good to me 👍

PR #934, opened 2018-10-17
Add test for disabled opcodes in binary reader

binji:
It's easy to forget to add a check for unexpected opcodes, and it's
tedious to have to write additional tests in `test/binary/*`.

This way we can test all potentially disabled instructions at once.

### DIFF
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -202,19 +202,23 @@ void WABT_PRINTF_FORMAT(2, 3) BinaryReader::PrintError(const char* format,
 }
 
 Result BinaryReader::ReportUnexpectedOpcode(Opcode opcode,
-                                            const char* message) {
-  const char* maybe_space = " ";
-  if (!message) {
-    message = maybe_space = "";
+                                            const char* where) {
+  std::string message = "unexpected opcode";
+  if (where) {
+    message += ' ';
+    message += where;
   }
-  if (opcode.HasPrefix()) {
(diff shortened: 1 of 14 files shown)


src/opcode.h
```diff
@@ -54,14 +57,19 @@ struct Opcode {
   bool HasPrefix() const { return GetInfo().prefix != 0; }
   uint8_t GetPrefix() const { return GetInfo().prefix; }
   uint32_t GetCode() const { return GetInfo().code; }
-  size_t GetLength() const { return HasPrefix() ? 2 : 1; }
+  size_t GetLength() const {
+    return HasPrefix() ? 1 + U32Leb128Length(GetCode()) : 1;
```
jgravelle-google:
```suggestion
    return GetBytes().size();
```

src/binary-reader.cc
```diff
@@ -806,6 +810,7 @@ Result BinaryReader::ReadFunctionBody(Offset end_offset) {
       case Opcode::F32Load:
       case Opcode::F64Load:
       case Opcode::V128Load: {
+        ERROR_UNLESS_OPCODE_ENABLED(opcode);
```
jgravelle-google: Any reason to not do this for every opcode, i.e. before the switch? Non-disableable opcodes will always pass, and we'll cut down on the code needed to add new disabled opcodes.

binji: Probably just a premature optimization. I'll take a look and see if it makes a difference.

PR #937, opened 2018-10-28
Check type use vs. explicit sig. in call_indirect

binji:
The `call_indirect` instruction can take a type use, but can also take
an explicit signature:

```
call_indirect (type $foo)               ;; type use
call_indirect (param i32)               ;; explicit signature
call_indirect (type $bar) (result f32)  ;; both
```

These type signatures must match, or the wat file is considered
malformed. This was properly checked in the spec tests, because it
was performed in a separate function `wabt::ValidateFuncSignatures`.
This wasn't validated in `wat2wasm`, because it performs the full module
validation via `wabt::ValidateModule`, which doesn't share code with
`ValidateFuncSignatures`.

It might be a good idea to share the code between these two, but for now
it's enough to fix this bug by performing the same check in the module
validation path.

Fixes issue #936.

### DIFF
--- a/src/validator.cc
+++ b/src/validator.cc
@@ -602,10 +602,7 @@ Result Validator::OnCallIndirectExpr(CallIndirectExpr* expr) {
   if (current_module_->tables.size() == 0) {
     PrintError(&expr->loc, "found call_indirect operator, but no table");
   }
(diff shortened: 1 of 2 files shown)


binji:
Travis failure is an unrelated emscripten error.

PR #938, opened 2018-10-29
Add support for the reference types proposal

alexcrichton:
This commit adds support for the reference types proposal to wabt.
Namely it adds new opcodes like `table.{get,set,grow}` as well as adds a
new `anyref` type. These are plumbed throughout for various operations
in relatively simple fashions, no support was added for a subtyping
relationship between `anyref` and `anyfunc` just yet.

This also raises the restriction that multiple tables are disallowed,
allowing multiple tables to exist when `--enable-reference-types` is
passed.


alexcrichton:
One final thing I know of that would want to be sorted out before merging is the binary representation of the opcodes. Currently the [reference types overview](https://github.com/WebAssembly/reference-types/blob/master/proposals/reference-types/Overview.md) doesn't go into detail about the binary encodings and the [table instructions specification](https://github.com/WebAssembly/reference-types/blob/master/document/core/binary/instructions.rst#table-instructions) doesn't currently mention `table.grow` (yet!). The [experimental Gecko implementation](https://bugzilla.mozilla.org/show_bug.cgi?id=1450263), however, chooses a different encoding for `table.{get,set}` as well as picks an encoding for `table.grow`. 

I think it should be easy to sort this out here whichever way we want it to be, I just want to make sure that this isn't merged before that's settled! I also don't mind sending a PR to the reference-types proposal with an update of encodings too!

alexcrichton:
Oh I should also mentioned that this goes ahead and uses the renamings proposed of `table.{get,set,grow}` rather than `get_table` and `set_table` and such

alexcrichton:
Oh hm I'm also not entirely sure where travis is going awry with the asmjs build (I think?). Is there perhaps a file I need to regenerate somewhere as well?

binji:
> Oh hm I'm also not entirely sure where travis is going awry with the asmjs build (I think?).

It's an unrelated error, I believe. I filed a bug here: https://github.com/WebAssembly/wabt/issues/939

src/binary-reader-ir.cc
```diff
@@ -172,6 +172,9 @@ class BinaryReaderIR : public BinaryReaderNop {
   Result OnTableCopyExpr() override;
   Result OnTableDropExpr(Index segment_index) override;
   Result OnTableInitExpr(Index segment_index) override;
+  Result OnTableGetExpr(Index segment_index) override;
```
binji: `table_index` for each of these

src/binary-reader-ir.cc
```diff
@@ -847,6 +851,18 @@ Result BinaryReaderIR::OnTableInitExpr(Index segment) {
   return AppendExpr(MakeUnique<TableInitExpr>(Var(segment)));
 }
 
+Result BinaryReaderIR::OnTableGetExpr(Index segment) {
```
binji: `table_index`

src/binary-reader-nop.h
```diff
@@ -244,6 +244,9 @@ class BinaryReaderNop : public BinaryReaderDelegate {
   Result OnTableCopyExpr() override { return Result::Ok; }
   Result OnTableDropExpr(Index segment_index) override { return Result::Ok; }
   Result OnTableInitExpr(Index segment_index) override { return Result::Ok; }
+  Result OnTableGetExpr(Index segment_index) override { return Result::Ok; }
```
binji: `table_index`

src/binary-reader.h
```diff
@@ -240,6 +240,9 @@ class BinaryReaderDelegate {
   virtual Result OnTableCopyExpr() = 0;
   virtual Result OnTableDropExpr(Index segment_index) = 0;
   virtual Result OnTableInitExpr(Index segment_index) = 0;
+  virtual Result OnTableGetExpr(Index segment_index) = 0;
```
binji: `table_index`

test/typecheck/bad-reference-types-no-table.txt
```diff
@@ -0,0 +1,20 @@
+;;; TOOL: wat2wasm
+;;; ARGS: --enable-bulk-memory
```
binji: --enable-reference-types

lars-t-hansen:
I think there are many open questions around encoding.  I have filed a bug here: https://github.com/WebAssembly/reference-types/issues/18 to get some eyes on this.  Basically, the bulk memory ops will require work for multi-table support and we could usefully coordinate the encodings between the two proposals.  I'll file a similar bug for bulk memory.

alexcrichton:
Ok I've rebased over the changes that have happened inbetween since this started, and I was curious if perhaps since the binary encoding is still in-the-air-but-implemented-in-a-few-places it might be good to get this in as a starting point and the binary format can be tweaked as the spec evolves?

binji:
Yes, I think we probably should (same as with other in-progress proposals). See further discussion about the instruction encoding here: https://github.com/WebAssembly/reference-types/issues/29

src/binary-writer.cc
```diff
@@ -601,6 +601,62 @@ void BinaryWriter::WriteExpr(const Func* func, const Expr* expr) {
       WriteU32Leb128(stream_, 0, "table.init reserved");
       break;
     }
+    case ExprType::TableGet: {
+      Index index =
+          module_->GetElemSegmentIndex(cast<TableGetExpr>(expr)->var);
```
binji: Should be `GetTableIndex` for these

src/wat-writer.cc
```diff
@@ -1008,12 +1048,15 @@ void WatWriter::WriteFoldedExpr(const Expr* expr) {
...
     case ExprType::MemorySize:
+    case ExprType::TableSize:
     case ExprType::Unreachable:
       PushExpr(expr, 0, 1);
       break;
 
     case ExprType::DataDrop:
     case ExprType::ElemDrop:
+    case ExprType::RefIsNull:
+    case ExprType::RefNull:
```
binji: Should be in the `PushExpr(expr, 0, 1)` group.

src/wat-writer.cc
```diff
@@ -1008,12 +1048,15 @@ void WatWriter::WriteFoldedExpr(const Expr* expr) {
...
     case ExprType::LocalGet:
     case ExprType::MemorySize:
+    case ExprType::TableSize:
     case ExprType::Unreachable:
       PushExpr(expr, 0, 1);
       break;
 
     case ExprType::DataDrop:
     case ExprType::ElemDrop:
+    case ExprType::RefIsNull:
```
binji: Should be in the `PushExpr(expr, 1, 1)` group.

src/wat-writer.cc
```diff
@@ -1031,9 +1074,15 @@ void WatWriter::WriteFoldedExpr(const Expr* expr) {
     case ExprType::LocalTee:
     case ExprType::MemoryGrow:
     case ExprType::Unary:
+    case ExprType::TableGrow:
+    case ExprType::TableGet:
       PushExpr(expr, 1, 1);
       break;
 
+    case ExprType::TableSet:
+      PushExpr(expr, 2, 0);
```
binji: Can be moved up to the existing `2, 0` group (w/ Store)

src/wat-writer.cc
```diff
@@ -1031,9 +1074,15 @@ void WatWriter::WriteFoldedExpr(const Expr* expr) {
     case ExprType::LocalTee:
     case ExprType::MemoryGrow:
     case ExprType::Unary:
+    case ExprType::TableGrow:
```
binji: Should be in the `PushExpr(expr, 2, 1)` group.

src/wast-parser.cc
```diff
@@ -604,7 +610,7 @@ bool WastParser::ParseVarListOpt(VarVector* out_var_list) {
 Result WastParser::ParseValueType(Type* out_type) {
   WABT_TRACE(ParseValueType);
   if (!PeekMatch(TokenType::ValueType)) {
-    return ErrorExpected({"i32", "i64", "f32", "f64", "v128"});
+    return ErrorExpected({"i32", "i64", "f32", "f64", "v128", "anyref", "anyfunc"});
```
binji: anyfunc (i.e. funcref) is not allowed as a value type in this proposal (it was split out).

alexcrichton:
Ok thanks for the review! I think I've covered all those points and increased the test coverage some more as well (embarassing bugs!)

I've also gone ahead and updated all the encodings to match the latest proposal in https://github.com/WebAssembly/reference-types/issues/29 which looks to be the way forward

PR #955, opened 2018-11-13
Add support for dylink section

sbc100:
See: https://github.com/WebAssembly/tool-conventions/blob/master/DynamicLinking.md


src/binary-reader.cc
```diff
@@ -1690,6 +1708,8 @@ Result BinaryReader::ReadCustomSection(Offset section_size) {
   if (options_.read_debug_names && section_name == WABT_BINARY_SECTION_NAME) {
     CHECK_RESULT(ReadNameSection(section_size));
     did_read_names_section_ = true;
+  } else if (section_name.rfind(WABT_BINARY_SECTION_DYLINK, 0) == 0) {
```
binji: should just be `section_name == WABT_BINARY_SECTION_DYLINK`, right? The reloc is special since they all start with `reloc.`

PR #957, opened 2018-11-15
Support for SIMD instructions in wasm-objdump

tlively:
Tests shamelessly borrowed with light editing from test/interp.

### DIFF
--- a/test/dump/simd-binary.txt
+++ b/test/dump/simd-binary.txt
@@ -1,0 +1,378 @@
+;;; TOOL: run-objdump
+;;; ARGS0: --enable-simd
+
+(module
+  ;; i8x16 add
+  (func (export "i8x16_add_0") (result v128)
+    v128.const i32 0x00000001 0x00000002 0x00000003 0x00000004
+    v128.const i32 0x00000001 0x00000002 0x00000003 0x00000004
+    i8x16.add)
+
+  ;; i16x8 add
+  (func (export "i16x8_add_0") (result v128)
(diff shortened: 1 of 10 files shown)


test/dump/simd-binary.txt
```diff
@@ -0,0 +1,378 @@
...
+ 0002ed: 0b                         | end
+0002ee <i32x4_add_0>:
+ 0002f1: fd 02 01 00 00 00 02 00 00 00 03 | v128.const 0x00000001 0x00000002 0x00000003 0x00000004
+ 000303: fd 02 01 00 00 00 02 00 00 00 03 | v128.const 0x00000001 0x00000002 0x00000003 0x00000004
+ 000315: fd 79                      | i32x4.add
+ 000316: 0b                         | end
+000317 <i64x2_add_0>:
+ 00031a: fd 02 01 00 00 00 02 00 00 00 03 | v128.const 0x00000001 0x00000002 0x00000003 0x00000004
+ 00032c: fd 02 01 00 00 00 02 00 00 00 03 | v128.const 0x00000001 0x00000002 0x00000003 0x00000004
+ 00033f: fd 8a                   | i64x2.add
```
binji: Hm, what's up with the `i64x2.*` instructions all being misaligned? Looks like `f32x4` and `f64x2` are too...

tlively: I think it's actually the v128.consts that are unaligned because the immediates push them to the right.

PR #970, opened 2018-12-05
Replace string_view with string-view-lite

binji:
There are many `string_view` implementations that we could use, but
`string-view-lite` is convenient since it is provided as a single header
file library.

There are also other `*-lite` libraries that we may want to add as well:
variant, expected, etc.


binji:
I'm not sure we should do this now... closing PR.

PR #971, opened 2018-12-05
Add Abseil; use its string_view implementation

binji:


binji:
Discussing this with some other folks, it sounds like introducing an abseil dependency might put too much burden on users.

PR #973, opened 2018-12-12
Fix SYNOPSIS in manpages

rrrapha:
Before: wasm-objdump [options] filename
After: wasm-objdump [options] file ...

Also document '-h' in wasm-objdump.1


binji:
Thanks!

PR #975, opened 2018-12-12
Pass function body size in BinaryReader callback

binji:
This is useful for finding large functions with wasm-objdump.


binji:
This also changes the address printed for a function; the previous location included the body size, this doesn't. I think it makes more sense not to include the size, similar to sections.

Issue #962, opened 2018-11-29
wasm-objdump aborts if passed a directory name


sbc100:
```
$ /dev/wasm/wabt/bin/wasm-objdump -d -x /tmp
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
Aborted
```

PR #976, opened 2018-12-14
error handling: corrected error with directory

deltova:
This commit is fixing the error with directory as parameter of objdump.
The error is fixed with a stat on the file and checking the type of file
that is given. #962


feedthejim:
Thanks a lot, keep it up dude!

PR #979, opened 2018-12-16
Add missing manpages, fix markup for references, links and examples

rrrapha:


rrrapha:
BTW, it is possible to generate nicely formatted markdown or html from the manpages.

For example:
`$ mandoc -T markdown wasm-interp.1`
`$ mandoc -T html -O man=%N.%S.html wasm-interp.1`

Issue #980, opened 2018-12-17
noexcept warning, treated as error on MSVC in Appveyor


sunfishcode:
When building [wasmtime](https://github.com/CraneStation/wasmtime), which uses wabt via [wabt Rust bindings](https://crates.io/crates/wabt-sys), on MSVC in Appveyor, I see warnings like this:

```
 C:\Users\appveyor\.cargo\registry\src\github.com-1ecc6299db9ec823\wabt-sys-0.5.2\wabt\src/string-view.h(165): warning C4577: 'noexcept' used with no exception handling mode specified; termination on exception is not guaranteed. Specify /EHsc [C:\projects\wasmtime\target\debug\build\wabt-sys-df9460bd8a8b54fb\out\build\wabt.vcxproj]
```
which are itnerpreted as errors:
```
C:\Users\appveyor\.cargo\registry\src\github.com-1ecc6299db9ec823\wabt-sys-0.5.2\wabt\src/string-view.h(165): error C2220: warning treated as error - no 'object' file generated [C:\projects\wasmtime\target\debug\build\wabt-sys-df9460bd8a8b54fb\out\build\wabt.vcxproj]
```
which if I'm reading things correctly comes from the -WX (/WX) [in wabt's CMakeLists.txt](https://github.com/WebAssembly/wabt/blob/master/CMakeLists.txt#L82).

See [here](https://ci.appveyor.com/project/CraneStation/wasmtime/build/job/8ovrt46vmwl5htb5) for the full log.

Would it be possible and reasonable to remove -WX outside of developer builds?

Alternatively, MSVC is suggesting adding `/EHsc`; would it make sense to add that?

binji:
I assume this fixed it, please reopen if not.

PR #982, opened 2018-12-17
Add WERROR option to CMake build

binji:
We shouldn't compile with `-Werror` or `-WX` by default, it is useful
for wabt developers but hostile to users.

Fixes issue #980.


sbc100:
But we should enable this option in all out CI, no?

binji:
Yes, I suppose so.

PR #985, opened 2018-12-19
 The great renaming

binji:
This huge PR does all the renaming as described in issue #933. It also
updates to the latest testsuite so the new names are used.

The old names of the MVP instructions are still supported for
convenience (though we should remove those too at some point), but the
old simd and atomic instruction names are no longer supported.


binji:
There isn't, but you're right that would be a convenient switch to add. I'll look into that today.

binji:
I think I'll handle that flag in another PR, to keep this change as being primarily renaming.

cggallant:
Are there plans to push these changes to the wat2wasm demo? https://cdn.rawgit.com/WebAssembly/wabt/aae5a4b7/demo/wat2wasm/

When I try using local.get rather than get_local, I receive the following error: 
`error: unexpected token local.get, expected get_local.`

binji:
@cggallant that is an old link, back when we used rawgit (which is being turned down). The new link is https://webassembly.github.io/wabt/demo/wat2wasm/, and supports the new and old names.

PR #1000, opened 2019-01-29
Rename {memory,table}.drop to {data,elem}.drop

alexcrichton:
Carrying over renames from WebAssembly/bulk-memory-operations#46


binji:
Thanks! I think you need to update the prebuilt lexer to fix these errors.

alexcrichton:
Oops, I think it should be updated now

PR #1014, opened 2019-02-12
Parse updated event text and binary format

binji:
An event has a type-section index, like a function definition. The
current proposal doesn't specify the text format, so I assumed that it
would match the format of the other sections that reference function
types. This means that the following declaration styles are allowed:

```
(type $t (func (param i32)))
(event $e1 (type $t))
(event $e2 (param f32))
```


rossberg:
Looks good to me, except that this probably is a good time to change the name "event". I don't know where that came from originally, but it doesn't have the right connotation and is potentially confusing. Why not simply "exception"?

binji:
@aheejin do you know where the name "event" came from?

src/wat-writer.cc
```diff
@@ -1382,17 +1381,18 @@ void WatWriter::WriteGlobal(const Global& global) {
...
 }
 
-void WatWriter::WriteBeginEvent(const Event& event) {
+void WatWriter::WriteEvent(const Event& event) {
   WriteOpenSpace("event");
   WriteNameOrIndex(event.name, event_index_, NextChar::Space);
   WriteInlineExports(ExternalKind::Event, event_index_);
   WriteInlineImport(ExternalKind::Event, event_index_);
-  WriteTypes(event.sig, nullptr);
+  if (event.decl.has_func_type) {
```
sbc100: Don't all events have a type?

binji: yes, this flag specifies whether the type is specified like `(event (type $t))`

binji:
Not sure about the `event` name, but we can change it back to exception later. Going to land this for now.

aheejin:
The term was first introduced by Karl. In the proposal:
> However, rather than limiting this new section to just defining exceptions, it defines a more general format that allows the declaration of other forms of events.

And [this](https://github.com/WebAssembly/exception-handling/blob/master/proposals/old/Level-1%2BN.md) was various bikeshedding on what kind of events we can support; no one is actively working on these though. And I heard @rossberg mentioned someone is also working at a prototype of effect handler..? Anyway I guess Karl's point was not to create a separate section each time we support a new kind of events.

sbc100:
I suggest we call it exceptions for now until we have a reason to call it something else.  If it turns out we can re-purpose it for other things we can rename it at that point, no?

aheejin:
I am not against to call it exceptions, but I'm not very excited to rename all of them back... 😨

PR #1016, opened 2019-02-13
 Add br_on_exn instruction

binji:
It takes two u32 immediates: the branch depth and an exception index. The
stack signature is `[expect_ref] -> [except_ref]`, so the `except_ref`
can be tested easily against multiple exception types.


aheejin:
It's stack signature is `[br_on_exn]` -> list of types of extracted exception values.
So for example in C++ the exception value is an `i32` because it is a pointer, then it is gonna be `[br_on_exn]` -> `[i32]`. But the spec also allows multiple values to be popped depending on its exception signature.

binji:
Right, it forwards those values to the label (like `br`) but if the exception doesn't match it propagates the `except_ref`.

aheejin:
Oh I think if it doesn't match it just consumes the `except_ref` value, no? Other branches don't really _take_ values in wasm stack as operands so they forward them according to the block signature, but `br_on_exn` _takes_ an `except_ref` as an operand.

binji:
I'm basing it on the wording [here](https://github.com/WebAssembly/exception-handling/blob/master/proposals/Exceptions.md#exception-data-extraction):

> The br_on_exn instruction checks the exception tag of an except_ref on top of the stack if it matches the given exception index. If it does, it branches out to the label referenced by the instruction (In the binary form, the label will be converted to a relative depth immediate, like other branch instructions), and while doing that, pops the except_ref value from the stack and instead pushes the exception's argument values on top of the stack. In order to use these popped values, the block signature of the branch target has to match the exception types - because it receives the exception arguments as branch operands. If the exception tag does not match, the except_ref value remains on the stack.

with the example (where I've added `<1>` and `<2>` -- I've also provided the exception type for `br_on_exn`, which was missing in the example):

```wasm
block $l (result i32 i64)
  ...
  ;; except_ref $e is on the stack at this point
  br_on_exn $l $e ;; branch to $l with $e's arguments
  <1>
  ...
end
<2>
```

So if the `except_ref` doesn't match `$e`, then it is still on the stack at `<1>`. But if it does match `$e`, then the stack contains `... i32 i64` at `<2>`.

aheejin:
Oh right I was remembering it wrong.. Thank you and sorry for the inconvenience :(

aheejin:
And thank you for implementing this! I thought of doing this but didn't quite find time...

binji:
No problem! I appreciate you double-checking this to make sure it's right :-)

binji:
Looks like the Travis failure is a CI flake, timeout querying the ubuntu keyserver. If passed on the other travis run, so it seems ok.

PR #1019, opened 2019-02-19
Correct `last_section_payload_offset_` calculation.

SanderSpies:
Move `last_section_payload_offset_ = stream_->offset();` to correct location. Otherwise the relocation offsets are incorrect.


SanderSpies:
Closing for now - need to investigate a bit further.

PR #1022, opened 2019-02-20
[wasm-objdump] Implemented R_WASM_EVENT_INDEX_LEB reloc type.

aardappel:
Also fixed 2 asserts that could trigger on user data.

### DIFF
--- a/src/binary-reader-objdump.cc
+++ b/src/binary-reader-objdump.cc
@@ -1470,7 +1470,6 @@ Result BinaryReaderObjdump::OnFunctionSymbol(Index index,
   if (sym_name.empty()) {
     sym_name = GetFunctionName(func_index);
   }
-  assert(!sym_name.empty());
   PrintDetails("   - [%d] F <" PRIstringview "> func=%" PRIindex, index,
                WABT_PRINTF_STRING_VIEW_ARG(sym_name), func_index);
   PrintSymbolFlags(flags);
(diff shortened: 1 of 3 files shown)


src/binary-reader-objdump.cc
```diff
@@ -1470,7 +1470,6 @@ Result BinaryReaderObjdump::OnFunctionSymbol(Index index,
   if (sym_name.empty()) {
     sym_name = GetFunctionName(func_index);
   }
-  assert(!sym_name.empty());
```
sbc100: I'm not sure about removing these.   It should not be possible to have a symbol table entry that doesn't have a name.  Perhaps we can push this change up into the parser?   Are you hitting these?

sbc100:
FYI, you don't need to fork wabt to push PRs, you can just push to the main repo, since you are member.

src/binary-reader-objdump.cc
binji: I think these seem OK -- can we guarantee that these never have a name?

sbc100: For each symbol table entry you either need to have an explicit name.. or it takes the name of the import it points too.   So yes, by that time we get here we should have a name.  Invalid symbol table enties should be handled in some way at the binary reader level no?  Presumbably by printing and error and skipping that entry?

aardappel: If you parse invalid modules, this can happen. The only alternative is for the caller of this function to detect this case and not call this function at all.. and presumably to abort. That may actually be nice since the caller should be in charge of parsing the binary, not this function.

But not sure if I want to do that in this change, just contributing to invalid binaries not asserting, as discussed with @binji

binji: The binary reader interface is low-level enough that it doesn't handle much validation. Looking up the name there seems like more work than it should do, so I'd say it's the responsibility of this code to be robust in that case.

PR #1026, opened 2019-02-25
Rename anyfunc -> funcref; parse reference types

binji:
Also:

* Add feature limits on using v128 and anyref types (requires
  --enable-simd and --enable-reference-types respectively).

* Separate out ParseValueType (used for params, locals, global types)
  from ParseRefType (used for table types).


sbc100:
If possible I think it would be good to do renames as separate PRs to any functional stuff.  I'm not sure it would have been possible in this case, but I would make it easier to read for people following along.

binji:
Sorry about that. Yeah, this case was a bit annoying to rename separately but I probably should have done it anyway.

PR #1027, opened 2019-02-25
[wabt-objdump] Refactored some functions to use string_view

aardappel:
The functions BinaryReaderObjdumpBase::GetSymbolName calls may
return null, so it should deal with that case.


aardappel:
Ok, `string_view` it is :)

binji:
Looks good, thanks. The `!((name = ...).empty())` stuff is kinda weird, but I don't have better suggestions without more refactoring :)

PR #1029, opened 2019-03-04
wasm-objdump: fix f64 init expressions

ohorn:
The `PrintInitExpr` function falsely treated f64 as a float instead of a double.

The test case `global.txt` also contained the wrong output, namely `0x0p+0` instead of `0x1p+2` for `(f64.const 4)`.


binji:
Thanks!

PR #1031, opened 2019-03-04
Make SIMD lane-wise relational operator implementations conform to the spec

gnzlbg:
The implementation of the SIMD lane-wise relational operators was incorrect. These operators were setting true lanes to `1`, but according to the spec they should _set all bits_ of true lanes. 

These operators create a vector mask that can be used in combination with bitselect to apply certain operations only to some vector lanes. However, because only 1 bit of each true lane was being set by these operators under wabt, the combination with `bitselect` was only selecting that one bit instead of whole lanes, leading to incorrect results.


binji:
lgtm, thanks! Have you joined the [WebAssembly community group](https://www.w3.org/community/webassembly/)?

gnzlbg:
@binji i've just done that

PR #1032, opened 2019-03-05
Update binary encoding of SIMD vector shuffle to conform to the latest draft

gnzlbg:
The textual (WAT) representation of the vector shuffle immediate argument was
underspecified. Recently, WebAssembly/simd#67 specified
it to be `i8x16.shuffle i5 i5 ... i5` (16 times `i5`), which is what all other
implementations were doing. WABT was reading the shuffle argument as 4 32-bit
hex integers that specified a `v128`, that is, `i8x16.shuffle 0x0 0x0 0x0 0x0`.

This PR updates WABT to conform to the modified draft of the SIMD extension.

Since there are more shuffle and permtue intrinsics in the WASM SIMD extension
pipeline, this PR adds a generic `iNxM` class implementing a packed array of `M`
`N`-bit wide integers.

The class can store the integer array in both textual and binary
representations. The textual repr is useful for WAT and type-checking, while the
binary repr is useful for the binary encoding.


gnzlbg:
So a couple of compilers choke with C++14. GCC 4.8 is 6 years old, and MSVC 2015 is not a C++11 compatible compiler either.. 

I'll remove the C++14 usage, and see if I still need to cut-off on some C++11 to make MSVC work.

src/c-writer.cc
```diff
@@ -2246,9 +2246,12 @@ void CWriter::Write(const SimdLaneOpExpr& expr) {
 void CWriter::Write(const SimdShuffleOpExpr& expr) {
   Type result_type = expr.opcode.GetResultType();
   Write(StackVar(1, result_type), " = ", expr.opcode.GetName(), "(",
-        StackVar(1), " ", StackVar(0), ", lane Imm: $0x%08x %08x %08x %08x",
-        expr.val.v[0], expr.val.v[1], expr.val.v[2], expr.val.v[3], ")",
-        Newline());
+        StackVar(1), " ", StackVar(0), ", lane Imm: %llu %llu %llu %llu %llu %llu %llu %llu %llu %llu %llu %llu %llu %llu %llu %llu",
```
gnzlbg: I'm not sure why "lane" was used here before, but I've left it as is. 

It might be worth it to change this to `i5x16 Imm:` instead or similar.

src/iNxM.h
```diff
@@ -0,0 +1,242 @@
...
+  /// Array length
+  static constexpr ptrdiff_t size() {
+    return Length;
+  }
+
+  /// In the textual encoding we treat the integers as Nat.
+  using TextValue = std::array<text_t, Length>;
+
+  static_assert(IntBits * Length % (sizeof(bin_t) * CHAR_BIT) == 0,
+                "The array length in bits must be a multiple of CHAR_BIT");
```
gnzlbg: This error message should be updated. For now the array has to precisely fit its layout, but we can choose `bin_t` programatically in the future if it doesn't fit.

src/interp/interp-disassemble.cc
```diff
@@ -518,12 +518,28 @@ void Environment::Disassemble(Stream* stream,
...
+          binary_value[l] = ReadU8(&pc);
+        }
+        i5x16 value{binary_value};
+        value.binary_to_text();
+
         stream->Writef(
-            "%s %%[-2], %%[-1] : (Lane imm: $0x%08x 0x%08x 0x%08x 0x%08x )\n",
-            opcode.GetName(), ReadU32(&pc), ReadU32(&pc), ReadU32(&pc),
-            ReadU32(&pc));
+          "%s %%[-2], %%[-1] : (Lane imm: \
```
gnzlbg: Same here, `Lane imm` should probably be `i5x16 imm`.

src/interp/interp-disassemble.cc
```diff
@@ -518,12 +518,28 @@ void Environment::Disassemble(Stream* stream,
...
+%llu %llu %llu %llu \
+%llu %llu %llu %llu \
+%llu %llu %llu %llu \
+%llu %llu %llu %llu)\n",
```
gnzlbg: This can be chunked in a loop.

src/interp/interp-trace.cc
```diff
@@ -575,16 +575,27 @@ void Thread::Trace(Stream* stream) {
...
+    case Opcode::V8X16Shuffle: {
       stream->Writef(
-          "%s $0x%08x %08x %08x %08x $0x%08x %08x %08x %08x : with lane imm: "
-          "$0x%08x %08x %08x %08x\n",
-          opcode.GetName(), Pick(2).v128_bits.v[0], Pick(2).v128_bits.v[1],
-          Pick(2).v128_bits.v[2], Pick(2).v128_bits.v[3],
-          Pick(1).v128_bits.v[0], Pick(1).v128_bits.v[1],
-          Pick(1).v128_bits.v[2], Pick(1).v128_bits.v[3], ReadU32At(pc),
-          ReadU32At(pc + 4), ReadU32At(pc + 8), ReadU32At(pc + 12));
+        "%s $0x%08x %08x %08x %08x $0x%08x %08x %08x %08x : with lane imm:",
```
gnzlbg: same here, with `lane` imm` should be `i5x16`

test/dump/simd-lane.txt
```diff
@@ -171,7 +171,7 @@ Code Disassembly:
  00035e: ff ff ff 00 ff 7f ff 00 ff | 
  000367: fd 02 55 00 55 00 55 00 55 | v128.const 0x00550055 0x00550055 0x00550055 0x00550155
  000370: 00 55 00 55 00 55 01 55 00 | 
- 000379: fd 03 10 01 12 03 14 05 16 | v8x16.shuffle 0x03120110 0x07160514 0x0b1a0918 0x0f1e0d1c
- 000382: 07 18 09 1a 0b 1c 0d 1e 0f | 
- 00038b: 0b                         | end
+ 000379: fd 03 30 c8 41 8b 3d 38 e9 | v8x16.shuffle 16 1 18 3 20 5 22 7 24 9 26 11 28 13 30 15
+ 000382: c5 9b 7f                   |
```
gnzlbg: I've verified these by hand.

test/parse/expr/bad-simd-shuffle-nat-expected.txt
```diff
@@ -9,7 +9,7 @@
     ))
 
 (;; STDERR ;;;
-out/test/parse/expr/bad-simd-shuffle-nat-expected.txt:8:19: error: unexpected token "$0x01020304", expected an Nat literal (e.g. 123).
+out/test/parse/expr/bad-simd-shuffle-nat-expected.txt:8:19: error: unexpected token "$0x01020304", expected a natural number (e.g. 123).
```
gnzlbg: This error will be a bit confusing, because while `123` is a natural number, it will be out-of-bounds for the `shuffle` instruction.

gnzlbg:
So I've backported the PR to C++11, but apparently GCC and MSVC are not C++11 compliant because they fail to normalize type aliases in function signatures, but not anywhere else for some reason. So I've removed all uses of type aliases in function signatures increasing their verbosity by 100000%, but now at least CI is green ¯\_(ツ)_/¯

binji:
Thanks for working on this! At a first glance, this PR seems to be changing a lot of code. Creating a generic `i5x16` struct doesn't seem necessary, since this type is only used for the shuffle immediate. I'd also prefer not to have it store both binary and text representations in the same class.

We can perhaps work toward something more comprehensive in the future, but for an initial PR, I'd like to keep this smaller. I think it would be a better start if we only fix the text and binary parsing, and continue to use `v128` as the immediate.

gnzlbg:
> and continue to use v128 as the immediate.

`v128` is 16 bytes wide, but the immediate is only 10 bytes wide, how can that work?

binji:
It's not necessary that the in-memory representation of the immediate matches the binary format representation. Since v128 is bigger, we can use one byte per lane idx, like the currently spec'd binary format. It's probably nicer to use that way anyway (aside from the fact that v128 doesn't have convenient accessors, though that would be relatively easy to add)

gnzlbg:
I think we maybe are talking past each other. Are you on IRC/Discord/Zulip/... were we could maybe chat?

binji:
I'm regularly on irc://irc.w3.org:6667/#webassembly

tlively:
> `v128` is 16 bytes wide, but the immediate is only 10 bytes wide, how can that work?

I think this was cleared up in https://github.com/WebAssembly/simd/pull/67, but repeating here for completeness: the current spec proposal specifies 16 bytes for the immediate (each of which can only have value 0-31), not 10 bytes.

gnzlbg:
@binji I've submitted #1034 , I think something in that spirit is what you were proposing doing as a first step. 

We can revisit this approach if the binary encoding gets changed in the spec to something more compact.

PR #1034, opened 2019-03-06
Update textual encoding of SIMD vector shuffle to conform to the latest SIMD draft

gnzlbg:
The textual (WAT) representation of the vector shuffle immediate argument was
underspecified. Recently, WebAssembly/simd#67 specified
it to be i8x16.shuffle i5 i5 ... i5 (16 times i5), which is what all other
implementations were doing. WABT was reading the shuffle argument as 4 32-bit
hex integers that specified a v128, that is, i8x16.shuffle 0x0 0x0 0x0 0x0.

This PR updates WABT to conform to the updated draft of the SIMD extension.

### DIFF
--- a/src/wast-parser.cc
+++ b/src/wast-parser.cc
@@ -1688,11 +1688,12 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
       ErrorUnlessOpcodeEnabled(token);
       v128 value;
       char* value_ptr = reinterpret_cast<char*>(&value);
-      for (int32_t lane = 0; lane < 16; ++lane) {
+      for (int lane = 0; lane < 16; ++lane) {
         Location loc = GetLocation();
 
-        if (!PeekMatch(TokenType::Nat))
-          return ErrorExpected({"an natural number in range [0, 32)"});
+        if (!PeekMatch(TokenType::Nat)) {
+          return ErrorExpected({"a natural number in range [0, 32)"});
+        }
 
         Literal literal = Consume().literal();
 
--- a/src/wast-parser.cc
+++ b/src/wast-parser.cc
@@ -1686,10 +1686,39 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
     case TokenType::SimdShuffleOp: {
       Token token = Consume();
       ErrorUnlessOpcodeEnabled(token);
-      Const const_;
-      CHECK_RESULT((ParseSimdConst(&const_, Type::I32, sizeof(v128))));
+      v128 value;
+      char* value_ptr = reinterpret_cast<char*>(&value);
+      for (int32_t lane = 0; lane < 16; ++lane) {
+        Location loc = GetLocation();
+
+        if (!PeekMatch(TokenType::Nat))
+          return ErrorExpected({"an natural number in range [0, 32)"});
+
+        Literal literal = Consume().literal();
+
+        string_view sv = literal.text;
+        const char* s = sv.begin();
+        const char* end = sv.end();
+        Result result;
+
+        uint32_t value = 0;
+
+        result = ParseInt32(s, end, &value, ParseIntType::UnsignedOnly);
+
+        if (Failed(result)) {
+          Error(loc, "invalid literal \"%s\"", literal.text.c_str());
+          return Result::Error;
+        }
+
+        if (value > 31) {
+          Error(loc, "shuffle index \"%s\" out-of-range [0, 32)", literal.text.c_str());
+          return Result::Error;
+        }
+
+        *(value_ptr + lane) = static_cast<uint8_t>(value);
+      }
       out_expr->reset(
-          new SimdShuffleOpExpr(token.opcode(), const_.v128_bits, loc));
+          new SimdShuffleOpExpr(token.opcode(), value, loc));
       break;
     }
 
--- a/src/wast-parser.cc
+++ b/src/wast-parser.cc
@@ -1686,8 +1686,7 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
     case TokenType::SimdShuffleOp: {
       Token token = Consume();
       ErrorUnlessOpcodeEnabled(token);
-      v128 value;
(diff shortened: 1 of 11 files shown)


src/wast-parser.cc
```diff
@@ -1686,10 +1686,39 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
     case TokenType::SimdShuffleOp: {
       Token token = Consume();
       ErrorUnlessOpcodeEnabled(token);
-      Const const_;
-      CHECK_RESULT((ParseSimdConst(&const_, Type::I32, sizeof(v128))));
+      v128 value;
+      char* value_ptr = reinterpret_cast<char*>(&value);
```
binji: Use the `Bitcast` function, and a `uint8_t[16]` destination instead.

src/wast-parser.cc
```diff
@@ -1686,10 +1686,39 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
...
+      for (int32_t lane = 0; lane < 16; ++lane) {
+        Location loc = GetLocation();
+
+        if (!PeekMatch(TokenType::Nat))
```
binji: nit: braces around if bodies please

src/wast-parser.cc
```diff
@@ -1686,10 +1686,39 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
...
+          return ErrorExpected({"an natural number in range [0, 32)"});
```
binji: nit: s/an/a/

src/wat-writer.cc
```diff
@@ -943,8 +943,10 @@ Result WatWriter::ExprVisitorDelegate::OnSimdLaneOpExpr(SimdLaneOpExpr* expr) {
 Result WatWriter::ExprVisitorDelegate::OnSimdShuffleOpExpr(
     SimdShuffleOpExpr* expr) {
   writer_->WritePutsSpace(expr->opcode.GetName());
-  writer_->Writef(" 0x%08x 0x%08x 0x%08x 0x%08x", (expr->val.v[0]), (expr->val.v[1]),
-                  (expr->val.v[2]), (expr->val.v[3]));
+  uint8_t* value_ptr = reinterpret_cast<uint8_t*>(&expr->val);
```
binji: Same here, please use `Bitcast`

src/wast-parser.cc
```diff
@@ -1686,10 +1686,39 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
     case TokenType::SimdShuffleOp: {
       Token token = Consume();
       ErrorUnlessOpcodeEnabled(token);
-      Const const_;
-      CHECK_RESULT((ParseSimdConst(&const_, Type::I32, sizeof(v128))));
+      v128 value;
+      char* value_ptr = reinterpret_cast<char*>(&value);
+      for (int32_t lane = 0; lane < 16; ++lane) {
```
binji: nit: it's ok to use `int` for bounded loops

src/wast-parser.cc
```diff
@@ -1686,10 +1686,39 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
...
+
+        result = ParseInt32(s, end, &value, ParseIntType::UnsignedOnly);
+
+        if (Failed(result)) {
+          Error(loc, "invalid literal \"%s\"", literal.text.c_str());
+          return Result::Error;
+        }
+
+        if (value > 31) {
+          Error(loc, "shuffle index \"%s\" out-of-range [0, 32)", literal.text.c_str());
```
xtuc:
```suggestion
          Error(loc, "shuffle index \"%s\" out-of-range [0, 32]", literal.text.c_str());
```

src/wast-parser.cc
gnzlbg: `Bitcast<uint8_t[16]>` is not supported.

src/wat-writer.cc
gnzlbg: Just replacing `reinterpret_cast` with `Bitcast`?

src/wast-parser.cc
gnzlbg: So I've used a `uint8_t[16]` to store the values, and `Bitcast<v128>` to bitcast to it from the `v128`, but see below, the other way around does not appear to be supported.

src/wat-writer.cc
gnzlbg: While `uint8_t[16] = Bitcast<v128>` works, the other way around does not appear to be supported.

gnzlbg: This is probably because `Bitcast` does not handle C arrays properly (it takes a `&&` but it has no overload for `T (&)[N]`.

gnzlbg: I've changed this to use a `std::array<uint8_t, 16>` instead, which does work with `Bitcast`.

src/wast-parser.cc
binji: Oh right, probably because of pointer decay.

gnzlbg: So.. kind of. Template argument deduction will deduce their type correctly as arrays, but as you mention, they decay into pointers, so overload resolution picks overloads taking pointers as well. 

What truly generic code does is add a couple of specific overloads to match array types before pointer decay: 

* range-v3 https://github.com/ericniebler/range-v3/blob/master/include/range/v3/begin_end.hpp#L305
* libc++ https://github.com/llvm-mirror/libcxx/blob/master/include/iterator#L1667

But treating C arrays properly kind of requires unit-testing all APIs against them, and unless one is writing a C++ standard library, this is probably not worth doing.

gnzlbg:
(let me know before merging so that I can rebase this)

binji:
still lgtm, why do you need to rebase? Looks like it will merge cleanly.

gnzlbg:
I thought the commit history would be clearer if I squashed the commits, but if the history is good for you then it is also good for me :)

binji:
Oh, I always just hit the button to squash and merge :)

gnzlbg:
:D

PR #1037, opened 2019-03-07
Make v128.const i32x4 op conform to the SIMD draft

gnzlbg:
Updates the `v128.const` WAT parsing to conform to the SIMD draft. 

* Update `v128.const i32` is now `v128.const i32x4`
* Add support for `v12.const {i8x16, i16x8, i64x2, f32x4, f64x2}`


gnzlbg:
cc @binji so i've opened a new PR for this one, since it made sense to directly add support for the whole syntax. 

I've tried to keep it as minimal as possible. That is, we are able to parse the `v128.const` syntax properly from a WAT file, and when deserializing to WAT we always use `v128.const i32x4`.

src/wast-parser.cc
```diff
@@ -1732,50 +1732,139 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
...
-
-  // Parse the Simd Consts according to input data type.
-  switch (in_type) {
-    case Type::I32: {
-      const_->loc = GetLocation();
-      int Count = nSimdConstBytes / sizeof(uint32_t);
-      // Meet expected "i32" token. start parse 4 i32 consts
-      for (int i = 0; i < Count; i++) {
-        Location loc = GetLocation();
+  Result WastParser::ParseSimdV128Const(Const* const_, TokenType token_type) {
```
binji: nit: indentation looks off here

src/wast-parser.cc
```diff
@@ -1732,50 +1732,139 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
...
+    }
+    default: break;
+  }
+
+  auto text = Consume().text();
+  // Look up the number of lanes and whether the v128 is an integer or
+  // floating-point vector:
+  uint8_t no_lanes = 0;
+  bool integer = true;
+  if (text == "i8x16") {
```
binji: Please add these as new tokens in the lexer rather raw text

src/wast-parser.cc
```diff
@@ -1732,50 +1732,139 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
...
+      );
+      return Result::Error;
+    }
+    default: break;
+  }
+
+  auto text = Consume().text();
+  // Look up the number of lanes and whether the v128 is an integer or
+  // floating-point vector:
+  uint8_t no_lanes = 0;
```
binji: nit: `no_lanes` reads like "no" instead of "number" to me, maybe just `lanes` or `lane_count`

src/wast-parser.cc
```diff
@@ -1732,50 +1732,139 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
...
+
+    // For each type, parse the next literal, bound check it, and write it to
+    // the array of bytes:
+    if (integer) {
+      switch(no_lanes) {
+        case 16: {
+          uint8_t value[4] = {0, 0, 0, 0};
+          result = ParseInt32(s, end, Bitcast<uint32_t*>(&value), ParseIntType::SignedAndUnsigned);
+          // If the literal fits in an i8, then all upper bytes of the parsed
+          // Int32 are zero (independently of how whether it is signed or unsigned):
```
binji: This doesn't seem right to me; shouldn't `-1` be allowed and converted to `0xff`?

src/wast-parser.cc
```diff
@@ -1732,50 +1732,139 @@ Result WastParser::ParsePlainInstr(std::unique_ptr<Expr>* out_expr) {
...
+    char* lane_ptr = &v128_bytes[lane_size * i];
+    Result result;
+
+    // For each type, parse the next literal, bound check it, and write it to
+    // the array of bytes:
+    if (integer) {
+      switch(no_lanes) {
+        case 16: {
+          uint8_t value[4] = {0, 0, 0, 0};
+          result = ParseInt32(s, end, Bitcast<uint32_t*>(&value), ParseIntType::SignedAndUnsigned);
```
binji: I think it might be easier to parse this as a `uint32_t`, then convert to `int32_t` and see if it is in the range `[-128,127]`.

test/parse/expr/bad-simd-shuffle-lane-index-overflow2.txt
```diff
@@ -9,6 +9,12 @@
     ))
 
 (;; STDERR ;;;
+out/test/parse/expr/bad-simd-shuffle-lane-index-overflow2.txt:6:16: error: Unexpected type at start of simd constant. Expected one of: i8x16, i16x8, i32x4, i64x2, f32x4, f64x2.
```
binji: update these to use `i32x4`?

test/parse/expr/bad-simd-shuffle-not-enough-indices.txt
```diff
@@ -9,6 +9,12 @@
     ))
 
 (;; STDERR ;;;
+out/test/parse/expr/bad-simd-shuffle-not-enough-indices.txt:6:16: error: Unexpected type at start of simd constant. Expected one of: i8x16, i16x8, i32x4, i64x2, f32x4, f64x2.
```
binji: these too

src/wast-parser.cc
gnzlbg: Uh, you are right. I think I just write these to an `int32_t`, check the bounds, and cast it to a `uint_8t`. How does this approach sound?

src/wast-parser.cc
gnzlbg: Yeah, this sounds good to me :D

gnzlbg:
@binji sorry it took me so long to pick these up, been busy with other stuff, will try to get these done today.

src/wast-parser.cc
gnzlbg: Indeed. Would it be possible to add a `.clang-format` file or similar to the repo (and use a particular clang-format version on CI?). Every line I change in this repo requires me to manually insert spaces to indent things because my IDE is set up for the styles of other projects I work on, and that makes it painful to contribute here. Not other projects I work on follow the same style that my ide is set up in, but all of them use clang format, and my IDE is also set up to run clang-format on save, which is also something I have to disable / re-enable every time I switch to working on this project.

src/wast-parser.cc
gnzlbg: So I think I've fixed this. Two changes were needed: 

* parsing these as both Int and Nat (since Nat does not support negative numbers)
* checking the range [signed::MIN, unsigned::MAX], otherwise 255 fails. 

I've added tests for the corner cases, except for i64x2 where the test suite does not appear to support the largest 64-bit values yet.

gnzlbg:
So I think I've fixed all issues except for adding the new tokens to the lexer.

gnzlbg:
So I've added the tokens as bare tokens to the lexer - i'm not sure if that's the right kind of token to use for these though.

For some reason, doing `make update-re2c` modified the `third-party/testsuite` submodule, and everything started to fail, which confused me quite a bit. After reverting those changes everything just worked =/

src/wast-parser.cc
gnzlbg: Ok I decided to look and for some reason the repo does have a `.clang-format` file, but running clang-format generates a huge diff :/

gnzlbg: There does not seem to be a `make format` command that does the right thing, but AFAIK my IDE just invokes clang-format on the file, and clang-format recursively search parent directories for its configuration file, so I suppose the right file is being picked. I suspect that it just hasn't been invoked on the whole repo for a long time because it does not seem to be an easy way to do so here (no `make format` or `make format-check`). It might also be that the `clang-format` binary that my IDE picks is different than the one the repo is formatted with, but the diffs are so big that I doubt that. Usually `make format` is in charge of picking the right clang-format version (e.g. `clang-format-5.0` or `clang-format-8.0`), I have a couple of them installed, but I suspect that `clang-format` is the one that my IDE picks (which should be `clang-format-8.0`).

tlively: Have you tried using `git clang-format`? That will only reformat lines that were touched in the commits you specify.

gnzlbg: I've used it in other projects, but i haven't tried to use it here. If that's the preferred way a note in the readme / contributing docs would be helpful. I think that just picking a particular clang-format version, and enforcing it on CI, would be easier though.

binji:
> So I've added the tokens as bare tokens to the lexer - i'm not sure if that's the right kind of token to use for these though.

I think that makes sense.

> For some reason, doing make update-re2c modified the third-party/testsuite submodule, and everything started to fail, which confused me quite a bit. After reverting those changes everything just worked =/

Oops, sorry about that! I really should clean up the lexer code...

Issue #1038, opened 2019-03-08
Binary reader errors on local count > 0


remko:
Due to the way my WASM code is generated, I have code bodies that explicitly list 0 locals of a certain type. It seems the binary reader (intentionally) fails on this:

    error: local count must be > 0

Is there a reason why this restriction is there? I couldn't immediately find something in the spec that constrains this, and tools like `wasm-dis` don't seem to mind (and neither do the engines).

binji:
You're right, this was a bug fix in wabt, but it incorrectly restricted the format. Thanks for catching! Working on a fix now.

PR #1039, opened 2019-03-08
Allow a local type count of 0

binji:
It's legal in the wasm format to have a local type count of 0. I
originally made it illegal to fix a wabt bug, and didn't notice because
there were no spec tests for this behavior.


binji:
Fixes #1038.

PR #1041, opened 2019-03-14
Fix bugs in integer parsing

binji:
* `ParseHexdigit` allowed `g` as a hex character

* Overflow checking can't just check `old_value > new_value`, that fails
  for many cases.


binji:
> But base-17 :(

Turns out it didn't work anyway, try running this in the [wat2wasm demo](https://webassembly.github.io/wabt/demo/wat2wasm/):
```
(func (export "foo") (result i32) i32.const 0xgabba)
```

Issue #1045, opened 2019-03-15
test failures on POWER8 (ppc64le)


rathann:
When building 1.0.10 on Fedora rawhide (development) with gcc-9.0.1 and RHEL/CentOS 7 with gcc-4.8.5, the following tests fail on POWER8 (ppc64le) architecture:
```
- test/spec/conversions.txt
  expected error code 0, got 2.
  STDOUT MISMATCH:
  --- expected
  +++ actual
  @@ -1 +1,3 @@
  -401/401 tests passed.
  +out/test/spec/conversions.wast:348: expected result to be nan, got f64:nan
  +out/test/spec/conversions.wast:350: expected result to be nan, got f64:-nan
  +399/401 tests passed.
- test/spec/conversions.txt
  expected error code 0, got 2.
  STDOUT MISMATCH:
  --- expected
  +++ actual
  @@ -1 +1,3 @@
  -401/401 tests passed.
  +out/test/spec/conversions.wast:348: expected result to be nan, got f64:nan
  +out/test/spec/conversions.wast:350: expected result to be nan, got f64:-nan
  +399/401 tests passed.
- test/wasm2c/spec/conversions.txt
  expected error code 0, got 1.
  STDERR MISMATCH:
  --- expected
  +++ actual
  @@ -0,0 +1,15 @@
  +Traceback (most recent call last):
  +  File "/builddir/build/BUILD/wabt-1.0.10/test/run-spec-wasm2c.py", line 407, in <module>
  +    sys.exit(main(sys.argv[1:]))
  +  File "/builddir/build/BUILD/wabt-1.0.10/test/run-spec-wasm2c.py", line 400, in main
  +    forward_stdout=True).RunWithArgs()
  +  File "/builddir/build/BUILD/wabt-1.0.10/test/utils.py", line 104, in RunWithArgs
  +    raise error
  +utils.Error: Error running "out/test/wasm2c/spec/conversions/conversions":
  +conversions-main.c:1118: assertion failed: in Z_conversions_0Z_f64Z2Epromote_f32Z_df(make_nan_f32(0x200000)): expected result to be a arithmetic nan, got 0x0000000000000000.
  +conversions-main.c:1124: assertion failed: in Z_conversions_0Z_f64Z2Epromote_f32Z_df(-make_nan_f32(0x200000)): expected result to be a arithmetic nan, got 0x0000000000000000.
  +During handling of the above exception, another exception occurred:
  +Traceback (most recent call last):
[...]

Issue #1043, opened 2019-03-15
test/interp/simd-unary.txt fails on ARM and POWER8


rathann:
Running the testsuite from 1.0.10 on ARM (armv7hl and aarch64) and POWER8 (ppc64le) fails:
```
- test/interp/simd-unary.txt
  STDOUT MISMATCH:
  --- expected
  +++ actual
  @@ -25,8 +25,8 @@
   f32x4_abs_0() => v128:0x00000000 0x7fc00000 0x449a5000 0x3f800000
   f64x2_abs_0() => v128:0x00000000 0x00000000 0x00000000 0x7ff80000
   f64x2_abs_1() => v128:0x00000000 0x40934a00 0x00000000 0x3ff00000
  -f32x4_sqrt_0() => v128:0xffc00000 0xffc00000 0x40000000 0x40400000
  -f64x2_sqrt_0() => v128:0x00000000 0xfff80000 0x00000000 0xfff80000
  +f32x4_sqrt_0() => v128:0x7fc00000 0xffc00000 0x40000000 0x40400000
  +f64x2_sqrt_0() => v128:0x00000000 0x7ff80000 0x00000000 0xfff80000
   f64x2_sqrt_1() => v128:0x00000000 0x40000000 0x00000000 0x40080000
   f32x4_convert_i32x4_s_0() => v128:0x3f800000 0xbf800000 0x00000000 0x40400000
   f32x4_convert_i32x4_u_0() => v128:0x3f800000 0x40000000 0x00000000 0x40400000
...
**** FAILED ******************************************************************
- test/interp/simd-unary.txt
    /builddir/build/BUILD/wabt-1.0.10/bin/wasm-interp out/test/interp/simd-unary/simd-unary.wasm --run-all-exports --enable-simd
```

This is the only failing test on ARM.

There are more test failures on POWER8 (ppc64le) and other architectures, but I'll open separate issues for that.

binji:
I see, looks like `sqrt(-1)` is +nan on these architectures.

rathann:
I can confirm #1046 fixes this and several others.

binji:
Thanks!

PR #1046, opened 2019-03-18
Canonicalize NaNs in the interpreter

binji:
Even though we store f32 and f64 values as their representation, we
still use the underlying system's implementations for float operations.
These have non-deterministic behavior w.r.t. NaNs.

This change canoncalizes all NaNs where it is allowed. This excludes
`fxx.abs`, `fxx.neg`, `fxx.copysign` and `fxx.reinterpret*`, which
always work on the floating-point representation directly, and the
`fxx.convert*` instructions, which cannot be NaN because the input is an
integer.


binji:
This should fix some issues in #1043 and #1045. The wasm2c failures still need to be addressed.

PR #1052, opened 2019-03-28
Add ParseInt{8,16} functions, for use with SIMD

binji:


binji:
@gnzlbg FYI, just a minor cleanup from your PR

PR #1053, opened 2019-03-29
Rename v8x16.shuffle to v8x16.shuffle2_imm

gnzlbg:
To conform to the latest SIMD draft, see: https://github.com/WebAssembly/simd/pull/71


tlively:
That SIMD spec proposal change was merged explicitly without consensus on the instruction names. I think toolchain updates could wait until we actually have a discussion on the new names, since I don’t think the current temporary names are any more stable than the previous names.

gnzlbg:
We can leave this on hold, I understood from the discussion that the new names had consensus, but if not, we can leave this PR open and maybe try to achieve consensus on the next meeting about the names.

binji:
I believe the changes in this PR have now landed here: https://github.com/WebAssembly/wabt/commit/c36c95d34aacced046314d444f89e1cd07fa4ba2. If not, feel free to send a new PR with the missing parts. Thanks!

PR #1060, opened 2019-04-05
Made common string lookups more robust.

aardappel:
Since the purpose of these strings is to display them to the user,
an "error string" is a more useful default than an assert, which
would require all callers to consistently check for the range
of valid values (which they didn't).

Tools like wasm-objdump should preferably be able to continue
outputting even in the face of illegal values.


binji:
lgtm, thanks!

PR #1069, opened 2019-04-24
wasm-objdump: Add function offset when disassembling

sbc100:
Currently we show file offset in hex, but backtraces in v8 show
functions offset in decimal

This change adds the decimal offset in addition.  Perhaps we could
drop the hex file offset?  Or make it an option?


src/binary-reader-objdump.cc
```diff
@@ -643,7 +646,8 @@ Result BinaryReaderObjdumpDisassemble::OnEndExpr() {
 
 Result BinaryReaderObjdumpDisassemble::BeginFunctionBody(Index index,
                                                          Offset size) {
-  printf("%06" PRIzx " func[%" PRIindex "]", state->offset, index);
+  current_function_offset = state->offset;
+  printf("%06" PRIzx " func[%02" PRIindex "]", state->offset, index);
```
binji: Seems a bit strange to me to use a leading zero for a decimal number. Also, none of the other indexes have leading zeroes.

sbc100: Its for alignment of the instructions which are on the right of this information.

sbc100:
So this a v8 bug I guess.

sbc100:
Opened a v8 issue: https://bugs.chromium.org/p/v8/issues/detail?id=9172

PR #1073, opened 2019-04-30
Fixing "fac" factorial function

igna31:


binji:
Thanks! Both of these forms are valid, though, so this change isn't needed.

kirjs:
image

I think this pull requests exists because the deployed version of demo is out of sync with the code.

Maybe redeploy?

binji:
Oops, you're right! I've pushed to gh-pages, should be fixed now.

PR #1076, opened 2019-05-08
Initial custom annotation support

jgravelle-google:
Lex custom annotations, but discard them in the parser. In the future we
should be able to parse some of them, but this is simple and
spec-compliant for now.

(Annotations proposal: https://github.com/WebAssembly/annotations)

### DIFF
--- a/src/wast-lexer.cc
+++ b/src/wast-lexer.cc
@@ -64,6 +64,9 @@ Token WastLexer::GetToken(WastParser* parser) {
             continue;
           }
           return BareToken(TokenType::Eof);
+        } else if (MatchString("(@")) {
+          ReadReservedChars();
+          return TextToken(TokenType::LparAnn);
         } else {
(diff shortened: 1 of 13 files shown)


src/wast-lexer.cc
```diff
@@ -64,6 +64,9 @@ Token WastLexer::GetToken(WastParser* parser) {
             continue;
           }
           return BareToken(TokenType::Eof);
+        } else if (MatchString("(@")) {
+          ReadReservedChars();
+          return TextToken(TokenType::LparAnn);
```
binji: Should be TextToken(..., 2), to skip the "(@"

jgravelle-google: done in https://github.com/WebAssembly/wabt/pull/1076/commits/03bdf8e5d7c91f10062ce767e09ce34a575bdcff

PR #1081, opened 2019-05-16
Fix up CMake warning

billti:
Building the project (on Windows) per the instructions I was seeing the below

```
CMake Warning (dev) at CMakeLists.txt:38 (elseif):
  Policy CMP0054 is not set: Only interpret if() arguments as variables or
  keywords when unquoted.  Run "cmake --help-policy CMP0054" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  Quoted variables like "MSVC" will no longer be dereferenced when the policy
  is set to NEW.  Since the policy is not set the OLD behavior will be used.
This warning is for project developers.  Use -Wno-dev to suppress it.
```

After some confusion attempts at a fix, I came across https://stackoverflow.com/a/31044116 which states:

```
But be aware this can give you a headache in if() commands. Here is an example where CMAKE_CXX_COMPILER_ID is "MSVC" and MSVC is "1":

  if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC") is true, because it evaluates to if("1" STREQUAL "1")
  if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC") is false, because it evaluates to if("MSVC" STREQUAL "1")

So the best solution here would be - see above - to directly check for if(MSVC)
```

So it turns out on older CMake versions that comparison isn't even checking what you think it is. "MSVC" is already a variable (see https://cmake.org/cmake/help/v3.0/variable/MSVC.html), and expanded in the `if` statement. Simplest check is just to check if it is "truthy" (i.e. 1).

I also removed the needless `${..}` in the `if` conditions, as `if` already expands variables (see https://stackoverflow.com/questions/45900159/how-to-use-variables-and-avoid-cmp0054-policy-violations). And switched the Clang check to a `MATCH`, as it can be `Clang` or `AppleClang` (according to https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html). See the note starting "As of CMake 3..." at https://stackoverflow.com/questions/10046114/in-cmake-how-can-i-test-if-the-compiler-is-clang.

I put a couple trivial additions in .gitignore also. I excluded `build`, as the readme instructrions generate the output to that dir. And excluded any `.vscode` dirs, as that's a common editor these days (and the one I'm using). Happy to separate out if you care out these things.


binji:
Wow, this is great, thanks!

PR #1082, opened 2019-05-16
Fix INSTALL_PREFIX in README

billti:
The `install` command in the `CMakeLists.txt` file already includes a `bin` destination (see https://github.com/WebAssembly/wabt/blob/master/CMakeLists.txt#L439), so the current readme instructions left me with the binaries in a `.\bin\bin` directory. Updating the README here to remove the unnecessary `bin`.

Also updated the readme to note that CMake (and Ninja) are in the box now with VS (and thus don't need to be installed as a separate dependencies).

Note: VS 2015 is a couple releases old now, so happy to update those instructions to 2019 too if you like :wink:

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,8 +100,12 @@ $ make gcc-debug-no-tests
 ## Building (Windows)
 
 You'll need [CMake](https://cmake.org). You'll also need
-[Visual Studio](https://www.visualstudio.com/) (2015 or newer) or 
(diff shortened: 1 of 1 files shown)


binji:
Thanks again!

> Note: VS 2015 is a couple releases old now, so happy to update those instructions to 2019 too if you like :wink:

Yes, but we test with VS2015 on the CI and it works, so may as well keep it :)

PR #1084, opened 2019-05-17
[wasm-objdump] Fix crash on invalid reloc sections

sbc100:


src/binary-reader-objdump.cc
```diff
@@ -1583,6 +1589,7 @@ Result BinaryReaderObjdump::OnGlobalSymbol(Index index,
 Result BinaryReaderObjdump::OnSectionSymbol(Index index,
                                             uint32_t flags,
                                             Index section_index) {
+  printf("%d\n", index);
```
binji: debugging?

PR #1099, opened 2019-06-28
Write text on assert_exhaustion

pepyakin:
Reported in https://github.com/pepyakin/wabt-rs/issues/33


binji:
This doesn't quite work, I've got a working PR here: https://github.com/WebAssembly/wabt/pull/1100

PR #1101, opened 2019-06-28
[interp] Add flag to provide dummy import funcs

binji:
You can now pass `--dummy-import-func` to `wasm-interp`, which will
provide a function that logs the call and returns zero.


fitzgen:
Awesome!

Issue #1103, opened 2019-07-02
Allow building with system gtest and ply


ids1024:
`gtest` and `ply `seem to be packaged by Arch and Alpine (and presumably other distros, though perhaps not all). Is it necessary have them vendored as submodules?

It seems desirable for packaging to either not provide a vendored version, or have an option to use the system version instead.

ids1024:
A test implementation for (unconditionally) using the system version of gtest: https://github.com/ids1024/wabt/commit/c497ff11b356720ca02cd63499673f03eacd87fb

This works on Arch. Not on Alpine, since `gtest_main.cc` isn't packaged. Perhaps it isn't possible to rely on that being packaged (and in a consistent place), but it's a small enough file to duplicate (7 non-comment lines).

binji:
Having gtest and ply as submodules makes it easier to build and test. Otherwise I may run into various version incompatibilities, etc. Both libraries are used only for development/testing of wabt, and not used for the final installed executables. So I'm not sure if it is all that useful to use the system versions here. If so, we can provide an option, though it wouldn't be the default so it's likely to break over time unless maintained.

PR #1105, opened 2019-07-04
Build against system gtest

ids1024:
Meant to address #1103.

`gtest` seems to be packaged on Arch Linux, Alpine Linux, Ubuntu, etc. It is generally preferable to use the system copy of a library instead of vendoring a copy.

Alternately, it be desirable to retain an option to build gtest, which could make things easier for some (for example, on Windows). But in my experience, building dependencies is generally left to the user.

I'm not particularly experience with `cmake`, so I don't necessarily know the preferred way to do things.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,6 @@
 [submodule "third_party/testsuite"]
 	path = third_party/testsuite
 	url = https://github.com/WebAssembly/testsuite
-[submodule "third_party/gtest"]
-	path = third_party/gtest
-	url = https://github.com/google/googletest
 [submodule "third_party/ply"]
 	path = third_party/ply
 	url = https://github.com/dabeaz/ply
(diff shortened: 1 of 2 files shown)


ids1024:
Apparently the gtest FAQ used to have an entry "Why is it not recommended to install a pre-compiled copy of Google Test (for example, into /usr/local)?", but this was removed in https://github.com/google/googletest/commit/5437926b2213b1c45c2f34bd858734de90e5fffd.

I assume the removal means they no longer advise against it?

binji:
I wouldn't want to remove gtest as a submodule; instead I'd prefer to add a CMake flag to optionally use the system gtest. The Travis and AppVeyor build scripts would have to be updated to support this too.

ids1024:
Overall, I'm inclined to agree that that's a sensible choice.

PR #1111, opened 2019-07-13
Update spec testsuite

binji:
* Remove passive keyword from bulk-memory
* Fix const parsing


binji:
This _almost_ works, but fails on two tests on MSVC 32-bit:

```
out/test/spec/conversions.wast:319: mismatch in result 0 of assert_return: expected f32:9223373136366403584.000000, got f32:9223372036854775808.000000
out/test/spec/conversions.wast:320: mismatch in result 0 of assert_return: expected f32:18446742974197923840.000000, got f32:18446741874686296064.000000
```

Which are these two cases from `conversions.wast`:

```wasm
(assert_return (invoke "f32.convert_i64_u" (i64.const 0x8000008000000001)) (f32.const 0x1.000002p+63))
(assert_return (invoke "f32.convert_i64_u" (i64.const 0xfffffe8000000001)) (f32.const 0x1.fffffep+63))
```

The part I don't quite understand is that the current implementation of `wabt_convert_uint64_to_float` basically matches the generated clang/gcc assembly:

msvc:

```
__inline float wabt_convert_uint64_to_float(unsigned __int64 x) {
  static const float c = 18446744073709551616.0f;
  float result;
  __asm fild x;
  if (x & 0x8000000000000000ULL) {
    __asm fadd c;
  }
  __asm fstp result;
  return result;
}
```

gcc: (from godbolt.org)

```
        sub     esp, 20
        mov     eax, DWORD PTR [esp+24]
        mov     edx, DWORD PTR [esp+28]
        movd    xmm0, eax
        movd    xmm1, edx
        punpckldq       xmm0, xmm1
        movq    QWORD PTR [esp+8], xmm0
        fild    QWORD PTR [esp+8]
        test    edx, edx
        jns     .L8
        fadd    DWORD PTR .LC1
.L8:
        fstp    DWORD PTR [esp+4]
        fld     DWORD PTR [esp+4]
        add     esp, 20
        ret
.LC1:
        .long   1602224128
```

I'm currently wondering if it's because the functions are inlined...

binji:
> I'm currently wondering if it's because the functions are inlined...

nope.

PR #1116, opened 2019-07-18
Add support for v8x16.swizzle and the load_splats.

nlewycky:
v8x16.shuffle2_imm is a rename of the previous v8x16.shuffle, but I add a copy of the code as if it were a new instruction in case the spec proposal makes further changes. The tests for old v8x16.shuffle remain in place and while there are new tests for the new v8x16.shuffle1, there are not for v8x16.shuffle2_imm. The behaviour and implementation are the same as for v8x16.shuffle, so we should simply search and replace the existing tests at some point, leaving one of them untested, probably the deprecated v8x16.shuffle.

I did test v8x16.shuffle1 against the SIMD spec test from WAVM and it passes. The WAVM spec tests for v8x16.shuffle2_imm parse but it has no invocations of the instruction.


tlively:
The SIMD group decided a little while ago to rename v8x16.shuffle1 and v8x16.shuffle2_imm to v8x16.swizzle and v8x16.shuffle_imm, respectively. We just haven’t reflected that change in our proposal repo yet 😳 I’ll make that update to the proposal repo today. Sorry for the confusion.

nlewycky:
@tlively New name, same SIMD flavour!

tlively:
Here's the PR to update the proposal, fyi: https://github.com/WebAssembly/simd/pull/84

src/interp/interp-trace.cc
```diff
@@ -586,6 +586,26 @@ void Thread::Trace(Stream* stream) {
           ReadU32At(pc + 4), ReadU32At(pc + 8), ReadU32At(pc + 12));
       break;
 
+    case Opcode::V8X16Swizzle:
```
binji: This looks to be the same as other v128 binary instructions, combine with the list above (`F64X2Mul`, etc)

tlively:
I had misremebered: the (officially merged) new names are `v8x16.swizzle` and `v8x16.shuffle`. Apologies for the turn!

src/interp/interp-trace.cc
nlewycky: Done.

test/interp/simd-load-store.txt
```diff
@@ -12,6 +12,11 @@
     i32.const 4
     v128.load)
 
+  ;; i8x16 load_splat
```
binji: Test other load_splat instructions too?

nlewycky: Done.

PR #1126, opened 2019-07-22
Change WABT_UNREACHABLE to call abort()

binji:
Using __builtin_unreachable instead will basically produce undefined
behavior if it is ever reached, which makes it very difficult to debug.
This is perhaps desired for an optimized build, but certainly not for a
debug build.

For now, we'll always convert these to abort().

### DIFF
--- a/src/config.h.in
+++ b/src/config.h.in
@@ -18,6 +18,7 @@
 #define WABT_CONFIG_H_
 
 #include <stdint.h>
+#include <stdlib.h>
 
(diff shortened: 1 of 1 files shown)


tlively:
Binaryen has some fairly fancy logic for this: https://github.com/WebAssembly/binaryen/blob/db9124f1de0478dcac525009b6f1589b44a7edd8/src/compiler-support.h#L28. Perhaps WABT could borrow some of that.

PR #1131, opened 2019-07-25
Add support for ref.null in global initializers

wingo:
As per the reference-types spec, ref.null is a constant expression.

I did try to add support for ref.func at the same time, but I didn't really understand the strategy of wabt and reference types; there seems to be strong intertwingliness between func.ref and the element section, whereas it does seem possible for a func.ref to be global.  Oh well, I didn't need it :)


binji:
Looks good, thanks!

> I did try to add support for ref.func at the same time, but I didn't really understand the strategy of wabt and reference types; there seems to be strong intertwingliness between func.ref and the element section, whereas it does seem possible for a func.ref to be global.

Yeah, this will have to be untangled a bit. Though that's not part of the reference-types proposal, but instead the function-references proposal, right?

PR #1132, opened 2019-07-26
Add --enable-all option

aheejin:
This adds `--enable-all` option that enables all existing features.


Horcrux7:
I like this feature. Make the development for future versions easier. Never use a feature accidentally.

PR #1136, opened 2019-08-05
wasm2c: wasm_rt_grow_memory should return -1 when OOM

nomeata:
the README states that wasm_rt_grow_memory should return -1 if it runs
out of memory, but the example C runtime does not actually do that.

This should fix that.


binji:
Oh, actually it looks like this fails on the bots. The issue seems to be that `realloc()` with a size of zero will free the block. I think the easiest solution here is to early out at the beginning of `wasm_rt_grow_memory` when `delta == 0`.

PR #1144, opened 2019-08-14
Run flake8 on the whole repository

quantum5:
Seems like flake8 was only run on the scripts directory.


test/update-spec-tests.py
```diff
@@ -75,13 +76,6 @@ def ProcessProposalDir(name, flags=None):
              flags)
 
 
-def ProcessTestsuite(wabt_test_dir, testsuite_dir, tool, flags=None):
-  testsuite_dir = os.path.join(TESTSUITE_DIR, reldir)
```
quantum5: Removed this function because it's not called anywhere and I have no idea what `reldir` is supposed to be. It's not defined anywhere.

PR #1145, opened 2019-08-15
Switch python indentation from 2-space to 4-space

sbc100:
pep8 specifies 4 space indentation.  The use of 2 spaces is, I believe,
a historical anomaly where certain large organizations such as google
chose 2 over 4 and have yet to make the switch.

For a project like wabt with little python code I think the cost of
switching is small enough to justify the churn.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,8 +1,5 @@
 [flake8]
 ignore =
-  E111, # indentation not a multiple of 4 (we use 2))
-  E114, # comment indentation not a multiple of 4
   E501, # line too long
-  E121, # continuation line under-indented for hanging indent
   W504  # line break after binary operator
-exclude = ./third_party
+exclude = ./third_party ./out
--- a/scripts/gen-emscripten-exported-json.py
+++ b/scripts/gen-emscripten-exported-json.py
@@ -32,78 +32,78 @@ from utils import Executable, Error  # noqa: E402
 
 
 def FindFiles(cmake_build_dir):
-  result = []
-  for root, dirs, files in os.walk(cmake_build_dir):
-    for file_ in files:
-      path = os.path.join(root, file_)
-      if file_ == 'libwabt.a' or re.search(r'emscripten-helpers', file_):
-        result.append(path)
-  return result
+    result = []
+    for root, dirs, files in os.walk(cmake_build_dir):
+        for file_ in files:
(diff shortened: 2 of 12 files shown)


Alhadis:
> I don't like 4-space much

Wouldn't it be great if there was a configurable yet unambiguous way to encode a tabstop directly inside a file without making assumptions about its width? And somehow have this semantic information understood by even the most rudimentary of text-editors?

Oh wait, tabs are bad... 😉 But people can never explain why, only remind you of the *main reason* to use them. *“They have different sizes in different editors…”*

But, like, tabs are also evil, and stuff. Probably also taboo in three world religions, banned in China, the real cause of the Bronze Age Collapse, and unequivocally spawned from parasitic alien life which [infected Earth](https://www.youtube.com/watch?v=SST5NjFSYEo) during the Stone Age... oh never mind, I'm thinking of EBCDIC.

> \*holds his breath and smashes that sexy little <kbd>SUBMIT</kbd> button\*

*Geronimoooooooo!*

tlively:
:joy::joy::joy:

... but seriously, tabs are evil :smiling_imp:

Alhadis:
👌 It's okay, pick whatever tabstop you want, because I'm reformatting your code before reading it anyway. No, seriously, I'm [habitually running this](https://github.com/Alhadis/Unfuck-Tabstops) any time I'm loading most projects on GitHub. 👍 😉 Say what you want about my intolerance, you can't fault me for channeling that anger into a solution, instead of, y'know, whining about it at [`/r/bikeshed`](https://www.reddit.com/r/bikeshed/).

Also, uh, that subreddit, [`/r/bikeshed`](https://www.reddit.com/r/bikeshed/)? Its posts seem to be a mix about hacker-related stuff and stuff about... actual bicycles. I really have no words for what this is making me feel right now.

sbc100:
Perhaps we should add an extra clause to our apache license that forbids refomatting of the code to contain tabs?    It would make the license non-free, but it would also help reduce evil in the world.

mohe2015:
https://dmitryfrank.com/articles/indent_with_tabs_align_with_spaces should be the correct way in my opinion.

Alhadis:
@mohe2015 That's what I do, but I always keep them separated by at least one non-whitespace character, ensuring there's no ambiguity.

PR #1151, opened 2019-09-04
WIP: Improve support for reference-types proposal

acfoltzer:
- Adds `table.fill` instruction
- Adds `assert_return_func` and `ref.host` for spec scripts
- Expands the IR `Const` type to include `ref.null` and `ref.host`
- Enables opcodes that are shared between reference-types and bulk-memory-operations as long as either is enabled
- Swaps the order of type and table arguments to `call_indirect` in the text format to match current proposals

A few caveats to this code:
- I am not a C++ programmer; I tried to follow the style of the surrounding code, but if you spot anything that seems questionable, it's probably not for any good reason other than me being boneheaded.
- My primary objective with this was to get `wast2json` working so that we could run the spec tests in Lucet. While I have made sure that `make test` passes, I may have some blind spots for other use cases.
- I'm not terribly satisfied with how `ref.null` and `ref.host` are included in `Const`, but this seemed like the most straightforward way to add support. In particular, since `ref.host` doesn't have an associated opcode, it's treated as a different token type despite eventually becoming a `Const`.


acfoltzer:
I just also added `Features` arguments to more of the C API functions in `emscripten-helpers.cc`, but I don't know how to test those as `make test` allowed non-compiling code through on my first attempt.

acfoltzer:
I discovered that the type checker and validator need to be extended to handle the subtyping relation for references before this will fully work on the spec tests; please don't merge this yet.

acfoltzer:
I haven't had too much time to work on this since the initial PR, but I'm getting wedged on what the right way is to implement the more complex typechecking rules present in the proposal. If someone who's more familiar with the project could give me a nudge in the right direction, I would appreciate it.

First, I'm not sure how best to add `nullref` as a type; currently I have added it to the `Type` enum, but I'm not sure whether this is the right thing to do for a type that can't actually appear in a Wasm module:

```c++
// Matches binary format, do not change.
enum class Type : int32_t {
  I32 = -0x01,      // 0x7f
  I64 = -0x02,      // 0x7e
  F32 = -0x03,      // 0x7d
  F64 = -0x04,      // 0x7c
  V128 = -0x05,     // 0x7b
  Funcref = -0x10,  // 0x70
  Anyref = -0x11,   // 0x6f
  Exnref = -0x18,   // 0x68
  Func = -0x20,     // 0x60
  Void = -0x40,     // 0x40
  ___ = Void,       // Convenient for the opcode table in opcode.h
  Any = 0,          // Not actually specified, but useful for type-checking
  Nullref = -0x12,  // Specified in reference-types as an internal-only type
};
```

Second, the current uses of functions like `TypeChecker::CheckType()` assume that there is no subtyping at work. Adding subtyping means that we need to add a notion of variance, because in some contexts we can accept subtypes, in others we can accept supertypes, and finally some might need types to still match exactly. I'm leaning towards adding variance as an `enum class` argument to the relevant typechecking functions, but again I'm not much of a C++ programmer so I'd like to check whether this is a reasonable plan.

Finally, we will need least-upper-bound and greatest-lower-bound operators on types. It seems clear that `Type::Any` is suitable to use as the top element for the lub, but is `Type::Void` suitable for bottom? Also, is there an implementation for these you'd recommend that'd be more readable/maintainable than just a couple of hairy `switch` statements?

binji:
Sorry, I haven't been able to look at this since I'm temporarily away on other projects. It seems like [this change](https://github.com/WebAssembly/wabt/pull/1156) may address some of the issues, can you confirm?

I'll try to find some time to do a more thorough review this weekend.

acfoltzer:
Ah, thanks for that pointer! I've rebased my work onto the latest changes. There are still some issues, though, with getting the reference-types spec tests to validate. I'll work on adding them to the test suite so we can all see what's happening more easily.

acfoltzer:
Well, that was easier than I was expecting thanks to the update spec tests script!

It looks like the type of `ref.null` is being treated as a type index, probably because `Nullref` is set to a positive integer in @yurydelendik's changes:
```
out/test/spec/reference-types/globals.wast:187:17: error: type mismatch for result 0 of action. got anyref, expected <type_index>
(assert_return (invoke "get-r") (ref.null))
                ^^^^^^
```

In my attempt (pasted upthread) I used `-0x12` to avoid being in the type index space, but this also is just making up a binary encoding which I'm not entirely comfortable with.

Secondly, we also need to add a lub operator and adapt the typechecking of instructions like branches and select that currently require their operands to have exactly equal types. For example:

```
out/test/spec/reference-types/br_table.wast:1257:10: error: br_table labels have inconsistent types: expected [funcref], got [anyref]
        (br_table $l2 $l1 $l2 (table.get 0 (i32.const 0)) (local.get 0))
         ^^^^^^^^
```

Also, it'll take quite a bit more work to get these tests to actually _pass_, as the proposal extensions are unimplemented in the interpreter, but it's still useful to see which of the spectests manage to make it past the `wast2json` step.

acfoltzer:
> In my attempt (pasted upthread) I used `-0x12` to avoid being in the type index space, but this also is just making up a binary encoding which I'm not entirely comfortable with.

This was a red herring; there just wasn't a case to properly print "nullref". What remains is the issue of needing different variances at different points in the typechecking: we usually want to check whether the actual type is a subtype of the expected type, but sometimes we want the relationship the other way around:

```
out/test/spec/reference-types/globals.wast:187:17: error: type mismatch for result 0 of action. got anyref, expected nullref
(assert_return (invoke "get-r") (ref.null))
```

binji:
Right, the current typechecking is pretty limited. I imagine it will need a rewrite for this. Unfortunately, I don't have the bandwidth to do that currently. It seems like you know how it should work, it would be awesome if you could make the change! Is there anything I could help with w.r.t. direction?

acfoltzer:
Unfortunately my bandwidth has dried up as well, probably for at least the next couple months. I'm happy to pick it up again later, but I also don't want to block progress for anyone else who may be inclined.

binji:
Gonna close this -- there's pretty good reference types support now, both in the interpreter and elsewhere.

PR #1153, opened 2019-09-06
Attempt to locate Emscripten if EMSCRIPTEN_DIR is not set

dcodeIO:
This PR is an attempt to locate Emscripten automatically if it is in path, like when using emsdk, instead of having to set `EMSCRIPTEN_DIR` manually in order to run `make emscripten-release`. Not super important, just stumbled over this while trying to compile libwabt.js in a docker container.


Makefile
```diff
@@ -23,6 +23,7 @@ USE_NINJA ?= 0
 FUZZ_BIN_DIR ?= ${ROOT_DIR}/afl-fuzz
 GCC_FUZZ_CC := ${FUZZ_BIN_DIR}/afl-gcc
 GCC_FUZZ_CXX := ${FUZZ_BIN_DIR}/afl-g++
+EMSCRIPTEN_DIR ?= $(dir $(shell which emcc))
```
binji: Will line 27 ever execute in this case? If not, we should remove it. Otherwise, lgtm. Thanks!

PR #1155, opened 2019-09-10
[WIP] Added initial skeleton code for wasm-decompile.

aardappel:
This is the skeleton code with some partial, very incomplete functionality, and will likely heavily change in the future.

@binji your review is optional, but welcome :)


src/decompiler.cc
```diff
@@ -0,0 +1,426 @@
...
+  string_view OpcodeToToken(Opcode opcode) {
+    return opcode.GetDecomp();
+  }
+
+  void IndentValue(Value &val, size_t amount, string_view first_indent) {
+    WABT_TRACE_ARGS(IndentValue, "\"" PRIstringview "\" - %d",
+                    WABT_PRINTF_STRING_VIEW_ARG(val.v[0]), val.v.size());
+    auto indent = Indent(amount);
+    for (auto& s : val.v) {
+      auto is = &s != &val.v[0] || first_indent.empty()
```
jgravelle-google: Parens around the condition would help me here (or extracting to its own var)

src/decompiler.cc
```diff
@@ -0,0 +1,426 @@
...
+      }
+    } else {
+      // Multiline with prefix on its own line.
+      IndentValue(child, indent_amount, {});
+      v.insert(v.begin(), std::string(prefix));
+      v.back().append(postfix.data(), postfix.size());
+    }
+  }
+
+  void BracketIfNeeded(Value &val) {
```
jgravelle-google: So we decouple creating bracketed expressions from actually bracketing them, to avoid brackets when they aren't needed. Nice.
Possibly worth a comment in the struct definition?

src/decompiler.cc
```diff
@@ -0,0 +1,426 @@
...
+      Value bin { {}, true };
+      std::move(left.v.begin(), left.v.end(), std::back_inserter(bin.v));
+      bin.v.back().append(infix.data(), infix.size());
+      if (indent_right) IndentValue(right, indent_amount, {});
+      std::move(right.v.begin(), right.v.end(), std::back_inserter(bin.v));
+      stack.push_back(std::move(bin));
+    }
+  }
+
+  std::string TypeFromLoadStore(Opcode opcode, string_view name) {
```
jgravelle-google: An example in a comment would be useful here. E.g., "i32.load8_u" -> "i32_8_u"

src/decompiler.cc
```diff
@@ -0,0 +1,426 @@
...
+            ss << *reinterpret_cast<const float *>(&c.f32_bits) << "f";
+            break;
+          case Type::F64:
+            ss << *reinterpret_cast<const double *>(&c.f64_bits) << "d";
+            break;
+          case Type::V128:
+            ss << "V128";  // FIXME
+            break;
+          default:
+            assert(false);
```
jgravelle-google: Nit: WABT_UNREACHABLE maybe?

src/decompiler.cc
```diff
@@ -0,0 +1,426 @@
...
+      case ExprType::GlobalGet: {
+        auto& gg = *cast<GlobalGetExpr>(&e);
+        ss << gg.var.name();
+        break;
+      }
+      case ExprType::LocalSet: {
+        auto& ls = *cast<LocalSetExpr>(&e);
+        WrapChild(stack.back(), ls.var.name() + " = ", "");
+        stack.back().needs_bracketing = true;
+        return;
```
jgravelle-google: Oof, some of these breaking and some returning is weird.

Can we split this into two switches, one where we always early-return, another where we always break (plus the logic in `default`)? That makes the control flow edges less overlapping.

Or simpler, wrap the end of this function with `if (ss.str().size() > 0)` and always break. We return in the cases where we don't use `ss`, are there any times where we rely on keeping state in `ss` across calls to `DecompileExpr`?

src/decompiler.cc
```diff
@@ -0,0 +1,426 @@
...
+        auto le = cast<LoopExpr>(&e);
+        DecompileExprs(le->block.exprs);
+        auto &val = stack.back();
+        IndentValue(val, indent_amount, {});
+        val.v.insert(val.v.begin(), "loop {");
+        val.v.push_back("}");
+        return;
+      }
+      default: {
+        std::string name = GetExprTypeName(e.type());
```
jgravelle-google: Nit: can move to default case of switch and avoid unconditionally calling `GetExprTypeName`

src/decompiler.cc
```diff
@@ -0,0 +1,426 @@
...
+            break;
+          }
+          case ExprType::Convert: {
+            name = std::string(OpcodeToToken(cast<ConvertExpr>(&e)->opcode));
+            break;
+          }
+          default:
+            break;
+        }
+        size_t twidth = 0;
```
jgravelle-google: total_width and max_width? These live just long enough that the extra clarity is worth it. Alternatively: tot_w and max_w

CMakeLists.txt
```diff
@@ -207,45 +208,87 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${WABT_SOURCE_DIR}/cmake)
 add_custom_target(everything)
 
 add_library(wabt STATIC
+  src/apply-names.h
```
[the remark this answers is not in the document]
aardappel: This is just for IDEs, who typically depend on CMake to tell it what files are part of a project. Without it, it thinks the headers don't exist.

src/tools/wasm-decompile.cc
```diff
@@ -0,0 +1,114 @@
...
+  if (Succeeded(result)) {
+    Errors errors;
+    Module module;
+    const bool kStopOnFirstError = true;
+    ReadBinaryOptions options(features, nullptr,
+                              true, kStopOnFirstError,
+                              fail_on_custom_section_error);
+    result = ReadBinaryIr(infile.c_str(), file_data.data(), file_data.size(),
+                          options, &errors, &module);
+    if (Succeeded(result)) {
```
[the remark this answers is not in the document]
aardappel: Yeah, this code is copied from `wasm2wat`, decided to leave the logic as-is.

src/opcode.def
```diff
@@ -1,7 +1,7 @@
 /*
  * Copyright 2017 WebAssembly Community Group participants
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
+ * Licensed under the Apache License, Version 2.0 (the "License", "");
```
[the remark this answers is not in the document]
aardappel: eek!

src/ir-util.h
```diff
@@ -0,0 +1,69 @@
...
+        param_types(param_types),
+        result_types(result_types) {}
+
+  std::string name;
+  LabelType label_type;
+  TypeVector param_types;
+  TypeVector result_types;
+};
+
+struct ModuleContext {
```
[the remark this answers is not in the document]
aardappel: A base class name doesn't necessarily need to express the smallest possible set of subclasses, it can be quite broad. Is an Animal an IntrusiveLinkedListNode? It sure can be.

This is a ModuleContext in the sense that it tracks additional information as you iterate thru a Module, but it is not necessarily specific to text writing.

Also, goal was to make as little changes to WatWriter by making it a base class, not a member.

src/ir-util.h
```diff
@@ -0,0 +1,69 @@
...
+
+  Index GetFuncParamCount(const Var& var);
+  Index GetFuncResultCount(const Var& var);
+
+  void BeginBlock(LabelType label_type, const Block& block);
+  void EndBlock();
+  void BeginFunc(const Func& func);
+  void EndFunc();
+
+  std::pair<Index, Index> GetExprArity(const Expr& expr);
```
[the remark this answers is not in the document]
aardappel: Sure

src/decompiler.cc
aardappel: Agree, though this code is heavily in flux, so I want to avoid spending too much time perfecting it.

Actually, the simplest is to always make them return, and to stick the finishing code in its own function.

PR #1163, opened 2019-09-19
Improved naming.

aardappel:


src/generate-names.h
```diff
@@ -23,7 +23,16 @@ namespace wabt {
 
 struct Module;
 
-Result GenerateNames(struct Module*);
+Result GenerateNames(struct Module*, bool numbered_names, bool dollar_prefix);
```
jgravelle-google: Default args maybe?

Also I'm never a fan of `foo(bar, true, true)` at the callsite. So,

```cpp
enum NameOptions {
  nameNumbered = 1 << 0,
  nameDollared = 1 << 1,
};

// ...

Result GenerateNames(struct Module*, NameOptions options = nameNumbered | nameDollared);
```

src/generate-names.h
```diff
@@ -23,7 +23,16 @@ namespace wabt {
...
+
+inline std::string IndexToAlphaName(Index index) {
+  std::string s;
+  do {
+    s += 'a' + (index % 26);
```
jgravelle-google: This builds LTR, so ... `x`, `y`, `z`, `aa`, `ba`, `ca`, `da` ...
That might be fine, because it puts the most frequently changing character at the front? But to have the most significant digit first, `s = ('a' + (index % 26)) + s;`. Or this needs a comment saying it was intended.

Ok actually trying this out (on https://binji.github.io/wasm-clang/ :D) it actually goes `z` `ab` `bb` `cb`. This turns out to be because the first time we get a new digit, it always starts with 1. We actually have an almost-base27 system because we may not have a digit.

So we really need:

```cpp
inline std::string IndexToAlphaName(int x) {
  std::string s;
  s += 'a' + x % 26;
  x /= 26;
  while(x) {
    s = (char)('a' + (x-1)%26) + s;
    x /= 27;
  }
  return s;
}
```

Test cases:
```
0: a
1: b
2: c
3: d
4: e
23: x
24: y
25: z
26: aa
27: ab
28: ac
29: ad
50: ay
51: az
52: ba
53: bb
54: bc
699: zx
700: zy
701: zz
702: aaa
703: aab
704: aac
18951: zzx
18952: zzy
18953: zzz
18954: aaaa
18955: aaab
18956: aaac
```

jgravelle-google: Just kidding that doesn't work for:
```
1376: azy
1377: azz
1378: aaa
1379: aab
1380: aac
1401: aax
1402: aay
1403: aaz
1404: bba
1405: bbb
1406: bbc
```

So, who knows :woman_shrugging:

src/generate-names.cc
```diff
@@ -104,30 +104,44 @@ class NameGenerator : public ExprVisitor::DelegateNop {
...
   return !str.empty();
 }
 
-// static
 void NameGenerator::GenerateName(const char* prefix,
                                  Index index,
                                  unsigned disambiguator,
                                  std::string* str) {
-  *str = prefix;
+  if (dollar_prefix_) *str = "$";
```
jgravelle-google: Kinda slick how this moves lots of hardcoded $'s from elsewhere.

src/generate-names.h
aardappel: Done

jgravelle-google:
Oh, and this could stand to have a better commit message. Improve its naming, if you will.
It needs more detail to be clearer as to what it's doing, and how the naming is improved.

src/generate-names.h
aardappel: Not that it matters a ton, but most frequently changing char first may well be a bit more readable. I hope not a lot of code needs it :)

    do {
      // For multiple chars, put most frequently changing char first.
      s += 'a' + (index % 26);
      index /= 26;
      // Continue remaining sequence with 'a' rather than 'b'.
    } while (index--);

It's funny, when I first wrote this code, I thought, ooh, finally a place where I can use do-while. Thanks `--` for allowing me to continue using it rather than resorting to `for(;;) { .. if() break; .. }`

src/generate-names.cc
aardappel: Actually I had first coded it to remove the `$` when necessary, then thought better of it :)

aardappel:
Improved commit (but not PR) message.

src/generate-names.h
jgravelle-google: Wow that's subtle. I love it.

PR #1165, opened 2019-09-23
wasm-decompile: Output Tee as Set+Get instead if possible.

aardappel:
This really de-tangles the code, as in-line assignments are
hard to read.

To make this possible, I had to track the current stack depth, take
into account unreachable paths and a few other support features.

Also added debug output upon assert.

### DIFF
--- a/src/decompiler.cc
+++ b/src/decompiler.cc
@@ -181,6 +181,16 @@ struct Decompiler : ModuleContext {
     ss.str({});
   }
 
+  template<ExprType T> void Get(const VarExpr<T>& ve) {
+    ss << ve.var.name();
+    PushSStream();
+  }
+
+  template<ExprType T> void Set(const VarExpr<T>& ve) {
+    WrapChild(stack.back(), ve.var.name() + " = ", "");
+    stack.back().needs_bracketing = true;
+  }
+
   void DecompileExpr(const Expr& e) {
     switch (e.type()) {
       case ExprType::Const: {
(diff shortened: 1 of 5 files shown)


src/ir-util.h
```diff
@@ -56,7 +56,13 @@ struct ModuleContext {
   void BeginFunc(const Func& func);
   void EndFunc();
 
-  struct Arities { Index nargs; Index nreturns; };
+  struct Arities {
```
jgravelle-google: Yeah I'm glad this is a struct instead of a `std::tuple<Index, Index, bool>` :)

src/decompiler.cc
```diff
@@ -208,33 +218,27 @@ struct Decompiler : ModuleContext {
...
-        stack.back().needs_bracketing = true;
+        Set(*cast<GlobalSetExpr>(&e));
         return;
       }
       case ExprType::LocalTee: {
         auto& lt = *cast<LocalTeeExpr>(&e);
-        WrapChild(stack.back(), lt.var.name() + " = ", "");
-        stack.back().needs_bracketing = true;
+        Set(lt);
+        if (stack_depth == 1) {  // Tee is the only thing on there.
```
jgravelle-google: Otherwise what happens? If `stack_depth > 1`, the value is still on the stack? We don't need to re-get here?

So if given
```
A:
i32.const 1
local.tee 0
i32.const 2
i32.add

B:
i32.const 1
i32.const 2
local.tee 0
i32.add
```

I would expect to see:
```
A:
a = 1
a + 2

B:
1 + (a = 2)
```

oh ok I answered my own question. The set still goes on the stack and gets parenthesized, so it just looks like an expression in that case. Nice!

aardappel: Will add a comment to clarify what happens in the `else` case :)

PR #1172, opened 2019-09-25
wasm-decompile: Implement br/br_if

aardappel:
br becomes break or continue, and br_if the same, but prefixed
by an if(..).

These refer to the enclosing loop/block by generated label name,
this syntax may change.

We may later want to replace these by while/do-while/switch and
other special cases, but for now this is decently readable.

Also added ; to statements.

### DIFF
--- a/src/decompiler.cc
+++ b/src/decompiler.cc
@@ -237,6 +237,16 @@ struct Decompiler : ModuleContext {
     stack.back().needs_bracketing = true;
   }
 
+  void Block(const Block& block, LabelType label, const char *name) {
+    BeginBlock(label, block);
+    DecompileExprs(block.exprs, block.decl.GetNumResults(), false);
+    EndBlock();
+    auto &val = stack.back();
+    IndentValue(val, indent_amount, {});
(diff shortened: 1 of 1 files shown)


src/decompiler.cc
```diff
@@ -373,21 +385,27 @@ struct Decompiler : ModuleContext {
...
-        DecompileExprs(le->block.exprs, le->block.decl.GetNumResults(), false);
-        auto &val = stack.back();
-        IndentValue(val, indent_amount, {});
-        val.v.insert(val.v.begin(), "loop {");
-        val.v.push_back("}");
+        Block(cast<LoopExpr>(&e)->block, LabelType::Loop, "loop");
+        return;
+      }
+      case ExprType::Br: {
+        auto bi = cast<BrExpr>(&e);
```
jgravelle-google: `bi`? Ah, copied from brIf below. Nit; `be` for Br-Expr? (and `bie` for Br-If-Expr)

Also `l` is a questionable name for a variable (looks too much like `1` or `I` at a quick glance), how about `lbl`?

aardappel: Done.

Issue #1106, opened 2019-07-08
bin/wasm2c --version # how to ?


drandreaskrueger:
```
bin/wasm2c --version
unknown option '--version'
```

How can I do that? If it doesn't exist yet then:

**feature request:** allow me to ask the binary about its version number.

(obviously, `--version` for all of them: `bin/spectest-interp  bin/wasm2c  bin/wasm2wat  bin/wasm-interp  bin/wasm-objdump  bin/wasm-opcodecnt  bin/wasm-strip  bin/wasm-validate  bin/wast2json  bin/wat2wasm  bin/wat-desugar`)

thanks.

binji:
Good point, these would be nice to add.

drandreaskrueger:
thumbs up

thanks!

PR #1175, opened 2019-09-26
Add `--version` to wabt tools

amirbawab:
Closes: #1106

Ported versioning system from [Binaryen CMakeLists.txt](https://github.com/WebAssembly/binaryen/blob/dc31b460fef47dfb3415b4ae6276fff4919a03e2/CMakeLists.txt#L10-L23)

```
bin/wasm2c --version
1.0.11-44-g71f883ad
```

Applied to (all) tools in `src/tools/`.

### DIFF
--- a/src/tools/spectest-interp.cc
+++ b/src/tools/spectest-interp.cc
@@ -71,6 +71,7 @@ static void ParseOptions(int argc, char** argv) {
     s_log_stream = FileStream::CreateStdout();
   });
   parser.AddHelpOption();
+  parser.AddVersionOption();
   s_features.AddOptions(&parser);
   parser.AddOption('V', "value-stack-size", "SIZE",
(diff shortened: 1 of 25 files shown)


src/tools/spectest-interp.cc
```diff
@@ -71,6 +71,7 @@ static void ParseOptions(int argc, char** argv) {
     s_log_stream = FileStream::CreateStdout();
   });
   parser.AddHelpOption();
+  parser.AddVersionOption();
```
sbc100: Since we are adding this to all tools, why not just call this in the OptionParser constructor?  

As a followup (to keep this change small) we could move AddHelpOption to the constructor too?

amirbawab: Both updated in 87c82544f5552e416c264ac348c743876c5309ea

PR #1178, opened 2019-09-30
Add a testcase reduction tool.

nlewycky:
Work in progress initial commit of a testcase reduction tool for wabt. This is similar to bugpoint for LLVM, or bugpoint for cranelift, C-Reduce or delta and multidelta from http://delta.tigris.org .

A testcase reduction tool for Wasm also exists as part of binaryen, named wasm-reduce. I started this implementation before I was aware of that. My main reason for choosing wabt was to build on the promise of full fidelity (1:1 round trips with no changes to instructions) and most spec conformance (spec proposals tend to supply implementations in wabt first). Is there any interest in having this tool as part of wabt?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -422,6 +422,13 @@ if (NOT EMSCRIPTEN)
         SOURCES src/tools/wasm-decompile.cc
         INSTALL
       )
+
+      # wasm-bugpoint
+      wabt_executable(
+        NAME wasm-bugpoint
+        SOURCES src/tools/wasm-bugpoint.cc
+        INSTALL
+      )
     endif()
   endif ()
 
--- a/src/tools/wasm-bugpoint.cc
+++ b/src/tools/wasm-bugpoint.cc
@@ -1,0 +1,139 @@
+/*
+ * Copyright 2019 WebAssembly Community Group participants
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


aardappel:
I think this would be cool to have. Are there any test cases, examples, or documentation that can be added to this PR?

nlewycky:
> I think this would be cool to have. Are there any test cases, examples, or documentation that can be added to this PR?

I know I need to add them to this PR before it would be approved to land, I just wanted to start the review process early. The example I've been using for testing is
```
$ cat x.sh
#! /bin/sh
! /home/nick/wabt/build/wasm2wat "$@" | grep -q 'f64.convert_i64_u'
$ time wabt/build/wasm-bugpoint x.sh cowsay.wasm
Reduced testcase written out to wasm-bugpoint.wasm

real    0m3.014s
user    0m3.005s
sys     0m0.084s
$ ls -l cowsay.wasm wasm-bugpoint.wasm
-rw-r--r-- 1 nick nick 164744 Sep 30 12:19 wasm-bugpoint.wasm
-rwxr-xr-x 1 nick nick 818623 Jul 22 17:32 cowsay.wasm
```

I plan to move the `Bisect` and `BisectWithFixedLinearRange` classes to their own header out of `tools/` so that I can write gtest tests for them. I'm not certain what to do about tests for `TryRemovingBlocksFromFunction` and `TryRemovingBodyFromFunctions`, those don't seem like utilities that should be shared but I still want tests for them. Maybe integration tests, ensuring the whole thing works, instead of unit tests? Also a man page.

I'll need the wabt developers to point out portability problems and help with some API issues. It would be nice if I could copy a module without going serializing it and deserializing it, especially if we could preserve a mapping from pointer to object in the old module to a pointer to the copy of it in the copy. Calling `WriteBinaryModule` on a `FileStream` fails because `FileWriter::MoveData` is not implemented, and I don't know what that method is supposed to do. I have a workaround for now.

keithw:
Closing as stale. :-(

PR #1179, opened 2019-10-03
wasm-objdump: Fix type signature for multi-value result types

ohorn:
Previously only the first result type was printed, even for multiple results.

This change prints multiple result types in parenthesized form whereas single result types are printed as before.

The `func-result-multi.txt` test case is modified to cover the Type section as well.

### DIFF
--- a/src/binary-reader-objdump.cc
+++ b/src/binary-reader-objdump.cc
@@ -1029,10 +1029,23 @@ Result BinaryReaderObjdump::OnType(Index index,
     printf("%s", GetTypeName(param_types[i]));
   }
   printf(") -> ");
-  if (result_count) {
-    printf("%s", GetTypeName(result_types[0]));
-  } else {
-    printf("nil");
+  switch (result_count) {
+    case 0:
+      printf("nil");
+      break;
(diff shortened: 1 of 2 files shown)


src/binary-reader-objdump.cc
```diff
@@ -1029,10 +1029,20 @@ Result BinaryReaderObjdump::OnType(Index index,
...
-    printf("nil");
+  switch (result_count) {
+    case 0:
+      printf("nil");
+      break;
+    case 1:
+      printf("%s", GetTypeName(result_types[0]));
+      break;
+    default:
+      printf("(%s", GetTypeName(result_types[0]));
```
aardappel: move this statement into the loop? check for `,` with `if (i) printf(", ");` ?

ohorn: What about the parenthesis? Since I need a print statement for `(` anyway, I combined it with the first element.

ohorn: I've modified the loop to make it more standard.

aardappel:
Thanks!

PR #1184, opened 2019-10-12
Fix unused lambda capture warning

sbc100:
Also, inline two trivial methods.


sbc100:
Should we add -DWERROR to the CI builds to prevent warning creeping in?

PR #1191, opened 2019-10-18
wasm-decompile: Refactored code to first build an AST.

aardappel:
This will pave the way for a better multi-pass analysis that
can collect information for the final output pass.

### DIFF
--- a/src/decompiler-ast.inl
+++ b/src/decompiler-ast.inl
@@ -1,0 +1,258 @@
+/*
+ * Copyright 2016 WebAssembly Community Group participants
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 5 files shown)


src/decompiler.cc
```diff
@@ -16,40 +16,17 @@
...
 
 #define WABT_TRACING 0
 #include "src/tracing.h"
 
 namespace wabt {
 
-namespace {
-
-struct Decompiler : ModuleContext {
+struct Decompiler {
```
sbc100: Why get rid of the anonymous namespace here?

aardappel: Since the code here and in `decompiler-ast.inl` need to be in the same namespace. I suppose wit would be possible to retain it here, but since I had to remove it in the other it seemed logical to not have it at all.

src/decompiler-ast.inl
```diff
@@ -0,0 +1,258 @@
+/*
+ * Copyright 2016 WebAssembly Community Group participants
```
jgravelle-google: nit: copyright year

src/decompiler.cc
```diff
@@ -16,54 +16,40 @@
...
-
-#include "src/cast.h"
-#include "src/common.h"
-#include "src/expr-visitor.h"
-#include "src/ir.h"
-#include "src/ir-util.h"
-#include "src/literal.h"
-#include "src/generate-names.h"
-#include "src/stream.h"
+#include "src/decompiler-ast.inl"
```
jgravelle-google: Why `.inl`? Just to move the AST out-of-line but still build it as an object file?

aardappel: Because this code is entirely private to the decompiler, putting it in an object file means pulling out an interface to it, which is of no use to anyone, etc.

PR #1192, opened 2019-10-18
wasm-decompile: improved Return & Drop.

aardappel:


aardappel:
Yes, I hope to add tests soon, once things settle down a bit.

Issue #1187, opened 2019-10-15
Shadowing elem segment names encodes incorrectly on macOS


alexcrichton:
I've been playing around with a custom wat parser and was using wabt as a reference test suite, and I think I came across some curious behavior!

The `tests/wabt/test/parse/expr/bulk-memory-named.txt` test ends up encoding differently on Linux and macOS. I think the Linux encoding is correct, and on macOS I've reduced the test case and am seeing:

```wat
(module                                                 
  (table 1 anyfunc)                                     
  (elem $elem funcref (ref.func 0))                     
  (elem $elem funcref 0)                                
  (func                                                 
    i32.const 0 i32.const 0 i32.const 0 table.init $elem
  )                                                     
)                                                       
```

```
$ wat2wasm foo.wat -o foo.wasm --enable-bulk-memory
$ wasm-objdump foo.wasm -d | grep table.init
 000035: fc 0c 00 00                | table.init 0 0
```

whereas on linux that prints:

```
$ wat2wasm foo.wat -o foo.wasm --enable-bulk-memory
$ wasm-objdump foo.wasm -d | grep table.init
 000035: fc 0c 01 00                | table.init 1 0
```

I think the test is a bit odd, but it looks like the second `$elem` is shadowing the first, but that seems to only work on Linux whereas on macOS the shadowing isn't working?

sbc100:
Obvious conclusion: Its time to switch to linux and be happy!

(seriously that looks weird)

sbc100:
I managed to reproduce this with a mac and linux machine.

I guessing we should not allow two `elem` segments with the same name.  For example trying to give two types of functions the same name results in: e.g.:

```foo.wat:16:4: error: redefinition of function "$t"
  (func $t (type $t)
   ^^^^
```

Would turning this into an error be a good enough fix for you?

alexcrichton:
Ah yeah this just came up in testing of a tool I've got, I don't mind whether it's an error or allowed, but yeah matching existing behavior for other items seems fine

PR #1193, opened 2019-10-18
Check for redefinitions of elem segments

sbc100:
Fixes: #1187


sbc100:
It looks like the current spec expects `(elem $foo)` to name a table but with bulk memory `(elem $foo)` is introducing a new name for the segment.   So as it stands perhaps we need to handle parsing differently when bulk memory is enabled?

sbc100:
I fixed the existing third_party/testsuite/elem.wast test by basically parsing the elem section differently when bulk memory is enabled.

aardappel:
lgtm.. I guess at some future date we can remove this flag.

PR #1196, opened 2019-10-22
Add option to wasm2wat for using names from custom linking section

TheFakeMontyOnTheRun:
This change introduces an option in wasm2wat for using the linking custom section for providing names for the functions present in the module. This was inspired by observing what wasm-objdump does and replicating it.

You might notice I've duplicated the code in `BinaryReaderIR::OnFunctionSymbol` from `BinaryReaderIR::OnFunctionName` - thats because while, on the surface level, both do the same thing, I expect this to change, as the handling of the flags might be expanded.

I was just "scratching my own itch", but this seems like something useful for other people as well (it surely makes it easier to read the generated WAT).

### DIFF
--- a/src/binary-reader-ir.cc
+++ b/src/binary-reader-ir.cc
@@ -222,6 +222,10 @@ class BinaryReaderIR : public BinaryReaderNop {
 
   Result OnModuleName(string_view module_name) override;
   Result OnFunctionNamesCount(Index num_functions) override;
+  Result OnFunctionSymbol(Index index,
+                          uint32_t flags,
+                          string_view name,
+                          Index function_index) override;
   Result OnFunctionName(Index function_index,
                         string_view function_name) override;
   Result OnLocalNameLocalCount(Index function_index, Index num_locals) override;
@@ -1107,6 +1111,28 @@ Result BinaryReaderIR::OnModuleName(string_view name) {
   }
 
   module_->name = MakeDollarName(name);
+  return Result::Ok;
+}
+
+
+Result BinaryReaderIR::OnFunctionSymbol(Index index,
+                                        uint32_t flags,
+                                        string_view name,
+                                        Index function_index) {
+  if ((flags & WABT_SYMBOL_FLAG_UNDEFINED) == 0 ||
+      (flags & WABT_SYMBOL_FLAG_EXPLICIT_NAME) != 0) {
+
+    if (name.empty()) {
+      return Result::Ok;
+    }
+
+    Func* func = module_->funcs[function_index];
+    std::string dollar_name =
+      GetUniqueName(&module_->func_bindings, MakeDollarName(name));
+    func->name = dollar_name;
+    module_->func_bindings.emplace(dollar_name, Binding(function_index));
+  }
+
   return Result::Ok;
 }
 
--- a/src/binary-reader-ir.cc
+++ b/src/binary-reader-ir.cc
@@ -1126,11 +1126,11 @@ Result BinaryReaderIR::OnFunctionSymbol(Index index,
       return Result::Ok;
     }
 
+    std::string func_name = name.to_string();
     Func* func = module_->funcs[function_index];
-    std::string dollar_name =
-      GetUniqueName(&module_->func_bindings, MakeDollarName(name));
-    func->name = dollar_name;
-    module_->func_bindings.emplace(dollar_name, Binding(function_index));
+    GetUniqueName(&module_->func_bindings, func_name);
+    func->name = func_name;
+    module_->func_bindings.emplace(func_name, Binding(function_index));
   }
 
   return Result::Ok;
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -31,6 +31,7 @@
 #include "src/leb128.h"
 #include "src/stream.h"
 #include "src/utf8.h"
+#include "src/generate-names.h"
 
 #if HAVE_ALLOCA
 #include <alloca.h>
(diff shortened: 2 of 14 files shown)


src/tools/wasm2wat.cc
```diff
@@ -41,6 +41,7 @@ static Features s_features;
 static WriteWatOptions s_write_wat_options;
 static bool s_generate_names;
 static bool s_read_debug_names = true;
+static bool s_read_linking_names = false;
```
aardappel: Can you add this option on `wasm-decompile` too?

src/binary-reader-ir.cc
```diff
@@ -1110,6 +1114,28 @@ Result BinaryReaderIR::OnModuleName(string_view name) {
...
+  if ((flags & WABT_SYMBOL_FLAG_UNDEFINED) == 0 ||
+      (flags & WABT_SYMBOL_FLAG_EXPLICIT_NAME) != 0) {
+
+    if (name.empty()) {
+      return Result::Ok;
+    }
+
+    Func* func = module_->funcs[function_index];
+    std::string dollar_name =
+      GetUniqueName(&module_->func_bindings, MakeDollarName(name));
```
aardappel: If this is forcing a `$` on the name, then that isn't great, `NameGenerator::GenerateName` is generally doing that optionally.

src/tools/wasm2wat.cc
TheFakeMontyOnTheRun: Gladly!
Any other tool that would be interesting to have it as well?

src/binary-reader-ir.cc
TheFakeMontyOnTheRun: NameGenerator::GenerateName  seems to have a lot more going on (as well as being private). Maybe I could add another function that checks for $ being already present and that's more streamlined to what wasm2wat (and wasm-decompile) needs?

src/tools/wasm2wat.cc
TheFakeMontyOnTheRun: Turns out I actually missed adding the 6th option to ReadBinaryOptions in wasm-decompiler.cc, so I would have to modify it anyway. From my search here, I couldn't find any other file I would need to change.

src/binary-reader-ir.cc
TheFakeMontyOnTheRun: (so, it seems that not generating names with $ will be useful for wasm-decompile. I will take this into consideration)

TheFakeMontyOnTheRun:
I'm a tabs-guy, so I strived very hard to keep coding standards consistent, but might have missed (or misunderstood) some parts. If so, please, let me know.

src/binary-reader.cc
```diff
@@ -1712,7 +1713,17 @@ Result BinaryReader::ReadLinkingSection(Offset section_size) {
                 CHECK_RESULT(ReadStr(&name, "symbol name"));
               switch (sym_type) {
                 case SymbolType::Function:
-                  CALLBACK(OnFunctionSymbol, i, flags, name, index);
+                  if (options_.read_linking_names != LinkingNameStrategy::DONT_USE_LINKING_NAMES) {
+                    bool prepend_dollar = (options_.read_linking_names == LinkingNameStrategy::USE_LINKING_NAMES_WITH_DOLLAR);
+
```
aardappel: No need for all these empty lines between statements, see surrounding code.

src/generate-names.cc
```diff
@@ -429,4 +430,23 @@ Result GenerateNames(Module* module, NameOpts opts) {
   return generator.VisitModule(module);
 }
 
+std::string GenerateNameWithDollar( const std::string& func_name, Index function_index) {
```
aardappel: This function seems to duplicate functionality from `NameGenerator::GenerateName` ? Also, whitespace. Also, weird to for the start with a `$` then have to remove it again in one of the cases, that could be done more elegantly.

aardappel:
@binji maybe you can have a quick look, not quite sure if this is the most elegant way to achieve this functionality.

src/generate-names.cc
TheFakeMontyOnTheRun: NameGenerator::GenerateName requires a lot of gymnastics to be used as it is and would still generate some extra information into the names ("print_and_compute" would become "print_and_compute_b" in my tests). What I could do is to extract the common functionality from ```NameGenerator::GenerateName``` and delegate it to ```GenerateNameWithDollar```. What do you think?

src/binary-reader.cc
TheFakeMontyOnTheRun: I was trying to make it easier to review, but can trim it if it's more desirable :+1:

src/generate-names.cc
aardappel: I'm not sure, hope at some point @binji  will chime in.

src/binary-reader.cc
aardappel: GitHub diffs are pretty easy to read even without extra blank lines.

src/generate-names.cc
binji: Hm, I'm not sure why it would be generating a different name unless that name already exists. In that case it would disambiguate with a number, not a letter. In any case, yeah, it'd be nicer to reuse the code in `GenerateName` if possible.

binji:
Sorry for the delay!

I think we should be able to use the linking section names, but I'd rather this not touch all the tools. It seems like the default should be to not use the linking section names. Similarly, I'm not sure I see why we should ever allow pure linking section names. The wat format requires names to have a leading `$`, so we shouldn't produce output without it.

TheFakeMontyOnTheRun:
To be honest, I don't need this feature anymore (not sure about other people). The proof-of-concept I had (using WebAssembly to generate BASIC code and legally run it on the Nintendo 3DS) worked, but I don't feel like pursuing it anymore.

I can withdraw this Pull Request and would gladly help in other fronts, if you guys point me towards what would be desirable.

aardappel:
This PR likely has some overlap with this one: https://github.com/WebAssembly/wabt/pull/1318

TheFakeMontyOnTheRun:
Yep - most likely. I would re-do it anyway. Thanks for the code reviews.

PR #1201, opened 2019-10-26
Fix warning in test-hexfloat.cc

### LINKED ISSUE #1200: implicit-int-float-conversion warnings in test-hexfloat.cc
With:

```
$ clang -v
clang version 10.0.0 (https://github.com/llvm/llvm-project/ 64a362e7216a43e3ad44e50a89265e72aeb14294)
Target: x86_64-unknown-linux-gnu
Thread model: posix
```

I'm seeing:

```
test-hexfloat.cc:109:7: warning: implicit conversion from 'unsigned int' to 'float' changes value from 4294967295 to 4294967296 [-Wimplicit-int-float-conversion]
      LOG_COMPLETION(bits);
      ^~~~~~~~~~~~~~~~~~~~
```

binji:
A `float` cant fit all the values of a `uint32_t` with full precision, since they are the same size. But a `double` has a 52-bit mantissa so it can represent a `uint32_t` with full precision. This fixes the compiler warning in test-hexfloat.cc.

Fixes #1200.

### DIFF
--- a/src/test-hexfloat.cc
+++ b/src/test-hexfloat.cc
@@ -35,15 +35,15 @@
   for (; bits >= last_bits;  \
        last_bits = bits, bits += num_threads_ * FOREACH_UINT32_MULTIPLIER)
 
-#define LOG_COMPLETION(bits)                                                 \
-  if (shard == 0) {                                                          \
-    int top_byte = bits >> 24;                                               \
-    if (top_byte != last_top_byte) {                                         \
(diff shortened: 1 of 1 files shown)


sbc100:
> A float can't always fit a uint32_t, but a double can.

Can you explain a little more... this seems very counter intuitive.

binji:
Updated

PR #1202, opened 2019-10-26
[wat2wasm] Don't crash on unsupported reloc type

binji:
TypeIndexLEB relocations (used by call_indirect) doesn't seem to be
supported.

### DIFF
--- a/test/regress/regress-35.txt
+++ b/test/regress/regress-35.txt
@@ -21,5 +21,5 @@
   (export "run" (func $run))
 )
 (;; STDERR ;;;
-warning: Unsupported relocation type: R_WASM_TYPE_INDEX_LEB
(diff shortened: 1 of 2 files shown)


test/regress/regress-35.txt
```diff
@@ -0,0 +1,25 @@
...
+    i32.add
+    return
+  )
+  (table 1 anyfunc)
+  (elem (i32.const 0) $add)
+  (func $run
+    i32.const 4444
+    i32.const 5555
+    i32.const 0
+    ;;call_indirect (type 0) ;; fails
```
sbc100: Should we also update the `test/dump/relocations.txt` to include this relocation type?    Or is that part of the TODO?

binji: Yeah, this was just to keep it from crashing. Actually supporting it is more work.

PR #1203, opened 2019-10-26
[wat2wasm] Passive segments require bulk memory

binji:


test/regress/regress-36.txt
```diff
@@ -0,0 +1,8 @@
+;;; TOOL: wat2wasm
+;;; ERROR: 1
+(data "hello")
+(;; STDERR ;;;
+out/test/regress/regress-36.txt:3:2: error: passive data segments are not allowed
```
sbc100: Do you need to add new regress test when you also updates the `test/parse` unit tests?  Seems like its redundant.

binji: Those tests weren't testing for the absence of the flag, though. I've added the flag so they are clearer.

sbc100: But doesn't `test/parse/expr/bulk-memory-disabled.txt` cover this now?  Why add this extra regress test?

PR #1204, opened 2019-11-05
Remove some used classes/members from the interpreter

sbc100:
This change also makes it easier to add an alternative import
resolver that is needed for implementing the C API.


sbc100:
Nope. Not used.

PR #1205, opened 2019-11-05
wasm-decompile: reworked how "stacky" code gets decompiled.

aardappel:
For example: multi-value, and void exps while there are non-void
exps on the stack.

It now uses temp variables instead of pseudo push/pop, as the latter
weren't particularly readable and had an ordering problem that was
hard to make intuitive.

The new system covers all possible situations, generates as few
variables as possible, has clearer comments, and tests.

### DIFF
--- a/src/decompiler-ast.inl
+++ b/src/decompiler-ast.inl
@@ -40,8 +40,8 @@
 namespace wabt {
 
 enum class NodeType {
-  PushAll,
-  Pop,
+  FlushToVars,
+  FlushedVar,
   Statements,
   EndReturn,
   Decl,
@@ -59,6 +59,7 @@ struct Node {
   union {
     LabelType lt;  // br/br_if target.
     const Var* var;  // Decl/DeclInit.
+    struct { Index var_start, var_count; };  // FlushedVar/FlushToVars
   };
 
   Node()
@@ -89,18 +90,20 @@ struct AST {
     if (f) mc.EndFunc();
   }
 
-  void NewNode(NodeType ntype, ExprType etype, const Expr* e, Index nargs) {
-    assert(stack.size() >= nargs);
+  // Create a new node, take nargs existing nodes on the exp stack as children.
+  Node& InsertNode(NodeType ntype, ExprType etype, const Expr* e, Index nargs) {
+    assert(exp_stack.size() >= nargs);
     Node n { ntype, etype, e, nullptr };
     n.children.reserve(nargs);
-    std::move(stack.end() - nargs, stack.end(),
+    std::move(exp_stack.end() - nargs, exp_stack.end(),
               std::back_inserter(n.children));
-    stack.resize(stack.size() - nargs);
-    stack.push_back(std::move(n));
+    exp_stack.erase(exp_stack.end() - nargs, exp_stack.end());
+    exp_stack.push_back(std::move(n));
+    return exp_stack.back();
   }
 
   template<ExprType T> void PreDecl(const VarExpr<T>& ve) {
-    stack.emplace(stack.begin() + pre_decl_insertion_point++,
+    exp_stack.emplace(exp_stack.begin() + pre_decl_insertion_point++,
                   NodeType::Decl, ExprType::Nop, nullptr, &ve.var);
   }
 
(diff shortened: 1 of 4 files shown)


src/decompiler-ast.inl
```diff
@@ -89,18 +90,20 @@ struct AST {
     if (f) mc.EndFunc();
   }
 
-  void NewNode(NodeType ntype, ExprType etype, const Expr* e, Index nargs) {
-    assert(stack.size() >= nargs);
+  Node& NewNode(NodeType ntype, ExprType etype, const Expr* e, Index nargs,
```
jgravelle-google: Oo that got weird. What is dip? (It seems to be an "n from the end" offset, but that's not completely obvious). This returning a reference to the newly inserted node feels convenient, but odd.
Partly the naming of "NewNode" makes it feel extra odd to return a value that is already in-place. I'd expect something more like `exp_stack.push_back(NewNode(...));`. It having returned void in the past made it more clear that that wasn't how "new" was being used, but now seeing `NewNode(...).foo = ...;` looks surprising.

src/decompiler-ast.inl
```diff
@@ -204,41 +206,102 @@ struct AST {
     NewNode(NodeType::Expr, e.type(), &e, arity.nargs);
   }
 
+  void FlushVars(int nargs, int num_vars, int dip, int num_vals) {
```
jgravelle-google: Odd that dip was a default value before, but is now always included and not the last argument.

src/decompiler-ast.inl
```diff
@@ -204,41 +206,102 @@ struct AST {
...
+        // We have a void item on top of the exp_stack, so we must "lift" the
+        // previous values around it.
+        // We track what part of the stack is in variables to avoid doing
+        // this unnecessarily.
+        if (num_vals > 0) {
+          // We have actual new values that need lifting.
+          // This puts the part of the stack that wasn't already a variable
+          // (before the current void exp) into a FlushToVars.
+          // Now re-generate these values after the void exp as vars.
+          FlushVars(num_vals, num_vars, 1, num_vals);
```
jgravelle-google: As far as I can tell this is the only use of non-0 dip? That feels like a disproportionate amount of complexity/machinery for something with one use. Is dip the right abstraction going forward, or should we special-case this flush?

src/decompiler-ast.inl
aardappel: `dip` in stack languages is exactly that: do an operation as if N from the top of the stack is the actual top of the stack :)

I'll rename it `InsertNode` which is closer to what it does. What it actually does is `InsertNewNodeAndTakeExistingNodesOnTheStackAsChildren` but I can put that in a comment.

src/decompiler-ast.inl
aardappel: To me, default values should preferrably not be used, unless the function in question is called many times and the vast majority of callers should want the default. That is the case for `InsertNode` and not `FlushVars`.

src/decompiler-ast.inl
aardappel: I could remove the `dip` arg by making only that case pop the stack..

aardappel:
Note: CI error appears unrelated to this PR.

src/decompiler-ast.inl
```diff
@@ -187,81 +189,160 @@ struct AST {
...
+      if (arity.nreturns != 1) {
+        auto num_vars = value_stack_in_variables - value_stack_depth_start;
+        auto num_vals = value_stack_depth - value_stack_in_variables;
+        auto GenFlushVars = [&](int nargs) {
+          auto& ftv = InsertNode(NodeType::FlushToVars, ExprType::Nop, nullptr,
+                                 nargs);
+          ftv.var_start = flushed_vars;
+          ftv.var_count = num_vals;
+        };
+        auto MoveStatementsBelowVars = [&](size_t amount) {
```
jgravelle-google: I appreciate that this is a lambda that describes what it's doing, rather than a std::rotate inline :+1:

src/decompiler-ast.inl
```diff
@@ -187,81 +189,160 @@ struct AST {
...
+          std::rotate(exp_stack.end() - num_vars - amount,
+                      exp_stack.end() - amount,
+                      exp_stack.end());
+        };
+        auto GenFlushedVars = [&]() {
+          // Re-generate these values as vars.
+          for (int i = 0; i < num_vals; i++) {
+            auto& fv = InsertNode(NodeType::FlushedVar, ExprType::Nop, nullptr,
+                                  0);
```
jgravelle-google: I assume 80 chars? Shame that this looks so dorky

src/decompiler-ast.inl
```diff
@@ -187,81 +189,160 @@ struct AST {
...
-        // All values become pops.
-        for (Index i = 0; i < arity.nreturns; i++)
-          NewNode(NodeType::Pop, ExprType::Nop, nullptr, 0);
-        // TODO: can also implement a push_all_but_one that returns the top,
-        // then insert N-1 pops below it. Or have a function that returns N
-        // values direcly to N arguments for when structs are passed on,
-        // etc.
+      assert(unreachable || value_stack_depth >= value_stack_depth_start);
+      value_stack_depth += arity.nreturns;
+      // We maintain the invariant that a value_stack_depth of N is represented
```
jgravelle-google: The clarity of comments overall here is really nice

src/decompiler-ast.inl
```diff
@@ -187,81 +189,160 @@ struct AST {
...
+          // Number of other non-variable values that may be present:
+          assert(num_vals >= static_cast<int>(arity.nreturns));
+          // Flush multi-value exp + others.
+          GenFlushVars(num_vals - arity.nreturns + 1);
+          // Put this flush statement before any existing variables.
+          MoveStatementsBelowVars(1);
+          GenFlushedVars();
+          value_stack_in_variables = value_stack_depth;
+        }
+      } else {
```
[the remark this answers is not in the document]
aardappel: yeah, kinda want this ordering on purpose, since it puts the core of the algo first

PR #1206, opened 2019-11-05
Implement initial reftype support in interpreter.

sbc100:
Initial implementation of reftype proposal in the interpreter.
    
- Implement ref.func everywhere.
- Implement table.get and table.set in the interpreter.


alexcrichton:
Sorry I don't have any experience with the interpreter here, so I'm probably not much help on review :(

sbc100:
I got carried away an implemented more of this.   I'm not sure how best to test it other then cobbling together some local testrs, since the spec tests upstream don't yet contain the reftypes tests.

Is there some mechanism to pull in the test suites from the proposals repos?

src/interp/interp.h
```diff
@@ -249,8 +259,6 @@ struct EventImport : Import {
       : Import(ExternalKind::Event, module_name, field_name) {}
 };
 
-struct Func;
```
jgravelle-google: I'm sure this made sense at the time

src/interp/interp.cc
```diff
@@ -1056,25 +1097,30 @@ Result Thread::ElemDrop(const uint8_t** pc) {
...
   // When copying backward, if the range is out-of-bounds, then no data will be
   // written.
   if (ok || !copy_backward) {
-    ok &= ClampToBounds(src, &size, table_size);
+    ok &= ClampToBounds(src, &size, table->size());
     if (size > 0) {
-      Index* data = table->func_indexes.data();
-      memmove(data + dst, data + src, size * sizeof(Index));
+      Ref* data = table->entries.data();
+      memmove(data + dst, data + src, size * sizeof(table->entries[0]));
```
jgravelle-google: Why not `sizeof(Ref)`?

src/interp/interp.cc
```diff
@@ -1397,6 +1443,10 @@ ValueTypeRep<R> IntEqz(ValueTypeRep<T> v_rep) {
   return ToRep(v_rep == 0);
 }
 
+ValueTypeRep<uint32_t> RefIsNull(ValueTypeRep<Ref> v_rep) {
+  return ToRep(v_rep.index == kInvalidIndex);
```
jgravelle-google: Are null and invalid distinct? Probably not so I think this is fine (in that, uninitialized == null, so there's no other state)

Makefile
```diff
@@ -28,14 +28,15 @@ CMAKE_CMD ?= cmake
 
 DEFAULT_SUFFIX = clang-debug
 
-COMPILERS := GCC GCC_I686 GCC_FUZZ CLANG EMCC
+COMPILERS := GCC GCC_I686 GCC_FUZZ CLANG CLANG_I686 EMCC
```
jgravelle-google: Seems unrelated? (I'm fine with it but it seems not-quite-small to my eye)

src/interp/interp.cc
sbc100: Done.  I was trying to make the code robust against the type of the table entries changing .. perhaps `sizeof(data[0])` would be more clear?

src/interp/interp.cc
sbc100: For now yes.  We could revisit this later.

Makefile
sbc100: Reverted.

src/interp/interp.h
sbc100: Removed

PR #1207, opened 2019-11-06
Rename v128_bits to simply vec128 since this is stored natively

sbc100:
The _bits suffix is used the floating point types where we use types
that differ from the native C types.  For v128 we have a native struct
type that we store and manipulate directly.

Also remove the unneeded bitcast operations.


sbc100:
ping..

PR #1208, opened 2019-11-06
Add CMAKE_EXPORT_COMPILE_COMMANDS to CMakeLists.txt

sbc100:
This allows code completion tools such as clangd to work.  I'm not sure
if there is any downside to enabling this in all cases?


sbc100:
ping

PR #1213, opened 2019-11-08
wasm-decompile: Fixed arbitrary characters appearing in function names.

aardappel:
In particular, LLD may stick entire C++ signatures in names, that
WABT then uses with wasm2wat, but are not appropriate for wasm-decompile
function names.

### DIFF
--- a/src/generate-names.h
+++ b/src/generate-names.h
@@ -26,7 +26,8 @@ struct Module;
 enum NameOpts {
   None = 0,
   AlphaNames = 1 << 0,
-  NoDollar = 1 << 0,
+  NoDollar = 1 << 1,
+  OnlyAlphaNum = 1 << 2,
 };
 
 Result GenerateNames(struct Module*, NameOpts opts = NameOpts::None);
--- a/test/decompile/basic.txt
+++ b/test/decompile/basic.txt
@@ -41,6 +41,10 @@
     end
     i32.const 1
   )
+  ;; LLD outputs a name section with de-mangled C++ function signatures as names,
+  ;; so have to make sure special chars get removed.
+  (func (export "signature-&<>()") (param) (result))
+  (func (export "signature-&[]()") (param) (result))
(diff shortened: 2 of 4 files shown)


src/generate-names.h
```diff
@@ -26,7 +26,8 @@ struct Module;
 enum NameOpts {
   None = 0,
   AlphaNames = 1 << 0,
-  NoDollar = 1 << 0,
+  NoDollar = 1 << 1,
```
jgravelle-google: (wait how was this working)
(oh I guess the two were always set together and this was essentially "is wasm-decompile")

src/generate-names.h
```diff
@@ -26,7 +26,8 @@ struct Module;
 enum NameOpts {
```
jgravelle-google: Nit: if these are always fully qualified, would we want to make this an enum class to enforce that? (and get things out of the global namespace)

test/decompile/basic.txt
```diff
@@ -41,6 +41,9 @@
     end
     i32.const 1
   )
+  ;; LLD outputs a name section with de-mangled C++ function signatures as names,
+  ;; so have to make sure special chars get removed.
+  (func (export "signature-&<>()") (param) (result))
```
jgravelle-google: I'd add another test case of like `export "signature[]{}+/"` or something that would collide with the rewritten `signature______` identifier to check deduping. And if that doesn't work yet leave a note about it

src/generate-names.h
aardappel: `enum class` is a pain with bit flags, as you need to define a custom `operator|` with lots of casts to make it work.

test/decompile/basic.txt
aardappel: We're not deduping yet. Let me add a FIXME

aardappel: actually we are, doh.

PR #1215, opened 2019-11-08
interpreter: Move `env` member up to Module base class.

sbc100:
This is needed as part of implementing the C API on top of the
interpreter.

Also remove the trailing `_` to match other struct members in this
file.

Also remove unneeded forward decls.


sbc100:
This change is one of several I'm splitting off my interpreter PR.

PR #1218, opened 2019-11-13
reference-types: Add multi-table support to the interpreter

sbc100:
This involved plumbing the table argument for `table.init` and
`table.copy` all the way through rather than assuming they were
always zero.

Update the rather complex logic for assigning names to elem segments to
handle active elem segments which both name themselves and the table
they apply too.


src/binary-reader-nop.h
```diff
@@ -252,9 +252,13 @@ class BinaryReaderNop : public BinaryReaderDelegate {
   Result OnMemoryGrowExpr() override { return Result::Ok; }
   Result OnMemoryInitExpr(Index segment_index) override { return Result::Ok; }
   Result OnMemorySizeExpr() override { return Result::Ok; }
-  Result OnTableCopyExpr() override { return Result::Ok; }
+  Result OnTableCopyExpr(Index dst_index, Index src_index) override {
```
aardappel: This is formatted differently from the surrounding code?

sbc100: I think its just because it overflowed 80 chars.  I ran clang format and this stayed the same (but other things I missed, so thanks!)

PR #1231, opened 2019-11-18
Add support for atomic.fence from the threads proposal

wingo:
See https://github.com/WebAssembly/threads/pull/141 for the binary encoding.  This patch does add a field to `AtomicFenceExpr` for the consistency model, though without a type for the time being.


src/token.def
```diff
@@ -134,7 +135,7 @@ WABT_TOKEN(Throw, "throw")
 WABT_TOKEN(Try, "try")
 WABT_TOKEN(Unary, "UNARY")
 WABT_TOKEN(Unreachable, "unreachable")
-WABT_TOKEN_FIRST(Opcode, AtomicLoad)
```
sbc100: Was AtomicLoad not supposed to be here at all?

src/type-checker.cc
```diff
@@ -366,6 +366,15 @@ Result TypeChecker::OnAtomicWait(Opcode opcode) {
...
 }
 
+Result TypeChecker::OnAtomicFence(uint32_t consistency_model) {
+  if (consistency_model != 0x0) {
+    PrintError("bad atomic.fence consistency model: %u",
+               consistency_model);
+    return Result::Error;
+  }
+  return Result::Ok;
+}
```
sbc100: Is the type checker the right place for this?  i.e. is this a more general validation thing than a type checking thing?

src/token.def
wingo: Mmm, I think this was a bad change that snuck in.  I will make sure it's not there in the revised patch and regenerate the lexer.

PR #1237, opened 2019-11-19
Update spec testsuite

sbc100:
The only major change to the interpreter is to move segment
initialization out `ReadBinaryInterp` (in the binary reader) and into
interp.cc.  This is because the test suite now expects out of bound
semgments to be reported during initialization rather than reported
as validation errors.

Fixes: #1223

### DIFF
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -99,6 +99,8 @@ class BinaryReader {
   Result ReadS32Leb128(uint32_t* out_value, const char* desc) WABT_WARN_UNUSED;
   Result ReadS64Leb128(uint64_t* out_value, const char* desc) WABT_WARN_UNUSED;
   Result ReadType(Type* out_value, const char* desc) WABT_WARN_UNUSED;
+  Result ReadExternalKind(ExternalKind* out_value,
+                          const char* desc) WABT_WARN_UNUSED;
   Result ReadStr(string_view* out_str, const char* desc) WABT_WARN_UNUSED;
   Result ReadBytes(const void** out_data,
                    Address* out_data_size,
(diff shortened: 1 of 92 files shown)


sbc100:
I've tried to land a lot of parts of this change upfront, and I can continue to split it up more if you like, but its fairly minimal at this point in terms of the changes need to pass the new set of tests.

src/binary-reader.cc
```diff
@@ -99,6 +99,7 @@ class BinaryReader {
   Result ReadS32Leb128(uint32_t* out_value, const char* desc) WABT_WARN_UNUSED;
   Result ReadS64Leb128(uint64_t* out_value, const char* desc) WABT_WARN_UNUSED;
   Result ReadType(Type* out_value, const char* desc) WABT_WARN_UNUSED;
+  Result ReadExtenKind(ExternalKind* out_value, const char* desc) WABT_WARN_UNUSED;
```
jgravelle-google: Typo: ExternalKind, or ExternKind

src/binary-reader.cc
```diff
@@ -2157,12 +2154,30 @@ Result BinaryReader::ReadElemSection(Offset section_size) {
       CALLBACK(EndElemSegmentInitExpr, i);
     }
 
+    // For backwards compat we support not declarign the element kind.
```
jgravelle-google: Typo: s/declarign/declaring/

src/binary-writer.cc
```diff
@@ -1060,11 +1071,11 @@ Result BinaryWriter::WriteModule() {
           WriteOpcode(stream_, Opcode::End);
         }
       } else {
-        // Active segment.
+
```
jgravelle-google: Nit: Extra newline?

PR #1250, opened 2019-11-22
Initial WASM C API implementation.

sbc100:


sbc100:
I think all the precursor PRs have now landed and this code is ready for review.

sbc100:
PTAL

sbc100:
If you are reviewing this then I recommend doing it in two phase:

1. First, everything outside of interp-wasm-c-api.cc.  This should be relatively simple.
2. This requires some understanding of the C-API.  However since the tests pass and this is the first pass at implementation you should be able to review this without looking too hard at all my bugs in your choose.

test/run-c-api-examples.py
```diff
@@ -0,0 +1,122 @@
...
+    sys.exit(1)
+
+
+def check_for_missing(found):
+    # Check that all the expected examples are found
+
+    basenames = [os.path.splitext(f)[0] for f in found]
+    for e in ALL_EXAMPLES:
+        if e not in basenames:
+            error('Example binary not found: %s' % e)
```
binji: this will exit on the first error, might be nicer to use set subtraction in both directions so you can get a full list instead.

src/interp/binary-reader-interp.h
```diff
@@ -24,20 +24,32 @@ namespace wabt {
...
+// Read and instantiate a module in the given environment.
 Result ReadBinaryInterp(interp::Environment* env,
                         const void* data,
                         size_t size,
                         const ReadBinaryOptions& options,
                         Errors*,
                         interp::DefinedModule** out_module);
 
+// Read and instantiate a module in the given environment. Similar to above but
+// using using fixed set of exports to resolve the module exports.
```
binji: Fixed set of imports?

src/interp/binary-reader-interp.cc
```diff
@@ -764,16 +771,35 @@ wabt::Result BinaryReaderInterp::FindRegisteredModule(string_view module_name,
...
-               WABT_PRINTF_STRING_VIEW_ARG(field_name));
-    return wabt::Result::Error;
+wabt::Result BinaryReaderInterp::ResolveImport(Index import_index,
+                                               ExternalKind kind,
+                                               string_view module_name,
+                                               string_view field_name,
+                                               Index sig_index,
+                                               Export** out_export) {
+  Export* export_ = nullptr;
+  if (imports_.size()) {
```
binji: I don't really like using size() as a bool; maybe `imports_.size() != 0` or `!imports_.empty()` instead?

src/interp/binary-reader-metadata.cc
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright 2016 WebAssembly Community Group participants
```
binji: 2020, other new files too

src/interp/interp-wasm-c-api.cc
```diff
@@ -0,0 +1,1430 @@
...
+struct wasm_memorytype_t : wasm_externtype_t {
+  wasm_memorytype_t(wasm_limits_t limits)
+      : wasm_externtype_t(WASM_EXTERN_MEMORY), limits(limits) {}
+  wasm_limits_t limits;
+};
+
+struct wasm_importtype_t {};
+
+struct wasm_exporttype_t {
+  ~wasm_exporttype_t() { delete type; }
```
binji: use unique_ptr instead?

src/interp/interp-wasm-c-api.cc
```diff
@@ -0,0 +1,1430 @@
...
+      return "trap";
+    case WasmRefType::Foreign:
+      return "foreign";
+  }
+  WABT_UNREACHABLE;
+}
+
+struct wasm_ref_t {
+  wasm_ref_t(WasmRefType kind) : kind(kind) {}
+  wasm_ref_t(const wasm_ref_t& other)
```
binji: assignment operator too? or delete it

src/interp/interp-wasm-c-api.cc
```diff
@@ -0,0 +1,1430 @@
...
+
+  virtual ~wasm_ref_t() {
+    if (finalizer) {
+      finalizer(host_info);
+    }
+  }
+
+  virtual wasm_ref_t* Copy() const { return new wasm_ref_t(kind); }
+
+  bool Same(const wasm_ref_t& other) const;
```
binji: Not implemented?

src/interp/interp-wasm-c-api.cc
```diff
@@ -0,0 +1,1430 @@
...
+                                FromWabtLimits(t->limits));
+  }
+};
+
+struct wasm_memory_t : wasm_extern_t {
+  wasm_memory_t(std::shared_ptr<WasmInstance> instance, Index index)
+      : wasm_extern_t(instance, ExternalKind::Memory, index) {}
+
+  interp::Memory* GetMemory() const {
+    auto* env = instance.get()->store->env;
```
binji: use Env()?

src/interp/interp-wasm-c-api.cc
```diff
@@ -0,0 +1,1430 @@
...
+    case WASM_ANYREF: {
+      wasm_ref_t* ref = value.of.ref;
+      if (!ref) {
+        out.set_ref(Ref{RefType::Null, kInvalidIndex});
+        break;
+      }
+      if (ref->kind == WasmRefType::Foreign) {
+        auto* f = static_cast<wasm_foreign_t*>(ref);
+        out.set_ref(Ref{RefType::Host, f->index});
+        out.type = Type::Hostref;
```
binji: hm, maybe set_ref should do this for you

src/interp/interp-wasm-c-api.cc
```diff
@@ -0,0 +1,1430 @@
...
+      TRACE("result: %p\n", &host_results[i]);
+      results[i] = ToWabtValue(host_results[i]);
+      TRACE("added result value: %s\n", TypedValueToString(results[i]).c_str());
+    }
+    delete[] host_args;
+    delete[] host_results;
+    return interp::ResultType::Ok;
+  };
+
+  type->Sig();
```
binji: ?

sbc100:
Thanks for the review!

I share you concern regarding memory management.   I am quite surprised that the sanitizers didn't find anything for me actually.. a little suspicious. 

Even if there are leaks here I'm keen to land this initial version and iterate.

test/run-c-api-examples.py
sbc100: Done

sbc100:
Switched to unique_ptr in several places.

binji:
unique_ptr stuff lgtm, thanks!

PR #1251, opened 2019-11-23
Add github actions badge to README.md

sbc100:


Horcrux7:
A link to the badge image does not make much sense. I would expect a link to the result. A mistake?

sbc100:
I added a link to the repo.  Sadly I can't find out how to deep link into the master branch.   Even github actions official repo just linked back to itself: https://github.com/actions/toolkit

Hopefully this is get fixes and we can update with a more useful link later.

sbc100:
Right, but the URL that clicking on the image links to should be something useful.   Currently I don't know how to construct that... but its easy to fix once we get the knowledge.

PR #1255, opened 2019-11-29
Split run-unittests out as seperate target

sbc100:
New `check` target now runs them both.  This allows for github actions to
show unittests and system tests as separate steps.

Also a couple of CMakeLists.txt cleanups:
- Don't use add_definition to add `-fno-exceptions`, this is a C++-only
  flag.
- Lowercase the name of the `sanitizer` function.
- Remove opcode.def from list of library input file.  On windows when
  building a DLL .def files are assumed to be windows DLL .def files,
  which this is not.

This change is split out from #1250

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,7 +122,7 @@ else ()
   endif ()
 
   if (NOT WITH_EXCEPTIONS)
-    add_definitions(-fno-exceptions)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-exceptions")
   endif ()
 
   # Need to define __STDC_*_MACROS because C99 specifies that C++ shouldn't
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -196,22 +196,22 @@ function(SANITIZER NAME FLAGS)
...
+sanitizer(USE_MSAN "-fsanitize=memory")
+sanitizer(USE_LSAN "-fsanitize=leak")
 
 if (USE_UBSAN)
   # -fno-sanitize-recover was deprecated, see if we are compiling with a newer
   # clang that requires -fno-sanitize-recover=all.
   set(UBSAN_BLACKLIST ${WABT_SOURCE_DIR}/ubsan.blacklist)
   include(CheckCXXCompilerFlag)
-  CHECK_CXX_COMPILER_FLAG("-fsanitize=undefined -fno-sanitize-recover -Wall -Werror" HAS_UBSAN_RECOVER_BARE)
+  check_cxx_compiler_flag("-fsanitize=undefined -fno-sanitize-recover -Wall -Werror" HAS_UBSAN_RECOVER_BARE)
```
jgravelle-google: Are these case-insensitive? I assume they are because CI is passing but, huh

sbc100: Yes.  The more modern style is lower case and that is the style used elsewhere in this file.

PR #1258, opened 2019-12-06
wasm-decompile: Load/Store tracking for struct output.

aardappel:
This tries to make code more readable by summarizing patterns of
load/store ops into "struct" declarations.

Initial version, can probably be improved, but has all essentials
of the idea in place.

### DIFF
--- a/src/decompiler-ls.h
+++ b/src/decompiler-ls.h
@@ -1,0 +1,218 @@
+/*
+ * Copyright 2019 WebAssembly Community Group participants
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef WABT_DECOMPILER_LS_H_
+#define WABT_DECOMPILER_LS_H_
+
+#include "src/decompiler-ast.h"
+
+#include <map>
+
+namespace wabt {
+
+// Names starting with "u" are unsigned, the rest are "signed or doesn't matter"
+inline const char *GetDecompTypeName(Type t) {
+  switch (t) {
+    case Type::I8: return "byte";
+    case Type::I8U: return "ubyte";
+    case Type::I16: return "short";
+    case Type::I16U: return "ushort";
+    case Type::I32: return "int";
+    case Type::I32U: return "uint";
+    case Type::I64: return "long";
+    case Type::F32: return "float";
+    case Type::F64: return "double";
+    case Type::V128: return "simd";
+    case Type::Anyref: return "anyref";
+    case Type::Func: return "func";
+    case Type::Funcref: return "funcref";
+    case Type::Exnref: return "exceptionref";
+    case Type::Void: return "void";
+    default: return "ILLEGAL";
+  }
+}
+
+inline Type GetMemoryType(Type operand_type, Opcode opc) {
+  // TODO: something something SIMD.
+  // TODO: this loses information of the type it is read into.
+  // That may well not be the biggest deal since that is usually obvious
+  // from context, if not, we should probably represent that as a cast around
+  // the access, since it should not be part of the field type.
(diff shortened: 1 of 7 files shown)


src/decompiler.cc
```diff
@@ -540,13 +529,19 @@ struct Decompiler {
       auto is_import =
           CheckImportExport(ExternalKind::Func, func_index, f->name);
       AST ast(mc, f);
-      if (!is_import) ast.Construct(f->exprs, f->GetNumResults(), true);
+      LoadStoreTracking ls;
```
jgravelle-google: This is weird. We take a pointer to a stack-allocated value, which has odd scopey lifetimey things.
Not actually sure what's ideal here. My gut would say have a class-scoped value, and then call a `reset` method for each function, but that requires explicitly clearing state and that seems actually-worse.
Maybe just null cur_ls out after each function?

test/decompile/basic.txt
```diff
@@ -79,9 +79,9 @@ data d_a(offset: 0) = "Hello, World!\0a\00";
 import function ns_fi();
 
 export function f(a:int, b:int):int {
-  var c:long = 8L;
-  var d:float = 6f;
-  var e:double = 7d;
+  c:long = 8L;
```
jgravelle-google: wait where'd the vars go

src/decompiler-ls.h
```diff
@@ -0,0 +1,211 @@
...
+    case Type::I64: return "long";
+    case Type::F32: return "float";
+    case Type::F64: return "double";
+    case Type::V128: return "simd";
+    case Type::Anyref: return "anyref";
+    case Type::Func: return "func";
+    case Type::Funcref: return "funcref";
+    case Type::Exnref: return "exceptionref";
+    case Type::Void: return "void";
+    default: return "illegal";
```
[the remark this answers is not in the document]
aardappel: I did `ILLEGAL` since `[ILLEGAL]` would make it look like a vector type :P

src/decompiler-ls.h
```diff
@@ -0,0 +1,211 @@
...
+        auto& access = var.accesses[offset];
+        // Check if previous access at this offset (if any) is of same size
+        // and type (see Checklayouts below).
+        if (access.byte_size &&
+            ((access.byte_size != byte_size) ||
+             (access.type != type) ||
+             (access.align != align)))
+          access.is_uniform = false;
+        // Also exclude weird alignment accesses from structs.
+        if (!opc.IsNaturallyAligned(align))
```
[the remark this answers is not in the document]
aardappel: Yay I'm learning!

src/decompiler-ls.h
```diff
@@ -0,0 +1,211 @@
...
+    auto byte_size = opc.GetMemorySize();
+    type = GetMemoryType(type, opc);
+    // We want to associate memory ops of a certain offset & size as being
+    // relative to a uniquely identifyable pointer, such as a local.
+    // TODO: expand this to more kinds of address expressions.
+    // NOTE: types supported here must be same as in GenAccess below.
+    switch (addr_exp.etype) {
+      case ExprType::LocalGet: {
+        auto& lg = *cast<LocalGetExpr>(addr_exp.e);
+        auto& var = vars[lg.var.name()];
```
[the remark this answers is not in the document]
aardappel: I think this is sufficiently known STL feature that that probably doesn't need a comment, yeah.

src/decompiler.cc
aardappel: Given that the state of this thing is a single map, `reset` seems the way to go.

test/decompile/basic.txt
aardappel: oops regression not caught by blind test re-gen..

src/decompiler.cc
```diff
@@ -572,8 +565,9 @@ struct Decompiler {
         }
         stream.Writef("}");
       }
-      mc.EndFunc();
       stream.Writef("\n\n");
+      mc.EndFunc();
+      lst.Clear();
```
jgravelle-google: Yeah that's nicer IMO

src/decompiler-ls.h
```diff
@@ -0,0 +1,218 @@
...
+  // TODO: something something SIMD.
+  // TODO: this loses information of the type it is read into.
+  // That may well not be the biggest deal since that is usually obvious
+  // from context, if not, we should probably represent that as a cast around
+  // the access, since it should not be part of the field type.
+  if (operand_type == Type::I32 || operand_type == Type::I64) {
+    auto name = string_view(opc.GetName());
+    // FIXME: change into a new column in opcode.def instead?
+    auto is_unsigned = name.substr(name.size() - 2) == "_u";
+    switch (opc.GetMemorySize()) {
```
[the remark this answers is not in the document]
aardappel: It's switching on an int type.

Issue #1228, opened 2019-11-15
Formatting of `ref.func` inserts an extra newline by accident


alexcrichton:
Given this module:

```wat
(module
  (func)
  (func
    ref.func 0
    drop))
```

this yields:

```
$ wat2wasm --enable-all foo.wat -o foo.wasm
$ wasm2wat foo.wasm --enable-all
(module
  (type (;0;) (func))
  (func (;0;) (type 0))
  (func (;1;) (type 0)
    ref.func
    0
    drop))
```

A pretty minor issue, but it'd be a bit nicer if the `ref.func` didn't have a newline after it!

PR #1261, opened 2019-12-09
Replace newline by a space

amirbawab:
Closes: #1228 

Simply replaced the newline by a space


sbc100:
I guess there are no effected tests?

amirbawab:
Looks like the first time it failed for timeout which already has been reported. Second time (empty commit) it passed.

sbc100:
What I mean is it looks like we are missing test coverage, if you can make this change without effecting any tests..

amirbawab:
@sbc100 Added a test case using the original issues' example.

test/desugar/ref-func.txt
```diff
@@ -0,0 +1,14 @@
+;;; TOOL: wat-desugar
```
sbc100: Thanks!

I think maybe this should live in test/roundtrip/reference-types.txt?

lgtm with that change.

amirbawab: Updated :+1:

PR #1264, opened 2019-12-12
reference-types: add final test: br_table

sbc100:
This requires some slighly different validation rules for MVP vs
interface-types.

There is a test in unreachable-invalid.wast that was removed in the
reference-types repo:
  type-br_table-label-num-vs-label-num-after-unreachable

This test depends on the old validation rules and still exists in the
master testsuite so we need to continue to support both sets of rules
for now.

### DIFF
--- a/src/type-checker.cc
+++ b/src/type-checker.cc
@@ -51,9 +51,6 @@ TypeChecker::Label::Label(LabelType label_type,
       type_stack_limit(limit),
       unreachable(false) {}
 
-TypeChecker::TypeChecker(const ErrorCallback& error_callback)
-    : error_callback_(error_callback) {}
-
 void TypeChecker::PrintError(const char* fmt, ...) {
   if (error_callback_) {
     WABT_SNPRINTF_ALLOCA(buffer, length, fmt);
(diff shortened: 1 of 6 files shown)


tlively:
Can you link to the spec of the updated validation rules?

sbc100:
The change that removed the test in question was:
https://github.com/WebAssembly/reference-types/pull/43

The new validation algorithm is at:
https://github.com/WebAssembly/reference-types/blob/master/document/core/appendix/algorithm.rst

I think this change relates to how the new bottom type and how the unreachable value stack works, but I don't fully understand the change.

sbc100:
Ah the arity check is specified here:
https://github.com/WebAssembly/reference-types/blame/master/document/core/appendix/algorithm.rst#L225

src/type-checker.cc
```diff
@@ -436,12 +433,23 @@ Result TypeChecker::OnBrTableTarget(Index depth) {
...
-    PrintError("br_table labels have inconsistent types: expected %s, got %s",
-               TypesToString(*br_table_sig_).c_str(),
-               TypesToString(label_sig).c_str());
+  } else {
+    if (features_.reference_types_enabled()) {
+      if (br_table_sig_->size() != label_sig.size()) {
+        result |= Result::Error;
+        PrintError("br_table labels have inconsistent arity: expected %" PRIzd
+                   " got %" PRIzd,
+                   br_table_sig_->size(), label_sig.size());
```
tlively: I'm confused about where we check that the types are compatible beyond their arity. In the algorithm.rst doc, this is done by a call to `pop_vals`, but where is it done here?

sbc100: That already happens above in the call to `CheckSignature` .. that basically ensure that the value stack contains the values that match the label.  That part doesn't change.

PR #1265, opened 2019-12-13
wabt-decompile: cleaned up string composition.

aardappel:
The code had 3 ways of doing string composition:
- Using + and += on string/string_view
- ostringstream
- wabt::Stream

Of these, the first was by far the most widely used, simply
because decompilation is a hierarchical process, which requires
storing intermediate strings before knowing what surrounds them
(thus unsuitable for streams).

To make the code more uniform, everything was converted to use
the first approach. To not get further performance degradations,
some more efficient concatenation methods were added, that also
work with wabt::string_view.

### DIFF
--- a/src/decompiler.cc
+++ b/src/decompiler.cc
@@ -24,15 +24,13 @@
 #define WABT_TRACING 0
 #include "src/tracing.h"
 
-#include <sstream>
 #include <inttypes.h>
 
 namespace wabt {
 
 struct Decompiler {
-  Decompiler(Stream& stream, const Module& module,
-             const DecompileOptions& options)
-      : mc(module), stream(stream), options(options) {}
+  Decompiler(const Module& module, const DecompileOptions& options)
+      : mc(module), options(options) {}
 
   struct Value {
     std::vector<std::string> v;
@@ -58,6 +56,14 @@ struct Decompiler {
     Value& operator=(const Value& rhs) = delete;
   };
 
+  std::string to_string(double d) {
+    auto s = std::to_string(d);
+    // Remove redundant trailing '0's that to_string produces.
+    while (s.size() > 2 && s.back() == '0' && s[s.size() - 2] != '.')
+      s.pop_back();
+    return s;
+  }
+
   std::string Indent(size_t amount) {
     return std::string(amount, ' ');
   }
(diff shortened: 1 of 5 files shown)


src/decompiler.cc
```diff
@@ -67,21 +73,16 @@ struct Decompiler {
...
     auto indent = Indent(amount);
-    for (auto& s : val.v) {
-      auto is = (&s != &val.v[0] || first_indent.empty())
-                ? string_view(indent)
-                : first_indent;
-      s.insert(0, is.data(), is.size());
+    for (auto& stat : val.v) {
+      auto is = (&stat != &val.v[0] || first_indent.empty())
+                    ? string_view(indent)
+                    : first_indent;
```
sbc100: Is this what clang-format says is correct?

src/decompiler.cc
```diff
@@ -58,6 +56,14 @@ struct Decompiler {
     Value& operator=(const Value& rhs) = delete;
   };
 
+  std::string to_string(double d) {
+    auto s = std::to_string(d);
+    // Remove redundant trailing '0's that to_string produces.
+    while (s.size() > 2 && s.back() == '0' && s[s.size() - 2] != '.')
+      s.pop_back();
```
sbc100: I think wabt style requires curlies here.

src/decompiler.cc
aardappel: Yup, this has been clang-formatted

src/decompiler.cc
aardappel: Looks like WABT uses `Chromium` style, which is an extension of `Google` style, which allows these curly-less statements.

test/decompile/basic.txt
```diff
@@ -80,9 +80,9 @@ import function ns_fi();
 
 export function f(a:int, b:int):int {
   var c:long = 8L;
-  var d:float = 6f;
-  var e:double = 7d;
-  if (e < 10d) { 1[4]:int = (2[3]:int + 5) }
+  var d:float = 6.0f;
+  var e:double = 7.0;
+  if (e < 10.0) { 1[4]:int = (2[3]:int + 5) }
```
sbc100: Which part caused this to change?

src/string-view.h
```diff
@@ -187,6 +187,59 @@ inline bool operator>=(string_view x, string_view y) noexcept {
   return x.compare(y) >= 0;
 }
 
+// non-member append.
+inline std::string& operator+=(std::string& x, string_view y) {
+  x.append(y.data(), y.size());
```
sbc100: maybe call the param `lhs` and `rhs`?  I don't know what the convention are here... just a thought.

src/string-view.h
```diff
@@ -187,6 +187,59 @@ inline bool operator>=(string_view x, string_view y) noexcept {
...
+inline std::string operator+(string_view x, const char* y) {
+  return x + string_view(y);
+}
+
+inline void cat_concatenate(std::string&) {}
+
+template<typename T, typename ...Ts>
+void cat_concatenate(std::string& s, const T& t, const Ts&... args) {
+    s += t;
+    cat_concatenate(s, args...);
```
sbc100: Recursive templating?  I guess this is standard practice these days?

src/decompiler.cc
```diff
@@ -467,142 +452,128 @@ struct Decompiler {
     return s;
   }
 
-  void Decompile() {
+  std::string Decompile() {
+    std::string s;
```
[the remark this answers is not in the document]
sbc100: Having said all that I'm not going to block this PR on that.. its just a suggestion, and its easy to reverse if we decide these points are valid/useful.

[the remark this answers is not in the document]
aardappel: Yes, we could at some point change this back to stream once per line or once per function, though for now its easier to work with if its all one system.

test/decompile/basic.txt
aardappel: It uses ostringstream before, which formats `1.0` as `1`. It now uses `std::to_string` which makes it into `1.000000`, which I decided to turn into `1.0` :)

src/string-view.h
aardappel: I was actually going to use exactly those names, until I saw the functions above used `x` and `y`.

src/string-view.h
aardappel: Yes, welcome to C++11 variadic templates :)
It's pretty neat, it can compute the size of an arbitrary set of strings of different types, and then concat them all in one alloc.

PR #1275, opened 2019-12-26
Update testsuite

sbc100:
The two primary changes involved are:

1. Removal of `assert_return_canonical_nan`/`arithetic nan` in favor of
   special `nan:canonical`/`nan:arithmetic` constants that can only be
   used in test expectations.
   See: https://github.com/WebAssembly/spec/pull/1104

2. New trapping behaviour for bulk memory operations.  Range checks are
   now performed up front for opterations such as memory.fill and
   memory.copy.
   See: https://github.com/webassembly/bulk-memory-operations/issues/111
   And: https://github.com/webassembly/bulk-memory-operations/pull/123
   The old behaviour is still kept around to support table.fill which
   is defined in reference-types proposal and has yet to be updated.


src/interp/interp.cc
```diff
@@ -1031,6 +1035,17 @@ bool ClampToBounds(uint32_t start, uint32_t* length, uint32_t max) {
   return true;
 }
 
+static bool CheckBounds(uint32_t start, uint32_t length, uint32_t max) {
+  if (start > max) {
+    return false;
+  }
+  uint32_t avail = max - start;
+  if (length > avail) {
```
aardappel: `return length <= avail` ?

test/gen-spec-js.py
```diff
@@ -165,14 +165,17 @@ def EscapeJSString(s):
 
 def IsValidJSConstant(const):
     type_ = const['type']
+    value = const['value']
+    if type_ in ('f32','f64') and value in ('nan:canonical', 'nan:arithmetic'):
```
aardappel: this seems to cause a CI error

test/gen-spec-js.py
```diff
@@ -463,13 +449,15 @@ def _Module(self, filename):
     def _Constant(self, const):
         assert IsValidJSConstant(const), 'Invalid JS const: %s' % const
         type_ = const['type']
-        value = int(const['value'])
+        value = const['value']
+        if type_ in ('f32','f64') and value in ('nan:canonical', 'nan:arithmetic'):
```
aardappel: and here

test/spec/reference-types/select.txt
```diff
@@ -2,69 +2,69 @@
 ;;; STDIN_FILE: third_party/testsuite/proposals/reference-types/select.wast
 ;;; ARGS*: --enable-reference-types
 (;; STDOUT ;;;
-out/test/spec/reference-types/select.wast:283: assert_trap passed: unreachable executed
```
aardappel: we have a lot of random numbers polluting our tests.. would be great to find ways in which we can make it less brittle?

vshymanskyy:
This is really important for https://github.com/wasm3/wasm3.
Thanks!

test/spec/reference-types/select.txt
binji: agreed, but it's the tradeoff for having the ability to automatically rebase tests. If we remove the random numbers, we'd probably have to do something more like LLVM.

src/tools/spectest-interp.cc
```diff
@@ -490,65 +497,131 @@ wabt::Result JSONParser::ParseConst(TypedValue* out_value) {
...
+      return wabt::Result::Error;
+    }
     out_value->type = Type::I32;
     out_value->value.i32 = value;
-    return wabt::Result::Ok;
   } else if (type_str == "f32") {
     uint32_t value_bits;
-    CHECK_RESULT(ParseInt32(value_start, value_end, &value_bits,
-                            ParseIntType::UnsignedOnly));
+    /*
```
binji: remove?

src/tools/spectest-interp.cc
```diff
@@ -490,65 +497,131 @@ wabt::Result JSONParser::ParseConst(TypedValue* out_value) {
...
+                                         string_view type_str,
+                                         string_view value_str) {
   const char* value_start = value_str.data();
   const char* value_end = value_str.data() + value_str.size();
-
   if (type_str == "i32") {
     uint32_t value;
-    CHECK_RESULT(
-        ParseInt32(value_start, value_end, &value, ParseIntType::UnsignedOnly));
+    if (Failed((ParseInt32(value_start, value_end, &value,
```
binji: why change this?

src/wast-parser.cc
```diff
@@ -2122,15 +2148,15 @@ Result WastParser::ParseHostRef(Const* const_) {
   return Result::Ok;
 }
 
-Result WastParser::ParseConstList(ConstVector* consts) {
+Result WastParser::ParseConstList(ConstVector* consts, bool is_assert_result) {
```
binji: enum is nicer here

src/tools/spectest-interp.cc
```diff
@@ -150,21 +150,23 @@ class ActionCommandBase : public CommandMixin<TypeEnum> {
...
-    AssertReturnArithmeticNanCommand;
 
 class RegisterCommand : public CommandMixin<CommandType::Register> {
  public:
   std::string as;
   std::string name;
 };
 
+struct ExpectedValue {
+  bool isExpectedNan;
```
binji: snake_case for member vars (here and a few other places too)

src/ir.h
```diff
@@ -96,13 +96,16 @@ struct Const {
 
   Location loc;
   Type type;
+
+  bool isExpectedNan = false;
```
binji: snake_case for member vars

test/run-spec-wasm2c.py
```diff
@@ -221,38 +219,40 @@ def _WriteActionCommand(self, command):
...
             type_ = expected[0]['type']
-            assert_macro = assert_map[type_]
-            self.out_file.write('%s(%s, %s);\n' %
-                                (assert_macro,
-                                 self._Action(command),
-                                 self._ConstantList(expected)))
+            value = expected[0]['value']
+            if value == 'nan:canonical':
+                assert_map = {
+                    ('f32'): 'ASSERT_RETURN_CANONICAL_NAN_F32',
```
binji: remove extra parens

sbc100:
This change for a little smaller thanks to https://github.com/WebAssembly/wabt/pull/1278

src/tools/spectest-interp.cc
sbc100: Oh.. these errors were not being reported.  This is a separate issue that I will split out.

vshymanskyy:
Thanks! The updated `wasm2json` worked fine for `wasm3` use case!

PR #1276, opened 2019-12-26
Change encoding of Exnref to -0x12

sbc100:
The spec for this was specified as -18 in decimal which is -0x12 in hex.

Fixes: https://github.com/WebAssembly/exception-handling/issues/92

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -214,7 +214,7 @@ enum class Type : int32_t {
   V128 = -0x05,     // 0x7b
   Funcref = -0x10,  // 0x70
   Anyref = -0x11,   // 0x6f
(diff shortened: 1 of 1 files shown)


Horcrux7:
This sounds wrong for me.
The value -0x12 or decimal -18 or 0x6E is the value of nullref
https://webassembly.github.io/reference-types/core/appendix/index-types.html

The value of exnref is -0x18 or decimal -24 or 0x68
https://github.com/WebAssembly/exception-handling/blob/master/proposals/Exceptions.md

aheejin:
No the current wabt implementation is correct. The correct value is hexadecimal -0x18.

PR #1277, opened 2019-12-26
wasm-decompile: added precedence support.

aardappel:
Previously it would simply bracket all binary exps. Now it has a
precedence system that is in line with what people know from most
programming languages.


src/decompiler.cc
```diff
@@ -313,11 +342,24 @@ struct Decompiler {
       }
       case ExprType::Binary: {
         auto& be = *cast<BinaryExpr>(n.e);
-        return WrapBinary(args, cat(" ", OpcodeToToken(be.opcode), " "), false);
+        auto opcs = OpcodeToToken(be.opcode);
+        // FIXME: do this more elegantly.
```
jgravelle-google: How so?
```c++
assert(opcs.length > 0);
switch(opcs[0]) {
case '*':
case '/':
case '%':
  prec = Precedence:::Multiply;
  break;
//...
case '<':
  if (opcs.length == 2 && opcs[1] == '<') {
    prec = Precedence::Shift;
  }
  break;
// ...
```

I don't think that's necessarily better.

Maybe an array keyed on the enum value with strings as the values? I don't think that's clearer, and definitely worse perf-wise.

My point being, I don't think the FIXME is worth it as is. With a more concrete plan for how to to better it would be though.

src/decompiler.cc
```diff
@@ -108,23 +133,26 @@ struct Decompiler {
     return std::move(child);
   }
 
-  void BracketIfNeeded(Value &val) {
-    if (!val.needs_bracketing) return;
+  void BracketIfNeeded(Value &val, Precedence parent_precedence) {
+    if (parent_precedence < val.precedence ||
```
jgravelle-google: Ok that's pretty slick. Nice

src/decompiler.cc
aardappel: Was thinking that this would be cleaner working directly on the `Opcode` type rather than on strings (or chars) derived from it. It's also brittle in the sense that if other operators are added, this just will bin them in the default case. I'll make a note in the code.

PR #1278, opened 2019-12-27
Several fixes for reference types

aheejin:
- Allow `ref.func` for global initialization expressions 
- Allow `nullref` as a full-fledged type, after
  WebAssembly/reference-types#66
- Allow all reference types to be table elements
- Enable reference types when exnref is used (The reference types
  proposal is a prerequisite of the EH proposal)


aheejin:
Hmm, I had an (incorrect) impression that the s-expression format Binaryen uses is the wast format and the straight-line format Wabt uses is the wat format. But it seems Wabt also can parse s-expression format Binaryen uses (although not all of them currently).

aheejin:
Ping :)

(Side note: Yesterday I commented on this thread about the difference between wast and wat or something, and that comment was meant for a completely different PR in Binaryen. Not sure why I ended up commenting here. If you saw it, please ignore it.)

test/spec/bulk-memory-operations/binary.txt
```diff
@@ -142,7 +142,7 @@ out/test/spec/bulk-memory-operations/binary.wast:725: assert_malformed passed:
   0000024: error: expected ref.null or ref.func in passive element segment
   0000025: error: expected END opcode after element expression
 out/test/spec/bulk-memory-operations/binary.wast:751: assert_malformed passed:
-  0000022: error: segment elem expr type must be funcref or anyref (got i32)
+  0000022: error: segment elem expr type must be reference types (got i32)
```
sbc100: "must be a reference type"?

aheejin: Fixed

PR #1279, opened 2020-01-02
wasm-decompile: supporting some more node types specifically.

aardappel:
This outputs some more WABT IR node types with special purpose
syntax, rather than the default catch-all of a function call.
Still incomplete (especially for >MVP), more later.

Reworking br_table will be a seperate PR.

### DIFF
--- a/test/decompile/basic.txt
+++ b/test/decompile/basic.txt
@@ -54,6 +54,48 @@
       br_if 0
     end
     i32.const 1
+    i32.const 2
+    i32.const 1
+    i32.const 1
+    i32.eq
+    select
+    drop
+    block
+      block
+        block
+          block
+            i32.const 0
+            drop
+            block
+              i32.const 0
+              drop
+              local.get 0
+              br_table 3 2 1 0 4
+              i32.const 99
+              return
+              unreachable
+            end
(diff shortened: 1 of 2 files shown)


test/decompile/basic.txt
```diff
@@ -91,7 +133,31 @@ export function f(a:int, b:int):int {
...
-  return 1;
+  select_if(1, 2, 1 == 1);
+  block B_e {
+    block B_f {
+      block B_g {
+        block B_h {
+          0;
+          block B_i {
+            0;
+            br_table[B_f, B_g, B_h, B_i, ..B_e](a);
```
jgravelle-google: Surprisingly readable already. An improvement over wat at least.
Dealing with 1000-wide br_tables will need something more custom, if that's possible to make nice at all. I'm thinking we can only nice-ify big br_tables if they have lots of repetition, something like:

```
switch N:
  cases 0-45:
    foo()
  case 46:
    bar()
  cases 47-99:
    foo() again
```

which I think is how I usually see 1K br_tables in the wild.

aardappel: One of the next PRs will change this syntax :)

PR #1282, opened 2020-01-06
wasm-decompile: blocks now represented as labels

aardappel:
What was before: `block L { STATS }`
is now `{ STATS; label L: }`
or when possible just: `STATS; label L:`
The latter having no indentation at all, and thus automatically
flattening all `br_table` nestings and other common patterns.

It was initially attempted to create a proper switch out of `br_table`,
but the typical LLVM output is so intertwined (with br/br_if jumping
in and out of the br_table targets etc) that a switch could have only
cleanly applied applied to a small subset of cases. The current
simple label flattening works with all wasm code equally, but is a
a bit more low level.

Also rename `break` into `goto`, reflecting what it is really doing.
Though here, `goto` only ever jumps downwards, backwards jumps to the
`loop` construct are still called `continue`.

### DIFF
--- a/src/decompiler.cc
+++ b/src/decompiler.cc
@@ -212,13 +212,6 @@ struct Decompiler {
     return WrapChild(child, ve.var.name() + " = ", "");
   }
 
-  Value Block(Value &val, const Block& block, LabelType label, const char *name) {
-    IndentValue(val, indent_amount, {});
-    val.v.insert(val.v.begin(), cat(name, " ", block.label, " {"));
-    val.v.push_back("}");
-    return std::move(val);
-  }
-
   std::string TempVarName(Index n) {
     // FIXME: this needs much better variable naming. Problem is, the code
     // in generate-names.cc has allready run, its dictionaries deleted, so it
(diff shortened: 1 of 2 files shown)


test/decompile/basic.txt
```diff
@@ -127,32 +121,23 @@ export function f(a:int, b:int):int {
...
-        }
-        return 101;
-      }
-      return 102;
-    }
-    return 103;
-  }
+  br_table[B_f, B_g, B_h, B_i, ..B_e](a);
+  unreachable;
+  label B_i:
```
jgravelle-google: Oh that's nicer.

Nit/Opinion: dedent the labels? Having them in the same column as the instructions makes it harder to see the control flow structure at a glance.

src/decompiler.cc
```diff
@@ -418,20 +411,37 @@ struct Decompiler {
         }
       }
       case ExprType::Block: {
-        return Block(args[0], cast<BlockExpr>(n.e)->block, LabelType::Block, "block");
+        auto& val = args[0];
+        val.v.push_back(
+              cat("label ", cast<BlockExpr>(n.e)->block.label, ":"));
+        if (parent && parent->ntype != NodeType::Statements
```
jgravelle-google: This isn't quite clear. This is the code that's responsible for wrapping `{}` around blocks? So long as we have a parent that isn't a statement, block, loop, or some ifs?

This is for blocks used as expressions, right? It would be good to see a test of this to see what it's supposed to do.

test/decompile/basic.txt
aardappel: Yeah, that wasn't trivial to do, so left it like this for now. That was also partially the motivation to add the `label` keyword. May revisit that later to dedent.

src/decompiler.cc
aardappel: Ok, added comment and test case.

aardappel:
Oh, @kripken this was the PR I was talking about :)

kripken:
Thanks @aardappel! Yes, this does look much nicer than a normal wat `br_table1` :smiley:

Issue #1268, opened 2019-12-17
Binary encoding of element segment uses non-MVP encoding when it could otherwise do so


alexcrichton:
Given a module that looks like:

```wat
(module
  (table 1 funcref)
  (elem (table 0) (i32.const 0)))
```

this uses the nifty new `(table 0)` syntax to clarify that the `elem` segment is part of a particular table. This `*.wat`, however, is MVP-compatible but ends up being encoded as:

```
$ wat2wasm foo.wat --enable-all -o foo.wasm
$ wasm-objdump -x foo.wasm

foo.wasm:	file format wasm 0x1

Section Details:

Table[1]:
 - table[0] type=funcref initial=1
Elem[1]:
 - segment[0] flags=2 table=0 count=0 - init i32=0
DataCount:
 - data count: 0
```

The data count section is covered by https://github.com/WebAssembly/wabt/issues/1189 but using `flags=2` here isn't quite necessary when it could otherwise use `flags=0` and be MVP compatible?

PR #1283, opened 2020-01-06
if bulk memory is not enabled, error if an element references a table…

### LINKED ISSUE #1280: Dynamic discovery of Wasm functions, params and result.
This is a suggestion for a feature request.

** Idea
Convert WAT, representation of a Wasm binary, into a language-independent data format so that the information can be transmitted across the web and also parsed (by any language) for inspection.

** Why
As far as I can tell, after unsuccessfully attempting to parse WAT files as S-expression data ( using tools like lexpr [1] and atoms [2]) there does not seem to be a way to parse WAT as a native object and then dynamically query function names (as well as param and result types) via code.

** Example
The following WAT file could be converted to a data interchange format that can be a) sent over the web and b) parsed as an object 
```
(module
  (type (;0;) (func (param i32 i32) (result i32)))
  (type (;1;) (func (param i32 i32 i32) (result i32)))
  (func (;0;) (type 0) (param i32 i32) (result i32)
    local.get 0
    local.get 1
    i32.add)
  (func (;1;) (type 1) (param i32 i32 i32) (result i32)
    local.get 0
    local.get 1
    i32.add
    local.get 2
    i32.add)
  (export "add" (func 0))
  (export "add_again" (func 1)))
```
** Arbitrary example of a format (in this case just a JSON file) 
```
{
	"functions": [{
			"name": "add",
			"type": "0",
			"params": ["i32", "i32"],
			"results": ["i32"]
		},
		{
			"name": "add_again",
			"type": "1",
			"params": ["i32", "i32", "i32"],
			"results": ["i32"]
		}
	]
}
```
** The parsed data object could be queried like this
```
data.functions[0].name
//returns "add"
data.functions[0].params[0]
//returns i32
```
Of course as with data objects the code could also iterate over the object and perform tasks such as determining how many params a function has etc. There are many use cases for this, including microservices where calling code wants to execute a Wasm function on a remote server via RPC etc.

** References
[1] https://docs.rs/lexpr/0.2.4/lexpr/
[2] https://docs.rs/atoms/2.2.3/atoms/

theor:
… != 0.

Fixes #1280 and #1268

### DIFF
--- a/src/wast-parser.cc
+++ b/src/wast-parser.cc
@@ -975,10 +975,14 @@ Result WastParser::ParseElemModuleField(Module* module) {
     }
   } else {
     if (has_name) {
-      field->elem_segment.table_var = Var(initial_name, loc);
-      field->elem_segment.flags |= SegExplicitIndex;
+      auto index = module->table_bindings.FindIndex(initial_name);
+      if(index != 0)
+        Error(loc, "Referencing a table different from the first table requires the bulk_memory_enabled feature");
+      field->elem_segment.table_var = Var(0, loc);
     } else {
       ParseVarOpt(&field->elem_segment.table_var, Var(0, loc));
+      if(field->elem_segment.table_var.is_index() && field->elem_segment.table_var.index() != 0)
+          Error(loc, "Referencing a table different from the first table requires the bulk_memory_enabled feature");
     }
   }
 


binji:
Thanks for taking a look at this! I took over this fix since it needed a little more work than seen here. Should be fixed in #1286.

Issue #1269, opened 2019-12-17
Binary encoding for empty funcref segment may use wrong flags?


alexcrichton:
Given this `*.wat`:

```wat
(module
  (table 1 funcref)
  (elem (table 0) (i32.const 0) funcref))
```

this encodes as:

```
$ wat2wasm foo.wat --enable-all -o foo.wasm
$ wasm-objdump -x foo.wasm

foo.wasm:	file format wasm 0x1

Section Details:

Table[1]:
 - table[0] type=funcref initial=1
Elem[1]:
 - segment[0] flags=7 table=0 count=0
DataCount:
 - data count: 0
```

Here the element segment uses `flags=7` but I think this one should more accurately use either 4 or 6 according to https://github.com/WebAssembly/bulk-memory-operations/blob/master/proposals/bulk-memory-operations/Overview.md#element-segments?

PR #1286, opened 2020-01-08
Segment flags should be MVP-compat when possible

### LINKED ISSUE #1280: Dynamic discovery of Wasm functions, params and result.

binji:
This should fix #1280, #1268, #1269.


src/binary-reader.cc
```diff
@@ -2172,7 +2172,7 @@ Result BinaryReader::ReadElemSection(Offset section_size) {
     ERROR_IF(flags > ~(~0u << SegFlagMax), "invalid elem segment flags: %#x",
              flags);
     Index table_index(0);
-    if (flags & SegExplicitIndex) {
+    if ((flags & (SegPassive | SegExplicitIndex)) == SegExplicitIndex) {
```
sbc100: Is this needed as part of this change?

Should we instead assert inside the if block if SegPassive is set?

src/binary-reader.cc
```diff
@@ -2186,8 +2186,7 @@ Result BinaryReader::ReadElemSection(Offset section_size) {
     }
 
     // For backwards compat we support not declaring the element kind.
-    bool legacy = !(flags & SegPassive) && !(flags & SegExplicitIndex);
-    if (!legacy) {
+    if (flags & (SegPassive | SegExplicitIndex)) {
```
sbc100: This is a no-op right?

src/ir.cc
```diff
@@ -651,4 +651,52 @@ Const::Const(F64Tag, uint64_t value, const Location& loc_)
 Const::Const(V128Tag, v128 value, const Location& loc_)
     : loc(loc_), type(Type::V128), vec128(value) {}
 
+uint8_t ElemSegment::GetFlags(const Module* module) const {
+  uint8_t flags = 0;
+
+  switch (kind) {
+    case SegmentKind::Active: {
+      Index table_index = module->GetTableIndex(table_var);
+      if (table_index != 0) {
```
sbc100: Doesn't this mean that specifying table_index of 0 explicitly doesn't round trip?

src/ir.cc
```diff
@@ -651,4 +651,52 @@ Const::Const(F64Tag, uint64_t value, const Location& loc_)
...
+      break;
+  }
+
+  bool all_ref_func = std::all_of(
+      elem_exprs.begin(), elem_exprs.end(), [](const ElemExpr& elem_expr) {
+        return elem_expr.kind == ElemExprKind::RefFunc;
+      });
+  if (!all_ref_func) {
+    flags |= SegUseElemExprs;
+  }
```
sbc100: Again doesn't this prevent round tripping?

src/validator.cc
```diff
@@ -1152,16 +1152,21 @@ void Validator::CheckTable(const Location* loc, const Table* table) {
...
+                             options_.features.reference_types_enabled();
+
   for (const ModuleField& field : module->fields) {
     if (auto elem_segment_field = dyn_cast<ElemSegmentModuleField>(&field)) {
       auto&& elem_segment = elem_segment_field->elem_segment;
       for (const ElemExpr& elem_expr : elem_segment.elem_exprs) {
         if (elem_expr.kind == ElemExprKind::RefFunc) {
           CheckFuncVar(&elem_expr.var, nullptr);
+        } else if (!bulk_or_ref_enabled) {
+          PrintError(&field.loc, "ref.null is not allowed");
```
sbc100: What if `elem_expr.kind` is not `RefNul`?

src/wast-parser.cc
```diff
@@ -934,7 +934,7 @@ Result WastParser::ParseDataModuleField(Module* module) {
     CHECK_RESULT(ParseOffsetExpr(&field->data_segment.offset));
   } else if (!ParseOffsetExprOpt(&field->data_segment.offset)) {
     if (options_->features.bulk_memory_enabled()) {
-      field->data_segment.flags |= SegPassive;
+      field->data_segment.kind = SegmentKind::Passive;
```
sbc100: How about flipping the condition which would result in an early return and removing the else block?

test/regress/regress-37.txt
```diff
@@ -0,0 +1,40 @@
+;;; TOOL: run-objdump
+;;; ARGS0: -v --enable-bulk-memory
```
sbc100: I'm not a big fan of these regress tests.  Wouldn't it be better to distil them into something more specific with a meaningful name?   Perhaps you are following some pattern that I'm not aware of?   Is there good reason for such a practice over just writing new unit tests?

src/binary-reader.cc
binji: We'll want to support declared segments soon, so yeah. Maybe should error instead of assert.

src/binary-reader.cc
binji: Yep, no difference. I found this a bit clearer to compare to the table in the doc, but it's still not super clear.

src/ir.cc
binji: Yes, but I think that's OK. We don't guarantee a perfect roundtrip (one example is if you use over-long LEB encoding). The semantics are otherwise unchanged.

src/ir.cc
binji: Yes, but it is always weakening the requirements (fewer features), without changing semantics.

src/validator.cc
binji: The enum only allows RefNull and RefFunc for now, so if it were something else there's an error earlier in the pipeline.

src/wast-parser.cc
binji: Done.

test/regress/regress-37.txt
binji: Yeah, you're right. They're fine for tests that are odd combinations, but probably shouldn't be added when the real test is natural.

src/validator.cc
sbc100: I still rather see an explicit check here followed by an "else { UNREACHABLE }" to catch the other reftypes when they start to make sense.

binji: Sorry, responded before pushing. :)

binji: Whoops, thought you were talking about something else. :)

PR #1287, opened 2020-01-09
Add default output name for wast2json

sbc100:
Without this the default was to write the wasm files but not the json.

This matches the behaviour of wat2wasm.

### DIFF
--- a/src/tools/wast2json.cc
+++ b/src/tools/wast2json.cc
@@ -39,7 +39,7 @@
 using namespace wabt;
 
 static const char* s_infile;
-static const char* s_outfile;
+static std::string s_outfile;
 static int s_verbose;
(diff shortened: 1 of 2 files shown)


binji:
I think this change may have started dropping wasm+json files in the project dir when I run tests:

```
	assert-after-module.0.wasm
	assert-after-module.json
	assertinvalid-binary-module.0.wasm
	assertinvalid-binary-module.1.wasm
	assertinvalid-binary-module.json
```

sbc100:
Oops.  Looking.

sbc100:
Right.. looks like the default behaviour before was to write files alongside the input.   This is not how other tools such as gcc, clang, wat2wasm work.  They wrong files in the CWD by default.  I've got a fix.

PR #1289, opened 2020-01-09
Apply start function name in wasm2wat

### LINKED ISSUE #1226: `wasm2wat` doesn't give the start function a pretty name
Given a module like so:

```wat
(module
  (func $foo)
  (func $bar
    call $foo)
  (start $foo))
```

it generates the following output:

```sh
$ wat2wasm foo.wat -o foo.wasm --debug-names
$ wasm2wat ./foo.wasm
(module
  (type (;0;) (func))
  (func $foo (type 0))
  (func $bar (type 0)
    call $foo)
  (start 0))
```

Here there's `start 0`, but it'd be a bit nicer if it were `start $foo`!

binji:
Fixes #1226.

### DIFF
--- a/src/apply-names.cc
+++ b/src/apply-names.cc
@@ -88,6 +88,7 @@ class NameApplier : public ExprVisitor::DelegateNop {
   Result VisitExport(Index export_index, Export* export_);
   Result VisitElemSegment(Index elem_segment_index, ElemSegment* segment);
   Result VisitDataSegment(Index data_segment_index, DataSegment* segment);
+  Result VisitStart(Index start_index, Var* start_var);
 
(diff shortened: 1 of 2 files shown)


src/apply-names.cc
```diff
@@ -453,6 +454,11 @@ Result NameApplier::VisitDataSegment(Index data_segment_index,
   return Result::Ok;
 }
 
+Result NameApplier::VisitStart(Index start_index, Var* start_var) {
+  CHECK_RESULT(UseNameForFuncVar(start_var));
```
sbc100: Why have the first arg if its not used?

binji: copy pasta :) removed

PR #1294, opened 2020-01-09
Add explicit -o flag to wast2json tests

sbc100:
Otherwise the new default behavior (as of #1287) is to write the CWD
which we don't want when running tests.

Writing to CWD rather than alongside the input file is deliberate and
desirable and matches the behavior of other compilers such as clang
gcc and wat2wasm.

### DIFF
--- a/test/run-tests.py
+++ b/test/run-tests.py
@@ -49,7 +49,7 @@ TOOLS = {
     ],
     'wast2json': [
         ('RUN', '%(wast2json)s'),
-        ('ARGS', ['%(in_file)s']),
+        ('ARGS', ['%(in_file)s', '-o', '%(out_dir)s/out.json']),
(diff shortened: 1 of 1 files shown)


test/run-tests.py
```diff
@@ -49,7 +49,7 @@
     ],
     'wast2json': [
         ('RUN', '%(wast2json)s'),
-        ('ARGS', ['%(in_file)s']),
+        ('ARGS', ['%(in_file)s', '-o', '%(out_dir)s/out.json']),
```
binji: Is the out_dir unique for each test? I forget. If it isn't, then we'll probably want the `out.json` file to be based on the input filename so the tests won't be screwed up when run in parallel.

sbc100: It is unique yes

PR #1296, opened 2020-01-10
wasm-decompile: support for pointers to single types.

aardappel:
If deriving a "struct" from load/store ops fails, the next
best thing is a typed pointer, if all accesses are to the
same type.

Also fixed some precedence related issues.

### DIFF
--- a/test/decompile/loadstore.txt
+++ b/test/decompile/loadstore.txt
@@ -108,7 +108,7 @@ export function f(a:{ a:float, b:float }, b:{ a:ushort, b:long }) {
   var e:int;
   var f:int;
   var g:int;
-  var h:int;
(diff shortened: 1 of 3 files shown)


test/decompile/loadstore.txt
```diff
@@ -117,7 +117,7 @@ export function f(a:{ a:float, b:float }, b:{ a:ushort, b:long }) {
   f[0]:short;
   g[0]:int;
   g[0]:int@1;
-  h[1]:float@1;
+  h[1];
```
binji: no type here?

aardappel: Yup that's exactly what the change is intended to do. Rather than have types on all accesses, only have them once on the pointer decl.

Issue #1176, opened 2019-09-27
spec-interp validation differs from wasm-validate


sbc100:
When a wasm file doesn't have a DataCount section I think validation of  memory.init should fail.  I have a wasm file that validates in wabt but in v8 reports:

`CompileError: WebAssembly.Module(): Compiling function #5840 failed: invalid data segment index: 2 @+2798285
`

amirbawab:
Hi @sbc100, can you please post the wasm file?

tlively:
Here's a trivial example.

```
(module
 (memory 1 1)
 (data passive "")
 (func (data.drop 0))
)
```
You can get a binary with no DataCount section using Binaryen's wasm-opt and passing the `--mvp-features --no-validation` flags.

sbc100:
Presumably we are missing a spec test case here.

sbc100:
Seems like there is a spec test for this

However we pass the test because we use binary-reader-interp.cc when loading all the spec tests and the interpreter has its own check for this.  However the check is missing from the validator itself.

binji:
Probably should have a fuller solution to this (difference between spec-interp and wasm-validate), but this particular bug is fixed in #1297.

binji:
spec-interp and wasm-validate now share a validator (`SharedValidator`), so they produce the same results.

PR #1297, opened 2020-01-10
Error on memory.init|data.drop without DataCount

binji:
See #1176. This was working in the spec interpreter because of the way
the segment indexes are handled, but it's better to handle it earlier in
the binary reader.

### DIFF
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -1433,8 +1433,8 @@ Result BinaryReader::ReadFunctionBody(Offset end_offset) {
 
       case Opcode::MemoryInit: {
         Index segment;
-        ERROR_UNLESS(data_count_ != kInvalidIndex,
-                     "memory.init requires data count section");
+        ERROR_IF(data_count_ == kInvalidIndex,
+                 "memory.init requires data count section");
         CHECK_RESULT(ReadIndex(&segment, "elem segment index"));
         uint8_t reserved;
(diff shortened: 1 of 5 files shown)


test/binary/bad-data-drop-no-data-count.txt
```diff
@@ -0,0 +1,23 @@
+;;; TOOL: run-gen-wasm-bad
+;;; ARGS1: --enable-bulk-memory
+;;; ARGS2: --enable-bulk-memory
```
sbc100: Why not ARGS*?

src/binary-reader.cc
```diff
@@ -1433,6 +1433,8 @@ Result BinaryReader::ReadFunctionBody(Offset end_offset) {
 
       case Opcode::MemoryInit: {
         Index segment;
+        ERROR_UNLESS(data_count_ != kInvalidIndex,
```
sbc100: How about `ERROR_IF(data_count_ == kInvalidIndex)`?

test/binary/bad-data-drop-no-data-count.txt
binji: Because this flag doesn't work with ARGS0, unfortunately. We probably should have a less brittle syntax for this.

src/binary-reader.cc
binji: Done.

PR #1301, opened 2020-01-13
[simd] Add support for narrow/widen instructions

binji:
* i{8x16,16x8}.narrow_i{16x8,32x4}_{s,u}
* i{16x8,32x4}.widen_{low,high}_i{8x16,16x8}_{s,u}

(though not all combinations are valid)

These are not currently supported in the interpreter.

TODO: Implement other new SIMD instructions too: load and extend,
bitwise and-not, lanewise rounding average


src/interp/interp.h
```diff
@@ -74,7 +74,9 @@ namespace interp {
   /* we tried to get an export by name that doesn't exist */                \
   V(UnknownExport, "unknown export")                                        \
   /* the expected export kind doesn't match. */                             \
-  V(ExportKindMismatch, "export kind mismatch")
+  V(ExportKindMismatch, "export kind mismatch")                             \
+  /* instruction not implemented in the interpreter */                      \
+  V(NotImplemented, "not implemented")
```
sbc100: Interesting.. is this really the first time that we've added instructions and left them as unimplemented in the interpreter?

binji: I think in the past I'd left them as WABT_UNREACHABLE, but that's dangerous, since it ends up being compiled to "this never happens", not an abort. Throwing an error is nicer, I think.

PR #1302, opened 2020-01-14
wasm-decompile: absolute accesses refer to data segments

aardappel:
This makes them easier to look up than the large integer
constants LLVM output is full of.

### DIFF
--- a/src/decompiler.cc
+++ b/src/decompiler.cc
@@ -224,7 +224,16 @@ struct Decompiler {
     return cat(name, ":", struc.empty() ? GetDecompTypeName(t) : struc);
   }
 
-  void LoadStore(Value &val, const Node& addr_exp, uint32_t offset,
+  bool ConstIntVal(const Expr* e, uint64_t &dest) {
+    dest = 0;
+    if (!e || e->type() != ExprType::Const) return false;
+    auto& c = cast<ConstExpr>(e)->const_;
+    if (c.type != Type::I32 && c.type != Type::I64) return false;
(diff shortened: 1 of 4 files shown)


src/decompiler.cc
```diff
@@ -224,7 +224,14 @@ struct Decompiler {
     return cat(name, ":", struc.empty() ? GetDecompTypeName(t) : struc);
   }
 
-  void LoadStore(Value &val, const Node& addr_exp, uint32_t offset,
+  std::pair<uint64_t, bool> ConstIntVal(const Expr* e) {
```
jgravelle-google: Simpler to just return `uint64_t`, and if it's non-0 then we have a result? This makes it so we can't have 0 as a constant value, but I'm kind of fine with that...

Although the name here is generic enough that this could be used in other contexts, so maybe it's fine as it is.

aardappel: Having 0 as an address or offset is totally possible though, since it is their sum that is the absolute address. Either one could be zero. Which means I'd need to cast it to signed and use -1, or use MAXUINT, or.. a bool. Which do you prefer? :)

jgravelle-google: Ah right. I kinda realized this and that's why I added the second line.

lgtm as-is then

binji: yep, optional would be nicer here. The rest of wabt would probably return bool and have uint64_t* as an out param, but I'm OK with this too.

aardappel: Ok, out param it is..

PR #1308, opened 2020-01-16
Update testsuite

binji:
* Remove `assert_return_func`. This is now handled by using
  `assert_return` with `(ref.func)`.
* The reference types proposal depends on the bulk memory proposal, so
  using `--enable-reference-types` automatically includes
  `--enable-bulk-memory`.
* `table.fill` no longer clamps to the valid range, and instead checks
  before writing anything. This matches the other bulk instructions.


src/interp/interp.cc
```diff
@@ -1144,15 +1127,16 @@ Result Thread::TableGet(const uint8_t** pc) {
...
   Table* table = ReadTable(pc);
-  uint32_t table_size = table->size();
   uint32_t size = Pop<uint32_t>();
   Ref value = static_cast<Ref>(Pop<Ref>());
   uint32_t dst = Pop<uint32_t>();
-  bool ok = ClampToBounds(dst, &size, table_size);
+  bool ok = CheckBounds(dst, size, table->size());
+  if (!ok) {
+    TRAP_MSG(TableAccessOutOfBounds, "table.fill out of bounds");
+  }
```
sbc100: Use `TRAP_UNLESS` and move `CheckBounds` inside it?

test/dump/reference-types.txt
```diff
@@ -84,10 +84,12 @@ Table[3]:
  - table[1] type=anyref initial=1
  - table[2] type=funcref initial=1
 Elem[2]:
- - segment[0] flags=2 table=2 count=1 - init i32=0
+ - segment[0] flags=0 table=0 count=1 - init i32=0
```
sbc100: This doesn't look right.  Did something change such that `(elem $baz (i32.const 0) $f1)` not long means that the elements are for table $baz?

sbc100: Oh I guess it pulled in bulk memory and that changes the meaning here?   In which case we now have elem and table with the same name.. is that allowed?

src/interp/interp.cc
binji: We could, but this matches the way it's written elsewhere.

test/dump/reference-types.txt
binji: Yeah, the names are kept in separate namespaces. We always know what kind of thing we are referring to.

binji: Oh yeah, and I guess the new syntax for referring to a table is `(table $baz)`

PR #1309, opened 2020-01-17
wasm-decompile: name data sections with their content

aardappel:
This is a fun way to give somewhat meaningful names to sections
containing strings. In the case of pure binary sections this likely
generates random characters, but that's not any worse than the
current generated names.

### DIFF
--- a/src/decompiler-naming.h
+++ b/src/decompiler-naming.h
@@ -110,6 +110,47 @@ void RenameToIdentifiers(std::vector<T*>& things, BindingHash& bh,
   }
 }
 
+enum {
+  // This a bit arbitrary, change at will.
+  min_content_identifier_size = 7,
+  max_content_identifier_size = 30
+};
+
+void RenameToContents(std::vector<DataSegment*>& segs, BindingHash& bh) {
+  std::string s;
+  for (auto seg : segs) {
+    s = "d_";
+    for (auto c : seg->data) {
+      if (isalnum(c) || c == '_') {
+        s += static_cast<char>(c);
+      }
+      if (s.size() >= max_content_identifier_size) {
+        // We truncate any very long names, since those make for hard to
+        // format output. They can be somewhat long though, since data segment
+        // references tend to not occur that often.
+        break;
+      }
+    }
+    if (s.size() < min_content_identifier_size) {
+      // It is useful to have a minimum, since if there few printable characters
+      // in a data section, that is probably a sign of binary, and those few
+      // characters are not going to be very significant.
+      continue;
+    }
+    // We could do the same disambiguition as RenameToIdentifier and
+    // GenerateNames do, but if we come up with a clashing name here it is
+    // likely a sign of not very meaningful binary data, so it is easier to
+    // just keep the original generated name in that case.
+    if (bh.count(s) != 0) {
+      continue;
+    }
+    // Remove original entry.
+    bh.erase(seg->name);
+    seg->name = s;
+    bh.emplace(s, Binding(static_cast<Index>(&seg - &segs[0])));
+  }
+}
+
 // Function names may contain arbitrary C++ syntax, so we want to
 // filter those to look like identifiers. A function name may be set
 // by a name section (applied in ReadBinaryIr, called before this function)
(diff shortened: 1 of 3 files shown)


src/decompiler-naming.h
```diff
@@ -110,6 +110,40 @@ void RenameToIdentifiers(std::vector<T*>& things, BindingHash& bh,
...
+    s = "d_";
+    for (auto c : seg->data) {
+      if (isalnum(c) || c == '_') {
+        s += static_cast<char>(c);
+      }
+      if (s.size() >= max_content_identifier_size) {
+        break;
+      }
+    }
+    if (s.size() < min_content_identifier_size) {
```
jgravelle-google: My first thought was "why have a minimum at all?" Probably worth a comment saying why too-short names aren't useful (which is alluded to in the following comment about meaningful binary data)

test/decompile/basic.txt
```diff
@@ -119,8 +119,8 @@ import table ns_tab3:funcref;
 table T_b:funcref(min: 0, max: 10);
 export table tab2:funcref(min: 0, max: 11);
 
-data d_a(offset: 0) = "Hello, World!\0a\00";
-data d_b(offset: 100) = 
+data d_HelloWorld(offset: 0) = "Hello, World!\0a\00";
+data d_abcdefghijklmnoqrstuvwxyzabc(offset: 100) =
```
jgravelle-google: Oh it wasn't obvious to me that the maximum would give a truncated identifier. I thought it would be symmetric with the minimum and not generate any name.

test/decompile/basic.txt
```diff
@@ -119,8 +119,8 @@ import table ns_tab3:funcref;
 table T_b:funcref(min: 0, max: 10);
 export table tab2:funcref(min: 0, max: 11);
 
-data d_a(offset: 0) = "Hello, World!\0a\00";
-data d_b(offset: 100) = 
+data d_HelloWorld(offset: 0) = "Hello, World!\0a\00";
```
jgravelle-google: Should add a data segment that tests the minimum case.

Also a test of the duplicate-identifier behavior.

src/decompiler-naming.h
aardappel: will do

test/decompile/basic.txt
aardappel: commented

test/decompile/basic.txt
aardappel: will do

aardappel:
@binji from personal testing with real world large wasm modules, most data segments are binary so it doesn't end up helping much, but it also doesn't hurt. A typical binary only segment would have been named `d_abc` before, and is now named `d_KkRR22Vl_` due to the accidental printable characters found inside, which ends up slightly more unique?

For sections containing any bit of text at all it seems really useful for example, so random picks from the binary I am look at:

```
data d_AZAD0truetype(offset: 365088) = 
  "\01\02\03\04\05\06\07\08\01\02\03\04\05\06\07\08\10\10\10\10\10\10\10\10"
  "\10\10\10\10\10\10\10\10\10\10\10\10\10\10\10\10\10\10\10\10\10\10\10\10"
  "                                \00 \00\00\00@\00\00\00\80\00\00\00@\00"
  "\00\a0\16\00\00A-\00\00\82Z\00\00A-\00\00\01\05\00\00D\00\00\00\f0\92\05"
  "\00\00\00\01\00\00\00\02\00\00\00\00\00\e0\07\00\00\e1\07\00\00\e2\07\00"
  "\00\18\03\00\000\01\00\00\a0\00\00\00\e3\07\00\00\e4\07\00\00\e5\07\00"
  "\00\e6\07\00\00\e7\07\00\00\00\00\00\00\e8\07\00\00\e9\07\00\00\00\00\00"
  "\00\ea\07\00\00\eb\07\00\00\ec\07\00\00truetype";
data d_fontformatTrueTypemultimaste(offset: 366048) = 
"font-format\00TrueType\00multi-masters";
data d_truetypeenginettglyfproperti(offset: 366092) = 
  "\ed\07\00\00\ee\07\00\00\ef\07\00\00truetype-engine\00\02\00\00\00tt-g"
  "lyf\00\f0\07\00\00properties\00\00\f1\07\00\00\f2\07\00\00interpreter-"
  "version\00\04\00\10\00\11\04\00\00\0c\02\04\00\0c\02\06\00\0c\02\08\00"
  "\0c\02\0a\00\0c\02\0c\00\0c\02\0e";
```

Clearly, seeing those names in code is going to be much more helpful than random :)

PR #1318, opened 2020-01-27
wasm-decompile: use symbols from linking section for names.

aardappel:
This allows wasm .o files to have more readable names, or even final
linked modules if the linking information is preserved (with e.g.
--emit-relocs in LLD).

This is implemented as part of the WABT IR representation, so
benefits wasm2wat as well.

Named obtained this way are only set for functions if the function
doesn't also have a name in the name section, but is preferred over
the export name if there is one.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -140,7 +140,9 @@ The currently supported list of tools (see
   binary files), then parse via `wasm2wat` and display the result
 - `run-gen-wasm-interp`: parse a "gen-wasm" text file, generate a wasm file,
   the run `wasm-interp` on it, which runes all exported functions in an
-  interpreter
+  interpreter.
+- `run-gen-wasm-decompile`: parse a "gen-wasm" text file (which can describe
+  invalid binary files), then parse via `wasm-decompile` and display the result.
 - `run-opcodecnt`: parse a wasm text file, convert it to binary, then display
   opcode usage counts.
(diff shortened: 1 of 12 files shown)


test/README.md
```diff
@@ -141,6 +141,8 @@ The currently supported list of tools (see
 - `run-gen-wasm-interp`: parse a "gen-wasm" text file, generate a wasm file,
   the run `wasm-interp` on it, which runes all exported functions in an
   interpreter
+- `run-gen-wasm`: parse a "gen-wasm" text file (which can describe invalid
```
binji: run-gen-wasm-decompile?

sbc100:
This seems to result in a segfault on the waterfall:
https://ci.chromium.org/p/wasm/builders/ci/linux

Gdb tells me the crash is on:

```
binary-reader-ir.cc:1263
1263	  DataSegment* seg = module_->data_segments[segment];
```

data_segments here is empty.

I'll attache the object file that causes this to fail if you like.

sbc100:
[torture__pr49394.C.o.zip](https://github.com/WebAssembly/wabt/files/4119953/torture__pr49394.C.o.zip)

aardappel:
@sbc100 fixed here, thanks! https://github.com/WebAssembly/wabt/pull/1319

PR #1319, opened 2020-01-28
wasm-decompile: Fix access of undefined symbols.

aardappel:


src/binary-reader-ir.cc
```diff
@@ -1255,6 +1255,10 @@ Result BinaryReaderIR::OnDataSymbol(Index index, uint32_t flags,
   if (name.empty()) {
     return Result::Ok;
   }
+  if (flags & WABT_SYMBOL_FLAG_UNDEFINED) {
```
binji: Is there a simple test we can add for this?

sbc100: Would need to be a gen-wasm-based test since we can't generate these symbols from wat files (yet).  

Maybe just modifying test/binary/linking-section.txt would be enough?

aardappel: It can't be `test/binary/linking-section.txt` since it has a data section, the test was triggered by the combination of a lack of data section and an undefined data symbol.

aardappel: btw should probably add to the waterfall to run the decompiler on all torture tests :)

sbc100: sgtm!

aardappel: Given that its already covered by the torture tests and this not likely to regress, I don't think a new test is needed.

PR #1320, opened 2020-01-28
Increase timeout; fix bug when a timeout occurs

binji:


test/run-tests.py
```diff
@@ -36,7 +36,7 @@
 TEST_DIR = os.path.dirname(os.path.abspath(__file__))
 REPO_ROOT_DIR = os.path.dirname(TEST_DIR)
 OUT_DIR = os.path.join(REPO_ROOT_DIR, 'out')
-DEFAULT_TIMEOUT = 10    # seconds
```
sbc100: This timeout actually helped me to figure out that I have made a performance regression a few weeks back.  However I guess there are probably better ways to measure that rather than measuring the speed of the CI machines.

PR #1322, opened 2020-01-30
wasm-decompile: escape hatch for variables used outside scope.

aardappel:
The decompiler assumes it can define a variable where it is first
assigned to, which works for almost all cases, but occasionally there
is a use of a variable outside of the scope where it was defined.

This detects that case, and makes sure that variable is pre-declared.

### DIFF
--- a/src/decompiler-ast.h
+++ b/src/decompiler-ast.h
@@ -22,7 +22,7 @@
 #include "src/ir.h"
 #include "src/ir-util.h"
 
-#include <set>
+#include <map>
 
 namespace wabt {
 
@@ -81,7 +81,7 @@ struct AST {
       mc.BeginFunc(*f);
       for (Index i = 0; i < f->GetNumParams(); i++) {
         auto name = "$" + IndexToAlphaName(i);
-        vars_defined.insert(name);
+        vars_defined.insert({ name, { 0, false }});
       }
     }
   }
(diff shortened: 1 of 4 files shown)


src/decompiler-ast.h
```diff
@@ -340,15 +355,21 @@ struct AST {
     if (size != 1) {
       InsertNode(NodeType::Statements, ExprType::Nop, nullptr, size);
     }
+    blocks_closed[cur_block_id] = true;
+    cur_block_id = block_stack.back();
+    block_stack.pop_back();
```
jgravelle-google: Wait, couldn't you just `cur_block_id = block_stack.pop_back();`?

Checking: Apparently not no. Dammit C++

src/decompiler-ast.h
```diff
@@ -340,15 +355,21 @@ struct AST {
...
   }
 
   ModuleContext& mc;
   std::vector<Node> exp_stack;
   std::vector<Node> predecls;
   const Func *f;
   int value_stack_depth = 0;
-  std::set<std::string> vars_defined;
+  std::map<std::string, std::pair<size_t, bool>> vars_defined;
```
jgravelle-google: Probably worth a comment what the pairs represent here. This is simple and scoped enough that a struct is probably overkill, but some clear reminder would be nice.

src/decompiler-ast.h
aardappel: Yes, the lack of that function has surprised me many times over.

src/decompiler-ast.h
aardappel: Actually, maybe make it a struct, given that I already had an occurrence of `ret.first->second.first` :)
Structs are so lightweight in recent C++, `pair` is almost an anti-pattern.

PR #1325, opened 2020-02-04
[interp] Unify active and passive segments

sbc100:
This simplifies the code and makes passive segments behave more like
active segments that are dropped on startup.

In also lays the ground work for the "declared" segments type which
is about be land in the next testsuite updated.


sbc100:
Turns out this wasn't needed for the declared elements change.. so I can probably abandon this change.

binji:
> Turns out this wasn't needed for the declared elements change.. so I can probably abandon this change.

Seems like a good change anyway.

binji:
Not needed anymore, since the interpreter rewrite.

PR #1327, opened 2020-02-04
Update testsuite

sbc100:
The main change here is the addition of declared elem sections.


src/interp/binary-reader-interp.cc
```diff
@@ -645,6 +649,17 @@ bool BinaryReaderInterp::OnError(const Error& error) {
   return true;
 }
 
+wabt::Result BinaryReaderInterp::EndModule() {
```
binji: could be `EndElemSection`, but this is fine too.

src/wast-parser.cc
```diff
@@ -646,23 +659,31 @@ Result WastParser::ParseVarList(VarVector* out_var_list) {
 
 bool WastParser::ParseElemExprOpt(ElemExpr* out_elem_expr) {
   Location loc = GetLocation();
-  if (MatchLpar(TokenType::RefNull)) {
+  if (MatchLpar(TokenType::Item)) {
+    if (!ParseElemExprOpt(out_elem_expr))
```
binji: This seems weird; we probably don't want this to be recursive, right?

src/wast-parser.cc
```diff
@@ -355,6 +355,19 @@ void ResolveFuncTypes(Module* module) {
   }
 }
 
+void MarkDeclaredFunctions(Module* module) {
```
binji: This logic should be in the validator, not the parser/binary-reader

src/interp/binary-reader-interp.cc
sbc100: Hmm, I tried that but it doesn't for if there are is no elem section at all.   Should we change the semantics so end for sections are called even if there is none?  I guess not.  I'll add a comment at least.

src/wast-parser.cc
sbc100: One problem with doing that is the validator generally doesn't mutate the IR (it takes a const Module).

I did consider doing this internally withing the validator using a side data structure rather then embedding this `is_declared` field in the IR itself.   Do you think that would be better approach?  I'll give that a go and see what you think..

src/interp/binary-reader-interp.cc
binji: Oh, good point. Yeah, it shouldn't call if the section is omitted.

src/wast-parser.cc
binji: Yeah, I'd prefer it that way I think. Storing it in the IR seems odd to me.

sbc100:
Should be good to go now.

PR #1330, opened 2020-02-07
New Interpreter

binji:
It's not quite ready yet, but I thought I should show the new design of the interpreter. Here are some of the goals:

* Match the structure of c-api + spec more closely (modules, instances, etc)
* Add support for garbage collected objects (currently just the same parts that c-api requires, but designed to be extensible)
* Unify decoding of the instruction stream (istream):
  * Used for interpreter, disassembler, tracer

For now I put everything under the `interp2` namespace, but I'm planning to move it over to `interp` when it is finished.


aardappel:
Hmm, I guess this is not that easy to review exactly because it is a copy, then again if it wasn't the diff would probably be crazy too. Anything in particular you want us to look at, or would you prefer this to be reviewed as if it was entirely new?

binji:
@aardappel yes, good point. Here's an overview of the directory layout:

* interp2.{h,cc,inl.h}
  * Main source file with new runtime. It's separated into three sections:
    * Types: this section includes the various "types" that can be exposed by the module interface (usually through imports/exports). e.g. FuncType, GlobalType, etc. It's meant to mirror the types in the wasm-c-api
    * Structure: this section describes the static structure of a Module. It has no "runtime" information, e.g. the current size of memory. All of the `struct`s are called `*Desc` to keep the names distinct from their runtime equivalents.
    * Runtime: this section is the new runtime of the interpreter, and includes most of the interesting parts of this PR. It is also modeled after the wasm-c-api. 
* istream.{h,cc}: These files contain the instruction stream, which is generated when "compiling" a module. The instruction stream does not follow the wasm binary format, and is designed to be easier to interpreter. For example, all control flow is lowered to `br` and `br_unless`, which are just goto instructions that jump to a location in the istream.
* read-module.{h,cc}: This is the equivalent of `binary-reader-interp` in the previous interpreter. It reads the binary format and converts it to a `ModuleDesc`, which includes an `Istream`. Unlike the previous interpreter, it does not instantiate the module -- that is a separate step. This means that the code is a lot simpler than the previous interpreter.

binji:
Here are the interesting parts you may want to look at:

 in `interp2.{h,cc}`:

`class Store`: this holds all garbage-collected objects, and has the mark-sweep GC code.
`class Instance`: this has new code for instantiation.
`class Thread`: This is similar to the thread implementation from the previous interpreter, but it's simplified to use float values directly (rather than use `ValueTypeRep` everywhere).

in `istream.{h,cc}`:

The disassemble/tracing code is unified, so it isn't as tedious to implement. They both use `Istream::Disassemble`, but provide a different `Istream::TraceSource`. They use the `InstrKind` to distinguish the structure of the instruction (in the istream), to reduce the number of different cases to handle.

You can take a look at `test-interp2.cc` to see what the new API looks like in practice.

sbc100:
Mostly locks logical and make sense to me.

Is there any way we can land it in pieces?  e.g. could we replace the binary reader and streamer while leaving everything else alone?  Or would that be way more work than is worth the effort?

binji:
> Is there any way we can land it in pieces?

I was trying to think of a way to do this, but I think it's all too interconnected. The old binary-reader-interp combines compiling + instantiation, so that can't really be replaced without changing the runtime. But the old runtime (`Environment`) also doesn't separate out modules + instances either, so that would have to be changed too. Maybe the `Istream` stuff could be done incrementally, but that's a pretty small part of this PR.

binji:
Yay, finally green!!

There are a few small changes that I can separate out, so I'll pull them into their own PRs.

binji:
OK, I've separated out and landed some of the independent changes, and squashed this PR. PTAL!

src/tools/wasm-interp.cc
```diff
@@ -16,46 +16,41 @@
...
-enum class RunVerbosity {
-  Quiet = 0,
-  Verbose = 1,
-};
+std::unique_ptr<FileStream> s_log_stream;
+std::unique_ptr<FileStream> s_stdout_stream;
+
+Store s_store;
+
+}  // namespace
```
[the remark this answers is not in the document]
binji: I thought I was following the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Unnamed_Namespaces_and_Static_Variables), but it seems to allow use of static too. I'll switch it back to static.

src/interp/istream.h
```diff
@@ -0,0 +1,151 @@
...
+    u32 imm_u32;
+    f32 imm_f32;
+    u64 imm_u64;
+    f64 imm_f64;
+    v128 imm_v128;
+    struct { u32 fst, snd; } imm_u32x2;
+  };
+};
+
+struct Istream {
```
sbc100: I would have though something like this with non-trivial methods and private fields would be a class?  Do you have set of rules for when you choose class vs struct?

src/interp/istream.h
```diff
@@ -0,0 +1,151 @@
...
+using u64 = uint64_t;
+using f32 = float;
+using f64 = double;
+
+using Buffer = std::vector<u8>;
+
+using ValueType = wabt::Type;
+
+enum class InstrKind {
+  Imm_0_Op_0,             // Nop
```
sbc100: Is the comment on the right here just an example?

src/opcode.cc
```diff
@@ -24,17 +24,23 @@ namespace wabt {
...
+     0,
+     0,
+     0,
+     0},
 };
 
 #define WABT_OPCODE(rtype, type1, type2, type3, mem_size, prefix, code, Name, \
-                    text, decomp)                                                     \
+                    text, decomp)                                             \
   /* static */ Opcode Opcode::Name##_Opcode(Opcode::Name);
```
sbc100: Changes to this file look like formatting only?   Separate PR?

src/interp/test-interp.cc
```diff
@@ -0,0 +1,549 @@
...
+
+  Values results;
+  Trap::Ptr trap;
+  ASSERT_EQ(Result::Ok, rot13->Call(store_, {}, results, &trap));
+
+  EXPECT_EQ("Uryyb, JroNffrzoyl!", string_data);
+
+  ASSERT_EQ(Result::Ok, rot13->Call(store_, {}, results, &trap));
+
+  EXPECT_EQ("Hello, WebAssembly!", string_data);
```
sbc100: Are the tests in this file basically unchanged in terms of what the actually assert?  (github isn't showing a diff from the old file).

src/opcode.def
```diff
@@ -48,16 +48,16 @@ WABT_OPCODE(___,  ___,  ___,  ___,  0,  0,    0x09, Rethrow, "rethrow", "")
 WABT_OPCODE(___,  ___,  ___,  ___,  0,  0,    0x0a, BrOnExn, "br_on_exn", "")
 WABT_OPCODE(___,  ___,  ___,  ___,  0,  0,    0x0b, End, "end", "")
 WABT_OPCODE(___,  ___,  ___,  ___,  0,  0,    0x0c, Br, "br", "")
-WABT_OPCODE(___,  ___,  ___,  ___,  0,  0,    0x0d, BrIf, "br_if", "")
```
sbc100: Remind me what these changes do again?  Would these have been useful outside of this change?  (I guess I'm saying can this be separated?)

src/interp/istream.h
binji: No, I should probably be more consistent about this. You're right that this should be a class.

src/interp/istream.h
binji: Yeah, I'll add a comment mentioning it.

src/opcode.cc
binji: No, there are new braces around the `type1`, `type2`, `type3` now. This makes it possible to access the types by index.

src/interp/test-interp.cc
binji: Some are new, I think it shows up different because the file is in a different location. I'll move it back to the old place.

src/opcode.def
binji: This makes it easier to handle tracing; these types can be extracted directly so the values can properly be displayed in the interpreter (e.g. what type to use for a given `Value`)

binji:
I tried to make more stuff similar to upstream (in particular `read-module.cc` is renamed back to `binary-reader-interp.cc`). Hopefully this is easier to see the differences now. Sorry for all the churn!

src/test-interp.cc
```diff
@@ -16,183 +16,391 @@
...
-interp::Result TrapCallback(const interp::HostFunc* func,
-                            const interp::FuncSignature* sig,
-                            const interp::TypedValues& args,
-                            interp::TypedValues& results) {
-  return interp::ResultType::TrapHostTrapped;
-}
+class InterpTest : public ::testing::Test {
+ public:
+  virtual void SetUp() {}
+  virtual void TearDown() {}
```
sbc100: Are these needed?

src/test-interp.cc
```diff
@@ -16,183 +16,391 @@
...
+      0x01, 0x7f, 0x01, 0x7f, 0x60, 0x01, 0x7d, 0x01, 0x7d, 0x60, 0x00, 0x00,
+      0x02, 0x0b, 0x01, 0x03, 0x66, 0x6f, 0x6f, 0x03, 0x62, 0x61, 0x72, 0x00,
+      0x00, 0x03, 0x03, 0x02, 0x01, 0x02, 0x04, 0x05, 0x01, 0x70, 0x01, 0x01,
+      0x02, 0x05, 0x03, 0x01, 0x00, 0x01, 0x06, 0x06, 0x01, 0x7f, 0x00, 0x41,
+      0x01, 0x0b, 0x07, 0x08, 0x01, 0x04, 0x71, 0x75, 0x75, 0x78, 0x00, 0x01,
+      0x08, 0x01, 0x02, 0x09, 0x08, 0x01, 0x00, 0x41, 0x00, 0x0b, 0x02, 0x00,
+      0x01, 0x0a, 0x0c, 0x02, 0x07, 0x00, 0x43, 0x00, 0x00, 0x28, 0x42, 0x0b,
+      0x02, 0x00, 0x0b, 0x0b, 0x0b, 0x01, 0x00, 0x41, 0x02, 0x0b, 0x05, 0x68,
+      0x65, 0x6c, 0x6c, 0x6f,
+  });
```
sbc100: I wonder if (in the long run) these would be better of as binary files on disk?  As it stands it quite hard debug these tests.   For example, I remember having an issue with them and wanting to be able to run this binary through other tools to inspect if. 

Just a thought.  Then we could keep the text versions alongside and pragmatically regenerate them periodically.  Not related to this PR.

src/test-interp.cc
```diff
@@ -16,183 +16,391 @@
...
+ public:
+  virtual void SetUp() {}
+  virtual void TearDown() {}
 
-Features s_features;
+  void ReadModule(const std::vector<u8>& data) {
+    Errors errors;
+    ReadBinaryOptions options;
+    Result result =
+        ::ReadModule(data.data(), data.size(), options, &errors, &module_desc_);
```
sbc100: I find the fact that `::` works here it be a little magically.   I'm not quite sure what C++ is doing here.. I guess its find :)

src/test-interp.cc
binji: Yeah, I saw another interpreter use this technique. Seemed pretty cool!

src/test-interp.cc
binji: Removed.

src/test-interp.cc
binji: Renamed back to `ReadBinaryInterp`, so this isn't necessary anymore.

PR #1332, opened 2020-02-11
Cygwin build fixes

okuoku:
Build fixes to run wabt tools on Cygwin64. I haven't tested it much, but at least it runs DOOM https://github.com/okuoku/freestandoom/issues/2 with the interpreter + WASM C-API :)

- CMakeLists.txt -- Cygwin's libc(newlib) doesn't export POSIX API `fileno` under `-std=c++11`; use `-std=gnu++11` instead.
- binary-reader-objdump.cc: Use proper format string. Using `%lu` here would take some garbage, as argument is `uint32_t`. I guess it can be `PRIu32` but followed other occurrences there.
- (not Cygwin specific) decompiler.cc: Use unsigned `1U`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,7 +118,15 @@ else ()
     -Wuninitialized
   )
 
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wold-style-cast")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wold-style-cast")
+
+  if(NOT CYGWIN)
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+  else()
+      # On Cygwin, POSIX extensions are not visible with std=c++11 
+      # because it defines __STRICT_ANSI__.
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11")
+  endif()
 
   if (WERROR)
     add_definitions(-Werror)
--- a/src/binary-reader-objdump.cc
+++ b/src/binary-reader-objdump.cc
@@ -598,7 +598,7 @@ Result BinaryReaderObjdumpDisassemble::OnOpcodeUint32(uint32_t value) {
 Result BinaryReaderObjdumpDisassemble::OnOpcodeUint32Uint32(uint32_t value,
                                                             uint32_t value2) {
   Offset immediate_len = state->offset - current_opcode_offset;
-  LogOpcode(immediate_len, "%lu %lu", value, value2);
+  LogOpcode(immediate_len, "%u %u", value, value2);
   return Result::Ok;
 }
 
--- a/src/decompiler.cc
+++ b/src/decompiler.cc
@@ -302,7 +302,7 @@ struct Decompiler {
         if (se.opcode == Opcode::I32Shl &&
             const_exp.etype == ExprType::Const) {
           auto& ce = *cast<ConstExpr>(const_exp.e);
-          if (ce.const_.type == Type::I32 && (1 << ce.const_.u32) == align) {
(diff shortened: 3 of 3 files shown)


sbc100:
Cool!  Can you point to the code where use used the C-API?  I'm keen to see it an action.

CMakeLists.txt
```diff
@@ -118,7 +118,15 @@ else ()
...
 
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wold-style-cast")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wold-style-cast")
+
+  if(NOT CYGWIN)
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+  else()
+      # On Cygwin, POSIX extensions are not visible with std=c++11 
+      # because it defines __STRICT_ANSI__.
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11")
```
sbc100: Would it be better to define _POSIX_SOURCE or something like that to signal that we want POSIX rather than asking for GNU extensions?  I also don't think its a great idea to use different `-std` on different platforms.

CMakeLists.txt
```diff
@@ -118,7 +118,15 @@ else ()
     -Wuninitialized
   )
 
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wold-style-cast")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wold-style-cast")
```
sbc100: Why this?  Should we fix the instances instead?

CMakeLists.txt
okuoku: Yup, there's room for improvement. Currently, the only POSIX consumers are https://github.com/WebAssembly/wabt/blob/4ba97c116d1417b2fb516299a37965820b93ced3/src/color.cc#L69 AND gtest. For the `color.cc` we'd just have `_POSIX_C_SOURCE` on it but I'm not sure how do we want to deal with gtest... 

So I choose to limit `gnu++11` to Cygwin and leave it as-is on the other (major) implementations such as glibc(Linux) or BSD libc.

CMakeLists.txt
okuoku: Sorry, I'm not sure I get what you mean; `-Wold-style-cast` is C++ only thus I think it could not be absorbed into `add_definitions` (as we're on CMake3, perhaps they meant `add_compile_options`) above. It seems it was introduced in #320.

okuoku:
> Can you point to the code where use used the C-API? 

Not more than copy/pasted from hello.c: https://github.com/okuoku/freestandoom/blob/9150923c78a169d1563100fbc85a1302c0bab5df/host-common/callwasm.c

I found current C API implementation lacks `wasm_module_imports` API so I had to rely import order to expose native APIs to compiled WASM module. It seems the upstream does not have any example on the API anyway.

Currently it does not have any proper build system/instructions(sorry!) video instead: https://www.youtube.com/watch?v=EBF7E6_qkn8

CMakeLists.txt
sbc100: Sorry I mis-read the diff.  Through you were adding the `-W` flag.

CMakeLists.txt
sbc100: I think it would make sense to define _POSIX_C_SOURCE for the whole project (assuming that works) than turn on GNU extensions.

okuoku: Okay, I'll try. I found gtest upstream also have a proposed patch https://github.com/google/googletest/pull/2653/files so we should be able to do the same here.

The only culprit seems `strcasecmp` which is also POSIX. I'll chase why Cygwin does not export it under `_POSIX_C_SOURCE=200809L`.

okuoku: Filed: #1333

PR #1333, opened 2020-02-15
Use standard C++11 on GNU/Clang compilers

okuoku:
Following up #1332 . Enforce standard C++11 in more sane way.

- `binary-reader-objdump.cc` -- it lacked `<strings.h>` to import `strcasecmp` which is not part of standard C++
- Adjust CMake variables to use `std=c++11` everywhere. `CMAKE_CXX_EXTENSIONS OFF` is required to teach CMake 3.1+ not to inject `std=gnu++11` reliably. `_POSIX` on MinGW is no longer required with `_POSIX_C_SOURCE`.

### DIFF
--- a/src/binary-reader-objdump.cc
+++ b/src/binary-reader-objdump.cc
@@ -22,6 +22,7 @@
 #include <cstdio>
 #include <cstring>
 #include <vector>
+#include <strings.h>
 
 #include "src/binary-reader-nop.h"
 #include "src/filenames.h"
--- a/src/binary-reader-objdump.cc
+++ b/src/binary-reader-objdump.cc
@@ -22,7 +22,10 @@
 #include <cstdio>
 #include <cstring>
 #include <vector>
+
+#if HAVE_STRCASECMP
 #include <strings.h>
+#endif
 
 #include "src/binary-reader-nop.h"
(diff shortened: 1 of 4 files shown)


okuoku:
Ugh `sysconf` ... `_SC_NPROCESSORS_ONLN` is not POSIX thus we need `_BSD_SOURCE` or something. 

Why not `std::thread::hardware_concurrency` being used here?

src/binary-reader-objdump.cc
```diff
@@ -23,6 +23,10 @@
 #include <cstring>
 #include <vector>
 
+#if HAVE_STRCASECMP
+#include <strings.h>
+#endif
```
sbc100: Given that we don't guard use of strcasecmp why guard the include?

okuoku: On MSVC, it's polyfilled with `_strcmpi` which is in `<string.h>` and it's included in `"config.h"` thus no additional include required.

https://github.com/WebAssembly/wabt/blob/4ba97c116d1417b2fb516299a37965820b93ced3/src/config.h.in#L288

I choose not to include `<strings.h>` in `"config.h"` because it's not a standard C header.

Issue #1339, opened 2020-02-21
Can't decode LLVM-built i8x16.shuffle


abrown:
LLVM still builds code using the old encoding for `i8x16.shuffle`--`0xfd03`. WABT expects code to have the [new encoding](https://github.com/WebAssembly/simd/blob/master/proposals/simd/BinarySIMD.md)--`0xfdc1`. Over in an [emscripten issue](https://github.com/emscripten-core/emscripten/issues/9281#issuecomment-589339146), @tlively mentioned that he would prefer to wait to modify all of the LLVM-emitted encodings until after all the encodings are renumbered. In the meantime, could we modify WABT to allow both `0xfd03` and `0xfdc1` to parse as `i8x16.shuffle`? Otherwise the WABT tools can't really be used to look at LLVM Wasm SIMD code (and I like WABT!).

I took a stab at this but quickly ran into things I don't yet understand about WABT. I wouldn't mind doing the work to get this working if 1) you all agree this type of temporary hack can be allowed (similar code exists in [wasmparser.rs](https://github.com/bytecodealliance/wasmparser/blob/4789a9cadd6390579df8574b4e2bba810d428fe9/src/binary_reader.rs#L1562-L1568)), and 2) someone can help me understand what I don't understand about `infos_` (and probably other locations that need changing):

```diff
diff --git a/src/opcode-code-table.c b/src/opcode-code-table.c
index c3e06d03..dcdbc060 100644
--- a/src/opcode-code-table.c
+++ b/src/opcode-code-table.c
@@ -38,4 +38,6 @@ uint32_t WabtOpcodeCodeTable[WABT_OPCODE_CODE_TABLE_SIZE] = {
   [(prefix << 8) + code] = Name,
 #include "opcode.def"
 #undef WABT_OPCODE
+  // FIXME: alias V8X16Shuffle (currently numbered as 0xfdc1) to its previous opcode (0xfd03); remove when SIMD instructions are renumbered
+  [(0xfd << 8) + 0x03] = V8X16Shuffle,
 };
diff --git a/src/opcode.cc b/src/opcode.cc
index 5327f4b4..46d91e0b 100644
--- a/src/opcode.cc
+++ b/src/opcode.cc
@@ -30,6 +30,8 @@ Opcode::Info Opcode::infos_[] = {
 #include "src/opcode.def"
 #undef WABT_OPCODE
 
[...]

PR #1340, opened 2020-02-21
Add the old alias for v8x16.shuffle

binji:
It used to have the encoding 0xfd03, but now is 0xfdc1. It is useful to
keep the old encoding around though.

### DIFF
--- a/src/opcode-code-table.c
+++ b/src/opcode-code-table.c
@@ -38,4 +38,8 @@ uint32_t WabtOpcodeCodeTable[WABT_OPCODE_CODE_TABLE_SIZE] = {
   [(prefix << 8) + code] = Name,
 #include "opcode.def"
 #undef WABT_OPCODE
+
+  // TODO(binji): LLVM still generates 0xfd03 for v8x16.shuffle, see
+  // https://github.com/WebAssembly/wabt/issues/1339
+  [(0xfd << 8) + 0x03] = V8X16Shuffle,
 };
--- a/test/binary/v8x16shuffle-alias.txt
+++ b/test/binary/v8x16shuffle-alias.txt
@@ -1,0 +1,24 @@
+;;; TOOL: run-gen-wasm
+;;; ARGS1: --enable-simd
+;;; ARGS2: --enable-simd
+magic
+version
+section(TYPE) { count[1] function params[0] results[1] v128 }
+section(FUNCTION) { count[1] type[0] }
+section(CODE) {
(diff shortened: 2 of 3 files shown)


binji:
@abrown I think this should be enough to make it work. If not, please let me know!

src/opcode-code-table.c
```diff
@@ -38,4 +38,8 @@ uint32_t WabtOpcodeCodeTable[WABT_OPCODE_CODE_TABLE_SIZE] = {
   [(prefix << 8) + code] = Name,
 #include "opcode.def"
 #undef WABT_OPCODE
+
+  // TODO(binji): LLVM still generates 0xfd03 for v8x16.shuffle, see
+  // https://github.com/WebAssembly/wabt/issues/1339
+  [(0xfd << 8) + 0x03] = V8X16Shuffle,
```
abrown: I tried this change over in #1339 but when I use it to decode old-numbered shuffles with ` make && bin/wasm-objdump -d --debug ../old.wasm` I get:
```
...
on_opcode: 0x3c82: v8x16.shuffle
warning: 0x3c82 missing opcode callback at 0x3c81 (0xfd=<invalid>)
```
This is what led me to believe that there might be a change needed somewhere else?

binji: Ah, I see what you mean. It's annoying because the new encoding is 3 bytes and the old encoding is 2 bytes. So it doesn't alias quite right. Not sure the least hacky way to fix this... :-)

binji: Sounds like the new renumbering should be happening soon -- perhaps the simplest solution is for you to recompile wabt with the old numbering. Sorry, but it seems like it's not quite as easy as I'd hope to add it in!

tlively:
It would be more turn, but it actually might be useful for WABT to switch back to the old numbering entirely. All other tools currently emit the old number and all engines also accept the old number, so I don't think that would cause issues.

binji:
Sounds good, it's not that much churn to renumber now. See https://github.com/WebAssembly/wabt/pull/1342

src/opcode-code-table.c
abrown: I see #1342... thanks! But am I confused? I thought both encodings were 2 bytes long?

tlively: The prefix byte is a single byte, but the opcode after that is LEB128 encoded, so numbers above 0x7f take more than one byte to encode.

PR #1346, opened 2020-02-27
Share validator between IR + binary-reader-interp

binji:
The TypeChecker was already shared, but the rest of the other validation
logic was duplicated. This change creates a SharedValidator which is
used by both the Validator and the BinaryReaderInterp classes.

The validator is structured similarly to TypeChecker as a collection of
functions. It's assumed that the functions will be called in the same
order as sections occur in the binary format. The IR valiator does this
too, even though it's possible to validate components out-of-order in
that case.

This change also splits Module validation and Script validation into two
different classes. It should have been written this way in the first
place, and it's nice to do as part of this change since the module
validator logic is mostly moved into the SharedValidator anyway.

Next steps:

* Remove all validation from BinaryReader and move it into the
  SharedValidator.
* Move the TypeChecker into the SharedValidator (maybe not necessary)
* Ensure that validation occurs before creating IR from binary (use
  SharedValidator in BinaryReaderIR? or maybe create
  BinaryReaderValidator passthru that both BinaryReaderIR and
  BinaryReaderInterp use?)
* Rename Validator -> IRValidator, SharedValidator -> Validator

### DIFF
--- a/test/spec/bulk-memory-operations/memory_init.txt
+++ b/test/spec/bulk-memory-operations/memory_init.txt
@@ -9,7 +9,7 @@ test() =>
 out/test/spec/bulk-memory-operations/memory_init.wast:189: assert_invalid passed:
   0000023: error: data.drop requires data count section
 out/test/spec/bulk-memory-operations/memory_init.wast:195: assert_invalid passed:
-  error: invalid data_segment_index: 4 (max 1)
+  0000000: error: data_segment variable out of range: 4 (max 0)
   000002c: error: OnDataDropExpr callback failed
 test() =>
 out/test/spec/bulk-memory-operations/memory_init.wast:216: assert_trap passed: out of bounds memory access: memory.init out of bounds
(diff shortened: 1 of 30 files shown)


test/spec/bulk-memory-operations/memory_init.txt
```diff
@@ -9,15 +9,15 @@ test() =>
 out/test/spec/bulk-memory-operations/memory_init.wast:189: assert_invalid passed:
   0000023: error: data.drop requires data count section
 out/test/spec/bulk-memory-operations/memory_init.wast:195: assert_invalid passed:
-  error: invalid data_segment_index: 4 (max 1)
+  0000000: error: data_segment variable out of range: 4 (max 0)
```
sbc100: What does variable here mean?  Is an immediate to an instruction?

test/spec/call_indirect.txt
```diff
@@ -80,7 +80,7 @@ out/test/spec/call_indirect.wast:701: assert_malformed passed:
   ...0 funcref)(func (result i32)  (call_indirect (type $sig) (param i32) (resu...
                                     ^^^^^^^^^^^^^
 out/test/spec/call_indirect.wast:716: assert_invalid passed:
-  error: found call_indirect operator, but no table
+  0000000: error: table variable out of range: 0 (max 4294967295)
```
sbc100: max seems strange here.  Maybe special case the error?

test/spec/local_set.txt
```diff
@@ -83,22 +83,22 @@ out/test/spec/local_set.wast:329: assert_invalid passed:
   error: type mismatch in implicit return, expected [f64] but got []
   0000025: error: EndFunctionBody callback failed
 out/test/spec/local_set.wast:337: assert_invalid passed:
-  error: invalid local_index: 3 (max 2)
+  0000000: error: local variable out of range (max 2)
```
sbc100: Can you print the actual value anymore?

binji:
Yeah most of the changed error output is because the validator is shared now. I had to pick either going with the binary-reader-interp messages or the the IR validator messages, and chose to use the IR validator ones. Ultimately we'll want to clean up these messages, but I wanted to do that in a later PR.

sbc100:
Sure, none of them are a big deal.

PR #1347, opened 2020-02-27
Always check index before accessing vectors

binji:
Since the validator tries to print as many errors as possible, it
shouldn't stop when the index is invalid (since this may produce future
errors). Instead it uses a default value.

This is encoded in a new function called CheckIndexWithValue, which
either accesses the vector, or uses a default value.


src/shared-validator.cc
```diff
@@ -801,10 +793,6 @@ Result SharedValidator::OnDrop(const Location& loc) {
 Result SharedValidator::OnElemDrop(const Location& loc, Var segment_var) {
   Result result = Result::Ok;
   expr_loc_ = &loc;
-  // TODO: Remove, elem.drop doesn't require a table.
-#if 1
-  result |= CheckTableIndex(Var(0, loc), Opcode::ElemDrop);
```
sbc100: I wonder if this means we are missing a couple of spec tests?

test/typecheck/atomic-no-shared-memory.txt
```diff
@@ -0,0 +1,83 @@
+;;; TOOL: wat2wasm
+;;; ARGS: --enable-threads
+
+(module
+  (memory 1)
```
sbc100: Is this because we are changing the spec?  I'd expect to see this change along with a testsuite update.  Don't we have threads spec tests that check this?

src/shared-validator.cc
binji: Hm, good point.

test/typecheck/atomic-no-shared-memory.txt
binji: Hm, looks like the tests weren't updated.

src/shared-validator.cc
binji: https://github.com/WebAssembly/bulk-memory-operations/pull/140

test/typecheck/atomic-no-shared-memory.txt
binji: https://github.com/WebAssembly/threads/pull/150

PR #1350, opened 2020-02-29
Convert Type from an enum into a class

binji:
This is similar to the way Opcode is structured, which allows us to hang
member functions off of the enumeration.

The primary motivator for this change is the GC proposal (and the
function-references proposal) where a Type can be parameterized:

     (type $T (struct ...))
     (func
       (local (ref $T)
       ...
     )

In this case the type is ref, with a parameter of the type index. Making
Type a class will make it easier to store this additional information.

### DIFF
--- a/src/type.h
+++ b/src/type.h
@@ -87,6 +87,18 @@ class Type {
     }
   }
 
+  // Functions for handling types that are an index into the type section.
+  // These are always positive integers. They occur in the binary format in
+  // block signatures, e.g.
+  //
+  //   (block (result i32 i64) ...)
+  //
+  // is encoded as
+  //
+  //   (type $T (func (result i32 i64)))
+  //   ...
+  //   (block (type $T) ...)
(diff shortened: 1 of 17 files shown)


src/type.h
```diff
@@ -22,103 +22,103 @@
...
+      case Type::Void:    return "void";
+      case Type::Any:     return "any";
+      case Type::Anyref:  return "anyref";
+      case Type::Nullref: return "nullref";
+      default:            return "<type_index>";
+    }
   }
-}
+
+  bool IsIndex() const { return static_cast<int32_t>(enum_) >= 0; }
```
sbc100: Does this need a better name now?   Or maybe a comment.  This overloading of the meaning of the enum_ seems very strange now.

binji: Added a comment. Maybe could be renamed to `IsBlockTypeIndex` or something?

PR #1351, opened 2020-02-29
Update testsuite (w/ reference-types changes)

binji:
The new table-sub test, checks whether the subtyping is handled
properly w/ table.init and table.copy instructions.

The BeginElemSegment callback can't pass the element type anymore, since
it's not known yet. The callback also can't be deferred, since the
BeginElemSegmentInitExpr callback has to happen after the
BeginElemSegment callback, but the reference type is not always known
until after the initializer expression is read. To work around this, I
added a new OnElemSegmentElemType callback.

Other element segment changes:

* The element type must be tracked in the SharedValidator
* A subtle fix: when writing out the segment flags, we need to take into
  account whether the element type of the segment is not funcref, even
  if there are no element expressions. In that case, we have to use flag
  bit 0x4 (SegUseElemExprs).

In addition, the TableCopy and TableInit instructions weren't handling
table indexes fully.

* TableCopy variables are read in the parser (both optional)
* TableCopy names are now resolved + applied
* TableCopy indexes are now validated
* TableInit table variables are read in the parser; this is subtle,
  since the text format has order $table $segment, but the $table is
  optional.

### DIFF
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -2220,7 +2220,7 @@ Result BinaryReader::ReadElemSection(Offset section_size) {
     }
     Type elem_type = Type::Funcref;
 
-    CALLBACK(BeginElemSegment, i, table_index, flags, elem_type);
+    CALLBACK(BeginElemSegment, i, table_index, flags);
 
     if (!(flags & SegPassive)) {
       CALLBACK(BeginElemSegmentInitExpr, i);
@@ -2243,7 +2243,10 @@ Result BinaryReader::ReadElemSection(Offset section_size) {
                      GetTypeName(elem_type));
         elem_type = Type::Funcref;
       }
+
     }
+
+    CALLBACK(OnElemSegmentElemType, i, elem_type);
 
     Index num_elem_exprs;
     CHECK_RESULT(ReadCount(&num_elem_exprs, "elem count"));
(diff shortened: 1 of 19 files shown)


src/binary-reader.cc
```diff
@@ -2243,8 +2243,11 @@ Result BinaryReader::ReadElemSection(Offset section_size) {
                      GetTypeName(elem_type));
         elem_type = Type::Funcref;
       }
+
```
sbc100: Did you mean to add this newline?

src/binary-reader.cc
```diff
@@ -2220,7 +2220,7 @@ Result BinaryReader::ReadElemSection(Offset section_size) {
     }
     Type elem_type = Type::Funcref;
 
-    CALLBACK(BeginElemSegment, i, table_index, flags, elem_type);
+    CALLBACK(BeginElemSegment, i, table_index, flags);
```
sbc100: So was this simple report the Funcref in all cases before?  It looks like it.  I think I remember having a mental TODO about this.

binji: Exactly, the elem_type was changed below, but the callback always would pass Funcref.

src/binary-reader.cc
binji: Nope. :)

PR #1352, opened 2020-03-03
Initial pass parsing/reading struct

binji:
This parses just the format `(struct)` as a new type. I added a test for
this using `wat2wasm`, but that requires a rudimentary binary format.
The test runner automatically attempts to rountrip all wat2wasm tests,
so this required implementing the wat writing and binary reading too.

Here's a summary of the changes:

* binary-reader:h: Rename `BinaryReader::OnType` callbacks to `OnFuncType`
* binary-reader.h: Add `BinaryReader::OnStructType`
* binary-reader.cc: Use a switch after reading the type form to
    determine whether we're reading a function or struct.
* tokens.def: Add new `TokenType::Struct`
* lexer-keywords.txt: Add new `struct` keyword
* type.h: Add `Type::Struct` type form
* wast-parser.cc: Parse `(struct)` in text format
* wat-writer.cc: Write Func or Struct type forms


src/binary-reader.h
```diff
@@ -82,11 +82,12 @@ class BinaryReaderDelegate {
...
-                        Index param_count,
-                        Type* param_types,
-                        Index result_count,
-                        Type* result_types) = 0;
+  virtual Result OnFuncType(Index index,
+                            Index param_count,
+                            Type* param_types,
+                            Index result_count,
+                            Type* result_types) = 0;
+  virtual Result OnStructType(Index index) = 0;  // TODO
```
jgravelle-google: Not sure what the TODO here is for?

binji: Oh yeah, this was a TODO to add information about the struct fields to the callback. I'll remove it in the next PR. :-)

PR #1354, opened 2020-03-04
Remove validation from the BinaryReader

binji:
Validation should only happen in ValidateModule, BinaryReader should
only check whether the binary is malformed.

This change also fixes a few places in BinaryReaderIR where an index is
assumed to be valid.

### DIFF
--- a/src/binary-reader-ir.cc
+++ b/src/binary-reader-ir.cc
@@ -270,6 +270,7 @@ class BinaryReaderIR : public BinaryReaderNop {
   Result TopLabel(LabelNode** label);
   Result TopLabelExpr(LabelNode** label, Expr** expr);
   Result AppendExpr(std::unique_ptr<Expr> expr);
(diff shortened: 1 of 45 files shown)


aardappel:
Cool, sounds like this would make it even less likely to hit asserts caused by module contents?

But now its more important than before that users validate before accessing.. are there any that don't?

binji:
I fixed bugs / removed assertions in `BinaryReaderIR`, so it will be relatively safe to use even without full validation. I don't think there are any current users who aren't validating but should be; the only ones are in `wasm-objdump`, `wasm-opcodecnt`, `wasm-strip`, and optionally `wasm2wat`. All of those tools don't need to validate though.

Since I separated out validation into a set of functions, it is possible to create a new pass-through reader that can read+validate directly into IR. That way validation only has to be run separately when parsing text to IR. I might do that as a next step.

PR #1355, opened 2020-03-09
Parse struct fields

binji:
This allows the following field formats:

* `(struct (field $name i32))`
* `(struct (field $name (mut i32)))`
* `(struct (field i32))`
* `(struct (field (mut i32)))`
* `(struct (mut i32))`
* `(struct i32)`


binji:
@jgravelle-google ping

binji:
Looks like the travis failures are flakes?

src/binary-reader.h
```diff
@@ -51,6 +51,12 @@ struct ReadBinaryOptions {
   bool fail_on_custom_section_error = true;
 };
 
+// TODO: Move somewhere else?
```
jgravelle-google: Probably yes, though I couldn't say where. Seems reasonable to leave it here at least.

src/shared-validator.h
```diff
@@ -169,10 +173,23 @@ class SharedValidator {
...
   struct FuncType {
+    FuncType() = default;
+    FuncType(const TypeVector& params, const TypeVector& results)
+        : params(params), results(results) {}
+
     TypeVector params;
     TypeVector results;
   };
 
+  using TypeMutVector = std::vector<TypeMut>;
```
jgravelle-google: Nit: this seems dubiously-useful, and we still have two `std::vector<TypeMut>`s in this PR. I'd either move this to be next to `TypeMut`'s declaration so it's usable everywhere, or just use naked std::vectors.

binji: Good point, I moved it to where `TypeMut` is defined, and used it everywhere. I like it a bit better than using `std::vector` everywhere.

PR #1357, opened 2020-03-13
Fix interpreter garbage collector

binji:
The interpreter has a garbage colletor (Store::Collect), that *almost*
worked.

In particular, the free list had no fast way to determine whether an
object was used or not. This is necessary when iterating over objects to
actually delete them. This commit adds an `is_free_` vector for this
purpose.

The `Store::Collect` method was also not correct; it was calling
`Store::Mark(Ref)` on each of the GC roots, but that only marks the ref
itself, it doesn't iterate over the Object's references. This commit
updates Collect to iterate over all marked objects iteratively until
steady-state is reached. It's potentially quadratic in the number of
objects as currently written, so there's definitely room for
improvement.

### DIFF
--- a/src/interp/interp.cc
+++ b/src/interp/interp.cc
@@ -254,7 +254,8 @@ void Store::DeleteRoot(RootList::Index index) {
 }
 
 void Store::Collect() {
-  marks_.resize(objects_.size());
+  size_t object_count = objects_.size();
+  marks_.resize(object_count);
   std::fill(marks_.begin(), marks_.end(), false);
 
   // First mark all roots.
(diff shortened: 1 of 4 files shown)


src/interp/interp.cc
```diff
@@ -254,10 +254,34 @@ void Store::DeleteRoot(RootList::Index index) {
 }
 
 void Store::Collect() {
+  marks_.resize(objects_.size());
```
sbc100: Maybe add `size_t object_count = objects_.size()` which you can then use here and below on lines 269, 273 and 383.

binji: Good idea, done.

PR #1364, opened 2020-03-20
Parse ArrayTypes

binji:
The following formats are supported:

* (type (array i32))
* (type (array (field i32)))
* (type (array (field (mut i32))))

This PR adds support for reading/writing binary and text, but no
interpreter support yet.

### DIFF
--- a/src/binary-reader-logging.cc
+++ b/src/binary-reader-logging.cc
@@ -99,6 +99,16 @@ void BinaryReaderLogging::LogTypes(TypeVector& types) {
   LogTypes(types.size(), types.data());
 }
 
+void BinaryReaderLogging::LogField(TypeMut field) {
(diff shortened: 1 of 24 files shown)


src/binary-reader-logging.cc
```diff
@@ -149,13 +159,7 @@ Result BinaryReaderLogging::OnStructType(Index index,
...
   LOGF_NOINDENT("[");
   for (Index i = 0; i < field_count; ++i) {
-    if (fields[i].mutable_) {
-      LOGF_NOINDENT("(mut ");
-    }
-    LogType(fields[i].type);
-    if (fields[i].mutable_) {
-      LOGF_NOINDENT(")");
-    }
+    LogField(fields[i]);
```
jgravelle-google: Nice

PR #1369, opened 2020-03-22
Always generate DataCount section, even if number of datas is 0

### LINKED ISSUE #1368: wast2json not generating wasm with DataCount section when data.drop is used in code
In https://github.com/WebAssembly/bulk-memory-operations there is a test that checks for a validation error against `data.drop 0`.

https://github.com/WebAssembly/bulk-memory-operations/blob/ffdbb6e357d79818c88c3557cd3325346a057d27/test/core/memory_init.wast#L188

This test fails with a malformed error instead because no `DataCount` section is generated from `wast2json` from the latest revision of wabt https://github.com/WebAssembly/wabt/commit/5ba0296190a077acfdc6c1de19510143b4dcc7c4, presumably because no `Data` section is present.

This violates the current version of the bulk memory spec, that a wasm binary without a `DataCount` section is malformed if `data.drop` or `memory.init` is used anywhere in the `Code` section. This is currently in review here https://github.com/WebAssembly/bulk-memory-operations/pull/143

The fix as I see it is that `wast2json` should always generate a `DataCount` section if `memory.init` or `data.drop` is used.

To save @binji some cycles, I'll provide a PR for this issue shortly.

rianhunter:
The bulk-memory spec requires that a DataCount section is always
present if a memory.init or data.drop instruction is present in
the following code section. This requirement remains even if the
number of datas is 0, so remove that condition.

Fixes #1368

### DIFF
--- a/src/binary-writer.cc
+++ b/src/binary-writer.cc
@@ -1133,8 +1133,7 @@ Result BinaryWriter::WriteModule() {
     EndSection();
   }
 
-  if (options_.features.bulk_memory_enabled() &&
-      module_->data_segments.size()) {
+  if (options_.features.bulk_memory_enabled()) {
     // Keep track of the data count section offset so it can be removed if
     // it isn't needed.
     data_count_start_ = stream_->offset();
--- a/test/dump/elem-mvp-compat.txt
+++ b/test/dump/elem-mvp-compat.txt
@@ -27,6 +27,12 @@
 0000014: 0b                                        ; end
 0000015: 00                                        ; num elems
 000000f: 06                                        ; FIXUP section size
+; section "DataCount" (12)
+0000016: 0c                                        ; section code
+0000017: 00                                        ; section size (guess)
+0000018: 00                                        ; data count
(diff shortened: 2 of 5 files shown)


sbc100:
But if data count is zero then wouldn't all memory.init and data.drop be invalid anyway?   i.e. any module with a memory.init or data.drop instruction but no segments is already invalid weather you write this datacount section or not.

rianhunter:
If `DataCount` section exists but number of datas is zero, then memory.init and data.drop would be invalid.

If `DataCount` section doesn't exist, then the use of memory.init and data.drop would result in a malformed binary (not an invalid module).

This change was made to accomodate this test case https://github.com/WebAssembly/bulk-memory-operations/blob/ffdbb6e357d79818c88c3557cd3325346a057d27/test/core/memory_init.wast#L188

binji:
Yep, this seems like the right fix. Can you fix the tests here too?

rianhunter:
Done, "-r" is very convenient

PR #1370, opened 2020-03-23
Avoid spurious copy-to-bin invocations

ajklein:
Switch to using add_custom_command with POST_BUILD option
instead of add_custom_target for copying files to bin.

This causes the copy to happen only after rebuilding the target,
rather than on every build invocation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -368,10 +368,11 @@ if (NOT EMSCRIPTEN)
       list(APPEND WABT_EXECUTABLES ${EXE_NAME})
       set(WABT_EXECUTABLES ${WABT_EXECUTABLES} PARENT_SCOPE)
 
-      add_custom_target(${EXE_NAME}-copy-to-bin ALL
+      add_custom_command(
+        TARGET ${EXE_NAME}
+        POST_BUILD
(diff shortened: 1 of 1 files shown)


ajklein:
Note that I'm a CMake noob, so this may be wrong. I was just surprised to see ninja constantly copying the same files everytime I ran it...

binji:
Yeah, I noticed that too. IIRC, the issue is that if you don't always copy it, then the behavior is a little bit strange if you switch build types; e.g. if you build `debug` then build `release`, but both are already built, then the `release` build won't be copied over. Maybe this is a sign that this pattern should be discouraged, though.

ajklein:
Ah, I didn't look very hard at the `make`-based build, so this didn't occur to me. Given that it doesn't take any time at all to do these copies, I'm fine with dropping this PR if it causes any problems (and it sounds like it does).

binji:
Yeah, I'd rather keep as-is. (Or do a bigger refactor to remove the hacky top-level Make)

sbc100:
I think I had the exact same PR up at some point... and also abandoned it.

ajklein:
Thanks for the feedback, closing. I at least learned a little bit of CMake :)

PR #1373, opened 2020-03-24
Update testsuite (for SIMD)

binji:
Lots of changes necessary to make this work, as well as some bug fixes.

The main change is allowing `nan:canonical` and `nan:arithmetic` as a
possible value for each lane of a `v128`. This needs to propogate
through the parser, IR, the JSON format, and the spec interpreter.

This also changes the format of the spec JSON file, where a SIMD value
is now stored as a list of values instead of a single u128:

```
{"type": "v128", "lane_type": "i32", "value": ["0", "0", "0", "0"]}
```

Since the lane type can be `i8` and `i16`, these types can now be used
in more places (not just the decompiler). They'll be used for the GC
proposal too (for packed values), so I've updated them to use the binary
value specified for that proposal.

Here are the actual SIMD fixes:

* SIMD lanes are malformed if they don't match the binary format, but
  invalid if they are smaller than the lane width. For example,
  `i8x16.extract_lane_s` is malformed if the lane is >= 256, because the
  lane is stored as a byte. But it is invalid if the lane is >= 16.
* The `i8x16.narrow_i16x8_u`, `i16x8.narrow_i32x4_u` and
  `i64x2.load_32x2_u` instructions were not handling sign-extension
  propoerly.

TODO: This code is pretty clumsy now; it would be better to have a
universal `Value` and `ExpectedValue` that can be used everywhere, so
the logic doesn't need to be duplicated.

### DIFF
--- a/src/binary-writer-spec.cc
+++ b/src/binary-writer-spec.cc
@@ -54,6 +54,8 @@ class BinaryWriterSpec {
   void WriteLocation(const Location& loc);
   void WriteVar(const Var& var);
   void WriteTypeObject(Type type);
+  void WriteF32(uint32_t, ExpectedNan);
+  void WriteF64(uint64_t, ExpectedNan);
   void WriteConst(const Const& const_);
   void WriteConstVector(const ConstVector& consts);
   void WriteAction(const Action& action);
@@ -160,6 +162,38 @@ void BinaryWriterSpec::WriteTypeObject(Type type) {
   json_stream_->Writef("}");
 }
 
+void BinaryWriterSpec::WriteF32(uint32_t f32_bits, ExpectedNan expected) {
+  switch (expected) {
+    case ExpectedNan::None:
+      json_stream_->Writef("\"%u\"", f32_bits);
+      break;
+
+    case ExpectedNan::Arithmetic:
+      WriteString("nan:arithmetic");
+      break;
+
+    case ExpectedNan::Canonical:
+      WriteString("nan:canonical");
+      break;
+  }
+}
+
+void BinaryWriterSpec::WriteF64(uint64_t f64_bits, ExpectedNan expected) {
+  switch (expected) {
+    case ExpectedNan::None:
+      json_stream_->Writef("\"%" PRIu64 "\"", f64_bits);
+      break;
+
+    case ExpectedNan::Arithmetic:
+      WriteString("nan:arithmetic");
+      break;
+
+    case ExpectedNan::Canonical:
+      WriteString("nan:canonical");
+      break;
+  }
+}
+
 void BinaryWriterSpec::WriteConst(const Const& const_) {
   json_stream_->Writef("{");
   WriteKey("type");
@@ -181,39 +215,19 @@ void BinaryWriterSpec::WriteConst(const Const& const_) {
       json_stream_->Writef("\"%" PRIu64 "\"", const_.u64());
       break;
 
-    case Type::F32: {
-      /* TODO(binji): write as hex float */
+    case Type::F32:
       WriteString("f32");
       WriteSeparator();
       WriteKey("value");
-      if (const_.is_expected_nan()) {
-        if (const_.expected() == ExpectedNan::Arithmetic) {
-          WriteString("nan:arithmetic");
-        } else {
-          WriteString("nan:canonical");
-        }
-      } else {
-        json_stream_->Writef("\"%u\"", const_.f32_bits());
-      }
+      WriteF32(const_.f32_bits(), const_.expected_nan());
       break;
-    }
 
-    case Type::F64: {
-      /* TODO(binji): write as hex float */
+    case Type::F64:
       WriteString("f64");
       WriteSeparator();
       WriteKey("value");
-      if (const_.is_expected_nan()) {
-        if (const_.expected() == ExpectedNan::Arithmetic) {
-          WriteString("nan:arithmetic");
-        } else {
-          WriteString("nan:canonical");
-        }
-      } else {
-        json_stream_->Writef("\"%" PRIu64 "\"", const_.f64_bits());
-      }
+      WriteF64(const_.f64_bits(), const_.expected_nan());
       break;
-    }
 
     case Type::Nullref:
       WriteString("nullref");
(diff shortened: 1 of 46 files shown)


src/binary-writer-spec.cc
```diff
@@ -160,6 +162,38 @@ void BinaryWriterSpec::WriteTypeObject(Type type) {
...
 }
 
+void BinaryWriterSpec::WriteF32(uint32_t f32_bits, ExpectedNan expected) {
+  switch (expected) {
+    case ExpectedNan::None:
+      json_stream_->Writef("\"%u\"", f32_bits);
+      break;
+
+    case ExpectedNan::Arithmetic:
+      WriteString("nan:arithmetic");
```
sbc100: Unrelated: Why is this called BinaryWriter?

src/interp/interp.cc
```diff
@@ -1597,7 +1597,7 @@ RunResult Thread::StepInternal(Trap::Ptr* out_trap) {
     case O::I32X4Load16X4S: return DoSimdLoadExtend<s32x4, s16x4>(instr, out_trap);
     case O::I32X4Load16X4U: return DoSimdLoadExtend<u32x4, u16x4>(instr, out_trap);
     case O::I64X2Load32X2S: return DoSimdLoadExtend<s64x2, s32x2>(instr, out_trap);
-    case O::I64X2Load32X2U: return DoSimdLoadExtend<s64x2, s32x2>(instr, out_trap);
```
sbc100: Is this simply a bugfix here?   Why didn't it show up before?

src/ir.h
```diff
@@ -101,40 +114,71 @@ struct Const {
...
+  void set_v128_u32(int lane, uint32_t x) { set_v128_lane(lane, Type::I32, x); }
+  void set_v128_u64(int lane, uint64_t x) { set_v128_lane(lane, Type::I64, x); }
+  void set_v128_f32(int lane, uint32_t x) { set_v128_lane(lane, Type::F32, x); }
+  void set_v128_f64(int lane, uint64_t x) { set_v128_lane(lane, Type::F64, x); }
 
   // Only used for expectations. (e.g. wast assertions)
-  void set_f32(ExpectedNan nan) { From<float>(Type::F32, 0); nan_ = nan; }
-  void set_f64(ExpectedNan nan) { From<double>(Type::F64, 0); nan_ = nan; }
+  void set_f32(ExpectedNan nan) { set_f32(0); set_expected_nan(0, nan); }
+  void set_f64(ExpectedNan nan) { set_f64(0); set_expected_nan(0, nan); }
```
sbc100: Is clang format OK with multiple statements on one line like this?

src/type.h
```diff
@@ -36,6 +36,8 @@ class Type {
     F32 = -0x03,      // 0x7d
     F64 = -0x04,      // 0x7c
     V128 = -0x05,     // 0x7b
+    I8 = -0x06,       // 0x7a  // packed-type only, used in gc and as v128 lane
+    I16 = -0x07,      // 0x79  // packed-type only, used in gc and as v128 lane
```
sbc100: How about `:` rather than a repeated comment prefix?

src/tools/spectest-interp.cc
```diff
@@ -160,11 +160,115 @@ class RegisterCommand : public CommandMixin<CommandType::Register> {
 };
 
 struct ExpectedValue {
-  bool is_expected_nan;
   TypedValue value;
-  ExpectedNan expectedNan;
+  Type lane_type;      // Only valid if value.type == Type::V128.
+  ExpectedNan nan[4];  // One for each v128 lane (up to 4 for f32x4).
```
sbc100: Is nan[0] valid for non-simd types?

src/tools/spectest-interp.cc
```diff
@@ -491,121 +613,236 @@ wabt::Result JSONParser::ParseTypeVector(TypeVector* out_types) {
...
 wabt::Result JSONParser::ParseConst(TypedValue* out_value) {
-  std::string type_str;
-  std::string value_str;
-  EXPECT("{");
-  PARSE_KEY_STRING_VALUE("type", &type_str);
-  EXPECT(",");
-  PARSE_KEY_STRING_VALUE("value", &value_str);
-  EXPECT("}");
+  ExpectedValue expected;
+  CHECK_RESULT(ParseExpectedValue(&expected));
```
sbc100: Are all the consts in the JSON expected values?  What about const argument values?

test/update-spec-tests.py
```diff
@@ -92,6 +92,7 @@ def main(args):
     ProcessProposalDir('sign-extension-ops', '--enable-sign-extension')
     ProcessProposalDir('bulk-memory-operations', '--enable-bulk-memory')
     ProcessProposalDir('reference-types', '--enable-reference-types')
+    ProcessProposalDir('simd', '--enable-simd')
```
sbc100: Oops.. is this why there were so many new tests added here?  I guess we were not updating the simd tests previously?

src/binary-writer-spec.cc
binji: Good question, it should be JSONWriter I suppose. I think it was historical, because the old binary-writer used to contain this stuff too, so I probably kept the name when I split it out.

src/interp/interp.cc
binji: There weren't any tests of the output before, this is a relatively new instruction.

src/ir.h
binji: No, it will expand, though I kind of prefer keeping it on a single line for readability. What do you think?

src/tools/spectest-interp.cc
binji: Yes, I'll add that to the comment.

src/ir.h
sbc100: Depends what you want the policy for wabt to be ?  Absolute deferral to clang-format, or programmer discretion.    I don't think is one correct answer.   It depends on the project.   We don't have a lot of formatting discussions so allowing discretion isn't currently effecting productivity.  So i'm with with either way right now.   We should probably document it though.

src/tools/spectest-interp.cc
binji: No, the expected values are more expressive. It's nice to be able to share the code since it's a significant amount of code. But yeah, I'll add an enum to make sure we don't parse stuff that isn't valid.

test/update-spec-tests.py
binji: Yep, I guess there weren't SIMD tests for a while...

src/ir.h
binji: Yeah, I'm thinking I'll spend some time writing up a C++ style guide finally this week.

PR #1380, opened 2020-03-26
Add Type::RefT to represent (ref $T)

binji:
This adds support for handling types of the form (ref $T), where $T is
an index into the type section. Rather than storing an additional field
in the Type struct, it's more efficient to use the extra bits in the
enum to encode the index. For example, RefT is encoded as -0x13
(0xffffffff6d):

```
   1111 1111 1111 1111 1111 1111 0110 1101
      f    f    f    f    f    f    6    d
```

We can store the index in the top bits (making sure to keep the MSB set so
the number remains negative), giving us 23 bits to use for the type index.

```
   1nnn nnnn nnnn nnnn nnnn nnnn 0110 1101
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -507,6 +507,7 @@ if (NOT EMSCRIPTEN)
       src/test-option-parser.cc
       src/test-string-view.cc
       src/test-filenames.cc
+      src/test-type.cc
       src/test-utf8.cc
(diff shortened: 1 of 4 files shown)


binji:
@jgravelle-google ping

keithw:
Closing as stale. :-(

PR #1381, opened 2020-03-27
Update testsuite

binji:
* Add i{8x16,16x8,32x4}.abs instructions
* Implement IntAbs in interp-math.h

### DIFF
--- a/test/wasm2c/spec/table.txt
+++ b/test/wasm2c/spec/table.txt
@@ -1,0 +1,5 @@
+;;; TOOL: run-spec-wasm2c
+;;; STDIN_FILE: third_party/testsuite/table.wast
+(;; STDOUT ;;;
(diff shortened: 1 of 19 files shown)


test/wasm2c/spec/table.txt
```diff
@@ -0,0 +1,5 @@
+;;; TOOL: run-spec-wasm2c
+;;; STDIN_FILE: third_party/testsuite/table.wast
+(;; STDOUT ;;;
+0/0 tests passed.
```
sbc100: Looks like there should be 3 tests in this file.  Does wasm2c not report this number correctly?

binji: Those are all `assert_malformed` tests, which wasm2c doesn't run (it's more a test of the binary reader than wasm2c).

PR #1384, opened 2020-04-09
[wasm2c] Add rot13 example

binji:
This example demonstrates how to use imported functions. The `rot13`
program takes each command line argument, and rot13-encodes it.

The exported `rot13` function has no arguments, and instead calls back
into the program (via `fill_buf`) with a buffer to fill in. When the
function finishes it calls `buf_done`.

(rot13.wat is the same as in src/test-interp.cc.)


.gitignore
```diff
@@ -7,3 +7,4 @@
 .idea/
 .vscode/
 /cmake-build-debug/
+*.o
```
sbc100, line 10: I'd rather be specific about where its ok to ignore `.o` files.

I often create temporary `.c` and `.o` file in my project directory and I want `git status` to show them to me and `git clean -f` to clean them up for me.   Maybe I'm strange like that.

wasm2c/examples/rot13/Makefile
```diff
@@ -0,0 +1,9 @@
+# Use implicit rules for compiling C files.
+CFLAGS=-I../..
+rot13: main.o rot13.o ../../wasm-rt-impl.o
```
sbc100, line 3: I'm not very familiar with the wasm2c stuff.  Why is this called `wasm-rt-impl.o`?  Wouldn't `wasm-rt.o` be clear enough?

wasm2c/examples/rot13/rot13.c
```diff
@@ -0,0 +1,258 @@
+#include <math.h>
```
sbc100, line 1: I guess its obvious by it might be nice to have a comment at the top along the lines of `Automatically generated by wasm2c ...` (with the full command line maybe?)`

wasm2c/examples/rot13/main.c
```diff
@@ -0,0 +1,77 @@
+#include <stdio.h>
```
sbc100, line 1: Top level comment describing that this is doing ?     Something like `entry point for rot13 program which loads and runs the wasm rot13 program as compiled into C by wasm2c`?

wasm2c/examples/rot13/Makefile
binji: My original thought was that `wasm-rt.h` just provides an interface, and could be implemented different ways. But I guess there's not much value in that.

sbc100: I see, but isn't object code by definition an implementation making this redundant?    If there were multiple implementations we might expect an adjective there such as wasm-rt-generic or wasm-rt-default maybe?  I might be missing something here.

wasm2c/examples/rot13/main.c
binji: Done.

wasm2c/examples/rot13/rot13.c
binji: Done.

.gitignore
binji: Done.

binji:
> Any reason to check in the wasm2c output of the wasm binary?

I included it for the fac example, since it's referenced in the documentation for wasm2c. But you're right, it's probably only worth including once.

PR #1387, opened 2020-04-12
Update wasm-interp.cc

Issue-03:
changed to "initializing"

### DIFF
--- a/src/tools/wasm-interp.cc
+++ b/src/tools/wasm-interp.cc
@@ -191,7 +191,7 @@ Result ReadAndInstantiateModule(const char* module_filename,
   *out_instance = Instance::Instantiate(s_store, module.ref(), imports, &trap);
   if (!*out_instance) {
     // TODO: change to "initializing"
-    WriteTrap(stream, "error initialiazing module", trap);
+    WriteTrap(stream, "error initializing module", trap);
     return Result::Error;
   }
 


binji:
Thanks for fixing! It looks good, but you need to fix the tests too. Take a look at https://github.com/WebAssembly/wabt/blob/master/test/README.md#writing-new-tests, specifically passing the `-r` flag to rebaseline the tests.

Issue-03:
Hello, @binji 
Sorry, I am a beginner at Open Source, so I did not know why it was failing. I read the the file indicated by you, but did not understand. Can you please help?
Thank you.

binji:
It's failing because when the code is compiled, it produces a new output with the correct spelling. There are tests in the repo that check the output of various tools, and will error-out when the output doesn't match the expected output.

In this case, the `bad-data-size.txt` test is failing. Here's the output from one of the CI bots:

```
  STDOUT MISMATCH:
  --- expected
  +++ actual
  @@ -1 +1 @@
  -error initialiazing module: out of bounds memory access: data segment is out of bounds: [0, 8) >= max value 0
  +error initializing module: out of bounds memory access: data segment is out of bounds: [0, 8) >= max value 0
```

You can fix this by compiling the code, then running the tests with the `-r` flag:

```
$ test/run-tests.py -r bad-data-size.txt
```

This will run the tests, capture the output, and replace the expected text automatically.

binji:
This was fixed here: https://github.com/WebAssembly/wabt/pull/1410

Issue #1388, opened 2020-04-13
locals of type exnref


Horcrux7:
If try to compile the follow sample code with the command line: `wat2wasm.exe --enable-all sample.wat`

```
(module
    (func $test(local $exn exnref)
        try
          i32.const 0
		  drop
        catch
		  local.tee $exn
          drop
        end
    )
)
```
then I receive the error:
```
sample.wat:2:28: error: unexpected token "exnref", expected i32, i64, f32, f64, v128 or anyref.
    (func $test(local $exn exnref)
                           ^^^^^^

```

Because try/catch from the exception proposal is supported WABT should also supports the variable type `exnref`.

I have use the master version from today on Windows.

binji:
Whoops, thanks for mentioning! Should be fixed in #1389.

Horcrux7:
Works.

PR #1389, opened 2020-04-13
Add exnref value type to parser

binji:
The exnref type was already supported in the type checker, and other
parts of the code, but there was no way to name the type in the text
format.

This PR also fixes makes binary-reader.cc check for just the
exceptions_enabled flag to enable exnref. The exception-handling
proposal depends on the reference types proposal, but that is now
handled at a higher level, in the `UpdateDependencies` function.

Fixes issue #1388.


binji:
It doesn't seem that the exception-handling repo has new tests yet.

Issue #1385, opened 2020-04-09
Don't require Python in build (if not found, skip those targets)


binji:
The current build requires python 3.5, which is commonly found on modern *nixes, but may not be installed on Windows, Mac, etc. It's only used for running tests, so it would be nice to fail gracefully if python isn't available instead of preventing the build to proceed.

lkarthee:
I have submitted a PR for this - https://github.com/WebAssembly/wabt/pull/1392

sbc100:
We already have `-DBUILD_TESTS=OFF` which should remove the requirement.  

We could make the error message nicer maybe?  Something like: `python3 not found, either install it or build with -DBUILD_TESTS=OFF`.

binji:
@aardappel originally mentioned the issue in #1385 to me, primarily because the version of Python3 was not immediately available on Windows, and isn't needed unless you want to test. I figured that it would be a nice fallback to automatically. But perhaps a better error message is OK too. Wouter, what do you think?

sbc100:
What do you mean by "not immediately available on Windows"?  Is there some version we could switch to that would be more immediately available?

I guess we could have `BUILD_TESTS` automatically default of `OFF` if python isn't found, but it seems reasonable to ask people to opt out explicitly.

binji:
> What do you mean by "not immediately available on Windows"? 

Not sure, I'm probably just incorrectly echoing Wouter's comments about this. I agree that a better error message is probably enough here.

aardappel:
This particular machine only had Python 2.7 installed, so CMake failed. It's a bit odd for CMake to fail on a tool that's only required if I decide to run tests.

I should generally be able to build C++ binaries with a C++ build system without requiring Python.

lkarthee:
Is there a specific reason why python > 3.5 is needed to run tests?  Will scripts fail if python3 < 3.5 is used ?

binji:
We switched from Python2 to Python3 recently, and I believe we chose Python 3.5 because of some nice feature it had... though to be honest I can't remember what that was. Something to do with formatting byte strings perhaps?

sbc100:
Python2 is officially deprecated and unmaintained, and that is a cost to maintaining support for both.

Dropping python2 support doesn't effect our actual users since they can just build with `-DBUILD_TESTS=OFF`.  It only effects core developers who actually want to run the tests.  So, the impact of switch here is minimal, compared a project that actually has python as part of the product.

lkarthee:
[...]

PR #1392, opened 2020-04-22
Improve error message when python 3 not available - Fix for  #1385

lkarthee:
https://github.com/WebAssembly/wabt/issues/1385

In this PR, I have made python optional in CMakeLists.txt . I have tested locally by bumping the required version to 3.8 which is not existing on my Mac.

I have noticed that the below tests are run from python file - test/run-c-api-examples.py. 
-  c_api_example(callback)
-   c_api_example(finalize)
-   c_api_example(global)
-   c_api_example(hello)
-   c_api_example(hostref)
-   c_api_example(multi)
-   c_api_example(memory)
-   c_api_example(reflect)
-   c_api_example(serialize)
-   c_api_example(start)
-   c_api_example(table)
-   c_api_example(trap)

Do we need to port run-c-api-examples.py to gtest file or cpp file?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -464,6 +464,13 @@ if (NOT EMSCRIPTEN)
     )
   endif ()
 
+  # Python 3.5 is the version shipped in Ubuntu Xenial
+  find_package(PythonInterp 3.5)
+  if(NOT PYTHONINTERP_FOUND)
+    set(BUILD_TESTS OFF) 
+    message(WARNING "Skipping tests. Python 3 is required for wabt testing. Please install python3 to run tests.")
+  endif()
+
   find_package(Threads)
   if (BUILD_TESTS)
     if (NOT USE_SYSTEM_GTEST)
@@ -520,11 +527,6 @@ if (NOT EMSCRIPTEN)
       set(USES_TERMINAL USES_TERMINAL)
     endif ()
 
-    # Python 3.5 is the version shipped in Ubuntu Xenial
-    find_package(PythonInterp 3.5)
-    if(NOT PYTHONINTERP_FOUND)
-      message(FATAL_ERROR "Python 3 is required for wabt testing. Please install it. \nIf you want to proceed without installing python 3, either add -no-tests to make target (example: make clang-debug-no-tests) or run cmake with -DBUILD_TESTS=OFF.")
-    endif()
     # test running
     set(RUN_TESTS_PY ${WABT_SOURCE_DIR}/test/run-tests.py)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -466,7 +466,7 @@ if (NOT EMSCRIPTEN)
 
   # Python 3.5 is the version shipped in Ubuntu Xenial
   find_package(PythonInterp 3.5)
-  if(NOT PYTHONINTERP_FOUND)
+  if(BUILD_TESTS AND (NOT PYTHONINTERP_FOUND))
     set(BUILD_TESTS OFF) 
     message(WARNING "Skipping tests. Python 3 is required for wabt testing. Please install python3 to run tests.")
   endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -520,68 +520,72 @@ if (NOT EMSCRIPTEN)
       set(USES_TERMINAL USES_TERMINAL)
     endif ()
 
-    # Python 3.5 is the version shipped in Ubuntu Xenial
-    find_package(PythonInterp 3.5 REQUIRED)
-    # test running
-    set(RUN_TESTS_PY ${WABT_SOURCE_DIR}/test/run-tests.py)
-
-    add_custom_target(run-tests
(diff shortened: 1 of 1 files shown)


sbc100:
Can't you already do this by building with `-DBUILD_TESTS=OFF`?  

wabt's primary test runner ./test/run-tests.py is also written in python.  If you don't have python the only possible tests you could run are the gtest unittests and i'm not sure it worth the extra option to allow just those tests to build built.

sbc100:
Also, if you do want to run tests, isn't it easy install python3 on mac?

lkarthee:
Installing python3 is easy. I use pyenv to manage python versions. I wanted to contribute to this project, so I picked up #1385 with `good-first-bug` tag.

To test this PR I used a python version 3.8 which is not installed in my MacBook as I don't have Windows as mentioned in #1385. 

All 5 CI checks have python3 enabled and run python tests - they won't test the scenario where python3 is not installed.

sbc100:
Would you mind instead improving the error message, e.g. "python3 is required to for wabt testing,  please either install it or build with -DBUILD_TESTS=OFF".

lkarthee:
I will modify the PR accordingly.

lkarthee:
I have made required changes. See sample error message (ignore exact versions like 3.7.3 and 3.8 in the error message).

```
-- Could NOT find PythonInterp: Found unsuitable version "3.7.3", but required is at least "3.8" (found /usr/bin/python3)
CMake Error at CMakeLists.txt:587 (message):
  Python 3 is required for wabt testing.  Please install it.

  If you want to proceed without installing python 3, either add -no-tests to
  make target (example: make clang-debug-no-tests) or run cmake with
  -DBUILD_TESTS=OFF.


-- Configuring incomplete, errors occurred!
```

CMakeLists.txt
```diff
@@ -521,67 +521,71 @@ if (NOT EMSCRIPTEN)
...
-    c_api_example(trap)
-    if (NOT WIN32)
-      # depends on pthreads
-      set(THREADS_PREFER_PTHREAD_FLAG ON)
-      find_package(Threads REQUIRED)
-      c_api_example(threads)
-    endif ()
+    else()
+      message(FATAL_ERROR "Python 3 is required for wabt testing. Please install it. \nIf you want to proceed without installing python 3, either add -no-tests to make target (example: make clang-debug-no-tests) or run cmake with -DBUILD_TESTS=OFF.")
+    endif(PYTHONINTERP_FOUND)
```
sbc100, line 588: Seem like you could just put this error in an if an "if (NOT PYTHONINTERP_FOUND) .. endif()" above that the rest of the file would be left unchanged?

lkarthee: Done

aardappel:
Would be better if instead it displayed that warning and then turned tests off for you, rather than failing.
Tests are mostly for contributors to WABT, and should not concern people simply trying to build the binaries.

binji:
I tend to agree with @aardappel here. @sbc100, does that seem OK to you?

sbc100:
Find with me.   We can set BUILD_TESTS to false is python is not found.

sbc100:
Or should we even just have BUILD_TESTS default to OFF?

binji:
I like having it default to on, personally.

lkarthee:
Should I revert the initial commit of this PR? 
-  BUILD_TESTS default is ON (No change in existing code)
-  skip tests when python3  >= 3.5.

sbc100:
I think the conclusion is:

1. BUILD_TESTS defaults to ON
2. BUILD_TESTS gets forced to OFF if python is now found

lkarthee:
> - skip tests when python3 >= 3.5.

Got it - Oops typo, it should have been < sign.

CMakeLists.txt
```diff
@@ -464,6 +464,13 @@ if (NOT EMSCRIPTEN)
     )
   endif ()
 
+  # Python 3.5 is the version shipped in Ubuntu Xenial
+  find_package(PythonInterp 3.5)
+  if(NOT PYTHONINTERP_FOUND)
```
sbc100, line 469: how about AND BUILD_TESTS == ON there?

Otherwise lgtm!

lkarthee:
Cmake prints a message when suitable version is not found. This will appear irrespective of BUILD_TEST is ON or OFF.

When BUILD_TESTS=OFF and python3 not found
```
.......
-- Check size of size_t - done
-- Could NOT find PythonInterp: Found unsuitable version "3.7.3", but required is at least "3.8" (found /usr/bin/python3)
-- Looking for pthread.h
.......
```

When BUILD_TESTS=ON and python3 not found
```
.......
-- Check size of size_t - done
-- Could NOT find PythonInterp: Found unsuitable version "3.7.3", but required is at least "3.8" (found /usr/bin/python3)
CMake Warning at CMakeLists.txt:471 (message):
  Skipping tests.  Python 3 is required for wabt testing.  Please install
  python3 to run tests.


-- Looking for pthread.h
.......
```

CMakeLists.txt
lkarthee: added this check.

PR #1403, opened 2020-05-01
Renumber SIMD opcodes

dtig:
This PR mirrors the updated SIMD opcodes in https://github.com/WebAssembly/simd/issues/216. Uploading for review now, but should be merged in coordination with LLVM, emscripten updates to the opcode space.


binji:
Thanks for doing this! See my comment [here](https://github.com/WebAssembly/simd/pull/209#issuecomment-623638499). `simd_const.wast` uses the `v128.const` encoding directly, so that will need to be fixed before these tests will pass. It probably is easiest to mark those tests as failing for now, then we can fix them when that PR lands.

dtig:
Thanks Ben, I've skipped the simd_const test for now. Let me know if you meant something different by marking the test as failing.

PR #1408, opened 2020-05-07
wasm2c: build library wasm-rt-impl

eadwu:
Also include header files needed for compilation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,7 @@
 
 cmake_minimum_required(VERSION 3.0.0)
 project(WABT VERSION 1.0.13)
+include(GNUInstallDirs)
 
 if (POLICY CMP0077)
   cmake_policy(SET CMP0077 NEW)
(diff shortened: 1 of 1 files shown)


eadwu:
I'm guessing this is failing because `CMAKE_INSTALL_*` isn't defined. I'll figure out how this works.

binji:
Cool, seems like a good idea!

CMakeLists.txt
```diff
@@ -324,6 +325,10 @@ set(WABT_LIBRARY_SRC
 
 add_library(wabt STATIC ${WABT_LIBRARY_SRC})
 
+add_library(wasm-rt-impl STATIC wasm2c/wasm-rt-impl.c wasm2c/wasm-rt-impl.h)
+INSTALL(TARGETS wasm-rt-impl DESTINATION ${CMAKE_INSTALL_LIBDIR})
+INSTALL(FILES wasm2c/wasm-rt.h wasm2c/wasm-rt-impl.h DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
```
sbc100, line 330: The convention in this file is to use lowercase function names

CMakeLists.txt
```diff
@@ -16,6 +16,7 @@
 
 cmake_minimum_required(VERSION 3.0.0)
 project(WABT VERSION 1.0.13)
+include(GNUInstallDirs)
```
sbc100, line 19: Why is these needed?

eadwu: `CMAKE_INSTALL_LIBDIR` isn't set without by default without them which would fail the install, not sure where the default location to install them is without it.

Not sure about the compatibility between OS outside of Linux either.

Binaries installed to `bin/` seem to be done manually here and I can try adapting that over to fit with `lib/`?

CMakeLists.txt
eadwu: Adjusted.

nomeata:
I noticed that nixpkgs is applying this patch. Would be nice to bless it official?

sbc100:
The patch is approved.. but it looks like it needs a rebase before we can land.

eadwu:
Rebased.

sbc100:
Looks like a real windows build failure.

binji:
Yep, wasm2c still isn't quite right on Windows (https://github.com/WebAssembly/wabt/tree/wasm2c-windows tries to fix, but is incomplete). So this library shouldn't be built there.

eadwu:
Fixed.

binji:
Thanks!

PR #1411, opened 2020-05-10
Add initial MVP of WASI API support to wasm-interp

### LINKED ISSUE #1409: Add WASI API to wasm-interp / wasm2c
Seems like adding a dependency on https://github.com/cjihrig/uvwasi might be a good start.

sbc100:
This is proof of concept that only implements the `proc_exit` API.
Extending this to the full WASI API will to follow assuming this
approach seems reasonable.

Fixes #1409

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,6 +46,9 @@ option(USE_UBSAN "Use undefined behavior sanitizer" OFF)
 option(CODE_COVERAGE "Build with code coverage enabled" OFF)
 option(WITH_EXCEPTIONS "Build with exceptions enabled" OFF)
 option(WERROR "Build with warnings as errors" OFF)
+# WASI support is still a work in progress.
+# Only a handful of syscalls are supported at this point.
+option(WITH_WASI "Build WASI support via uvwasi" OFF)
 
 if (MSVC)
   set(COMPILER_IS_CLANG 0)
@@ -419,10 +422,21 @@ if (NOT EMSCRIPTEN)
       INSTALL
     )
 
+    if(WITH_WASI)
+      add_subdirectory("third_party/uvwasi" EXCLUDE_FROM_ALL)
+      include_directories("${libuv_SOURCE_DIR}/include")
+      include_directories(third_party/uvwasi/include)
+      add_definitions(-DWITH_WASI)
+      set(INTERP_LIBS uvwasi_a)
+      set(EXTRA_INTERP_SRC src/interp/interp-wasi.cc)
+    endif()
+
     # wasm-interp
+
     wabt_executable(
       NAME wasm-interp
-      SOURCES src/tools/wasm-interp.cc
+      SOURCES src/tools/wasm-interp.cc ${EXTRA_INTERP_SRC}
+      LIBS ${INTERP_LIBS}
       WITH_LIBM
       INSTALL
     )
--- a/src/interp/interp-wasi.cc
+++ b/src/interp/interp-wasi.cc
@@ -21,6 +21,7 @@
 
 #include "uvwasi.h"
 
+#include <cinttypes>
 #include <unordered_map>
 
 using namespace wabt;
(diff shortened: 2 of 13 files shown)


CMakeLists.txt
```diff
@@ -419,10 +421,19 @@ if (NOT EMSCRIPTEN)
...
+      include_directories(third_party/uvwasi/include)
+      add_definitions(-DWITH_WASI)
+      set(INTERP_LIBS uvwasi_a)
+    endif()
+
     # wasm-interp
     wabt_executable(
       NAME wasm-interp
-      SOURCES src/tools/wasm-interp.cc
+      SOURCES src/tools/wasm-interp.cc src/interp/interp-wasi.cc
```
binji, line 436: Why not make this an optional source rather than a source that is always included w/ #ifdef around it?

test/wasi/write_stdout.txt
```diff
@@ -0,0 +1,37 @@
...
+  (call $fd_write (i32.const 1) (i32.const 12) (i32.const 1) (i32.const 20))
+  drop
+)
+(;; STDOUT ;;;
+#0.    0: V:0  | i32.const 1
+#0.    8: V:1  | i32.const 12
+#0.   16: V:2  | i32.const 1
+#0.   24: V:3  | i32.const 20
+#0.   32: V:4  | call_import $0
+>>> running wasi function "fd_write":
```
binji, line 35: I was expecting to see "hello" here somewhere...

src/interp/wasi-api.def
```diff
@@ -0,0 +1,2 @@
+
```
binji, line 1: File currently unused?

src/interp/interp-wasi.cc
```diff
@@ -0,0 +1,216 @@
...
+    return Result::Ok;
+  }
+
+ protected:
+  template <typename T>
+  T* getMemPtr(uint32_t address, uint32_t size) {
+    printf("getMemPtr address=%d size=%d(%lu)\n", address, size,
+           size * sizeof(T));
+    if (!memory->IsValidAccess(address, 0, size * sizeof(T))) {
+      printf("invalid access\n");
```
binji, line 96: Probably should trap here.

sbc100:
Hello world now works :)

CMakeLists.txt
sbc100: Done.

test/wasi/write_stdout.txt
sbc100: Done

src/interp/wasi-api.def
sbc100: Done.

test/run-tests.py
```diff
@@ -913,12 +916,17 @@ def main(args):
...
     if options.patterns:
+        exclude_dirs = []
         pattern_re = '|'.join(
             fnmatch.translate('*%s*' % p) for p in options.patterns)
     else:
         pattern_re = '.*'
+        # By default, exclude wasi tests because WASI support is not include
+        # by int the build by default.
+        # TODO(sbc): Find some way to detect the WASI support.
+        exclude_dirs = ['wasi']
```
binji, line 927: nice, seems like a good localized solution.

sbc100:
OK I think this is pretty good starting point from which to evaluate/evolve.   I added trap handling and test for that.

PTAL.

src/interp/interp-wasi.cc
```diff
@@ -0,0 +1,250 @@
...
+    return Result::Ok;
+  }
+
+  // The trace stream accosiated with the instance.
+  Stream* trace_stream;
+
+  Instance::Ptr instance;
+ private:
+  template <typename T>
+  T* getMemPtr(uint32_t address, uint32_t size, Trap::Ptr* trap) {
```
binji, line 101: Hm, we may want to add this to the interp API instead, now that I think of it. But it's OK to have here for now.

src/interp/interp-wasi.cc
```diff
@@ -0,0 +1,250 @@
...
+                    uvwasi_s* uvwasi,
+                    Stream* stream,
+                    Stream* trace_stream) {
+  Store* store = instance.store();
+  auto module = store->UnsafeGet<Module>(instance->module());
+  auto&& module_desc = module->desc();
+
+  Func::Ptr start;
+  Memory::Ptr memory;
+  for (auto&& export_ : module_desc.exports) {
```
binji, line 193: Seems like we may want a convenient API for this in interp too.

sbc100: You mean something like a `getExportByName` function?

binji: Yeah. It'll be slightly less efficient (since it has to look through twice), but probably not a big deal for most users.

PR #1418, opened 2020-05-11
Fix bug when writing grouped named parameters

binji:
A named parameter (or local) must always be in its own `param` block, so
this syntax is not allowed:

`(param i32 $b i32)`

And must instead be converted to:

`(param i32) (param $b i32)`

### DIFF
--- a/src/wat-writer.cc
+++ b/src/wat-writer.cc
@@ -1125,26 +1125,27 @@ void WatWriter::WriteTypeBindings(const char* prefix,
    *   (param $foo i32)
    *   (param i32 i64 f32)
    */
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks for the quick fix!

FWIW as I've been poking at this some more locally I think there's the same issue with `local` declarations?

binji:
Yep, locals are handled in the same way as parameters in this code. But I should probably add a test for that too.

binji:
This has failures on Travis CI, but I believe this is because of pep8 issues which @sbc100 fixed. Gonna land for now, and follow up w/ other issues in another PR.

PR #1423, opened 2020-05-13
Implement more WASI APIs

sbc100:
This is almost enough to pass all the tiny tests in the wasi-sdk
repo.

### DIFF
--- a/src/interp/interp-wasi.cc
+++ b/src/interp/interp-wasi.cc
@@ -14,6 +14,20 @@
  * limitations under the License.
  */
 
+/*
+ * This is an experiment and currently extremely limited implementation
+ * of the WASI syscall API.  The implementation of the API itself is coming from
+ * uvwasi: https://github.com/cjihrig/uvwasi.
+ *
+ * Most of the code in the file is mostly marshelling data between the wabt
+ * interpreter and uvwasi.
+ *
+ * For details of the WASI api see:
+ * https://github.com/WebAssembly/WASI/blob/master/phases/snapshot/docs.md
+ * and the C headers version:
+ * https://github.com/WebAssembly/wasi-libc/blob/master/libc-bottom-half/headers/public/wasi/api.h
+ */
+
 #include "src/interp/interp-wasi.h"
 #include "src/interp/interp-util.h"
 
@@ -49,6 +63,8 @@ _Static_assert(_Alignof(__wasi_iovec_t) == 4, "witx calculated align");
 _Static_assert(offsetof(__wasi_iovec_t, buf) == 0, "witx calculated offset");
 _Static_assert(offsetof(__wasi_iovec_t, buf_len) == 4, "witx calculated offset");
 
+#define __WASI_ERRNO_NOENT (UINT16_C(44))
+
 class WasiInstance {
  public:
   WasiInstance(Instance::Ptr instance,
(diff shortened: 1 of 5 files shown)


sbc100:
Rather than add more of our own tests in wat form I'm currently relying on the tests in wasi-sdk and hoping to integrate them directly somehow.

src/interp/interp-wasi.cc
```diff
@@ -66,8 +76,49 @@ class WasiInstance {
     return Result::Ok;
   }
 
+  Result fd_fdstat_get(const Values& params, Values& results, Trap::Ptr* trap) {
```
binji, line 79: might be nice to duplicate the WASI function signature here, with the WASI type names

src/interp/interp-wasi.cc
```diff
@@ -66,8 +76,49 @@ class WasiInstance {
...
+  }
+
+  Result fd_read(const Values& params, Values& results, Trap::Ptr* trap) {
+    int32_t fd = params[0].i32_;
+    int32_t iovptr = params[1].i32_;
+    int32_t iovcnt = params[2].i32_;
+    if (trace_stream) {
+      trace_stream->Writef("fd_read %d [%d]\n", fd, iovcnt);
+    }
+    __wasi_iovec_t* wasm_iovs = getMemPtr<__wasi_iovec_t>(iovptr, iovcnt, trap);
```
binji, line 95: `getMemPtr` could use an out parameter and return `Result`, then we can use `CHECK_RESULT` here rather than duplicating the `if` statement below.

src/interp/interp-wasi.cc
```diff
@@ -76,27 +127,112 @@ class WasiInstance {
...
+  }
+
   // The trace stream accosiated with the instance.
   Stream* trace_stream;
 
   Instance::Ptr instance;
+
  private:
+  template <typename T>
+  Result writeValue(T value, uint32_t target_address, Trap::Ptr* trap) {
```
binji, line 227: nit: style is `WriteValue` (should probably fix `GetMemPtr` too). The WASI names are fine to keep as is.

src/tools/wasm-interp.cc
```diff
@@ -100,6 +101,9 @@ static void ParseOptions(int argc, char** argv) {
                    "Assume input module is WASI compliant (Export "
                    " WASI API the the module and invoke _start function)",
                    []() { s_wasi = true; });
+  parser.AddOption('e', "env", "ENV",
+                   "Pass the given environment string in the WASI runtime",
+                   [](const std::string& argument) { s_wasi_env = argument; });
```
binji, line 106: Probably want this to append to a vector, right? What is the expected format for the ENV string?

src/interp/interp-wasi.cc
sbc100: Yeah, your right.   I'm was just being laz.....  i was just doing TTD and test only use one :)

src/interp/interp-wasi.cc
```diff
@@ -66,8 +76,49 @@ class WasiInstance {
...
+    int32_t iovptr = params[1].i32_;
+    int32_t iovcnt = params[2].i32_;
+    if (trace_stream) {
+      trace_stream->Writef("fd_read %d [%d]\n", fd, iovcnt);
+    }
+    __wasi_iovec_t* wasm_iovs = getMemPtr<__wasi_iovec_t>(iovptr, iovcnt, trap);
+    if (!wasm_iovs) {
+      return Result::Error;
+    }
+    uvwasi_iovec_t* iovs = new uvwasi_iovec_t[iovcnt];
```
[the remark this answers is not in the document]
sbc100: hmm.. i can just use std::vector on the stack too.. don't know why I used new at all here.

src/interp/interp-wasi.cc
sbc100: You mean in comments?   I added links at the top of the docs and headers for the wasi API.

src/interp/interp-wasi.cc
sbc100: Done.  Nice.

PR #1424, opened 2020-05-13
Update testsuite

binji:
Includes the three merged proposals (nontrapping-float-to-int,
sign-extension, multi-value).

It also has one bug fix when parsing table limits (when the min or max
size does not fit in a 32-bit int).

### DIFF
--- a/test/wasm2c/spec/i32.txt
+++ b/test/wasm2c/spec/i32.txt
@@ -1,5 +1,6 @@
 ;;; TOOL: run-spec-wasm2c
 ;;; STDIN_FILE: third_party/testsuite/i32.wast
+;;; SKIP: sign-extend, multi-value not supported yet
 (;; STDOUT ;;;
 360/360 tests passed.
 ;;; STDOUT ;;)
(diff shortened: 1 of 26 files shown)


test/wasm2c/spec/i32.txt
```diff
@@ -1,5 +1,6 @@
 ;;; TOOL: run-spec-wasm2c
 ;;; STDIN_FILE: third_party/testsuite/i32.wast
+;;; SKIP: sign-extend, multi-value not supported yet
```
sbc100, line 3: Does wabt not support these?

sbc100: I don't get how these are marked as SKIP but still seem to report passing ?

binji: wasm2c doesn't support this yet. And `SKIP` doesn't run the test, but it also doesn't check the output. I guess that's a bit confusing, but it means the diffs here are smaller.

sbc100: Sorry I missed the sub directory name there.

PR #1427, opened 2020-05-15
Mangle wasm2c names in a simpler way

### LINKED ISSUE #1426: wasm2c: function name mangling?
Right now compiling a wasm with a function called `main` leads to clang complaining about
```cpp
static u32 main(u32, u32);

static u32 main(u32, u32) {
  .. impl ..
}
```

```cpp
a.c:179:1: warning: 'main' should not be declared static [-Wmain]
static u32 main(u32, u32);
^~~~~~~
a.c:179:8: warning: return type of 'main' is not 'int' [-Wmain-return-type]
static u32 main(u32, u32);
       ^
a.c:179:8: note: change return type to 'int'
static u32 main(u32, u32);
       ^~~
       int
a.c:179:12: error: first parameter of 'main' (argument count) must be of type 'int'
static u32 main(u32, u32);
           ^
a.c:179:12: error: second parameter of 'main' (argument array) must be of type 'char **'
```

Despite `main` being declared static, and in its own file where there isn't an actual `main`, it thinks it's a bad declaration of `main`. Similar things can probably happen with other libc things that clang warns about redeclarations of.

Perhaps all function names should be mangled, say by adding a `_` prefix to them? As this is just for the `static` impls, and not seen by the outside, this doesn't seem too bad maybe, but it would make it a little less readable.

kripken:
This replaces the explicit list of names to avoid with a simple prefixing
scheme. This is safer as it handles more cases, however, it does make
the code slightly less fun to read. Perhaps there's a nicer prefixing?

Fixes #1426 and with this I can compile a "hello world" program with
`emcc => wasm2c => clang` and it all works :tada:

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -220,7 +220,6 @@ class CWriter {
   void Write(const ResultType&);
   void Write(const Const&);
   void WriteInitExpr(const ExprList&);
-  void InitGlobalSymbols();
   std::string GenerateHeaderGuard() const;
   void WriteSourceTop();
   void WriteFuncTypes();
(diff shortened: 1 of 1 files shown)


src/c-writer.cc
```diff
@@ -539,6 +453,12 @@ std::string CWriter::LegalizeName(string_view name) {
   for (size_t i = 1; i < name.size(); ++i)
     result += isalnum(name[i]) ? name[i] : '_';
 
+  // In addition to containing valid characters for C, we must also avoid
+  // colliding with things C cares about, such as reserved words (e.g. "void")
+  // or a function name like main() (which a compiler will  complain about if we
+  // define it with another type). To avoid such problems, prefix.
+  result = "__w2c_" + result;
```
binji, line 460: Although it's unlikely to conflict with anything, this is technically a reserved identifier (anything with `__`).  Maybe just `w2c_` as the prefix?

aardappel:
Hmm.. I have similar but less comprehensive filtering going on here, wonder if I should instead use the list you are deleting.. or not at all. These identifiers can get really long though..
https://github.com/WebAssembly/wabt/blob/ebbf625e15b844d43feb24134b9a9cd3ea91daca/src/decompiler-naming.h#L175-L191

src/c-writer.cc
kripken: Oh heh, I was hoping by using reserved ids we'd be less likely to conflict with user code. But yeah, there's a risk with system code. `w2c_` sgtm, but it is short enough that I imagine someone will eventually collide with it...

binji: Should be OK though, since the names here only need to avoid colliding with system headers. If  someone actually uses the name `foo` and `w2c_foo` in the same program, it should be fine since the first will be mangled to `w2c_foo` and the second to `w2c_foo_1`, I believe.

kripken: Oh I see, yes, that sounds safe then.

binji:
@aardappel you could, though this list of identifiers was already too small for what it was trying to accomplish. I basically went through the headers that were being included (only a few, `math.h`, `string.h`, etc) and looked for any identifiers and added them to the list. But for wasm_decompiler, you'll want a lot more than that.

PR #1430, opened 2020-05-18
wasi: Implement more of the wasi API

sbc100:
I've been used the tests from wasi-sdk to get started:
https://github.com/WebAssembly/wasi-sdk/tree/master/tests

All these tests now pass.  Still this isn't saying much, there are still 
some big missing pieces.  Started using the new serdes 
(serialize/deserialze) API in uvwasi.  I think we are almost at the
point were we can look at auto-generating some of this stuff from
witx to avoid having the hand code all this marshelling stuff.

Add support for ArgumentCount::ZeroOrMore to OptionParser and
also the ability to force the end of option processing using `--`. 
This allows is to pass options through the underlying wasi program
when running wasm-interp.


sbc100:
I can split out that OptionParser changes if you like?

src/interp/interp-wasi.cc
```diff
@@ -76,18 +120,247 @@ class WasiInstance {
...
+    /* __wasi_errno_t __wasi_path_open(__wasi_fd_t fd,
+                                       __wasi_lookupflags_t dirflags,
+                                       const char *path,
+                                       size_t path_len,
+                                       __wasi_oflags_t oflags,
+                                       __wasi_rights_t fs_rights_base,
+                                       __wasi_rights_t fs_rights_inherting,
+                                       __wasi_fdflags_t fdflags,
+                                       __wasi_fd_t *opened_fd) */
+    uvwasi_fd_t dirfd = params[0].i32_;
```
binji, line 172: OK for now, but we'll want to check the number of params before accessing them, add a TODO?

src/interp/interp-wasi.cc
```diff
@@ -53,6 +53,49 @@ _Static_assert(_Alignof(__wasi_size_t) == 4, "witx calculated align");
...
+};
+
+typedef struct __wasi_fdstat_t {
+  __wasi_filetype_t fs_filetype;
+  __wasi_fdflags_t fs_flags;
+  __wasi_rights_t fs_rights_base;
+  __wasi_rights_t fs_rights_inheriting;
+} __wasi_fdstat_t;
+
+_Static_assert(sizeof(__wasi_fdstat_t) == 24, "witx calculated size");
```
binji, line 88: nit: nicer to use C++ `static_assert`

src/interp/interp-wasi.cc
```diff
@@ -76,18 +120,247 @@ class WasiInstance {
...
+    uint32_t filestat_ptr = params[4].i32_;
+    char* path;
+    CHECK_RESULT(getMemPtr<char>(path_ptr, path_len, &path, trap));
+    if (trace_stream) {
+      trace_stream->Writef("path_filestat_get : %d %s\n", fd, path);
+    }
+    uvwasi_filestat_t buf;
+    results[0].i32_ =
+        uvwasi_path_filestat_get(uvwasi, fd, flags, path, path_len, &buf);
+    uvwasi_serdes_write_filestat_t(memory->UnsafeData(), filestat_ptr, &buf);
```
binji, line 219: Check that this is a valid range first? Or is that tricky since it's not known what the serialized size is?

src/interp/interp-wasi.cc
```diff
@@ -76,18 +120,247 @@ class WasiInstance {
...
     int32_t fd = params[0].i32_;
+    uint32_t stat_ptr = params[1].i32_;
     if (trace_stream) {
       trace_stream->Writef("fd_fdstat_get %d\n", fd);
     }
-    results[0].i32_ = __WASI_ERRNO_NOENT;
+    CHECK_RESULT(getMemPtr<__wasi_fdstat_t>(stat_ptr, 1, nullptr, trap));
+    uvwasi_fdstat_t statbuf;
+    results[0].i32_ = uvwasi_fd_fdstat_get(uvwasi, fd, &statbuf);
+    uvwasi_serdes_write_fdstat_t(memory->UnsafeData(), params[1].i32_,
```
binji, line 362: use `stat_ptr` instead of `params[1].i32_`?

src/option-parser.cc
```diff
@@ -162,107 +162,117 @@ void OptionParser::HandleArgument(size_t* arg_index, const char* arg_value) {
 
 void OptionParser::Parse(int argc, char* argv[]) {
   size_t arg_index = 0;
+  bool processing_options = true;
```
binji, line 165: Add a test for this in `test-option-parser.cc`

src/interp/interp-wasi.cc
sbc100: This change adds such a TODO in WasiBindImports (which is where I think we would want to check such things right?

src/interp/interp-wasi.cc
sbc100: I'm going to leave as is because these copied directly of wasi-libc api.h and I plan on automating all of this soon.   Will clarify in the comment.

src/interp/interp-wasi.cc
sbc100: Updated all these to use getMemPtr.

src/interp/interp-wasi.cc
sbc100: Done

src/option-parser.cc
sbc100: Done

sbc100:
Got offline lgtm from binji..so landing this.

PR #1431, opened 2020-05-18
Complete conversion to github actions

sbc100:


.github/workflows/build.yml
```diff
@@ -46,3 +60,74 @@ jobs:
...
+    - uses: actions/checkout@v1
+      with:
+        submodules: true
+    - run: sudo apt-get install ninja-build
+    - run: make clang-debug-asan
+    - run: make clang-release-asan
+    - run: make test-clang-debug-asan
+    - run: make test-clang-release-asan
+
+  # Duplicates build-asan.  Please keep in sync
```
binji, line 95: probably can just remove this one, since asan includes lsan now (IIRC)

.github/workflows/build_release.yml
```diff
@@ -0,0 +1,97 @@
...
+    types:
+      - created
+
+jobs:
+  build:
+    name: build
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [linux-16.04, macos-latest, windows-latest]
```
binji, line 15: why is this one not `ubuntu-latest`?

.github/workflows/build_release.yml
```diff
@@ -0,0 +1,97 @@
...
+        upload_url: ${{ github.event.release.upload_url }}
+        asset_path: ./${{ steps.archive.outputs.tarball }}
+        asset_name: ${{ steps.archive.outputs.tarball }}
+        asset_content_type: application/gzip
+
+    # We do the shasum as a separate step because this tool doesn't
+    # exist on windows.
+    - name: shasum
+      id: shasum
+      run: shasum -a 256 ${{ steps.archive.outputs.tarball }} > ${{ steps.archive.outputs.shasum }}
```
binji, line 85: Use `scripts/sha256sum.py` for this on Windows?

.github/workflows/build_release.yml
```diff
@@ -0,0 +1,97 @@
...
+      run: find bin/ -type f -perm -u=x -exec strip {} +
+      if: matrix.os != 'windows-latest'
+
+    - name: archive
+      id: archive
+      run: |
+        OSNAME=$(echo ${{ matrix.os }} | sed 's/-latest//')
+        VERSION=${{ github.event.release.tag_name }}
+        PKGNAME="wabt-$VERSION-$OSNAME"
+        TARBALL=$PKGNAME.tar.gz
```
binji, line 64: we currently do `.zip` files on Windows; would be nice to do that here too.

.github/workflows/build_release.yml
```diff
@@ -0,0 +1,97 @@
...
+        shell: bash
+    steps:
+    - uses: actions/setup-python@v1
+      with:
+        python-version: '3.x'
+    - uses: actions/checkout@v1
+      with:
+        submodules: true
+
+    - name: install ninja (linux)
```
binji, line 27: currently we have archives for linux32/64 and win32/64, I assume these are all 64-bit?

.github/workflows/build_release.yml
sbc100: I figured for max binary compat we should to the release build on an older distro

.github/workflows/build_release.yml
sbc100: Yeah.  I don't see any point in supporting lesser used architectures like that here.  People on platforms other than the 3 most common and always build from source right?   What do you think?

binji: Seems reasonable to me.

.github/workflows/build_release.yml
sbc100: This turns out to be a little tricky.  Perhaps we could find a way as a followup?

binji: Sounds good

binji:
@sbc100 looks like you have changes to uvwasi submodule in this PR too, is that deliberate/

PR #1436, opened 2020-05-21
Allow building all tools with emscripten

kripken:
This allows building say wasm2c.js, and not just wabt.js.

(99% of this is a change to indentation, so please view with `?w=1`)

Also remove `--llvm-lto 1` which is a no-op with the upstream
backend anyhow (however, to get LTO we'd need to build sources
with `-flto`).

On a related note, I'd like to use wasm2c this way, which means I
also need the wasm2c/ directory. Perhaps that could be part of
wabt releases on npm? And would we want `wasm2c.js` to be
part of that package?

cc @dcodeIO @binji


kripken:
Build fails here.. is it using fastcomp maybe?

kripken:
Another option for the npm package might be to keep wabt as it is, and add a new wasm2c package?

sbc100:
Annoying that we don't see the failing command line.. since we are using make and not ninja :(
Looks like the error is:

```
error: DISABLE_EXCEPTION_THROWING was set (likely due to -fno-exceptions), which means no C++ exception throwing support code is linked in, but such support is required by symbol __cxa_begin_catch. Either do not set DISABLE_EXCEPTION_THROWING (if you do want exception throwing) or compile all source files with -fno-except (so that no exceptions support code is required); also make sure DISABLE_EXCEPTION_CATCHING is set to the right value - if you want exceptions, it should be off, and vice versa.
```

You don't see that when building locally?  Maybe we need switch the docker image we are using?

sbc100:
Yeah.. its fastcomp :(  I'll update

kripken:
It builds ok for me locally with both upstream and fastcomp... strange.

kripken:
Oh... it's using 1.38.43. Almost a year old.

sbc100:
Lgtm.. want to land this or are you able to?

kripken:
I thought to wait to see if there are other comments, but this seems low-risk and we can always follow up later and revert. Landing.

PR #1442, opened 2020-05-28
[wasm2c] Implement trap handler for linux

binji:


binji:
@kripken Here's an initial pass at adding signal handling for Linux. PTAL!

kripken:
Very nice!

How would this interact with other signal handlers the host has? I seem to recall wasm VMs have a bunch of complexity there. Not sure if they chain them, or if they enable/disable them at the boundary..?

binji:
> How would this interact with other signal handlers the host has? I seem to recall wasm VMs have a bunch of complexity there. Not sure if they chain them, or if they enable/disable them at the boundary..?

It's a good question, we'll probably want a better way to configure this at some point. Right now, this just overwrites the signal handler and ignores the old one.

syrusakbary:
On the VM side we usually allow the user to set it's own signalhandler via our API.

Once a trap occurs, this is what happens:
* Does the address belong to wasm-generated function? We check the address of the trap, and ask the engine to check if the address is in any of the generated code for the published functions
* In case it doesn't we fallback to the user-provided signalhandler

wasm2c/wasm-rt-impl.c
```diff
@@ -91,13 +102,47 @@ uint32_t wasm_rt_register_func_type(uint32_t param_count,
...
+  if (!g_signal_handler_installed) {
+    g_signal_handler_installed = true;
+    struct sigaction sa;
+    sa.sa_flags = SA_SIGINFO;
+    sigemptyset(&sa.sa_mask);
+    sa.sa_sigaction = signal_handler;
+
+    /* Install SIGSEGV and SIGBUS handlers, since macOS seems to use SIGBUS. */
+    if (sigaction(SIGSEGV, &sa, NULL) == -1 ||
+        sigaction(SIGBUS, &sa, NULL) == -1) {
```
sbc100, line 125: Would checking for `!= 0` be little more precise?

wasm2c/wasm-rt-impl.c
```diff
@@ -91,13 +102,47 @@ uint32_t wasm_rt_register_func_type(uint32_t param_count,
...
+      perror("sigaction failed");
+      abort();
+    }
+  }
+
+  /* Reserve 8GiB. */
+  void* addr =
+      mmap(NULL, 0x200000000ull, PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
```
sbc100, line 133: Isn't `ul` enough for that constant?

wasm2c/wasm-rt-impl.h
```diff
@@ -31,6 +31,14 @@ extern jmp_buf g_jmp_buf;
 /** Saved call stack depth that will be restored in case a trap occurs. */
 extern uint32_t g_saved_call_stack_depth;
 
+#if WASM_RT_MEMCHECK_SIGNAL_HANDLER_POSIX
+#define WASM_RT_SETJMP(buf) sigsetjmp(buf, 1)
+#define WASM_RT_LONGJMP(buf, val) siglongjmp(buf, val)
+#else
+#define WASM_RT_SETJMP(buf) setjmp(buf)
+#define WASM_RT_LONGJMP(buf, val) longjmp(buf, val)
```
sbc100, line 39: What was this needed?   It seems like it would be needed if the sigmask was modified between the a try and a catch, but is that possible?   Isn't the sigmask just set once startup and never again?

wasm2c/wasm-rt.h
```diff
@@ -35,6 +35,26 @@ extern "C" {
...
+/** Whether to enable memory checking via a signal handler.
+ *
+ * This is usually 10%-25% faster, but requires OS-specific support.
+ * */
+#ifndef WASM_RT_MEMCHECK_SIGNAL_HANDLER
+#define WASM_RT_MEMCHECK_SIGNAL_HANDLER 1
+#endif
+
+#if WASM_RT_MEMCHECK_SIGNAL_HANDLER
+#if defined(__linux__) || defined(__unix__) || defined(__APPLE__)
```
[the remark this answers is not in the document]
binji: Apparently not: https://developer.apple.com/library/archive/documentation/Porting/Conceptual/PortingUnix/compiling/compiling.html#//apple_ref/doc/uid/TP40002850-SW13

wasm2c/wasm-rt-impl.h
binji: I'm not 100% sure, but using regular setjmp/longjmp didn't work

wasm2c/wasm-rt-impl.c
binji: I guess that depends whether `long` is 64-bit. I suppose that's true pretty much everywhere though...?

PR #1451, opened 2020-05-29
[wasm2wat] Write select type immediate

binji:
The main fix is in `wat-writer.cc`, where the type immediate was never
being printed. But I've also included a change to how `select` type
immediates are represented in wabt.

Previously, a bare `select` instruction would be stored with the type
`Type::Any`. This is not a real wasm type, and is primarily used for
type validation. The spec instead considers this form of `select` to
have an empty type immediate, which is closer to the `Type::Void` type.

This commit now uses `Type::Void` (or an empty `TypeVector`) to
represent the bare `select` instruction.

### DIFF
--- a/test/typecheck/bad-select-value0.txt
+++ b/test/typecheck/bad-select-value0.txt
@@ -8,10 +8,10 @@
     select
     drop))
 (;; STDERR ;;;
-out/test/typecheck/bad-select-value0.txt:8:5: error: type mismatch in select, expected [f64, f64, i32] but got [i32, f64, f32]
+out/test/typecheck/bad-select-value0.txt:8:5: error: type mismatch in select, expected [any, any, i32] but got [i32, f64, f32]
     select
     ^^^^^^
-out/test/typecheck/bad-select-value0.txt:8:5: error: type mismatch in function, expected [] but got [f64]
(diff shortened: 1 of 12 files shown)


test/typecheck/bad-select-value0.txt
```diff
@@ -8,10 +8,10 @@
     select
     drop))
 (;; STDERR ;;;
-out/test/typecheck/bad-select-value0.txt:8:5: error: type mismatch in select, expected [f64, f64, i32] but got [i32, f64, f32]
+out/test/typecheck/bad-select-value0.txt:8:5: error: type mismatch in select, expected [any, any, i32] but got [i32, f64, f32]
     select
     ^^^^^^
-out/test/typecheck/bad-select-value0.txt:8:5: error: type mismatch in function, expected [] but got [f64]
+out/test/typecheck/bad-select-value0.txt:8:5: error: type mismatch in function, expected [] but got [any]
```
sbc100, line 14: This seems odd because the code moves away from `Any` and towards `Void` but these test results seems to move towards "any"   .. are `Void` and the string "any" somehow releated?

sbc100: Out of curiosity do you know why did the old code includes `f64` here?

binji: Oops, I forgot to change the text parser too. This is back to the old output now.

Issue #1464, opened 2020-06-13
Desugar confuses named variables


pkel:
Consider this example:

```
$ cat module0.wat
(module
  (type (func (param i32) (result i32)))
  (func (type 0) (local $var i32)
        (local.get 0)
        (local.tee $var)))
$ wat-desugar module0.wat
(module
  (type (;0;) (func (param i32) (result i32)))
  (func (;0;) (type 0) (param $var i32) (result i32)
    (local i32)
    local.get $var
    local.tee $var))
$ wat-desugar --version
1.0.15
```

I expect `$var` to be 1, but it is desugared to 0.

The following snippet does what I expect. 

```
$ cat module1.wat
(module
  (type (func (param i32) (result i32)))
  (func (param i32) (result i32) (local $var i32)
        (local.get 0)
        (local.tee $var)))
$ wat-desugar module1.wat
(module
  (type (;0;) (func (param i32) (result i32)))
  (func (;0;) (param i32) (result i32)
    (local $var i32)
    local.get 0
    local.tee $var))
$ wat-desugar --version
1.0.15
```

binji:
Thanks for the report! Fixed in #1466.

PR #1466, opened 2020-06-17
Update local bindings on function with type use

binji:
When a function is specified like this:

    (func (type $t) ...)

the type can have any sequence of params and results, but it is not
known what that signature is until it is resolved. However, locals that
are parsed will be given a binding index under the assumption that the
signature is empty, e.g.

    (func (type $t)
      (local $v0 i32)
      (local $v1 i32)
      ...
    )

So `$v0` will have index 0, and `$v1` will have index 1. If it turns out
later that type `$t` has any parameters, then these indexes need to be
updated.

### DIFF
--- a/src/wast-parser.cc
+++ b/src/wast-parser.cc
@@ -243,7 +243,7 @@ bool IsEmptySignature(const FuncSignature& sig) {
   return sig.result_types.empty() && sig.param_types.empty();
 }
 
-void ResolveFuncTypeWithEmptySignature(const Module& module,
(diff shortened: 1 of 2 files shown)


binji:
Fixes #1464.

rossberg:
Are there any tests in the spec test suite exercising this case? If not, do you have some that you could add?

binji:
@rossberg good point, I thought that this was a roundtripping issue only, but it actually occurs when just parsing too, and a quick test shows that the reference interpreter has the same bug. See https://github.com/WebAssembly/spec/issues/1212.

PR #1468, opened 2020-06-18
Update libwabt.js; fix some runtime issues

binji:
libwabt.js is meant to be included in multiple environments (web,
node.js, etc.) so it shouldn't be using NODERAWFS. A recent change to
the build system include `-s NODERAWFS` for all link steps, but those
are only needed for the wabt executables (e.g. wasm2wast.js).


CMakeLists.txt
```diff
@@ -649,8 +652,11 @@ if (EMSCRIPTEN)
     -s EXPORTED_FUNCTIONS=\"@${EMSCRIPTEN_EXPORTED_JSON}\"
```
kripken, line 652: the `--memory-init-file 0` a few lines above this might be `-s SINGLE_FILE` which will do the same, but also ensure a single file even if we change to wasm later (or other things which might add a new file). but not sure that's the intention here or not.

binji: I think so, yeah. I'll switch to using SINGLE_FILE here too.

PR #1474, opened 2020-06-23
Remove ref.is_null type parameter

binji:
See https://github.com/WebAssembly/reference-types/issues/99.

### DIFF
--- a/src/type-checker.cc
+++ b/src/type-checker.cc
@@ -420,21 +420,11 @@ Result TypeChecker::OnBrTableTarget(Index depth) {
   if (br_table_sig_ == nullptr) {
     br_table_sig_ = &label_sig;
   } else {
-    if (features_.reference_types_enabled()) {
-      if (br_table_sig_->size() != label_sig.size()) {
-        result |= Result::Error;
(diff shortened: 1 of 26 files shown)


src/type-checker.cc
```diff
@@ -420,21 +420,11 @@ Result TypeChecker::OnBrTableTarget(Index depth) {
...
-        PrintError(
-            "br_table labels have inconsistent types: expected %s, got %s",
-            TypesToString(*br_table_sig_).c_str(),
-            TypesToString(label_sig).c_str());
-      }
+    if (*br_table_sig_ != label_sig) {
+      result |= Result::Error;
+      PrintError("br_table labels have inconsistent types: expected %s, got %s",
+                 TypesToString(*br_table_sig_).c_str(),
+                 TypesToString(label_sig).c_str());
```
sbc100, line 427: Is this chunk unrelated?

binji: Yeah, you're right. I think I fixed it because it was caught when I updated the tests. I'll pull into a separate PR.

PR #1485, opened 2020-07-16
Add LoadSplat case to GetExprArity

ngzhian:


ngzhian:
Not sure where tests should go, let me know if needed.

binji:
Thanks! I actually was already working on fix for this and found another case, so I added some tests too: https://github.com/WebAssembly/wabt/pull/1486

PR #1486, opened 2020-07-16
Fix folding for *.load_splat and table.fill

binji:
Also remove default case from `ir-util.cc` so we get compile warnings in
the future.


ngzhian:
Very nice 👍

PR #1488, opened 2020-07-16
Completely disable signal handler on 32-bit

binji:
The previous change prevented `WASM_RT_MEMCHECK_SIGNAL_HANDLER_POSIX`
from being defined, but `WASM_RT_MEMCHECK_SIGNAL_HANDLER` was still
defined, which would prevent the memory bounds check.

### DIFF
--- a/wasm2c/wasm-rt.h
+++ b/wasm2c/wasm-rt.h
@@ -47,6 +47,8 @@ extern "C" {
 #if defined(__linux__) || defined(__unix__) || defined(__APPLE__)
 #if defined(__WORDSIZE) && __WORDSIZE != 64
 #warning "Signal handler is only supported on 64-bit architectures"
+#undef WASM_RT_MEMCHECK_SIGNAL_HANDLER
+#define WASM_RT_MEMCHECK_SIGNAL_HANDLER 0
 #else
 #define WASM_RT_MEMCHECK_SIGNAL_HANDLER_POSIX 1
 #endif
--- a/wasm2c/wasm-rt.h
+++ b/wasm2c/wasm-rt.h
@@ -39,26 +39,32 @@ extern "C" {
  *
  * This is usually 10%-25% faster, but requires OS-specific support.
  * */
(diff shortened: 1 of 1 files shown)


wasm2c/wasm-rt.h
```diff
@@ -47,6 +47,8 @@ extern "C" {
 #if defined(__linux__) || defined(__unix__) || defined(__APPLE__)
 #if defined(__WORDSIZE) && __WORDSIZE != 64
 #warning "Signal handler is only supported on 64-bit architectures"
+#undef WASM_RT_MEMCHECK_SIGNAL_HANDLER
+#define WASM_RT_MEMCHECK_SIGNAL_HANDLER 0
```
sbc100, line 51: Perhaps better not to enable it in the first place just above?

binji: Yeah, but that's only done if it's not already defined, so it could be defined externally too. Though in that case, I guess we'd want to error out instead of warn.

sbc100: I think an error would make more sense yeah.  `undef` seems little odd.

Obviously this just a nit.

binji:
@sbc100 PTAL

wasm2c/wasm-rt.h
```diff
@@ -39,24 +39,32 @@ extern "C" {
  *
  * This is usually 10%-25% faster, but requires OS-specific support.
  * */
+// #define WASM_RT_MEMCHECK_SIGNAL_HANDLER 1
```
sbc100, line 42: remove comment?

binji: I was using it as documentation of what the flag is. Maybe I could make that more clear?

sbc100: Maybe it inside the docstring? without the 1?

Something like `Set the default value for WASM_RT_MEMCHECK_SIGNAL_HANDLER which determines ...`?

binji: Done.

PR #1489, opened 2020-07-16
Fix gcc warning in GetExprArity

binji:
All enumeration values are handled in the switch, but it's still
possible to produce an invalid value, which would not return.


src/ir-util.cc
```diff
@@ -260,4 +259,8 @@ ModuleContext::Arities ModuleContext::GetExprArity(const Expr& expr) const {
     case ExprType::SimdShuffleOp:
       return { 2, 1 };
   }
+
+  fprintf(stderr, "bad expr type: %s\n", GetExprTypeName(expr));
```
sbc100, line 263: Is it that this case is fully covered?   In that case I would use something like LLVM_UNREACHABLE here.

binji: Yeah, good point.

Issue #1492, opened 2020-07-21
Numbering of `local.get` may not account for parameters


alexcrichton:
The test located at [`test/desugar/locals.txt`](https://github.com/WebAssembly/wabt/blob/cd5ff133f84854f0b269f5cb06193ad8205f05d3/test/desugar/locals.txt) which looks like:

```wasm
(module
  (type (func (param i32) (result i32)))
  (func (type 0) (local $var i32)
        (local.get 0)
        (local.tee $var)))
```

will give `$var` local index 0, but in this case shouldn't it get local index 1? I saw this test was added recently 6bff9f079 which looks like it was intended to address the issue, but is this an accident?

tested with:

```
$ wat2wasm test/desugar/locals.txt -o foo.wasm
$ wasm2wat foo.wasm
...
    local.tee 0))
```

alexcrichton:
Fixed in #1494

PR #1494, opened 2020-07-21
Fix resolving local names w/ empty func signatures

binji:
This was originally landed in 6bff9f0, but was incorrect (the test
didn't even do the right thing!)

The problem was that the local index was correctly being updated, but it
happened _after_ the names had already been resolved. To fix it,
`ResolveFuncTypes` and `ResolveNamesModule` needed be called in that
order. There was an erroneous assertion that fired as a result ("This
should only be run after resolving names"), but that's incorrect -- the
code works properly before resolving names too, since
`Module::GetFuncType` can properly look up function types by name.

Thanks to @alexcrichton for pointing this out.


binji:
Fixes #1492.

PR #1495, opened 2020-07-21
Print newline after ref.is_null instruction

binji:


binji:
Fixes #1495.

test/roundtrip/fold-reference-types.txt
```diff
@@ -41,7 +41,8 @@
     (drop
       (ref.null extern))
     (drop
-      (ref.is_null (local.get 0)))
+      (ref.is_null
+        (local.get 0)))
```
sbc100, line 45: Hmm.. isn't the old form more readable?   I guess its not as flexible?

binji: It's more readable here, but it looks pretty weird in flat format (see bug 1493):

```wasm
(module
  (type (;0;) (func))
  (func (;0;) (type 0)
    ref.null func
    ref.is_null drop))
```

PR #1497, opened 2020-07-23
[decompiler] fixed blocks with params.

aardappel:
It would previously assume the blocktype is "simple" (at most a single result value), but now also supports function signatures.
Also fixed it ignoring the validator result.


aardappel:
it should automatically generate variables for any "multi-value" situation, but I am going to leave trying that out that to some other day :)

PR #1500, opened 2020-07-28
Added initial "memory64" proposal support

aardappel:
This is a preview without tests, do not merge


binji:
Looks pretty good! Everything looks about how I would imagined it would be. I'm not sure about the `memory64` keyword, but it's fine for now (maybe use the index type instead, `i64` (or optional `i32`?)).

src/type-checker.cc
```diff
@@ -255,23 +255,23 @@ Result TypeChecker::PopAndCheck3Types(Type expected1,
   return result;
 }
 
-Result TypeChecker::CheckOpcode1(Opcode opcode) {
-  Result result = PopAndCheck1Type(opcode.GetParamType1(), opcode.GetName());
+Result TypeChecker::CheckOpcode1(Opcode opcode, const Limits *limits) {
```
binji, line 258: Maybe add an overload, or use a default value of `limits=nullptr`

aardappel: I tend to use default values sparingly, since I wanted to make sure I considered all the cases where it needs an arg. But I guess I can add that after some further tests :)

aardappel:
I'd be happy to change `memory64` to whatever.. just `i64` seems a little too generic though? `i64` has its place in wasm32 as well. I guess the `memory` part is a bit redundant. `addr64` ? Anyone opinions?

aardappel:
`i64` it is for now :)

src/interp/interp-inl.h
```diff
@@ -631,19 +631,19 @@ inline Memory::Ptr Memory::New(interp::Store& store, MemoryType type) {
   return store.Alloc<Memory>(store, type);
 }
 
-inline bool Memory::IsValidAccess(u32 offset, u32 addend, size_t size) const {
-  return u64{offset} + addend + size <= data_.size();
+inline bool Memory::IsValidAccess(u64 offset, u64 addend, u64 size) const {
+  return offset + addend + size <= data_.size();
```
binji, line 635: Hm, this makes me think we need to be more careful w.r.t. memory accesses that wrap across the 64-bit boundary. It probably needs to check both ends.

src/type-checker.cc
```diff
@@ -598,24 +604,24 @@ Result TypeChecker::OnLoop(const TypeVector& param_types,
...
 }
 
 Result TypeChecker::OnMemoryGrow() {
   return CheckOpcode1(Opcode::MemoryGrow);
 }
 
-Result TypeChecker::OnMemoryInit(uint32_t segment) {
-  return CheckOpcode3(Opcode::MemoryInit);
+Result TypeChecker::OnMemoryInit(uint32_t segment, const Limits& limits) {
+  return CheckOpcode3(Opcode::MemoryInit, &limits, &limits, &limits);
```
binji, line 624: I think we want the size and source offset to be `i32` still, since the data segments are still limited to 32-bits.

aardappel: This particular piece of code does not have any 32->64 bit changes, what code are your referring to?

src/interp/interp-inl.h
aardappel: That in part depends how we define wrapping in the standard, right?
How do we want to do this for now? Check that all 3 components are <=62 bits before the final test? (`size` can be arbitrary as well for one of the bulk memory ops I think).

aardappel: It actually gets interesting: what if `offset + addend` doesn't wrap, but `offset + addend + size` does? :) All pretty hypothetical, but this code does run on every load, so can't be too complicated.

src/type-checker.cc
aardappel: At least in LLVM all 3 operands are now 64-bit, so that would need fixing too then: https://reviews.llvm.org/D82821
I guess I wasn't aware these can only come from data segments, and that we have no plans to make those 64-bit ever..

aardappel: Are the other bulk memory ops correct?

binji: I think so, I just wrote up the new ones here: https://github.com/WebAssembly/memory64/blob/master/proposals/memory64/Overview.md#validation

test/binary/bad-memory-limits-flag.txt
```diff
@@ -3,9 +3,9 @@ magic
 version
 section(MEMORY) {
   count[1]
-  flags[4]
+  flags[8]
```
chfast, line 6: These tests were here on purpose. It is surprising that you are modifying them without any considerations.
Fixed by https://github.com/WebAssembly/wabt/pull/1547.

binji: Unfortunately, the way testing is done in wabt does make this more likely to do -- it's very easy to rebase the tests and accidentally hide a bug.

aardappel: No, I changed these by hand, and I did consider it.. I just somehow thought it was the right thing to do.

PR #1501, opened 2020-07-29
Fix reading of section code

chfast:
The section code is just a single byte, not a LEB128-encoded value.
With this fix the section code is now read correctly with ReadU8().

The issue has been found by fuzzing wabt wasm validation against [Fizzy](https://github.com/wasmx/fizzy).

The wasm spectests reproducing the issue have been submitted to https://github.com/WebAssembly/spec/pull/1230.

### DIFF
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -2360,9 +2360,9 @@ Result BinaryReader::ReadSections() {
   bool seen_section_code[static_cast<int>(BinarySection::Last) + 1] = {false};
 
   for (; state_.offset < state_.size; ++section_index) {
-    uint32_t section_code;
+    uint8_t section_code;
     Offset section_size;
-    CHECK_RESULT(ReadU32Leb128(&section_code, "section code"));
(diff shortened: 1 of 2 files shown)


chfast:
The binary writer may want to get similar change for consistency.

binji:
It looks like the binary writer already writes a u8. Could you add a test for this too? Something like this should be OK (in `test/binary/bad-section-id-leb128.txt`):

```
;;; TOOL: run-gen-wasm-bad
magic
version
section_id[0x81 0]  ;; Section ID cannot be encoded as LEB128
section_size[1]
type_count[0]
```

You'll need to re-run the tests with `-r` to regenerate the expected output for this test.

chfast:
I will do. Thanks for the hint.

chfast:
Done.

chfast:
What's up with this one @binji?

binji:
@chfast was waiting for the CI to run, thanks for updating!

PR #1502, opened 2020-07-29
Update testsuite; fix conversion errors

binji:
The interpreter and wasm2c were incorrectly handling some float-to-int
conversions. For clarity, the wasm2c implementations of these
conversions now matches the implementation in interp-math.h more closely
(e.g. the numeric ranges are written as `x >= min && x <= max` in both
cases).

Quite a few wasm2c tests were previously being skipped, since wasm2c
doesn't currently support multi-value; it's better instead to duplicate
the tests here and disable the parts that are not supported so we don't
lose test coverage.


src/prebuilt/wasm2c.include.c
```diff
@@ -121,25 +121,25 @@ const char SECTION_NAME(declarations)[] =
...
-"#define I32_TRUNC_U_F32(x) TRUNC_U(u32, f32, UINT32_MAX, >=, x)\n"
-"#define I64_TRUNC_U_F32(x) TRUNC_U(u64, f32, UINT64_MAX, >=, x)\n"
-"#define I32_TRUNC_U_F64(x) TRUNC_U(u32, f64, UINT32_MAX, >,  x)\n"
-"#define I64_TRUNC_U_F64(x) TRUNC_U(u64, f64, UINT64_MAX, >=, x)\n"
+"#define TRUNC_S(ut, st, ft, min, minop, max, x)                             \\\n"
+"  ((UNLIKELY((x) != (x)))                        ? TRAP(INVALID_CONVERSION) \\\n"
+"   : (UNLIKELY(!((x)minop(min) && (x) < (max)))) ? TRAP(INT_OVERFLOW)       \\\n"
+"                                                 : (ut)(st)(x))\n"
+"\n"
+"#define I32_TRUNC_S_F32(x) TRUNC_S(u32, s32, f32, (f32)INT32_MIN, >=, 2147483648.f, x)\n"
```
kripken, line 129: shouldn't the minimum be `> 2147483649.f`?

binji: Not for the f32 -> i32 conversion, I believe. 2147483649.f cannot be represented in f32.

kripken: (oops, I meant negative before)

but `-2147483648.8` rounds up to a valid number, an i32 of `-2147483648`?

binji: See https://float.exposed/0xcf000000. The next number that can be represented is https://float.exposed/0xcf000001, which is -2147483904.0.

kripken: I see, thanks!

src/prebuilt/wasm2c.include.c
```diff
@@ -121,25 +121,25 @@ const char SECTION_NAME(declarations)[] =
...
+"#define I64_TRUNC_S_F64(x) TRUNC_S(u64, s64, f64, (f64)INT64_MIN, >=, (f64)INT64_MAX, x)\n"
+"\n"
+"#define TRUNC_U(ut, ft, max, x)                                            \\\n"
+"  ((UNLIKELY((x) != (x)))                       ? TRAP(INVALID_CONVERSION) \\\n"
+"   : (UNLIKELY(!((x) > (ft)-1 && (x) < (max)))) ? TRAP(INT_OVERFLOW)       \\\n"
+"                                                : (ut)(x))\n"
+"\n"
+"#define I32_TRUNC_U_F32(x) TRUNC_U(u32, f32, 4294967296.f, x)\n"
+"#define I64_TRUNC_U_F32(x) TRUNC_U(u64, f32, (f32)UINT64_MAX, x)\n"
+"#define I32_TRUNC_U_F64(x) TRUNC_U(u32, f64, 4294967296.,  x)\n"
```
kripken, line 141: I'm sure I'm wrong here as well, but shouldn't that be `4294967297.` so that fractional values above `2^32` are accepted?

kripken: Ah, no, of course I'm wrong, it's `2^32 - 1`...

binji:
Thanks for double-checking the bounds! I had to think about them a bit to convince myself these are the correct values, but I think they are. For the 8 cases, (s32/u32/s64/u64 <- f32/f64), if we consider which values should be bounds if we had infinite floating-point precision, we would have the following (where _MAX values are the maximum value that is representable type):

| to | from | min (excl.) | max (excl.) |
| - | - | - | - |
| s32 | f32 or f64 | INT32_MIN - 1 |  INT32_MAX + 1 |
| u32 | f32 or f64 |            -1 | UINT32_MAX + 1 |
| s64 | f32 or f64 | INT64_MIN - 1 |  INT64_MAX + 1 |
| u64 | f32 or f64 |            -1 | UINT64_MAX + 1 |

Now we have to consider which of these values are representable as f32 or f64. The values that had to be truncated are in **bold**:

| value | int | f32 | f64 |
| - | - | - | - |
| INT32_MIN - 1 | -2147483649 | **-2147483648.0** | -2147483649.0 |
| INT32_MAX + 1 | 2147483648 | 2147483648 | 2147483648 |
| UINT32_MAX + 1 | 4294967296 | 4294967296 | 4294967296 |
| INT64_MIN - 1 | -9223372036854775809 | **-9223372036854775808.0** | **-9223372036854775808.0** |
| INT64_MAX + 1 | 9223372036854775808 | 9223372036854775808 | 9223372036854775808 |
| UINT64_MAX + 1 | 18446744073709551616 | 18446744073709551616.0 | 18446744073709551616.0 |

And finally, we can determine the bounds, where the ranges that had to be modified are in **bold**:

| instr | to | from | valid bounds |
| - | - | - | - |
| i32.trunc_f32_s | s32 | f32 | **-2147483648.f <=** x < 2147483648.f |
| i32.trunc_f64_s | s32 | f64 | -2147483649.0 < x < 2147483648.0 |
| i32.trunc_f32_u | u32 | f32 | -1.f < x < 4294967296.f |
| i32.trunc_f64_u | u32 | f64 | -1.0 < x < 4294967296.0 |
| i64.trunc_f32_s | s64 | f32 | **-9223372036854775808.f <=** x < 9223372036854775808.f |
| i64.trunc_f64_s | s64 | f64 | **-9223372036854775808.f <=** x < 9223372036854775808.f |
| i64.trunc_f32_u | u64 | f32 | -1.f < x < 18446744073709551616.f |
| i64.trunc_f64_u | u64 | f64 | -1.0 < x < 18446744073709551616.0 |

PR #1503, opened 2020-07-30
Print newline after rethrow instruction

takikawa:
This PR fixes a minor printing consistency issue for the `rethrow` instruction and adds a new test for it.

e.g., in this diff, `wasm2wat` currently prints the `rethrow end` version, which isn't consistent with other instructions I think.

```
   (func (;1;) (type 0)
     try  ;; label = @1
     catch
-      rethrow end)
+      rethrow
+    end)
```


binji:
Thanks!

PR #1505, opened 2020-07-30
Detect malformed memory/table limits flag 

chfast:
First of all the limits flag is just single byte, not a LEB128 encoded value.

Moreover, this detects any unsupported "flags". Apparently, memory supports something like "shared flag" what is not part of wasm 1.0 spec, so this malformed binary will not be detected.


src/binary-reader.cc
```diff
@@ -539,14 +539,16 @@ Result BinaryReader::ReadInitExpr(Index index, bool require_i32) {
...
 
-  uint32_t flags;
+  uint8_t flags;
   uint32_t initial;
   uint32_t max = 0;
-  CHECK_RESULT(ReadU32Leb128(&flags, "table flags"));
-  CHECK_RESULT(ReadU32Leb128(&initial, "table initial elem count"));
+  CHECK_RESULT(ReadU8(&flags, "table flags"));
   bool has_max = flags & WABT_BINARY_LIMITS_HAS_MAX_FLAG;
   bool is_shared = flags & WABT_BINARY_LIMITS_IS_SHARED_FLAG;
```
binji, line 547: The shared flag should only be allowed if threads are enabled (`options_.features.threads_enabled()`), can you change the check here for that?

chfast: Added a check.

PR #1507, opened 2020-07-31
Add tools to use LLVM's libFuzzer

binji:
This is useful for reproducing bugs found by oss-fuzz (see
https://bugs.chromium.org/p/oss-fuzz/issues/list?q=wabt)


sbc100:
Are there docs already ?  Maybe nice to mention in the readme?

PR #1512, opened 2020-08-08
Made the interpreter "type-safe" in debug mode

aardappel:
Adds a type field to `Value` to verify the right union member is being read.
A Wasm module that has been validated shouldn't need this, but any code reading the wrong value would trigger UB, but still often work (like a uint32_t read from a uint64_t on little endian machine), and mask bugs.
Turning this on found bugs in the PR I just landed (as I suspected) but also in older code.

### DIFF
--- a/src/interp/interp-inl.h
+++ b/src/interp/interp-inl.h
@@ -371,53 +371,54 @@ inline bool TypesMatch(ValueType expected, ValueType actual) {
 }
 
 //// Value ////
-inline Value WABT_VECTORCALL Value::Make(s32 val) { Value res; res.i32_ = val; return res; }
-inline Value WABT_VECTORCALL Value::Make(u32 val) { Value res; res.i32_ = val; return res; }
-inline Value WABT_VECTORCALL Value::Make(s64 val) { Value res; res.i64_ = val; return res; }
-inline Value WABT_VECTORCALL Value::Make(u64 val) { Value res; res.i64_ = val; return res; }
-inline Value WABT_VECTORCALL Value::Make(f32 val) { Value res; res.f32_ = val; return res; }
-inline Value WABT_VECTORCALL Value::Make(f64 val) { Value res; res.f64_ = val; return res; }
-inline Value WABT_VECTORCALL Value::Make(v128 val) { Value res; res.v128_ = val; return res; }
-inline Value WABT_VECTORCALL Value::Make(Ref val) { Value res; res.ref_ = val; return res; }
+inline Value WABT_VECTORCALL Value::Make(s32 val) { Value res; res.i32_ = val; res.SetType(ValueType::I32); return res; }
+inline Value WABT_VECTORCALL Value::Make(u32 val) { Value res; res.i32_ = val; res.SetType(ValueType::I32); return res; }
+inline Value WABT_VECTORCALL Value::Make(s64 val) { Value res; res.i64_ = val; res.SetType(ValueType::I64); return res; }
+inline Value WABT_VECTORCALL Value::Make(u64 val) { Value res; res.i64_ = val; res.SetType(ValueType::I64); return res; }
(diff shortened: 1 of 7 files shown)


src/interp/interp-inl.h
```diff
@@ -371,53 +371,56 @@ inline bool TypesMatch(ValueType expected, ValueType actual) {
...
+template <> inline s16x8 WABT_VECTORCALL Value::Get<s16x8>() const { CheckType(ValueType::V128); return Bitcast<s16x8>(v128_); }
+template <> inline u16x8 WABT_VECTORCALL Value::Get<u16x8>() const { CheckType(ValueType::V128); return Bitcast<u16x8>(v128_); }
+template <> inline s32x4 WABT_VECTORCALL Value::Get<s32x4>() const { CheckType(ValueType::V128); return Bitcast<s32x4>(v128_); }
+template <> inline u32x4 WABT_VECTORCALL Value::Get<u32x4>() const { CheckType(ValueType::V128); return Bitcast<u32x4>(v128_); }
+template <> inline s64x2 WABT_VECTORCALL Value::Get<s64x2>() const { CheckType(ValueType::V128); return Bitcast<s64x2>(v128_); }
+template <> inline u64x2 WABT_VECTORCALL Value::Get<u64x2>() const { CheckType(ValueType::V128); return Bitcast<u64x2>(v128_); }
+template <> inline f32x4 WABT_VECTORCALL Value::Get<f32x4>() const { CheckType(ValueType::V128); return Bitcast<f32x4>(v128_); }
+template <> inline f64x2 WABT_VECTORCALL Value::Get<f64x2>() const { CheckType(ValueType::V128); return Bitcast<f64x2>(v128_); }
+
+template <> inline v128& WABT_VECTORCALL Value::GetRef<v128>() { return v128_; }
```
binji, line 414: `GetRef` is only implemented for v128? Maybe we should just remove instead (see below)

src/interp/interp.h
```diff
@@ -523,14 +522,35 @@ union Value {
...
+  };
+
+ public:
+#ifndef NDEBUG
+  Value() : v128_(0, 0, 0, 0), type(ValueType::Any) {}
+  void SetType(ValueType t) { type = t; }
+  void CheckType(ValueType t) const {
+    // Sadly we must allow Any here, since locals may be uninitialized.
+    // Alternatively we could modify InterpAlloca to set the type.
+    assert(t == type || type == ValueType::Any);
```
binji, line 546: This is going to start to get trickier with newer proposals (comparing references, etc.), but for now it's a good check.

src/interp/interp.cc
```diff
@@ -2027,7 +2027,7 @@ RunResult Thread::DoSimdShift(BinopFunc<R, T> f) {
   using ST = typename Simd128<T>::Type;
   using SR = typename Simd128<R>::Type;
   static_assert(ST::lanes == SR::lanes, "SIMD lanes don't match");
-  auto amount = Pop<T>();
+  auto amount = Pop<u32>();
```
binji, line 2030: Nice!

src/tools/spectest-interp.cc
```diff
@@ -703,7 +703,7 @@ wabt::Result JSONParser::ParseLaneConstValue(Type lane_type,
                                              ExpectedValue* out_value,
                                              string_view value_str,
                                              AllowExpected allow_expected) {
-  v128& v = out_value->value.value.v128_;
+  v128& v = out_value->value.value.GetRef<v128>();
```
binji, line 706: Maybe do `Get` here, then `Set` at the bottom instead, then we can remove `GetRef`.

aardappel: That's the first thing I had tried and for some reason it originally didn't work.. but now it does :)

src/interp/interp-inl.h
aardappel: ok

Issue #1513, opened 2020-08-10
Clarify how Supported Proposals work.


carlsmith:
In the project README, under *Supported Proposals*, it lists the flags for the various proposals, and says the following:

* Proposal: Name and link to the WebAssembly proposal repo
* flag: Flag to pass to the tool to enable support for the feature
* binary: Whether wabt can read/write the binary format
* text: Whether wabt can read/write the text format
* validate: Whether wabt can validate the syntax
* interpret: Whether wabt can execute these operations in wasm-interp or spectest-interp

This implies that the `multi-value` flag (for example) is required for the tools to parse multi-value WAT syntax or to generate the corresponding binary code. However, I am able to compile and run WAT code that uses the multi-value feature, without passing any flags to the tool (obviously, the feature is enabled in my browser).

In short, the README doesn't explain what the flags do very clearly.

binji:
Good point, now that a few of these proposals are standardized, they have been enabled by default in wabt. The README should be updated to make that clearer.

carlsmith:
Ah, that makes sense. Thanks, @binji.

carlsmith:
It makes much more sense now. Thank you, @binji.

PR #1515, opened 2020-08-10
Update README.md to include default-on proposals

binji:
Some proposals are enabled by default now, so the README should reflect that. Also, the flag was flipped to disable (rather than enable), so change that here too.

See issue #1513.


sbc100:
We've been talking about the tooling-conventions document that we can point to for rational as to when to enable a feature by default.    Once that exists maybe we can link to it from here?

binji:
> We've been talking about the tooling-conventions document that we can point to for rational as to when to enable a feature by default. Once that exists maybe we can link to it from here?

Sounds good, let me know when that's ready.

PR #1517, opened 2020-08-12
Fix relocations emitted by wat2wasm -r

sammax:
Adds relocations for block types and fixes the relocation offsets. 
I'll add tests for it once I figure out how that works.


sammax:
The failing regression test was meant to ensure that type relocs cause an error message instead of crashing. Since that is no longer relevant, should I remove the test or update it to conform with the new behaviour (i.e. make it expect no output)?

It did make me realize though that I didn't consider type relocs from `call_indirect` at all. I think they work fine, but I should probably integrate them in my tests as well.

src/binary-writer.cc
```diff
@@ -1188,8 +1189,17 @@ Result BinaryWriter::WriteModule() {
       Offset body_size_offset =
           WriteU32Leb128Space(leb_size_guess, "func body size (guess)");
       WriteFunc(func);
-      WriteFixupU32Leb128Size(body_size_offset, leb_size_guess,
+      auto func_start_offset = body_size_offset - last_section_payload_offset_;
+      auto func_end_offset = stream_->offset() - last_section_payload_offset_;
+      auto delta = WriteFixupU32Leb128Size(body_size_offset, leb_size_guess,
                               "FIXUP func body size");
+      if (current_reloc_section_ && delta != 0) {
```
binji, line 1196: Why are the relocations only written at the end of the functions section? Can you explain the bug that this is fixing a bit more?

test/dump/relocations-long-func-bodies.txt
```diff
@@ -0,0 +1,98 @@
+;;; TOOL: run-objdump
```
binji, line 1: Is there something about long function bodies that was producing incorrect output before?

sammax:
@binji Replying here instead of directly to the comments since they are both related.
The relocations are only written in the relocation section, in the code section whenever something that needs to be relocated is encountered, a 5-byte placeholder is written and the offset of that placeholder and the index of the symbol/type that belongs there is added to the relocation list. 
In the relocation custom section those relocations from the list are then written.
The problem that arises with long function bodies is that the size of a function is written at the beginning of it, when it is not yet known. So instead a placeholder is written, which is only one byte. At the end of the function, the size is known and the previously written size is updated. In a long function that size will need more than one byte to encode, which moves the entire function body by the difference between the length of the new size encoding and the old one (which is just 1 byte long). Because the function body moves, all the relocation offsets within that function need to be updated. This did not previously happen, which caused the relocation offsets to be wrong in certain circumstances.
The same thing happens with the function section: its size is written at the beginning as a one-byte placeholder and replaced with the actual size at the end of the section. This moves the entire section if it is big enough that the encoding of its size needs more than one byte. 
However, the relocation offsets are counted from the beginning of the payload, not the beginning of the section, so changing the length of the size encoding does not affect them. Previously they were updated at the end of the function section the same way they are now at the end of a function, which caused the offsets to be wrong.

The bug was not apparent in isolated tests of a single function, since in that case the function size and section size are likely to need the same number of bytes to encode, so updating the offsets at the end of the section made up for not updating them at the end of the function.

sammax:
Any idea what made the windows build fail? It says the test failed after running for 38 minutes, but there are no logs at all.

binji:
@sammax Maybe just a CI flake? I'm rerunning the jobs now. And thanks for explaining, this makes more sense to me now.

src/binary-writer.cc
```diff
@@ -245,7 +245,7 @@ void BinaryWriter::WriteBlockDecl(const BlockDeclaration& decl) {
   Index index = decl.has_func_type ? module_->GetFuncTypeIndex(decl.type_var)
                                    : module_->GetFuncTypeIndex(decl.sig);
   assert(index != kInvalidIndex);
-  WriteS32Leb128(stream_, index, "block type function index");
+  WriteU32Leb128WithReloc(index, "block type function index", RelocType::TypeIndexLEB);
```
sbc100, line 248: I think this should be conditional on `decl.has_func_type`?   Normal block with a single return value don't need relocations, right?

sammax: I haven't checked the spec, but from the existing code I assumed that we need a type with reloc for all blocks that have more than zero params or more than one result. The method returns early in these cases.

sbc100: There are two type is block signature ... either the block has a basic value type .. or it has a function type.  You can see this being checked above with `decl.has_func_type`.

If you do this unconditionally then *every* block will end up with a relocation.  This is not what we want.  Ordinary block (all blocks that predate the multi value proposal) should *not* have relocations for their types.

Its not just an optimization is simply won't work.  Each TypeIndexLEB relocation must correspond for an entry in the type section, and there are no such entry for blocks with value types.

sbc100: Apologies I didn't fully read your comment.  You are correct, regular blocks should return early here and this path should only be for multivalue enabled blocks.

sammax:
@sbc100 Yeah I can do that. Should I split out the line that adds the block type relocs together with the parts that make type relocs work, or put the latter in another separate PR since it affects indirect calls in addition to block types?

sbc100:
Good point.  How about we land the `RelocType::TypeIndexLEB` fix on its own first?  Then we can add block relocs as a followup.

sammax:
Yeah that sounds good, I'll split it up later today.

sammax:
I'm closing this in favour of the three new PRs (see above).

Issue #1249, opened 2019-11-22
Run CI with -Werror (cmake -DWERROR=ON)


sbc100:

sbc100:
Just got bitten by this #1521

PR #1522, opened 2020-08-18
Enable -Werror during CI

sbc100:
Fixes: #1249


sbc100:
Now contains a couple of actual fixed too.   FYI.

sbc100:
Can I get another LGTM before I land this since I need to do fair bit more hacking ..

PR #1523, opened 2020-08-18
[decompiler] Fix integer comparison sign warning

aardappel:


binji:
Looks like this was fixed in https://github.com/WebAssembly/wabt/pull/1521 (which has other fixes too).

Issue #1520, opened 2020-08-17
wasm-objdump/wabt::ObjdumpNames::Set integer overflow after reading export items' index


dwfault:
PoC:
```
dwfault-mbp:~ dwfault$ hexdump PoC.wasm.objdump.1ff 
0000000 00 61 73 6d 01 00 00 00 01 05 01 60 01 7f 00 03
0000010 03 02 00 00 05 03 01 00 00 07 16 03 03 6d 65 6d
0000020 02 00 04 6c 6f 57 64 00 [ff ff ff ff 0f] 1a 0b 0d
0000030 00 20 1c 02 86 80 80 80 00 02 ff ff ff ff 0f 1a
0000040 0b 0d 00 20 00 41 00 36 23 f6 ff ff ff 0f 0b   
000004f
```
The 0x0f ff ff ff ff in the brackets is the index of an export item in LEB128 unsigned form. And there's a crashing stack trace when parsing it:
```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x17ffffffe8)
  * frame #0: 0x00000001000027d5 wasm-objdump`std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::__is_long(this="") const at string:1420:22
    frame #1: 0x0000000100011115 wasm-objdump`std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::clear(this="") at string:3063:9
    frame #2: 0x000000010001104c wasm-objdump`std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::__clear_and_shrink(this="") at string:3820:5
    frame #3: 0x0000000100010fe0 wasm-objdump`std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::__move_assign(this="", __str="loWd", (null)=std::__1::true_type @ 0x00007ffeefbfec98) at string:2292:5
    frame #4: 0x00000001000020e4 wasm-objdump`std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::operator=(this="", __str="loWd") at string:2304:5
    frame #5: 0x0000000100001fef wasm-objdump`wabt::ObjdumpNames::Set(this=0x00007ffeefbff5d8, index=4294967295, name=(data_ = "loWd", size_ = 4)) at binary-reader-objdump.cc:1864:16
[...]

PR #1524, opened 2020-08-19
Fix integer overflow in objdump name tracking

sbc100:
Fixes: #1520

### DIFF
--- a/test/binary/bad-export-out-of-range.txt
+++ b/test/binary/bad-export-out-of-range.txt
@@ -1,0 +1,11 @@
+;;; TOOL: run-gen-wasm-bad
+magic
+version
+section(TYPE) { count[1] function params[0] results[0] }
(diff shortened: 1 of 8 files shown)


test/binary/bad-export-out-of-range.txt
```diff
@@ -0,0 +1,11 @@
+;;; TOOL: run-gen-wasm-bad
+magic
+version
+section(TYPE) { count[1] function params[0] results[0] }
+section(FUNCTION) { count[1] sig[0] }
+section(EXPORT) { count[1] str("foo") func_kind func[0xff] func[0xff] func[0xff] func[0xff] func[0x0f] }
```
binji, line 6: you can change this to `func[leb_u32(4294967295)]` to make the test a bit clearer

PR #1525, opened 2020-08-19
wat2wasm: type relocations

sammax:
split off from #1517


src/binary-writer.cc
```diff
@@ -323,8 +323,9 @@ Index BinaryWriter::GetSymbolIndex(RelocType reloc_type, Index index) {
       type = SymbolType::Global;
       name = module_->globals[index]->name;
       break;
+    case RelocType::TypeIndexLEB:
```
sbc100, line 326: Maybe just return `index` here with a comment about that fact that TypeIndexLEB doing use symbol indexes at all?  Would make this change smaller I think.

As it stands this change isn't having any effect at all right?  since it will drop through the default case and print the warning just like before.

test/dump/relocations-indirect-calls-and-table-elements.txt
```diff
@@ -0,0 +1,70 @@
...
+ - func[0] size=8 <$add>
+ - func[1] size=18 <run>
+Custom:
+ - name: "linking"
+  - symbol table [count=1]
+   - 0: F <$add> func=0 binding=global vis=default
+Custom:
+ - name: "reloc.Elem"
+  - relocations for section: 4 (Elem) [1]
+   - R_WASM_FUNCTION_INDEX_LEB offset=0x000006(file=0x000030) symbol=0 <$add>
```
sbc100, line 48: Interesting.   In llvm we don't currently use any relocations in the Elem section,  This is because the linker creates the elem section from scratch based on the functions that get called indirectly.

So I don't think we need this.

Maybe just call this new test `relocations_type_index.txt`?  Or just add to the existing `relocations.txt`

sbc100: Yeah I looks like we only currently support relocations in DATA, CODE, and custom sections:

https://github.com/llvm/llvm-project/blob/ca688ae4975932f0566b113dad10a66db852259e/lld/wasm/Writer.cpp#L168

src/binary-writer.cc
sammax: yeah it doesn't have any effect, I just put it there so the comment makes sense. We could return the index, but personally I would prefer `assert`ing that it's not a type index. Or otherwise change the function name to reflect that it doesn't necessarily deal with symbols.
This is not a very strong preference though.

test/dump/relocations-indirect-calls-and-table-elements.txt
sammax: The elem reloc section was already being created, I didn't introduce it. Does the linker correctly create the elements with relocated indices? If so, I'll look into removing the whole section. And yeah I'll rename the test.

src/binary-writer.cc
sbc100: I think having it return the right thing here make sense.

If you want to rename the function thats is fine, but I think a comment explaining that the "symbol index" for this type of relocation is not actually a symbol should be OK.   At least I can't think of a more descriptive name for this function. 

The return value of this function is assigned to the symbol index of the relocation so in some sense we are returning something that well be treated like a symbol index.

sammax: Yeah that makes sense, I'll do that.

test/dump/relocations-indirect-calls-and-table-elements.txt
sammax: Ah, reloc sections are automatically created based on the name of the section in which a reloc is added. So we would just remove the relocs in the elem section then to remove the elem reloc section.

sbc100: Yup!

PR #1527, opened 2020-08-19
wat2wasm: fix relocation offsets

sammax:
split off from #1517

### DIFF
--- a/src/binary-writer.cc
+++ b/src/binary-writer.cc
@@ -277,8 +277,8 @@ void BinaryWriter::BeginCustomSection(const char* name) {
 
 void BinaryWriter::EndSection() {
   assert(last_section_leb_size_guess_ != 0);
-  Offset delta = WriteFixupU32Leb128Size(
-      last_section_offset_, last_section_leb_size_guess_, "FIXUP section size");
+  WriteFixupU32Leb128Size(last_section_offset_, last_section_leb_size_guess_,
+                          "FIXUP section size");
   last_section_leb_size_guess_ = 0;
   section_count_++;
 }
--- a/src/binary-writer.cc
+++ b/src/binary-writer.cc
@@ -279,11 +279,6 @@ void BinaryWriter::EndSection() {
   assert(last_section_leb_size_guess_ != 0);
   Offset delta = WriteFixupU32Leb128Size(
       last_section_offset_, last_section_leb_size_guess_, "FIXUP section size");
-  if (current_reloc_section_ && delta != 0) {
-    for (Reloc& reloc : current_reloc_section_->relocations) {
-      reloc.offset += delta;
-    }
-  }
   last_section_leb_size_guess_ = 0;
   section_count_++;
 }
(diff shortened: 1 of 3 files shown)


sammax:
I just realized I wrote "function section" in the commit message, but I meant "code section".

pmatos:
I just hit this bug - it would be really useful to have this merged.

src/binary-writer.cc
```diff
@@ -1188,8 +1183,17 @@ Result BinaryWriter::WriteModule() {
...
+      auto func_end_offset = stream_->offset() - last_section_payload_offset_;
+      auto delta = WriteFixupU32Leb128Size(body_size_offset, leb_size_guess,
                               "FIXUP func body size");
+      if (current_reloc_section_ && delta != 0) {
+        for (Reloc& reloc : current_reloc_section_->relocations) {
+          if (reloc.offset >= func_start_offset && reloc.offset <= func_end_offset) {
+            reloc.offset += delta;
+          }
+        }
+      }
```
sbc100, line 1196: The complexity needed to handle these adjustments is a little sad.  I wonder if there is a better way we can do this.  I can think of a few options:

1. Write a patchable 5-byte LEB here
2. Have some way of precalualting the size of a function.
3. WriteFunc to a temporary location and them memmove into place once we know the size?

I'm not suggesting these for this PR but perhaps for future works.

I can't remember why we didn't choose any of these originally.  @binji what do you think?

test/dump/relocations-long-func-bodies.txt
```diff
@@ -0,0 +1,98 @@
+;;; TOOL: run-objdump
+;;; ARGS0: -r
+;;; ARGS1:
+(module
+  (global $g i32 (i32.const 0))
+  (func $long_func
```
sbc100, line 6: Can we have a preceding function there that also contains a relocation so check that the relocations in preceding functions are not also adjusted?

sbc100 (approved):
lgtm % comments

src/binary-writer.cc
sammax: Considering that we use patchable 5-byte LEBs in relocatable code anyways and it gets optimized later, I think that might be the easiest and most sensible solution.

test/dump/relocations-long-func-bodies.txt
sammax: sure, I'll add one

src/binary-writer.cc
binji: Yeah, the complexity here is annoying. I want to make sure we don't write a larger size (just in case optimization doesn't happen), but perhaps we can do it by default when the output is relocatable.

sbc100: Yes exactly we would only write wide-LEBs for relocatable output which will be used as input wasm-ld

sammax: If you want I can do that in this PR, it seems kinda pointless to introduce complexity just to remove it right afterwards.

sbc100: I'm fine with landing this PR as is.  It doesn't really add much.. just move the existing loop and adds a condition.

We can reason about using wide LEBs as a followup.

sammax: although it might also make sense to just make a new PR for that and close this one, because it is a bit broader in scope.

sbc100:
OK to land this now?

sammax:
Ah sorry I missed your comment, let me add the test I promised first.
Do I need to rebase?

binji:
> Do I need to rebase?

Doesn't look like it, I'll merge. Thanks for the patch!

Issue #1531, opened 2020-09-02
Incorrect encoding of multi-value block types


alexcrichton:
This text file:

<details>

```wasm
(module
  ;; 64 random types to force the block type below to get encoded as sleb
  (func (param f32 f32 f32 f32 f32 f32))
  (func (param i32 f32 f32 f32 f32 f32))
  (func (param f32 i32 f32 f32 f32 f32))
  (func (param i32 i32 f32 f32 f32 f32))
  (func (param f32 f32 i32 f32 f32 f32))
  (func (param i32 f32 i32 f32 f32 f32))
  (func (param f32 i32 i32 f32 f32 f32))
  (func (param i32 i32 i32 f32 f32 f32))
  (func (param f32 f32 f32 i32 f32 f32))
  (func (param i32 f32 f32 i32 f32 f32))
  (func (param f32 i32 f32 i32 f32 f32))
  (func (param i32 i32 f32 i32 f32 f32))
  (func (param f32 f32 i32 i32 f32 f32))
  (func (param i32 f32 i32 i32 f32 f32))
  (func (param f32 i32 i32 i32 f32 f32))
  (func (param i32 i32 i32 i32 f32 f32))
  (func (param f32 f32 f32 f32 i32 f32))
  (func (param i32 f32 f32 f32 i32 f32))
  (func (param f32 i32 f32 f32 i32 f32))
  (func (param i32 i32 f32 f32 i32 f32))
  (func (param f32 f32 i32 f32 i32 f32))
  (func (param i32 f32 i32 f32 i32 f32))
  (func (param f32 i32 i32 f32 i32 f32))
  (func (param i32 i32 i32 f32 i32 f32))
  (func (param f32 f32 f32 i32 i32 f32))
  (func (param i32 f32 f32 i32 i32 f32))
  (func (param f32 i32 f32 i32 i32 f32))
  (func (param i32 i32 f32 i32 i32 f32))
  (func (param f32 f32 i32 i32 i32 f32))
  (func (param i32 f32 i32 i32 i32 f32))
  (func (param f32 i32 i32 i32 i32 f32))
  (func (param i32 i32 i32 i32 i32 f32))
  (func (param f32 f32 f32 f32 f32 i32))
  (func (param i32 f32 f32 f32 f32 i32))
  (func (param f32 i32 f32 f32 f32 i32))
  (func (param i32 i32 f32 f32 f32 i32))
  (func (param f32 f32 i32 f32 f32 i32))
[...]

PR #1533, opened 2020-09-03
Fix bug when writing multi-value block type

binji:
The block type must be written as an SLEB, not an LEB. We also want it
to be relocatable, so I added a way to write a fixed-size SLEB too
(`WriteS32Leb128WithReloc` and `WriteFixedS32Leb128`).


binji:
Fixes #1531.

test/roundtrip/multi-value-block-type.txt
```diff
@@ -0,0 +1,149 @@
+;;; TOOL: run-roundtrip
+;;; ARGS: --stdout --reloc
+(module
+  ;; 64 random types to force the block type below to get encoded as sleb
```
sbc100, line 4: Why 64?    I guess because once we hit a 7 bit number the signed vs unsigned encoding differ?

sbc100 (approved):

test/roundtrip/multi-value-block-type.txt
binji: Yes, that's why. 64 must be encoded as two bytes as an SLEB, but can be encoded in just one byte as LEB.

PR #1535, opened 2020-09-08
[NFC] Refactor symbol table creation in binary writer

### LINKED ISSUE #1534: wat2wasm --relocatable should record exports in linking symtab
Consider `a.c` which exports a function that calls something from `b.c`:

```
$ cat a.c
void b(void);
__attribute__((export_name("a"))) void a(void) { b(); }

$ cat b.c
void b(void) {}
```

With clang and wasm-ld, we can make a wasm file:

```
$ clang -Oz --target=wasm32 -nostdlib -c -o a.o a.c
$ clang -Oz --target=wasm32 -nostdlib -c -o b.o b.c
$ wasm-ld --no-entry -o clang.wasm a.clang.o b.o
```

Let's say we want to replace `a.c` with a wat file.  

```
$ wasm2wat -o a.wat a.o
$ wat2wasm --relocatable -o a.wat.o a.wat
$ wasm-ld --no-entry -o wat.wasm a.wat.o b.o
```

OK great.  However:

```
$ wasm2wat wat.wasm
(module
  (import "env" "memory" (memory (;0;) 2))
  (table (;0;) 1 1 funcref)
  (global (;0;) (mut i32) (i32.const 66560)))
```

That is to say: dude, where's my export (of `a`)?

The problem is that in the a.o as produced by clang, the linking and reloc sections look like this:

```
Custom:
 - name: "linking"
  - symbol table [count=2]
   - 0: F <a> func=1 exported no_strip binding=global vis=hidden
   - 1: F <env.b> func=0 undefined binding=global vis=default
Custom:
 - name: "reloc.CODE"
   - R_WASM_FUNCTION_INDEX_LEB offset=0x000004(file=0x000077) symbol=1 <env.b>
Custom:
 - name: "producers"

Code Disassembly:

000075 func[1] <a>:
 000076: 10 80 80 80 80 00          | call 0 <env.b>
 00007c: 0b                         | end
```

Whereas as produced by `wat2wasm`, they look like this:

```
Custom:
 - name: "linking"
  - symbol table [count=1]
   - 0: F <env.b> func=0 undefined binding=global vis=default
Custom:
 - name: "reloc.Code"
   - R_WASM_FUNCTION_INDEX_LEB offset=0x000004(file=0x000063) symbol=0 <env.b>

Code Disassembly:

000061 func[1] <a>:
 000062: 10 80 80 80 80 00          | call 0 <env.b>
 000068: 0b                         | end
```

I know that on #1137 there is a question about whether to keep up the functionality of --relocatable at all, but I think that with a couple of bug-fixes it can be usable enough for me, so I will take a look at fixing this unless there's an objection.

wingo:
This commit refactors the representations of symbols in the binary writer.  It will allow us to fix #1534 in a followup.

### DIFF
--- a/src/binary-writer.cc
+++ b/src/binary-writer.cc
@@ -97,10 +97,97 @@ struct RelocSection {
   std::vector<Reloc> relocations;
 };
 
-struct Symbol {
-  Index symbol_index;
-  SymbolType type;
-  Index element_index;
+class Symbol {
+ public:
+  struct Function {
+    static const SymbolType type = SymbolType::Function;
+    Index index;
+  };
+  struct Data {
+    static const SymbolType type = SymbolType::Data;
+    Index index;
+    Offset offset;
+    Address size;
+  };
+  struct Global {
+    static const SymbolType type = SymbolType::Global;
+    Index index;
+  };
+  struct Section {
+    static const SymbolType type = SymbolType::Section;
+    Index section;
+  };
+  struct Event {
+    static const SymbolType type = SymbolType::Event;
+    Index index;
+  };
+
+ private:
+  SymbolType type_;
+  std::string name_;
+  uint8_t flags_;
+  union {
+    Function function_;
+    Data data_;
+    Global global_;
+    Section section_;
+    Event event_;
+  };
+
+ public:
+  Symbol(const std::string& name, uint8_t flags, const Function& f)
+      : type_(Function::type), name_(name), flags_(flags), function_(f) {}
+  Symbol(const std::string& name, uint8_t flags, const Data& d)
+      : type_(Data::type), name_(name), flags_(flags), data_(d) {}
+  Symbol(const std::string& name, uint8_t flags, const Global& g)
+      : type_(Global::type), name_(name), flags_(flags), global_(g) {}
+  Symbol(const std::string& name, uint8_t flags, const Section& s)
+      : type_(Section::type), name_(name), flags_(flags), section_(s) {}
+  Symbol(const std::string& name, uint8_t flags, const Event& e)
+      : type_(Event::type), name_(name), flags_(flags), event_(e) {}
+
+  SymbolType type() const { return type_; }
+  const std::string& name() const { return name_; }
+  uint8_t flags() const { return flags_; }
+
+  SymbolVisibility visibility() const {
+    return static_cast<SymbolVisibility>(flags() & WABT_SYMBOL_MASK_VISIBILITY);
+  }
+  SymbolBinding binding() const {
+    return static_cast<SymbolBinding>(flags() & WABT_SYMBOL_MASK_BINDING);
+  }
+  bool undefined() const { return flags() & WABT_SYMBOL_FLAG_UNDEFINED; }
+  bool defined() const { return !undefined(); }
+  bool exported() const { return flags() & WABT_SYMBOL_FLAG_EXPORTED; }
+  bool explicit_name() const { return flags() & WABT_SYMBOL_FLAG_EXPLICIT_NAME; }
+  bool no_strip() const { return flags() & WABT_SYMBOL_FLAG_NO_STRIP; }
+
+  bool IsFunction() const { return type() == Function::type; }
+  bool IsData() const { return type() == Data::type; }
+  bool IsGlobal() const { return type() == Global::type; }
+  bool IsSection() const { return type() == Section::type; }
+  bool IsEvent() const { return type() == Event::type; }
+
+  const Function& AsFunction() const {
+    assert(IsFunction());
+    return function_;
+  }
+  const Data& AsData() const {
+    assert(IsData());
+    return data_;
+  }
+  const Global& AsGlobal() const {
+    assert(IsGlobal());
+    return global_;
+  }
+  const Section& AsSection() const {
+    assert(IsSection());
+    return section_;
+  }
+  const Event& AsEvent() const {
+    assert(IsEvent());
+    return event_;
+  }
 };
 
 class BinaryWriter {
@@ -125,6 +212,8 @@ class BinaryWriter {
   void EndSection();
   void BeginSubsection(const char* name);
   void EndSubsection();
+  template <typename T>
+  Index InternSymbol(const std::string& name, uint8_t flags, const T& arg);
   Index GetLabelVarDepth(const Var* var);
   Index GetEventVarDepth(const Var* var);
   Index GetLocalIndex(const Func* func, const Var& var);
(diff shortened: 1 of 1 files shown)


src/binary-writer.cc
```diff
@@ -97,10 +97,97 @@ struct RelocSection {
...
+  bool explicit_name() const { return flags() & WABT_SYMBOL_FLAG_EXPLICIT_NAME; }
+  bool no_strip() const { return flags() & WABT_SYMBOL_FLAG_NO_STRIP; }
+
+  bool isFunction() const { return type() == Function::type; }
+  bool isData() const { return type() == Data::type; }
+  bool isGlobal() const { return type() == Global::type; }
+  bool isSection() const { return type() == Section::type; }
+  bool isEvent() const { return type() == Event::type; }
+
+  const Function& asFunction() const {
```
wingo, line 171: Hum, I see that I have SpiderMonkey conventions at my fingers; would need to fix this to AsFunction I guess (and same for IsFunction etc above)

binji: Yep, that would be nice, thanks!

src/binary-writer.cc
```diff
@@ -312,34 +401,43 @@ Index BinaryWriter::GetEventVarDepth(const Var* var) {
   return var->index();
 }
 
+template <typename T>
+Index BinaryWriter::InternSymbol(const std::string& name, uint8_t flags,
+                                 const T& arg) {
+  auto iter = symtab_.find(name);
+  if (iter != symtab_.end()) {
+    Index sym_index = iter->second;
+    assert(symbols_[sym_index].type() == T::type);
```
binji, line 410: I'm not sure whether this assert is safe; seems like it could be invalid based on malformed data in wasm module. @sbc100 do you know?

src/binary-writer.cc
```diff
@@ -312,34 +401,43 @@ Index BinaryWriter::GetEventVarDepth(const Var* var) {
...
+
 Index BinaryWriter::GetSymbolIndex(RelocType reloc_type, Index index) {
-  std::string name;
-  SymbolType type = SymbolType::Function;
+  uint8_t flags = 0;
   switch (reloc_type) {
     case RelocType::FuncIndexLEB:
-      name = module_->funcs[index]->name;
-      break;
+      if (index < module_->num_func_imports)
```
binji, line 425: nit: {} braces around if

src/binary-writer.cc
```diff
@@ -312,34 +401,43 @@ Index BinaryWriter::GetEventVarDepth(const Var* var) {
...
+        flags |= WABT_SYMBOL_FLAG_UNDEFINED;
+      return InternSymbol(module_->funcs[index]->name, flags, Symbol::Function{index});
     case RelocType::GlobalIndexLEB:
-      type = SymbolType::Global;
-      name = module_->globals[index]->name;
-      break;
+      if (index < module_->num_global_imports)
```
binji, line 429: same here

binji (approved):
Looks good, but let @sbc100 take a look too. Can you modify/add on to one of the existing tests (perhaps `test/binary/linking_section.txt`?)

src/binary-writer.cc
wingo: Good point; let's print a warning in that case instead of failing.

In any case, the current logic doesn't actually follow the toolchain-conventions invariants; symbols that are neither imported nor exported by a module have no name uniqueness requirements AFAIU.  I will fix the logic in a followup.

wingo:
Updated patch addresses review comments, except regarding tests: because this patch has no functional change, no test needs to be modified.  I'll extend the `linking-section.txt` test in the followup though; thanks for the pointer!

wingo:
(And apologies wrt force-pushes; I gather the convention is to push commits to fix nits, then you rebase/squash when landing -- will do that next time.)

sbc100:
> (And apologies wrt force-pushes; I gather the convention is to push commits to fix nits, then you rebase/squash when landing -- will do that next time.)

I don't think we have any rules around this.   I basically always use a rebase flow which means I often/always have to force push to github PRs (even if i don't squash, which I also occasionally do too).

src/binary-writer.cc
```diff
@@ -97,10 +97,97 @@ struct RelocSection {
...
+    Offset offset;
+    Address size;
+  };
+  struct Global {
+    static const SymbolType type = SymbolType::Global;
+    Index index;
+  };
+  struct Section {
+    static const SymbolType type = SymbolType::Section;
+    Index section;
```
sbc100, line 118: Why not call this `index` like the others?   Or the other way around and don't call any of them `index`?

src/binary-writer.cc
```diff
@@ -97,10 +97,97 @@ struct RelocSection {
...
+  SymbolBinding binding() const {
+    return static_cast<SymbolBinding>(flags() & WABT_SYMBOL_MASK_BINDING);
+  }
+  bool undefined() const { return flags() & WABT_SYMBOL_FLAG_UNDEFINED; }
+  bool defined() const { return !undefined(); }
+  bool exported() const { return flags() & WABT_SYMBOL_FLAG_EXPORTED; }
+  bool explicit_name() const { return flags() & WABT_SYMBOL_FLAG_EXPLICIT_NAME; }
+  bool no_strip() const { return flags() & WABT_SYMBOL_FLAG_NO_STRIP; }
+
+  bool IsFunction() const { return type() == Function::type; }
```
sbc100, line 165: Why not just use `SymbolType::Function` here?  What purpose does the `Function::type` serve?

sbc100 (approved):
LGTM.  Maybe add "NFC" to the PR title?

src/binary-writer.cc
wingo: I was using the names from https://github.com/WebAssembly/tool-conventions/blob/master/Linking.md, where this field is called "section".

src/binary-writer.cc
wingo: The purpose of the `::type` definitions is to allow `InternSymbol` to know what kind of symbol to make.  I admit that I wasn't sure which name to use here in `IsFunction` and friends; if you prefer that I change it, I am happy to do so.

wingo:
Thanks for the review @binji and @sbc100 :)  AFAICS this can be merged.  (Cheeky request: may I have merging privileges?  Very OK to say no.)

src/binary-writer.cc
sbc100: I guess it will make more sense if/when we add data symbol but for now it looks like you could avoid a lot of complexity and avoid InternSymbol being a template at all by just passing the SymbolType to InternSymbol.

It seems like having a separate class nested inside of `Symbol`, for each different type of symbol payload adds a fair amount of complexity, but perhaps that is justified and will make more sense once the followups land?

lgtm aside from this minor concern.

wingo: I see what you are saying, and if every symbol just had an `Index` payload it would be entirely clear-cut.  I guess since I saw the SymbolType::Data symbols having additional fields, I found it confusing for Symbol to have unused fields for e.g. SymbolType::Function symbols, which would propagate around as arguments to e.g. InternSymbol and related functions.  But that is a theoretical concern given that wabt doesn't know anything about data symbols, and my patches don't change that.

The followup #1537 doesn't do anything amazing here; this is just a question of data representation.  If you like I can change to a less structured `class Symbol { string name; uint8_t flags; SymbolType type; Index index; Offset offset; Address size; }` and assert that offset/size accessors only happen for data symbols (of which we will have none), or even exclude data symbols from the representation.  WDYT?

sbc100: I think I'm good with this approach.  What do you think binji?

sbc100 (approved):

src/binary-writer.cc
binji: I'm happy either way, tbh. I like having more explicit types (i.e. more structured) since it seems like it will be easier to expand  if/when we add new symbol values. OTOH, YAGNI and all :)

binji:
> may I have merging privileges?

Fine with me, I kinda thought you already could as long as there was a reviewer. Lemme check the settings.

PR #1539, opened 2020-09-09
Fix reloc section references when --enable-bulk-memory

### LINKED ISSUE #1538: Enabling bulk memory borks --relocatable
Consider `c.wat`:
```wat
(module
  (type (;0;) (func))
  (import "env" "b" (func (;0;) (type 0)))
  (func $a (type 0)
    call 0)
  (export "a" (func $a)))
```

If we `wat2wasm --relocatable` it and then `wasm-objdump -dx` it, we get:

```
...
Custom:
 - name: "reloc.Code"
   - R_WASM_FUNCTION_INDEX_LEB offset=0x000004(file=0x00002a) symbol=0 <env.b>
```

However if instead we do `wat2wasm --enable-bulk-memory`, we get

```
...
Custom:
 - name: "reloc.Code"
  - relocations for section: 5 (linking) [1]
   - R_WASM_FUNCTION_INDEX_LEB offset=0x000004(file=0x00004e) symbol=0 <env.b>
```

i.e. the relocs are targetting the wrong section.

wingo:
Fixes #1538.

### DIFF
--- a/src/binary-writer.cc
+++ b/src/binary-writer.cc
@@ -1214,6 +1214,16 @@ Result BinaryWriter::WriteModule() {
       stream_->MoveData(data_count_start_, data_count_end_, size);
     }
     stream_->Truncate(data_count_start_ + size);
+
+    --section_count_;
+
+    // We just effectively decremented the code section's index; adjust anything
+    // that might have captured it.
+    for (RelocSection& section : reloc_sections_) {
+      if (section.section_index == section_count_) {
+        --section.section_index;
+      }
+    }
   }
 
   if (module_->data_segments.size()) {
--- a/src/binary-writer.cc
+++ b/src/binary-writer.cc
@@ -171,7 +171,8 @@ class BinaryWriter {
   size_t last_subsection_payload_offset_ = 0;
 
   // Information about the data count section, so it can be removed if it is
-  // not needed.
+  // not needed, and relocs relative to the code section patched up.
+  size_t code_start_ = 0;
   size_t data_count_start_ = 0;
   size_t data_count_end_ = 0;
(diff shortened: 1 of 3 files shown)


src/binary-writer.cc
```diff
@@ -1214,6 +1214,16 @@ Result BinaryWriter::WriteModule() {
       stream_->MoveData(data_count_start_, data_count_end_, size);
     }
     stream_->Truncate(data_count_start_ + size);
+
+    --section_count_;
+
+    // We just effectively decremented the code section's index; adjust anything
+    // that might have captured it.
+    for (RelocSection& section : reloc_sections_) {
+      if (section.section_index == section_count_) {
```
binji, line 1223: I think this only works because the data count section comes immediately before the code section -- maybe we want to make that more explicit here? Or keep track of the data count section's index above and decrement any section_index that is after it.

binji (approved):
Good catch! Thanks for the fix.

lgtm

src/binary-writer.cc
wingo: You are correct regarding the limitation, but it is also as specified: see https://github.com/WebAssembly/bulk-memory-operations/blob/master/proposals/bulk-memory-operations/Overview.md#datacount-section.  The data count section can only appear between the element and code sections.

binji: Good point, I guess I was concerned that this might break in a non-obvious way if we add a new section between the two. Maybe there's an assertion that could added?

wingo: Sure, PTAL.  Thanks :)

binji: Great, thanks!

PR #1540, opened 2020-09-10
hint when section requires disabled feature

ddcc:


binji:
This is a good start, but I'd prefer a more holistic approach here, where all disabled features found in the module are tracked and communicated as a separate error, not integrated directly into the error message.

keithw:
Closing as stale (and superseded by #1595), but this would still be nice to have.

PR #1545, opened 2020-09-17
Fix up reloc-related tests after #1537

wingo:
PRs #1527 and #1539 needed their test expectations updated after #1537
was merged; this patch does that.  It also renames a test for
consistency.


binji (approved):
Thanks!

PR #1546, opened 2020-09-17
Disallow non-zero memidx when bulk memory disabled

chfast:
If bulk memory extension is disabled, do not allow memory index in the
data segment other than zero.


binji (approved):
thanks, lgtm!

chfast:
The spectest additions will follow soon.

PR #1547, opened 2020-09-18
Disallow "64-bit" flag if memory64 is disabled

chfast:
Fixes regressions in binary reader introduced by the implementation of
the memory64 extension: https://github.com/WebAssembly/wabt/pull/1500.


binji (approved):
lgtm, thanks!

aardappel:
This is not great actually, it be better for it to be enabled by the flag in the binary. Unlike new instructions, which you may have a reason for not wanting them to appear in the binary by accident, this feature is pretty unambiguous.

But I guess we'll have to live with this flag until it becomes default.

binji:
In general, I like that wabt doesn't include non-standardized features by default. Some tools will include all features (like `wasm-objdump`, but most probably shouldn't I think.

chfast:
Here is one more #1548.

PR #1548, opened 2020-09-18
Disallow non-zero table idx if reference types disabled

chfast:
If the reference types extension is disabled, do not allow table index
in the element segment other than zero.

Here I'm not 100% sure I identified the extension right.
And considering the complexity of the `ReadElemSection`, I decided to just check `flags !=0` at the beginning, so the error message is less precise than it could be.

### DIFF
--- a/test/binary/bad-elem-flags-no-reftype.txt
+++ b/test/binary/bad-elem-flags-no-reftype.txt
@@ -1,0 +1,19 @@
+;;; TOOL: run-gen-wasm-bad
+magic
+version
+section(TABLE) {
+  count[1]
+  anyfunc
+  has_max[0]
+  initial[0]
+}
+section(ELEM) {
+  count[1]
+  flags[1]
+  kind[0]
+  count[0]
+}
+(;; STDERR ;;;
+0000012: error: invalid elem segment flags: 0x1: reference types not allowed
+0000012: error: invalid elem segment flags: 0x1: reference types not allowed
+;;; STDERR ;;)
--- a/test/binary/bad-elem-flags.txt
+++ b/test/binary/bad-elem-flags.txt
@@ -1,4 +1,6 @@
 ;;; TOOL: run-gen-wasm-bad
+;;; ARGS1: --enable-reference-types
+;;; ARGS2: --enable-reference-types
 magic
 version
 section(TABLE) {
--- a/test/spec/bulk-memory-operations/binary.txt
+++ b/test/spec/bulk-memory-operations/binary.txt
@@ -1,6 +1,6 @@
 ;;; TOOL: run-interp-spec
 ;;; STDIN_FILE: third_party/testsuite/proposals/bulk-memory-operations/binary.wast
-;;; ARGS*: --enable-bulk-memory
+;;; ARGS*: --enable-bulk-memory --enable-reference-types
 (;; STDOUT ;;;
 out/test/spec/bulk-memory-operations/binary.wast:6: assert_malformed passed:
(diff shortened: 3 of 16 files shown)


test/binary/bad-elem-flags-no-reftype.txt
```diff
@@ -0,0 +1,19 @@
+;;; TOOL: run-gen-wasm-bad
```
chfast, line 1: I probably should start adding explicit `--disable-reference-types` so we don't have to change it after the extension gets enabled by default.

test/spec/bulk-memory-operations/binary.txt
```diff
@@ -89,15 +89,16 @@ out/test/spec/bulk-memory-operations/binary.wast:272: assert_malformed passed:
 out/test/spec/bulk-memory-operations/binary.wast:282: assert_malformed passed:
   000000e: error: unable to read i64 leb128: init_expr i64.const value
 out/test/spec/bulk-memory-operations/binary.wast:295: assert_malformed passed:
-  0000022: error: call_indirect reserved value must be 0
+  0000000: error: table variable out of range: 1 (max 1)
```
chfast, line 92: I don't know why the output changes. Still all the tests pass, but it is suspicious. Please double check, this is unknown territory for me.

test/binary/bad-elem-flags-no-reftype.txt
```diff
@@ -0,0 +1,19 @@
...
+}
+section(ELEM) {
+  count[1]
+  flags[1]
+  kind[0]
+  count[0]
+}
+(;; STDERR ;;;
+0000012: error: invalid elem segment flags: 0x1: reference types not allowed
+0000012: error: invalid elem segment flags: 0x1: reference types not allowed
```
sbc100, line 18: What does the error appear twice?

test/binary/bad-elem-flags.txt
```diff
@@ -1,4 +1,6 @@
 ;;; TOOL: run-gen-wasm-bad
+;;; ARGS1: --enable-reference-types
```
sbc100, line 2: Can you just use `ARGS`?

test/spec/bulk-memory-operations/binary.txt
```diff
@@ -1,6 +1,6 @@
 ;;; TOOL: run-interp-spec
 ;;; STDIN_FILE: third_party/testsuite/proposals/bulk-memory-operations/binary.wast
-;;; ARGS*: --enable-bulk-memory
```
sbc100, line 3: This seems maybe wrong.. unless bulk memory depends on reference types?   But if it did this flag would not be needed.

test/spec/bulk-memory-operations/bulk.txt
```diff
@@ -1,6 +1,6 @@
 ;;; TOOL: run-interp-spec
 ;;; STDIN_FILE: third_party/testsuite/proposals/bulk-memory-operations/bulk.wast
-;;; ARGS*: --enable-bulk-memory
```
sbc100, line 3: Ditto for all tests in the `spec/bulk-memory-operations` directory.

keithw:
Closing as stale. My understanding is that the current code with `--disable-reference-types` will correctly fail validation for an elem segment with a non-zero tableidx, just because the tableidx will be out of range. So I'm not sure this has to be a parse failure vs. a validation failure. Please reopen if you feel differently; I know there's a lot of interests in being able to enforce the 1.0/MVP spec.

PR #1549, opened 2020-09-22
Add --relocatable support for tables

wingo:
This patch adds a table symbol type and plumbs it through everywhere.

To do:

- [x] Add tests
- [x] Determine what the right thing is for modules that implicitly use table 0 but don't declare a table (see https://github.com/WebAssembly/wabt/compare/master...wingo:relocatable-tables?expand=1#diff-4803dd58b37da26b663dfaa7d34ba35e)
  - The answer here is a bit more straightforward: the initial patch wasn't emitting symtab entries for anonymous definitions.  The test case under consideration did in fact define the table.  The updated patch will emit relocs for all table number references, unless the reference is invalid/broken (the user must have passed --no-check).
- [x] Finalize the reloc kind for `R_WASM_TABLE_NUMBER_LEB` in [Linking.md](https://github.com/WebAssembly/tool-conventions/blob/master/Linking.md); @pmatos is on this.
- [x] Only emit relocs when reftypes are enabled (see https://github.com/WebAssembly/wabt/pull/1549#issuecomment-697480009)

Cc @pmatos and @sbc100

### DIFF
--- a/src/binary-reader-objdump.cc
+++ b/src/binary-reader-objdump.cc
@@ -58,6 +58,7 @@ class BinaryReaderObjdumpBase : public BinaryReaderNop {
   string_view GetEventName(Index index) const;
   string_view GetSymbolName(Index index) const;
   string_view GetSegmentName(Index index) const;
+  string_view GetTableName(Index index) const;
   void PrintRelocation(const Reloc& reloc, Offset offset) const;
   Offset GetSectionStart(BinarySection section_code) const {
     return section_starts_[static_cast<size_t>(section_code)];
@@ -152,6 +153,10 @@ string_view BinaryReaderObjdumpBase::GetSegmentName(Index index) const {
   return objdump_state_->segment_names.Get(index);
 }
 
+string_view BinaryReaderObjdumpBase::GetTableName(Index index) const {
+  return objdump_state_->table_names.Get(index);
+}
+
 string_view BinaryReaderObjdumpBase::GetSymbolName(Index symbol_index) const {
   if (symbol_index >= objdump_state_->symtab.size())
     return "<illegal_symbol_index>";
@@ -167,6 +172,8 @@ string_view BinaryReaderObjdumpBase::GetSymbolName(Index symbol_index) const {
       return GetSectionName(sym.index);
     case SymbolType::Event:
       return GetEventName(sym.index);
+    case SymbolType::Table:
+      return GetTableName(sym.index);
   }
   WABT_UNREACHABLE;
 }
@@ -283,6 +290,18 @@ class BinaryReaderObjdumpPrepass : public BinaryReaderObjdumpBase {
     return Result::Ok;
   }
 
+  Result OnTableSymbol(Index index,
+                       uint32_t flags,
+                       string_view name,
+                       Index table_index) override {
+    if (!name.empty()) {
+      SetTableName(table_index, name);
+    }
+    objdump_state_->symtab[index] = {SymbolType::Table, name.to_string(),
+                                     table_index};
+    return Result::Ok;
+  }
+
   Result OnImportFunc(Index import_index,
                       string_view module_name,
                       string_view field_name,
@@ -351,6 +370,7 @@ class BinaryReaderObjdumpPrepass : public BinaryReaderObjdumpBase {
   void SetFunctionName(Index index, string_view name);
   void SetGlobalName(Index index, string_view name);
   void SetEventName(Index index, string_view name);
+  void SetTableName(Index index, string_view name);
   void SetSegmentName(Index index, string_view name);
 };
 
@@ -365,6 +385,10 @@ void BinaryReaderObjdumpPrepass::SetGlobalName(Index index, string_view name) {
 
 void BinaryReaderObjdumpPrepass::SetEventName(Index index, string_view name) {
   objdump_state_->event_names.Set(index, name);
+}
+
+void BinaryReaderObjdumpPrepass::SetTableName(Index index, string_view name) {
+  objdump_state_->table_names.Set(index, name);
 }
 
 void BinaryReaderObjdumpPrepass::SetSegmentName(Index index, string_view name) {
@@ -891,6 +915,10 @@ class BinaryReaderObjdump : public BinaryReaderObjdumpBase {
                        uint32_t flags,
                        string_view name,
                        Index event_index) override;
+  Result OnTableSymbol(Index index,
+                       uint32_t flags,
+                       string_view name,
+                       Index table_index) override;
   Result OnSegmentInfoCount(Index count) override;
   Result OnSegmentInfo(Index index,
                        string_view name,
@@ -1781,8 +1809,20 @@ Result BinaryReaderObjdump::OnEventSymbol(Index index,
   if (name.empty()) {
     name = GetEventName(event_index);
   }
-  PrintDetails("   - [%d] E <" PRIstringview "> event=%" PRIindex, index,
+  PrintDetails("   - %d: E <" PRIstringview "> event=%" PRIindex, index,
                WABT_PRINTF_STRING_VIEW_ARG(name), event_index);
+  return PrintSymbolFlags(flags);
+}
+
+Result BinaryReaderObjdump::OnTableSymbol(Index index,
+                                          uint32_t flags,
+                                          string_view name,
+                                          Index table_index) {
+  if (name.empty()) {
+    name = GetTableName(table_index);
+  }
+  PrintDetails("   - %d: T <" PRIstringview "> table=%" PRIindex, index,
+               WABT_PRINTF_STRING_VIEW_ARG(name), table_index);
   return PrintSymbolFlags(flags);
 }
 
--- a/src/binary-reader-ir.cc
+++ b/src/binary-reader-ir.cc
@@ -260,6 +260,8 @@ class BinaryReaderIR : public BinaryReaderNop {
                           Index section_index) override;
   Result OnEventSymbol(Index index, uint32_t flags, string_view name,
                         Index event_index) override;
+  Result OnTableSymbol(Index index, uint32_t flags, string_view name,
+                       Index table_index) override;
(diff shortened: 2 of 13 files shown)


sbc100:
This generally looks great!

Personally I'd rather see these changes land in llvm / wasm-ld before we land them in wabt since as far as I know wasm-ld is the only consumer of these relocations right now.  Are you planning a corresponding llvm change?

Perhaps we can discuss this more on Monday with @pmatos.. but I was wondering if we should handle table 0 and memory 0 specially.   In llvm memory 0 is currently special in that it represents the linear memory of the program and table 0 is special in that it holds all the indirect function addresses.   I could imagine world where we could avoid relocations if the index is 0, avoiding the need for all call_indirect instructions to have an extra relocation, but I'm not set on this.

pmatos:
> This generally looks great!
> 
> Personally I'd rather see these changes land in llvm / wasm-ld before we land them in wabt since as far as I know wasm-ld is the only consumer of these relocations right now. Are you planning a corresponding llvm change?
> 

Yes - I plan to work on this as soon as I finish a couple of other small tasks.

> Perhaps we can discuss this more on Monday with @pmatos.. but I was wondering if we should handle table 0 and memory 0 specially. In llvm memory 0 is currently special in that it represents the linear memory of the program and table 0 is special in that it holds all the indirect function addresses. I could imagine world where we could avoid relocations if the index is 0, avoiding the need for all call_indirect instructions to have an extra relocation, but I'm not set on this.

Lets have a chat about this on Monday as planned so we can lay the ground for future work.

wingo:
> This generally looks great!

Thanks for taking a look!

> Personally I'd rather see these changes land in llvm / wasm-ld before we land them in wabt since as far as I know wasm-ld is the only consumer of these relocations right now. Are you planning a corresponding llvm change?

Agreed, and hopefully we can get that patch up this week too.

> I was wondering if we should handle table 0 and memory 0 specially. In llvm memory 0 is currently special in that it represents the linear memory of the program and table 0 is special in that it holds all the indirect function addresses. I could imagine world where we could avoid relocations if the index is 0, avoiding the need for all call_indirect instructions to have an extra relocation, but I'm not set on this.

I was wondering about this too.  Concretely, looking at the compilation of:

```c
void f(void) {}
void* g(void) { return f; }
```

I see that the current result objdumps as:

```
Type[2]:
 - type[0] () -> nil
 - type[1] () -> i32
Import[2]:
 - memory[0] pages: initial=0 <- env.__linear_memory
 - table[0] type=funcref initial=1 <- env.__indirect_function_table
Function[2]:
 - func[0] sig=0 <f>
 - func[1] sig=1 <g>
Elem[1]:
 - segment[0] flags=0 table=0 count=1 - init i32=1
  - elem[1] = func[0] <f>
Code[2]:
 - func[0] size=2 <f>
 - func[1] size=8 <g>
Custom:
 - name: "linking"
  - symbol table [count=2]
   - 0: F <f> func=0 binding=global vis=hidden
   - 1: F <g> func=1 binding=global vis=hidden
Custom:
 - name: "reloc.CODE"
   - R_WASM_TABLE_INDEX_SLEB offset=0x000007(file=0x000079) symbol=0 <f>
Custom:
 - name: "producers"

Code Disassembly:

000074 func[0] <f>:
 000075: 0b                         | end
000077 func[1] <g>:
 000078: 41 81 80 80 80 00          | i32.const 1
 00007e: 0b                         | end
```

If we want to omit relocs for the indirect function table, we can do so if the referred-to table has index 0, is imported, has type `funcref`, and the import is from the `env` module, and the name of the import is `__indirect_function_table`.

The advantage of doing so is that you would have fewer relocs to deal with, which probably don't need to be renumbered from 0 in the "normal" case, and for which you wouldn't have to reserve a 5-byte LEB.

The disadvantage is that the filter logic described above is a bit of a hack.  Also I would imagine that in the future as CFI techniques start to be more standard, and with the advent of typed function references, we'll start to see more partitioning of callables into multiple tables, so this optimization is not one for the long term, IMO.

For me, I think I would lean towards adding `R_WASM_TABLE_NUMBER_LEB` relocs for all table number references.  If short-term efficiency is still a consideration, we can only emit such relocs when reftypes are enabled, in wabt anyway; not sure how feature selection works in LLVM yet.

sbc100:
> > This generally looks great!
> 
> Thanks for taking a look!
> 
> > Personally I'd rather see these changes land in llvm / wasm-ld before we land them in wabt since as far as I know wasm-ld is the only consumer of these relocations right now. Are you planning a corresponding llvm change?
> 
> Agreed, and hopefully we can get that patch up this week too.
> 
> > I was wondering if we should handle table 0 and memory 0 specially. In llvm memory 0 is currently special in that it represents the linear memory of the program and table 0 is special in that it holds all the indirect function addresses. I could imagine world where we could avoid relocations if the index is 0, avoiding the need for all call_indirect instructions to have an extra relocation, but I'm not set on this.
> 
> I was wondering about this too. Concretely, looking at the compilation of:
> 
> ```c
> void f(void) {}
> void* g(void) { return f; }
> ```
> 
> I see that the current result objdumps as:
> 
> ```
> Type[2]:
>  - type[0] () -> nil
>  - type[1] () -> i32
> Import[2]:
>  - memory[0] pages: initial=0 <- env.__linear_memory
>  - table[0] type=funcref initial=1 <- env.__indirect_function_table
> Function[2]:
>  - func[0] sig=0 <f>
>  - func[1] sig=1 <g>
> Elem[1]:
>  - segment[0] flags=0 table=0 count=1 - init i32=1
>   - elem[1] = func[0] <f>
> Code[2]:
>  - func[0] size=2 <f>
>  - func[1] size=8 <g>
> Custom:
>  - name: "linking"
>   - symbol table [count=2]
>    - 0: F <f> func=0 binding=global vis=hidden
>    - 1: F <g> func=1 binding=global vis=hidden
> Custom:
>  - name: "reloc.CODE"
>   - relocations for section: 4 (Code) [1]
>    - R_WASM_TABLE_INDEX_SLEB offset=0x000007(file=0x000079) symbol=0 <f>
> Custom:
>  - name: "producers"
> 
> Code Disassembly:
> 
> 000074 func[0] <f>:
>  000075: 0b                         | end
> 000077 func[1] <g>:
>  000078: 41 81 80 80 80 00          | i32.const 1
>  00007e: 0b                         | end
> ```
> 
> If we want to omit relocs for the indirect function table, we can do so if the referred-to table has index 0, is imported, has type `funcref`, and the import is from the `env` module, and the name of the import is `__indirect_function_table`.
> 
> The advantage of doing so is that you would have fewer relocs to deal with, which probably don't need to be renumbered from 0 in the "normal" case, and for which you wouldn't have to reserve a 5-byte LEB.
> 
> The disadvantage is that the filter logic described above is a bit of a hack. Also I would imagine that in the future as CFI techniques start to be more standard, and with the advent of typed function references, we'll start to see more partitioning of callables into multiple tables, so this optimization is not one for the long term, IMO.
> 
> For me, I think I would lean towards adding `R_WASM_TABLE_NUMBER_LEB` relocs for all table number references. If short-term efficiency is still a consideration, we can only emit such relocs when reftypes are enabled, in wabt anyway; not sure how feature selection works in LLVM yet.

I like this last idea.  I think I agree that for consistency we should probably use relocations everywhere and that enabling this only when reftypes are enabled is a good way to transition to this new world.  

Also we look at adding `R_WASM_MEMORY_NUMBER_LEB` at the same time support multi-memory, or maybe this can wait until we have more concrete users of the multi memory proposal?

wingo:
Thanks for the review!

Regarding multi-memory, it's enough boilerplate that it probably warrants a separate PR.  I filed #1552.

I updated the TODO at the top and added an entry for only emitting table relocs if reftypes are enabled.  I also added a basic test.

pmatos:
>     * [ ]  Finalize the reloc kind for `R_WASM_TABLE_NUMBER_LEB` in [Linking.md](https://github.com/WebAssembly/tool-conventions/blob/master/Linking.md); @pmatos is on this.
> 

See https://github.com/WebAssembly/tool-conventions/pull/147

binji (approved):
lgtm

wingo:
Ready to merge, I think.  I know that wasm-ld work is ongoing but there is also a bug-fix in this patch; the logic at https://github.com/WebAssembly/wabt/pull/1549/files#diff-f69899e3f8be9636bcd8d71e903ea8a0R238 wasn't right before.  (All function references need relocs, so all functions should be in the symbol table, regardless of whether they have a name or not.)  If wabt needs followup fixes for the table_number reloc, I will make sure to take care of it.

src/binary-reader-objdump.cc
```diff
@@ -1781,11 +1809,23 @@ Result BinaryReaderObjdump::OnEventSymbol(Index index,
   if (name.empty()) {
     name = GetEventName(event_index);
   }
-  PrintDetails("   - [%d] E <" PRIstringview "> event=%" PRIindex, index,
+  PrintDetails("   - %d: E <" PRIstringview "> event=%" PRIindex, index,
```
sbc100, line 1812: Looks like we don't have any test that tickles this?  (Since I don't see any corresponding test change due to this?)

sbc100:
Oh .. we like to do a single commit per PR in this repo.   It looks like you have diligently split you your into commits with nice separate messages.

Would you mind creating a single message that you would like to use for the final commit message?  You could edit the PR description to include this and I can copy and paste it into the squashed commit?

src/binary-reader-objdump.cc
wingo: Indeed -- just noticed this as I had copied the table implementation from events, and events were printing inconsistently with regards to everything else, so I just fixed this one up.  Indeed, apparently no tests.

wingo:
@sbc100 I squashed, rebased, and updated the commit message, though I see now that you were just asking for an updated message and that you would do the squashing :grimacing: sorry about that.  In any case it's the same code, rebased cleanly onto today's master -- no manual edits.

PR #1554, opened 2020-09-30
Add initial support for extended names sections

sbc100:
See: https://github.com/WebAssembly/extended-name-section

Although this is only a phase 1 proposal its seems pretty
straight forward, and is already being implemented in
binaryen.

### DIFF
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -1737,8 +1737,8 @@ Result BinaryReader::ReadNameSection(Offset section_size) {
         }
         break;
       case NameSectionSubsection::Label:
-        // TODO(sbc): Implemnt label names. These are slightly more complicated
-        // since they refere to offers in the code section / instruction stream.
+        // TODO(sbc): Implement label names. These are slightly more complicated
+        // since they refer to offsets in the code section / instruction stream.
         state_.offset = subsection_end;
         break;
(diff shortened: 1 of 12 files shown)


src/binary-reader.cc
```diff
@@ -1731,6 +1736,32 @@ Result BinaryReader::ReadNameSection(Offset section_size) {
           }
         }
         break;
+      case NameSectionSubsection::Label:
+        // TODO(sbc): Implemnt label names. These are slightly more complicated
```
aardappel, line 1740: `Implemnt`, `offers`

src/binary-writer.cc
```diff
@@ -1514,6 +1516,155 @@ Result BinaryWriter::WriteModule() {
       }
     }
     EndSubsection();
+
+    size_t named_types = 0;
```
aardappel, line 1520: the code blocks below all look very similar.. is it practical to pull this into a function?

aardappel (approved):

src/binary-writer.cc
sbc100: Ahh yes... I guess I'll have to you one of those new-fangled template things I've heard about.

sbc100: Done!

PR #1559, opened 2020-10-14
Fix compile failed on Ubuntu 18.04 aarch64 cmake3.10.2

bxq2011hust:
I try to compile wabt on Ubuntu 18.04 aarch64 with gcc 7.5.0 and cmake 3.10.2, failed with the error below:
image

the wabt source code is downloaded from release page v1.0.19 tar package which without the .git dir.

I made a tiny modify to fix this


binji (approved):
Thanks! lgtm

Issue #1375, opened 2020-03-25
Fix Various TODOs


binji:
There are other TODOs in the source, but most of these should be small. 

- [ ] Update output from "type" to "func_type" [TODO](https://github.com/WebAssembly/wabt/blob/master/src/binary-reader-logging.cc#L146) [TODO](https://github.com/WebAssembly/wabt/blob/master/src/binary-writer.cc#L888)
- [ ] Rename LoadStoreExpr (maybe MemoryAccessExpr?) [TODO](https://github.com/WebAssembly/wabt/blob/master/src/ir.h#L528) [TODO](https://github.com/WebAssembly/wabt/blob/master/src/binary-writer.cc#L380)
- [ ] Use proper location in binary-reader-interp.cc [TODO](https://github.com/WebAssembly/wabt/blob/master/src/interp/binary-reader-interp.cc#L299)
- [ ] Implement event matching in interpreter [TODO](https://github.com/WebAssembly/wabt/blob/master/src/interp/interp.h#L167)
- [ ] Clean up "out of bounds" memory/table access messages [TODO](https://github.com/WebAssembly/wabt/blob/master/src/interp/interp.h#L1835)
- [ ] Use `u16` for SerializedOpcode in interpreter [TODO](https://github.com/WebAssembly/wabt/blob/master/src/interp/istream.h#L89)
- [ ] Move TypeChecker into SharedValidator [TODO](https://github.com/WebAssembly/wabt/blob/master/src/shared-validator.h#L48) [TODO](https://github.com/WebAssembly/wabt/blob/master/src/shared-validator.h#L279)
- [ ] Change `assert_unlinkable` to a single-line error [TODO](https://github.com/WebAssembly/wabt/blob/master/src/tools/spectest-interp.cc#L1277)
- [ ] Print `assert_uninstantiable` and `assert_exhaustion` errors when assertion passes [TODO](https://github.com/WebAssembly/wabt/blob/master/src/tools/spectest-interp.cc#L1318) [TODO](https://github.com/WebAssembly/wabt/blob/master/src/tools/spectest-interp.cc#L1474)
[...]

PR #1570, opened 2020-11-14
Update output from type to func type in binary writer and reader

radu-matei:
This commit addresses the first item of #1375 - specifically, it changes the output from `type` to `func type` in the binary writer, and updates the tests.

There are also changes in the third party test suite - what is the procedure for updating the upstream test suite repository? Should there be a PR to its repo first with these changes?

```
$ cd third_party/testsuite && git status
HEAD detached at 18f8340
        modified:   binary.wast
        modified:   proposals/bulk-memory-operations/binary.wast
        modified:   proposals/exception-handling/binary.wast
        modified:   proposals/function-references/binary.wast
        modified:   proposals/reference-types/binary.wast
        modified:   proposals/simd/simd_const.wast
```

### DIFF
--- a/src/binary-reader-logging.cc
+++ b/src/binary-reader-logging.cc
@@ -144,8 +144,7 @@ Result BinaryReaderLogging::OnFuncType(Index index,
                                        Type* param_types,
                                        Index result_count,
                                        Type* result_types) {
-  // TODO: switch to "OnFuncType"?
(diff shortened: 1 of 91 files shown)


binji (approved):
Thanks, lgtm! I don't think we need to update the testsuite now; it's OK to lag behind a bit. Either way, this change shouldn't affect it.

Issue #1571, opened 2020-11-19
How to build wabt in release mode on Linux


Praveer1981:
Hi,
I want to build all the tools in the release mode on Linux. I followed the instructions mentioned in the readme.md . but I am not able to get release version of binaries.
can someone help me

binji:
You need to add `-DCMAKE_BUILD_TYPE=Release` to your cmake build configuration (see https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html) for more info. If you're using the top-level `Makefile`, then you can build a target like `clang-release` instead.

Praveer1981:
The command I used to generate the binaries:

```
$ cmake .. -DCMAKE_BUILD_TYPE=Release
$ cmake --build .
```
It generates the wasm-objdump but when I run below command It seems to be debug binary

`$file wasm-objdump`
O/P:

`wasm-objdump: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, BuildID[sha1]=a7194a75b0079b4278a2cb380e140876071b8642, **with debug_info**, not stripped
`
To be more specific, I want release version of wasm-objdump.

is there anything I am missing here ? @binji

Praveer1981:
I built in verbose mode to see the exact commands executed:
Below are the command I ran:

```
$ cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON
$ cmake --build .
```
O/P:
```
Scanning dependencies of target wasm-objdump
make[2]: Leaving directory '/home/kumarpr/Praveer_wabt/wabt/build'
/usr/bin/make -f CMakeFiles/wasm-objdump.dir/build.make CMakeFiles/wasm-objdump.dir/build
make[2]: Entering directory '/home/kumarpr/Praveer_wabt/wabt/build'
[ 72%] Building CXX object CMakeFiles/wasm-objdump.dir/src/tools/wasm-objdump.cc.o
[...]

PR #1576, opened 2020-11-26
Let cmake add "-g" flag when it is needed

Praveer1981:
Reference: [Issue#1571](https://github.com/WebAssembly/wabt/issues/1571)
Problem: 
It used to add the debug info in the binary even if below command was used.

```
$ cmake .. -DCMAKE_BUILD_TYPE=Release
$ cmake --build .

```
Solution :
Let the cmake add `-g` flag when it is needed.

Test 1: run below command 
```
$ cmake .. -DCMAKE_BUILD_TYPE=Release
$ cmake --build .

```
It will generate binary without debug info

Test 2: 
```
$ cmake .. -DCMAKE_BUILD_TYPE=Debug
$ cmake --build .
```
It will generate binary with debug info

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -139,8 +139,7 @@ else ()
   #   interfaces, etc.
   # disable -Wpointer-arith: this is a GCC extension, and doesn't work in MSVC.
   add_definitions(
-    -Wall -Wextra -Wno-unused-parameter -Wpointer-arith -g
-    -Wuninitialized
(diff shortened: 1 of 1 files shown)


sbc100:
Certainly seems reasonable to me.  @binji do you remember why the explicit `-g` was added here in the first place?

sbc100:
It looks like the `-g` goes right back to the initial commit so I guess its a historical artifact: c91538cd3

sbc100 (approved):

PR #1590, opened 2020-12-14
Avoid warning when building with -Wsign-compare

steven-johnson:
Just a trivial warning fix.


binji:
Looks like this was already fixed here: https://github.com/WebAssembly/wabt/pull/1580.

steven-johnson:
(ah, sorry, I was building against 1.0.20 and neglected to check master!)

PR #1592, opened 2020-12-17
Add script to generate html from man, and generate wasm-decompile html

ngzhian:
The script regenerates html from all man pages. I ran it once, and only
checked in wasm-decompile.1.html.


aardappel:
Awesome, thanks for fixing :)

aardappel (approved):

Issue #1197, opened 2019-10-24
error: invalid section code: 12


godmar:
Hi, I'm new to WASM.

I've installed emscripten using the [installation instructions](https://emscripten.org/docs/getting_started/downloads.html) and tried it out on a simple program:
```
#include <cstdlib>
#include <cstring>

char *
malloc_demo()
{
    char *p = (char *)malloc(20);
    strcpy(p, "Hello, World\n");
    return p;
}
```
this is with
```
$ emcc -v
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 1.39.0
clang version 10.0.0 (/b/s/w/ir/cache/git/chromium.googlesource.com-external-github.com-llvm-llvm--project e44524736c4a97ae4fb37193e58647f838f6d36a)
```
when I run wabt on it (the current build), I see:
```
$ wasm2wat malloc.wasm 
000009d: error: invalid section code: 12
```
I googled the [specification](https://webassembly.github.io/spec/core/bikeshed/index.html#sections%E2%91%A0) - indeed, section numbers don't extend past 11.

However, I also found an issue in the [LLVM Phabricator](https://reviews.llvm.org/D60623) that introduced a `DataCount` section, which tools such as binaryen [apparently support](https://github.com/WebAssembly/binaryen/blob/master/src/wasm/wasm-binary.cpp#L859).

The Phabricator issue notes that:
_Engines that don't support bulk memory operations will not recognize the DataCount section and will report validation errors, but that's ok because object files aren't supposed to be run directly anyway._

I suppose my confusion is why I tapped into something that's apparently in development. Does wabt tracked 'tagged releases' of emscripten?  Should wabt be able to convert files containing such "linking sections" to the .wat format, and this particular feature just has not been implemented?  Or am I mistaken in the idea to be able to convert files written by emscripten using wabt?

Thank you.

godmar:
PS: running `wasm2wat -v`, I found:
```
$ wasm2wat --enable-bulk-memory malloc.wasm -o malloc.wat
```
or
```
$ wasm2wat --enable-all malloc.wasm -o malloc.wat
```
which make it parse this file.

I suppose this makes my issue a usability suggestion.  How about instead of outputting `invalid section code`, output `invalid section code - did you forget to enable a feature you want`?

binji:
Agreed, that would be a nice fix. It would be even nicer to look for these known but disabled features and mentions the flag explicitly.

PR #1595, opened 2021-01-04
Report missing section features in wasm2wat (#1197)

rusty122:
Ran into https://github.com/WebAssembly/wabt/issues/1197 when using `wasm2wat` so I took a stab at adding hints for missing features like `--enable-bulk-memory`. 

Previously, it seems like looping through WASM sections would return early for these errors (ignoring [`stop_on_first_error`](https://github.com/WebAssembly/wabt/blob/0b80b37c5f99102c5b1c669191ff7b48436875d9/src/binary-reader.cc#L2531)) since the `ERROR_UNLESS` macro will cause `ReadSections` to return Error for missing section features. 

With this PR, we accumulate a list of the missing features in `missing_section_features` and print them out with a helpful error message when returning from `ReadSections` - this makes it easy to extend in the future. Took some inspiration from [the thread on this un-merged PR](https://github.com/WebAssembly/wabt/pull/1540#issuecomment-690535601) tackling the same issue. We shouldn't run into any cases of reporting the same feature missing multiple times because there's already [a guard for multiple section definitions](https://github.com/WebAssembly/wabt/blob/0b80b37c5f99102c5b1c669191ff7b48436875d9/src/binary-reader.cc#L2508-L2514). 

See the updated test cases for what this looks like - happy to update the messaging if there are any suggestions

### DIFF
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -85,6 +85,7 @@ class BinaryReader {
   };
 
   void WABT_PRINTF_FORMAT(2, 3) PrintError(const char* format, ...);
+  void ReportMissingSectionFeatures(std::vector<std::string> missing_features);
   Result ReadOpcode(Opcode* out_value, const char* desc) WABT_WARN_UNUSED;
   template <typename T>
   Result ReadT(T* out_value,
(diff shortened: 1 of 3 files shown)


binji:
Thanks for looking into this, and sorry for the very long review delay. Like the previous PR, I think I'd prefer something a bit more comprehensive to handle missing features (i.e. collecting all `_enabled()` checks that lead to errors in a separate `Features` struct, then display those to the user before exiting).

That said, it seems like this error hits a lot of new users, so maybe a simpler solution for the common case would be OK. In that case, I think integrating the flag directly into the error (as was done in the previous PR) would be best.

rusty122:
> Like the previous PR, I think I'd prefer something a bit more comprehensive to handle missing features (i.e. collecting all `_enabled()` checks that lead to errors in a separate `Features` struct, then display those to the user before exiting).

Oh that's sort of what I was going for - collecting the necessary but not enabled features in a vector and reporting them at the end of parsing - this should catch both `--enable-bulk-memory` and `--enable-exceptions` and report them together before ending the parsing stage. Are there other features that I missed? I assumed this we could catch all the missing features in `binary-reader.cc` but maybe there needs to be a more extensive solution if missing features get caught in other parts of the codebase

binji:
> I assumed this we could catch all the missing features in binary-reader.cc but maybe there needs to be a more extensive solution if missing features get caught in other parts of the codebase

Yeah, features are checked in a lot of places. `binary-reader.cc` checks features in quite a few other places (search for `_enabled()` in that file). Then features are also checked when reading the text format too.

keithw:
Closing as stale, but this would still be very nice to have in some form. The error messages when a feature is not enabled are still unhelpful.

PR #1602, opened 2021-02-05
Create a static wasm C API instead of shared

basvandijk:
This enables building the `wabt-sys` Rust crate statically using musl.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -367,7 +367,11 @@ endif ()
 
 # libwasm, which implenents the wasm C API
 if (BUILD_LIBWASM)
-  add_library(wasm ${WABT_LIBRARY_SRC} src/interp/interp-wasm-c-api.cc)
+  if (WIN32)
+    add_library(wasm SHARED ${WABT_LIBRARY_SRC} src/interp/interp-wasm-c-api.cc)
+  else ()
+    add_library(wasm STATIC ${WABT_LIBRARY_SRC} src/interp/interp-wasm-c-api.cc)
+  endif ()
   target_link_libraries(wasm wabt)
   target_include_directories(wasm PUBLIC third_party/wasm-c-api/include)
   if (COMPILER_IS_MSVC)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -367,7 +367,7 @@ endif ()
 
 # libwasm, which implenents the wasm C API
 if (BUILD_LIBWASM)
-  add_library(wasm SHARED ${WABT_LIBRARY_SRC} src/interp/interp-wasm-c-api.cc)
+  add_library(wasm STATIC ${WABT_LIBRARY_SRC} src/interp/interp-wasm-c-api.cc)
   target_link_libraries(wasm wabt)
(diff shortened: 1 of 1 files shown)


binji (approved):
lgtm, @sbc100 wdyt?

CMakeLists.txt
```diff
@@ -367,7 +367,11 @@ endif ()
 
 # libwasm, which implenents the wasm C API
 if (BUILD_LIBWASM)
-  add_library(wasm SHARED ${WABT_LIBRARY_SRC} src/interp/interp-wasm-c-api.cc)
+  if (WIN32)
+    add_library(wasm SHARED ${WABT_LIBRARY_SRC} src/interp/interp-wasm-c-api.cc)
+  else ()
+    add_library(wasm STATIC ${WABT_LIBRARY_SRC} src/interp/interp-wasm-c-api.cc)
```
sbc100, line 373: I'm not sure how cmake normally handles this but I would have thought this would be some kind of build option.   Most libraries build as both dynamic and static right?    We should try to follow cmake best practices whatever they are.

CMakeLists.txt
```diff
@@ -367,7 +367,11 @@ endif ()
...
+  endif ()
```
basvandijk, line 374: To quote the [cmake docs](https://cmake.org/cmake/help/latest/command/add_library.html):
>  If no type is given explicitly the type is STATIC or SHARED based on whether the current value of the variable [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS) is ON

So maybe we should just remove the `SHARED` type and make sure `BUILD_SHARED_LIBS` is correctly set to `false` when building statically:

```suggestion
  add_library(wasm ${WABT_LIBRARY_SRC} src/interp/interp-wasm-c-api.cc)
```

@sbc100 WDYT?

sbc100: sgtm

basvandijk: When setting `add_library(wasm ${WABT_LIBRARY_SRC} src/interp/interp-wasm-c-api.cc)` the build crashes on Windows. To fix that we need to:

```
if (WIN32)
  set(BUILD_SHARED_LIBS ON)
endif ()
```

However I think that's worse from my previous solution since the previous solution only made a local change while this makes a global change which could have adverse effects. The previous solution is also more consistent with all other calls of `add_library` since they all specify the SHARED/STATIC type.

So I'm going back to my original solution. @sbc100 @binji WDYT? It would be great to get this merged.

sbc100: Why does the static build not work on windows?

basvandijk: The CI reported a bunch of symbol not found errors IIRC. Maybe the logs still exists but I don't know you folks CI well enough to locate them.

keithw:
Hi folks -- this has been dormant for almost two years. Is it ok to close or can we get it in mergeable state?

keithw:
> Hi folks -- this has been dormant for almost two years. Is it ok to close or can we get it in mergeable state?

Closing as nothing heard. Happy to revisit if this would be helpful.

PR #1621, opened 2021-03-03
Add and update final SIMD opcodes

ngzhian:
This adds all the new SIMD opcodes and also update names of certain
instructions enough to get compiling again. Partially addresses #1617.

Note that third_party/testsuite is updated to TOT, and I ran `tests/update-spec-tests.py`, but manually removed some simd test cases because we cannot handle all the new opcodes yet, they will require implementation in the binary-reader and interp. Support will come later, as with checking in the new SIMD tests.

PR #1622, opened 2021-03-04
Bump testsuite version and fix tests

ngzhian:
This is an incremental bump of testsuite, the TOT has SIMD tests which won't work yet.

The main fix is for a test added to the spec repo that has an invalid
functype in the type section. Based on the spec, the functype should be
a single byte 0x60. The leb encoding only comes into effect with the GC
proposal.

### DIFF
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -2089,7 +2089,15 @@ Result BinaryReader::ReadTypeSection(Offset section_size) {
 
   for (Index i = 0; i < num_signatures; ++i) {
     Type form;
-    CHECK_RESULT(ReadType(&form, "type form"));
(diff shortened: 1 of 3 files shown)


sbc100:
Is this supposed to be an incremental bump rather than a bump to ToT?  I see one more revision there form last month.   Maybe worth mentioning in the description if so.

sbc100 (approved):

ngzhian:
> Is this supposed to be an incremental bump rather than a bump to ToT? I see one more revision there form last month. Maybe worth mentioning in the description if so.

Yup, the TOT has SIMD fixes that are not compatible. Those will be addressed in #1621 .

Issue #1617, opened 2021-03-01
Support for final wasm-simd128 specin the wabt interpreter?


steven-johnson:
Is there work underway to bring the WABT interpreter in line with the final proposal for the wasm-simd128 spec? If so, is there any rough ETA? If not, is this work scheduled to be done? (Or has it already been done and I somehow missed it?)

Halide currently uses the WABT interpreter for a fair number of our correctness tests (where performance is unimportant), and simd support is pretty crucial for most Halide users.

tlively:
No, I don't know of any work underway for this right now. I can take a crack at it this week, though.

steven-johnson:
Excellent. Let me know if I can assist.

tlively:
If you're interested, PRs would be super welcome! If not, no problem :)

steven-johnson:
> If you're interested, PRs would be super welcome! 

Not sure where to begin (ie, code structure, testing requirements, etc) & am juggling several things at the moment, but if you can open a PR doing one of the ops (as a useful reference/example) I will try to use that as a basis to help fill in the gaps.

ngzhian:
@steven-johnson I can work on the opcode updates, don't worry for now, I should get something up today. :)

repi:
Noticed that `wasm-strip` fails to strip modules with SIMD instructions in it, is this the same issue or should we file a separate issue for that?

ngzhian:
> Noticed that `wasm-strip` fails to strip modules with SIMD instructions in it, is this the same issue or should we file a separate issue for that?

what's the error message? it's likely a new issue, the opcodes have already been updated.

repi:
filed separate issue #1803

keithw:
Done in 1ba6f750a50c9b033a2309f979edeb0b4bcbf1ef

PR #1623, opened 2021-03-05
Bump third_party/testsuite to TOT

ngzhian:
This bumps third_party/testsuite to the latest version.

We fix 1 wasm2c test in 6b15477, and do a bunch of rebasing in cf80765.

We also skip some simd tests that use unimplemented instructions (c66f87e), or instructions with updated names (72a4400).

As we continue to update the SIMD support, more and more tests will be unskipped.

(Part 1 of #1621, #1617.)


sbc100 (approved):
Nice!

PR #1624, opened 2021-03-05
[simd] Rename any_true, implement i64x2 bitmask and all_true

ngzhian:
After some rebasing of tests, we can unskip simd_boolean.txt, simd_i16x8_arith2.txt, and simd_lane.txt.


sbc100 (approved):
rubberstamp lgtm

Issue #1627, opened 2021-03-05
WITH_WASI=ON is broken


steven-johnson:
At recent top-of-tree, building with `-DWITH_WASI=ON` fails, with many errors of the form

```
_deps/wabt-src/src/interp/interp-wasi.cc:255:35: error: 'i32_' is a private member of 'wabt::interp::Value'
```

Is WITH_WASI (still) supported?

PR #1628, opened 2021-03-05
Fix compile errors for WITH_WASI=ON (#1627)

steven-johnson:


sbc100:
I can't remember why we don't have CI for this... but it seems we should.

sbc100 (approved):

steven-johnson:
> I can't remember why we don't have CI for this... but it seems we should.

I'll leave that part for someone on the team to add :-)

Issue #1629, opened 2021-03-06
Compilation is broken for 32-bit targets


steven-johnson:
Various compilation errors prevent compilation on 32-bit targets (e.g., x86-32). Sure would be handy to be able to build and use this tool on those systems.

steven-johnson:
To be a little more clear: most of the failures are under WITH_WASI=ON, which has (eg) structs that aren't the sizes that are expected, e.g. `static_assert(sizeof(__wasi_fdstat_t) == 24, "witx calculated size");` and many other static_asserts. I'm not quite sure of the proper fix here, since at least some of this code seems to be autogenerated, based on the comments at the top of the file.

(There is also a rogue warning/failure in BinaryReader::ReadAddress in which a format string uses `%lu` for a `size_t` value, which of course isn't quite right for 32-bit)

sbc100:
Oh I assumed you had fixed this with #1628,  sounds like there are a separate set of issues here?

sbc100:
Strange.. I would have expected the disparity to be on 64-bit hosts because was itself, like wasm, has 32-bit pointers (its ILP32).

steven-johnson:
> Oh I assumed you had fixed this with #1628, sounds like there are a separate set of issues here?

Correct (for #1628 I had not tried building on a 32-bit target)

> Strange.. I would have expected the disparity to be on 64-bit hosts because was itself, like wasm, has 32-bit pointers (its ILP32).

Yeah, I can't explain it either, but I didn't spend a lot of time looking into it after seeing that the code was autogenerated-then-copy-pasted.

steven-johnson:
From a casual inspection, it looks like the issue is just one of struct padding, eg `__wasi_fdstat_t` is basically `{ u8; u16; u64; u64; }`, which requires less padding on x86-32. I'm just not sure of the proper fix: should we edit these in place? Or should we regenerate the code from whatever generated it?

steven-johnson:
Yeah, this is puzzling; whatever tool generates this code makes some pretty specific assumptions about struct alignment and padding (apparently, that they always follow rules for x86-86, which seems odd)

sbc100:
I think the layout is determined by that the wasm32 target produces (i.e. its a header mostly designed for code that is running in WebAssembly not the implementation side).   At one point I looked into having the generator generate the host-side code for packing / unpacking these structs, but IIRC I never finished it because I got something that worked in the ways that I needed it to.

[...]

PR #1630, opened 2021-03-06
Fix format-string warning for 32-bit compilation (#1629)

steven-johnson:


src/binary-reader.cc
```diff
@@ -629,7 +629,7 @@ Result BinaryReader::ReadAddress(Address* out_value,
                                  const char* desc) {
   ERROR_UNLESS(memory < memories.size(),
                "load/store memory %u out of range %lu", memory,
-               memories.size());
+               static_cast<unsigned long>(memories.size()));
```
sbc100, line 632: How about changing the format string to "%zu"?

steven-johnson: Is that 100% portable? I thought it wasn't supported in MSVC.

sbc100: We have at least one other occurrence already so I guess its portable enough:

```
$ git grep "%zu"
src/wat-writer.cc:  WABT_TRACE_ARGS(FlushExprTreeVector, "%zu", expr_trees.size());
```

If it proves non-portable then the correct fix is probably to add a PRI macro like we do for `PRIaddress`

sbc100: According to this discussion MS added support for %z in VS2013: https://stackoverflow.com/questions/15610053/correct-printf-format-specifier-for-size-t-zu-or-iu

steven-johnson: Done.

sbc100 (approved):

PR #1632, opened 2021-03-10
Take care of opcodes whose bytes are non-canonical leb128

### LINKED ISSUE #1631: wasm-objdump cannot handle prefixed instructions with non-canonical leb128 opcodes
Uploaded a test case:

[wasm-objdump-fails.txt](https://github.com/WebAssembly/wabt/files/6110933/wasm-objdump-fails.txt)

```
$ wasm-objdump -d wasm-objdump-fails.txt

clusterfuzz-testcase-minimized-v8_wasm_async_fuzzer-5664681130524672:   file format wasm 0x1

Code Disassembly:

00002d func[0] <main>:
 00002e: 03 7e                      | local[0..2] type=i64
 000030: 20 00                      | local.get 0
 000032: 41 d6 49                   | i32.const 4294960342
 000035: 02 7f                      | block i32
 000037: 41 fe 20                   |   i32.const 4222
 00003a: 0c 00                      |   br 0
 00003c: 68                         |   i32.ctz
 00003d: fc 0e 00 00                |   table.copy 0 0
 000041: 0b                         | end
 000042: 20 00                      | local.get 0
 000044: 68                         | i32.ctz
 000045: fc 10 00                   | table.size 0
 000048: 68                         | i32.ctz
warning: 0x4d missing opcode callback at 0x4a (0xfc=<invalid>)
```

wasm2wat works

```
$ wasm2wat wasm-objdump-fails.txt --enable-all
(module
  (type (;0;) (func (result i32)))
  (type (;1;) (func (param i32) (result i32)))
  (func (;0;) (type 1) (param i32) (result i32)
    (local i64 i64 i64)
    local.get 0
    i32.const -6954
    block (result i32)  ;; label = @1
      i32.const 4222
      br 0 (;@1;)
      i32.ctz
      table.copy
    end
    local.get 0
    i32.ctz
    table.size 0
    i32.ctz
    table.size 0
    unreachable
    table.set 0)
  (table (;0;) 1 9 funcref)
  (export "main" (func 0)))
```

The problematic input is the second `table.size 0` instruction, it is encoded as `fc 90 80 00 00`, the opcode is 4 bytes, the canonical binary should be simply `fc 16`.

I think this is failing because https://github.com/WebAssembly/wabt/blob/master/src/binary-reader-objdump.cc#L511 calls GetLength, which writes the canonical encoding in https://github.com/WebAssembly/wabt/blob/master/src/opcode.cc#L374 then count the bytes.

ngzhian:
Fixed #1631.


keithw:
This works for me (and fixes the crash in the linked issue) but doesn't seem to actually dump the non-canonically-encoded opcode (because it still only prints GetLength bytes from the opcode). Something like this seems to work for me:

```
commit 45d794f3c711642bd0442c88bf74dddcfeadc83b
Author: Keith Winstein <keithw@cs.stanford.edu>
Date:   Mon Feb 13 01:11:56 2023 -0800

    objdump: Print bytes of non-canonical LEB128 opcode

diff --git a/src/binary-reader-objdump.cc b/src/binary-reader-objdump.cc
index 23f2016f..0f093d7a 100644
--- a/src/binary-reader-objdump.cc
+++ b/src/binary-reader-objdump.cc
@@ -658,7 +658,9 @@ void BinaryReaderObjdumpDisassemble::LogOpcode(const char* fmt, ...) {
     return;
   }
   const Offset immediate_len = state->offset - current_opcode_offset;
-  const Offset opcode_size = current_opcode.GetLength();
+  const Offset opcode_size = last_opcode_end
+                                 ? (current_opcode_offset - last_opcode_end)
+                                 : current_opcode.GetLength();
   const Offset total_size = opcode_size + immediate_len;
   // current_opcode_offset has already read past this opcode; rewind it by the
   // size of this opcode, which may be more than one byte.
```

keithw:
It would also be nice if we could have a test for this, but it doesn't seem trivial because wat2wasm seems to canonicalize the LEB128 (even in a "module binary").

keithw (approved):
This lgtm and will be nice to get this PR off the pending list. I added the test case from https://github.com/WebAssembly/wabt/issues/1631 as a regression test. Because of https://github.com/WebAssembly/wabt/pull/1632#issuecomment-1427616900, wasm-objdump doesn't fully print out the non-canonically-encoded opcode in the test case, which we can address in a follow-on PR if we care. I think getting rid of the crash is the main benefit here.

PR #1633, opened 2021-03-11
[simd] Rename widen to extend, implement some double precision ops

ngzhian:
4 double precision conversion instructions are implemented:

- f32x4.demote_f64x2_zero
- f64x2.demote_low_f32x4
- f64x2.convert_low_i32x4_s
- f64x2.convert_low_i32x4_u

This is now sufficient to unskip simd_conversions.txt.

Rebase a bunch of tests due to the rename from widen to extend.


sbc100 (approved):

ngzhian:
I'm getting this in the ci:

- test/spec/simd/simd_conversions.txt
  expected error code 0, got -6.
  STDERR MISMATCH:
  --- expected
  +++ actual
  @@ -0,0 +1 @@
  +*** stack smashing detected ***: terminated

any idea what's going on?

sbc100:
> I'm getting this in the ci:
> 
> * test/spec/simd/simd_conversions.txt
>   expected error code 0, got -6.
>   STDERR MISMATCH:
>   --- expected
>   +++ actual
>   @@ -0,0 +1 @@
>   +*** stack smashing detected ***: terminated
> 
> any idea what's going on?

Looks like the compiler has detected some kind of overrun..  you don't see this locally I guess?

ngzhian:
Nope, running this locally is fine. Where is the error message coming from? I don't see anything from Wabt, seems like it's from the VM or something?

sbc100:
> Nope, running this locally is fine. Where is the error message coming from? I don't see anything from Wabt, seems like it's from the VM or something?

I think its coming from gcc or clang which have stack protection mechanisms.   Try matching he exact compiler version used in the github actions maybe?  Or look for options to enable stack protection in your local compiler?

ngzhian:
The CI uses gcc-9, i tried using that and cannot reproduce.

ngzhian:
That caught a real bug! Fixed in latest commit.

PR #1639, opened 2021-03-12
[simd] Implement q15mulrsat_s

ngzhian:
Rebase and unskip simd_i16x8_q15mulr_sat_s.


ngzhian:
@sbc100 ptal, thanks!

sbc100 (approved):
rubberstamp lgtm (i.e. I'll let the unittest be the judge of the implementation part).

ngzhian:
I copied this from the v8 side https://source.chromium.org/chromium/chromium/src/+/master:v8/src/utils/utils.h;l=170;drc=5a4394ac9d354e619f8e179a7f5e7f7528317a3c so we have more tests to judge this :)

PR #1640, opened 2021-03-16
[simd] Implement i32x4.dot_i16x8_s

ngzhian:
Rebase and unskip simd_i32x4_dot_i16x8.


sbc100 (approved):

ngzhian:
It looks like Add in interp-math.h doesn't handle overflow. Need to find another way.

PR #1646, opened 2021-03-17
[simd] Implement load lane

ngzhian:
This is a new kind of ir/ast node/instruction. It has 3 immediates:
memarg align, memarg offset, and lane index. This required new visitor
functions in all the places.

Drive-by cleanup to share the simd lane parsing logic between shuffle,
lane op and this new load lane instructions. This requires rebasing some
tests because the error messages are slightly different now.


src/binary-reader.cc
```diff
@@ -1318,6 +1318,21 @@ Result BinaryReader::ReadFunctionBody(Offset end_offset) {
...
+      case Opcode::V128Load64Lane: {
+        Address alignment_log2;
+        CHECK_RESULT(ReadAlignment(&alignment_log2, "load alignment"));
+        Address offset;
+        CHECK_RESULT(ReadAddress(&offset, 0, "load offset"));
+        uint8_t lane_val;
+        CHECK_RESULT(ReadU8(&lane_val, "Lane idx"));
+
+        CALLBACK(OnSimdLoadLaneExpr, opcode, alignment_log2, offset, lane_val);
+        CALLBACK(OnOpcodeUint32Uint32, alignment_log2, offset);
```
sbc100, line 1333: Shouldn't this be `OnOpcodeUint32Uint32Uint32`  and include `lave_val` ?   (I guess that doesn't exist yet?)

ngzhian: Yea you're right, added and fixed.

ngzhian:
@sbc100 ptal, thanks!

sbc100 (commented):
Should there be an objdump change here?  Should we have test that we can disassembly/dump this new instruction?

ngzhian:
Good idea, found a bug in the dumping, wasn't overriding the new uint32uint32uint32 correctly. Added the test. Ptal, thanks!

ngzhian:
@sbc100 ptal, thanks!

sbc100 (approved):

PR #1649, opened 2021-03-23
Verify roundtrip for alignment stored in simd load/store lane

alexcrichton:
This should fix an issue where `wasm2wat` was printing the wrong
alignment for the new simd load/store lane instructions.


keithw (approved):
This fix was also applied in #1699. Taking the additional roundtrip test to prevent regression.

PR #1652, opened 2021-03-24
Add _free to list of exports

ngzhian:
_free is used in src/wabt.post.js.


ngzhian:
Not sure if there are tests for this, I encountered this while trying to update https://github.com/AssemblyScript/wabt.js.

sbc100 (approved):

PR #1657, opened 2021-03-30
Enable multi value by default

UltimatePro-Grammer:
[Multi value](https://github.com/WebAssembly/multi-value) was merged into the spec almost a year ago, and all browsers support it. I think it is time to check the multi value checkbox by default.

Reference types were merged recently but support among browsers is limited at the moment, so I don't think it is time to check it by default yet.

Node stable does not support multi value, but this is a web based tool.

### DIFF
--- a/docs/demo/wat2wasm/index.html
+++ b/docs/demo/wat2wasm/index.html
@@ -46,7 +46,7 @@
     <div>
       <input type="checkbox" id="simd"><label for="simd">simd</label>
       <input type="checkbox" id="threads"><label for="threads">threads</label>
-      <input type="checkbox" id="multi_value"><label for="multi_value">multi value</label>
+      <input type="checkbox" id="multi_value" checked><label for="multi_value">multi value</label>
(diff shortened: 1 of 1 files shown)


tlively:
@sbc100 what do you think of this? Perhaps we should discuss our policy on default features for all the tools again.

sbc100:
We should be consistent with whatever policy we land on in https://github.com/WebAssembly/tool-conventions/issues/158.

Does wabt currently enable any non-MVP features by default?  (e.g. non-trapping float to int?)

sbc100:
Actually mutli-value was enabled back in #1405.

It looks like these html documents all needed updating.. I think they are automatically generated.    Perhaps close this PR and open a new one that updates them all?

sbc100:
Oh wait, that page is manually maintained.  lgtm.. plus we should probably update the other checkbox that are on by default these days.

sbc100 (approved):

Issue #1651, opened 2021-03-23
`wasm2wat --generate-names` produces invalid names


fitzgen:
### Test Case

```wat
;; test.wat
(module
  (table (;0;) 102 102 funcref)
  (elem (;0;) (i32.const 1) func))
```

### Steps to Reproduce

```shell
$ wat2wasm test.wat -o test.wasm
$ wasm2wat --generate-names test.wasm -o with-names.wat
$ wat2wasm with-names.wat -o with-names.wasm
with-names.wat:3:4: error: undefined table variable "$e0"
  (elem $e0 (i32.const 1) func))
   ^^^^ 
```

### Expected Results

The Wasm file is round trippable through `wasm2wat --generate-names` without any errors.

### Actual Results

`wasm2wat --generate-names` produces invalid element segment names (I think they are in the place where the table name belongs?)

sbc100:
Does it make sense to name active segments?   Can the name be used anywhere?

fitzgen:
Oh another idea: WABT is in oss-fuzz right? It may make sense to add a fuzz target that does something like:

```rust
void fuzz_target(uint8_t* ptr, size_t len) {
    if !is_valid_wasm(ptr, len) {
        return;
    }

    // Do the equivalent of `wasm2wat --generate-names`.
    size_t wat_len = 0;
    char* wat = wasm2wat_with_generate_names(ptr, len, &wat_len);

    // Convert it back to Wasm.
    size_t wasm_len = 0;
    uint8_t* wasm = wat2wasm(wat, wat_len, &wasm_len);

    // The conversion back to Wasm had better have succeeded without
    // errors and produced a valid Wasm buffer.
    assert(is_wasm_valid(wasm, wasm_len));
}
```

This should quickly track down this whole class of bug.

fitzgen:
> Does it make sense to name active segments? Can the name be used anywhere?

AFAIK, the name cannot be used anywhere.

Maybe fixing this bug is just deleting a line of code? If so, I'm happy to make a PR if someone points out the relevant line of code and which tests might need updating :)

sbc100:
Actually in wasm-ld / emscripten we use make use of names for active elem segments so I think for completeness its worth fixing.

sbc100:
The issue relates to whether bulk memory is enabled.   If I enable bulk memory on the failing command the problem goes away:

```
$ wat2wasm with-names.wat -o with-names.wasm --enable-bulk-memory
```

[...]

PR #1659, opened 2021-04-07
Add more support to extended names section.

sbc100:
I noticed we lacked support here while debugging #1651.

### DIFF
--- a/src/binary-reader-ir.cc
+++ b/src/binary-reader-ir.cc
@@ -248,6 +248,9 @@ class BinaryReaderIR : public BinaryReaderNop {
   Result OnLocalName(Index function_index,
                      Index local_index,
                      string_view local_name) override;
+  Result OnNameEntry(NameSectionSubsection type,
+                     Index index,
+                     string_view name) override;
 
   Result BeginEventSection(Offset size) override { return Result::Ok; }
   Result OnEventCount(Index count) override { return Result::Ok; }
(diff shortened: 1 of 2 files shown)


test/roundtrip/debug-names.txt
```diff
@@ -1,6 +1,10 @@
 ;;; TOOL: run-roundtrip
-;;; ARGS: --debug-names
+;;; ARGS: --debug-names --enable-bulk-memory
```
ngzhian, line 2: Do we want a new test with this flag and the additions below? Otherwise we only test with this flag enabled.

ngzhian (approved):
LGTM, the repetition looks ripe for extracting into a helper, but can be done in a follow-up.

1 inline question on whether we want to add a test case.

test/roundtrip/debug-names.txt
sbc100: The reason I need to add this flag is that naming elem sections isn't possible without bulk memory enabled.   If we added another test it would looks just the same but I couldn't include the `elem` section name (its a parse error without bulk memory).

WDYT, should add a the second (module duplicate) test?  I'm happy to to if you think its worth it.

src/binary-reader-ir.cc
```diff
@@ -1273,6 +1301,105 @@ Result BinaryReaderIR::OnFunctionName(Index index, string_view name) {
   return Result::Ok;
 }
 
+Result BinaryReaderIR::SetTableName(Index index, string_view name) {
+  if (name.empty()) {
+    return Result::Ok;
+  }
+  if (index >= module_->tables.size()) {
+    PrintError("invalid table index: %" PRIindex, index);
```
sbc100, line 1309: Turns out that templatizing the error message here is not easy in C++... https://stackoverflow.com/questions/1826464/c-style-strings-as-template-arguments

I'll keep trying but will probably land as is and followup with a templetization.

ngzhian: Could we template on a `typename T`, then pass the entire string `"Invalid table index: %" PRIindex` as an arg, and maybe let some deduction/conversion happen?
Follow-up (with a TODO to indicate what we would like to do here) would be totally fine.

test/roundtrip/debug-names.txt
ngzhian: Ic, I can't imagine many cases where the duplicate case will be useful, since bulk-memory flag shouldn't affect the way we handle names for other sections. LGTM to not add a test. Thanks for explaining!

PR #1660, opened 2021-04-07
Implement wasm2c for SIMD expressions

ngzhian:
Note that all SIMD expressions are untested in wasm2c, since it doesn't support any feature flags yet. I followed existing implementation for other SIMD expressions, like LoadSplat.

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -264,8 +264,11 @@ class CWriter {
   void Write(const UnaryExpr&);
   void Write(const TernaryExpr&);
   void Write(const SimdLaneOpExpr&);
+  void Write(const SimdLoadLaneExpr&);
+  void Write(const SimdStoreLaneExpr&);
   void Write(const SimdShuffleOpExpr&);
   void Write(const LoadSplatExpr&);
+  void Write(const LoadZeroExpr&);
 
   const WriteCOptions& options_;
   const Module* module_ = nullptr;
@@ -1572,6 +1575,16 @@ void CWriter::Write(const ExprList& exprs) {
         break;
       }
 
+      case ExprType::SimdLoadLane: {
+        Write(*cast<SimdLoadLaneExpr>(&expr));
+        break;
+      }
+
+      case ExprType::SimdStoreLane: {
+        Write(*cast<SimdStoreLaneExpr>(&expr));
+        break;
+      }
+
       case ExprType::SimdShuffleOp: {
         Write(*cast<SimdShuffleOpExpr>(&expr));
         break;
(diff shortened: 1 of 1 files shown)


src/c-writer.cc
```diff
@@ -2171,6 +2188,30 @@ void CWriter::Write(const SimdLaneOpExpr& expr) {
...
 
+void CWriter::Write(const SimdLoadLaneExpr& expr) {
+  assert(module_->memories.size() == 1);
+  Memory* memory = module_->memories[0];
+  Type result_type = expr.opcode.GetResultType();
+  Write(StackVar(0, result_type), " = ", expr.opcode.GetName(), "(",
+        ExternalPtr(memory->name), ", (u64)(", StackVar(1), "), ");
+  if (expr.offset != 0)
+    Write(" + ", expr.offset, "u");
+  Write("< ", StackVar(0), ", lane Imm: %d", expr.val, ");", Newline());
```
kripken, line 2199: I'm not sure how much of this code is functional atm. This `, lane Imm: ` stuff (already existing before your PR) doesn't seem like it can be valid C. So I assume in addition to not being tested that this code is wip?

src/c-writer.cc
```diff
@@ -2195,6 +2236,19 @@ void CWriter::Write(const LoadSplatExpr& expr) {
...
 
+void CWriter::Write(const LoadZeroExpr& expr) {
+  assert(module_->memories.size() == 1);
+  Memory* memory = module_->memories[0];
+  Type result_type = expr.opcode.GetResultType();
+  Write(StackVar(0, result_type), " = ", expr.opcode.GetName(), "(",
+        ExternalPtr(memory->name), ", (u64)(", StackVar(0), ")");
+  if (expr.offset != 0)
+    Write(" + ", expr.offset, "u");
+  Write(");", Newline());
```
kripken, line 2247: this differs from the one above it, which includes the offset in the u64 cast. I guess the result is the same, but it would be good to be consistent.

kripken (commented):
I'd actually recommend getting some sort of basic testing first before adding any code. Which depends on feature support I guess. Without those we are working blindly here.

But I'm not advocating for working on testing and feature support here atm - I'm not sure how much work that would be, and have not heard of an urgent reason for wasm2c+simd. Is there?

ngzhian:
Thanks for the review Alon. In that case, Sam also suggested adding a TODO or assertion there, so that if we do run tests with SIMD, we can implement it then. Do you think that will be a better way to go?

sbc100:
> Thanks for the review Alon. In that case, Sam also suggested adding a TODO or assertion there, so that if we do run tests with SIMD, we can implement it then. Do you think that will be a better way to go?

I can't see any downside to asserting here .. given that its missing functionality.

kripken:
Assert sgtm.

ngzhian:
I left the functions there with `UNIMPLEMENTED` in the body. Ptal, thanks!

kripken (approved):

PR #1661, opened 2021-04-08
Always do a full roundtrip in run-roundtrip.py

sbc100:
Even when the result is to be printed rather than compared byte for byte
with the first version its still good to process the resulting wat
output file so that we know we can parse what we generate.

Case in point, this changed caused me to fix two latent bugs:

1. We were not correctly parsing events with inline import/export.
2. We were output element segment names even when bulk memory
   was not enabled (See #1651)

The fix for (2) is a little more involved that we might like since for
the first time the wat writer needs to know what features are enabled.

Fixes: #1651

### DIFF
--- a/src/wat-writer.cc
+++ b/src/wat-writer.cc
@@ -1435,7 +1435,15 @@ void WatWriter::WriteTable(const Table& table) {
 
 void WatWriter::WriteElemSegment(const ElemSegment& segment) {
   WriteOpenSpace("elem");
-  WriteNameOrIndex(segment.name, elem_segment_index_, NextChar::Space);
+  // The first name we encounter here, pre-bulk-memory, was intended to refer to
+  // the table while segment names were not supported at all.  For this reason
+  // we cannot emit a segment name here without bulk-memory enabled, otherwise
+  // the name will be assumed to be the name of a table and parsing will fail.
+  if (options_.features.bulk_memory_enabled()) {
+    WriteNameOrIndex(segment.name, elem_segment_index_, NextChar::Space);
+  } else {
+    Writef("(;%u;)", elem_segment_index_);
+  }
 
   uint8_t flags = segment.GetFlags(&module);
 
--- a/include/wabt/wat-writer.h
+++ b/include/wabt/wat-writer.h
@@ -18,6 +18,7 @@
 #define WABT_WAT_WRITER_H_
 
 #include "wabt/common.h"
+#include "wabt/feature.h"
 
 namespace wabt {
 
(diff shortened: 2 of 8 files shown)


src/wat-writer.cc
```diff
@@ -1380,7 +1380,13 @@ void WatWriter::WriteTable(const Table& table) {
 
 void WatWriter::WriteElemSegment(const ElemSegment& segment) {
   WriteOpenSpace("elem");
-  WriteNameOrIndex(segment.name, elem_segment_index_, NextChar::Space);
+  // The name there pre-bulk-memory was intended to refer to the table
```
ngzhian, line 1383: typo here? I can't really parse this sentence, s/there/that?

ngzhian (approved):
I'm not too familiar with the event field parsing part, the rest lgtm, nice clean up in the run-roundtrip.py script!

src/wat-writer.cc
sbc100: I re-write this .. let me know if it make more sense now.

sbc100:
@binji .. do you think there is any way we can avoid having to pass features to the wat writer here?

sbc100:
> @binji .. do you think there is any way we can avoid having to pass features to the wat writer here?

I guess the alternative would be instead to pass `Features` to `GenerateNames` and in that case avoid generating segment names at all?

src/wat-writer.cc
```diff
@@ -1380,7 +1380,15 @@ void WatWriter::WriteTable(const Table& table) {
 
 void WatWriter::WriteElemSegment(const ElemSegment& segment) {
   WriteOpenSpace("elem");
-  WriteNameOrIndex(segment.name, elem_segment_index_, NextChar::Space);
+  // The first name we encounter here, pre-bulk-memory, was intended to refer to
+  // the table while segment names were not supported at all.  For this reason
+  // we cannot emit a segment name here without bulk-memory enabled, otherwise
+  // the name will be assumed to be the ame of a table and parsing will fail.
```
ngzhian, line 1386: LGTM, typo here, s/ame/name.

src/wat-writer.cc
ngzhian: Thanks!

sbc100:
@binji any objections to this approach?

keithw:
I took the liberty of rebasing this on top of the current main branch, which caused it to start crashing on a bug that is fixed by #1889, which I also rebased on top of the current main branch and then rebased this PR on top of it.

@takikawa @sbc100 any objections to merging both of these in their present form?

keithw:
Longer-term, it could be nice to try to enforcing roundtripping on every (valid?) module in the spec tests.

keithw:
It doesn't look like #1889 will be merged soon, but it seems like it would still be nice to get this PR in after almost two years and benefit from the enhanced testing. How about if we disable the `roundtrip/fold-function-references.txt` test for now? (Thanks to this PR's full roundtripping, that test file starts exposing a bug that #1889 would fix.)

I just rebased this on the current main branch (and added a SKIP to that test file). @sbc100 Any objections to merging at this point?

sbc100 (commented):
Great!  Lets land it

Issue #1665, opened 2021-04-23
[wasm-decompile] Same variable declaration gets emitted multiple times


RReverser:
Check the attached example. For some reason `wasm-decompile` produces multiple identical declarations for the same variables:

```
  var b:{ a:int, b:int, c:int, d:int, e:int, f:int, g:int, h:int, i:int, j:int, k:int, l:int }
  var d:int;
  var d:int;
  var c:{ a:int, b:int, c:int, d:int, e:int, f:int, g:int, h:int, i:int, j:int, k:int, l:int }
  var d:int;
  var b:{ a:int, b:int, c:int, d:int, e:int, f:int, g:int, h:int, i:int, j:int, k:int, l:int }
  var e:int;
  var f:int;
  var g:int;
  var h:int;
  var i:int;
  var j:int;
  var k:int;
  var l:int;
  var m:int;
  var c:{ a:int, b:int, c:int, d:int, e:int, f:int, g:int, h:int, i:int, j:int, k:int, l:int }
  var c:{ a:int, b:int, c:int, d:int, e:int, f:int, g:int, h:int, i:int, j:int, k:int, l:int }
  var c:{ a:int, b:int, c:int, d:int, e:int, f:int, g:int, h:int, i:int, j:int, k:int, l:int }
  var d:int;
  var c:{ a:int, b:int, c:int, d:int, e:int, f:int, g:int, h:int, i:int, j:int, k:int, l:int }
  var e:int;
  var c:{ a:int, b:int, c:int, d:int, e:int, f:int, g:int, h:int, i:int, j:int, k:int, l:int }
  var f:int;
  var c:{ a:int, b:int, c:int, d:int, e:int, f:int, g:int, h:int, i:int, j:int, k:int, l:int }
  var g:int;
  var c:{ a:int, b:int, c:int, d:int, e:int, f:int, g:int, h:int, i:int, j:int, k:int, l:int }
  var h:int;
  var c:{ a:int, b:int, c:int, d:int, e:int, f:int, g:int, h:int, i:int, j:int, k:int, l:int }
  var i:int;
  var c:{ a:int, b:int, c:int, d:int, e:int, f:int, g:int, h:int, i:int, j:int, k:int, l:int }
  var j:int;
  var c:{ a:int, b:int, c:int, d:int, e:int, f:int, g:int, h:int, i:int, j:int, k:int, l:int }
  var k:int;
[...]

PR #1666, opened 2021-04-26
[wasm-decompile] Fix repeated pre-declarations

aardappel:
See https://github.com/WebAssembly/wabt/issues/1565
And https://github.com/WebAssembly/wabt/issues/1665
This could be done better, but for now is sufficient.


aardappel:
No test, since I haven't verified what causes this, and didn't appear in my own testing with LLVM-produced object files and modules, and don't have a clear repro other than the test files in the above issues. My guess is multiple uses of the same uninitialized or deeper-declared variables.

sbc100 (approved):

PR #1675, opened 2021-06-12
Fix resolution of delegate labels

takikawa:
This PR has two commits for fixing bugs related to try-delegate labels.

One fixes the printing in certain cases (similar to #1609).

The other fixes the label resolution code. Example from the spec (https://github.com/WebAssembly/exception-handling/blob/master/proposals/exception-handling/Exceptions.md):

```
try $l1
  try
    call $foo
  delegate $l1  ;; (= delegate 0)
catch
  ...
catch_all
  ...
end
```

The label `$l1` should resolve to `0` in the delegate target, which is 1 less than how typical block constructs work. The current code mis-resolves this to `1`. It should also error in cases like the following:

```
try $l1
  nop
delegate $l1
```

Because it's not possible to represent the label `$l1` in the binary format, as a `try-delegate` cannot target itself.

### DIFF
--- a/src/resolve-names.cc
+++ b/src/resolve-names.cc
@@ -412,7 +412,21 @@ Result NameResolver::OnCatchExpr(TryExpr*, Catch* catch_) {
 }
 
 Result NameResolver::OnDelegateExpr(TryExpr* expr) {
-  ResolveLabelVar(&expr->delegate_target);
+  Var* delegate_target = &expr->delegate_target;
+  bool was_name = delegate_target->is_name();
+  ResolveLabelVar(delegate_target);
+  if (was_name) {
+    // Invalid label (of the immediate try-delegate block) used.
+    if (delegate_target->index() == 0) {
+      PrintError(&delegate_target->loc, "invalid label variable for try-delegate");
+      return Result::Ok;
+    }
+    // Adjust label by one, because a delegate instruction cannot target its
+    // immediately enclosing try block.
+    delegate_target->set_index(delegate_target->index() - 1);
+  }
+  // Pop the label here as a try-delegate has no `end` instruction.
+  PopLabel();
   return Result::Ok;
 }
 
--- a/src/resolve-names.cc
+++ b/src/resolve-names.cc
@@ -412,21 +412,13 @@ Result NameResolver::OnCatchExpr(TryExpr*, Catch* catch_) {
 }
 
 Result NameResolver::OnDelegateExpr(TryExpr* expr) {
-  Var* delegate_target = &expr->delegate_target;
-  bool was_name = delegate_target->is_name();
-  ResolveLabelVar(delegate_target);
-  if (was_name) {
-    // Invalid label (of the immediate try-delegate block) used.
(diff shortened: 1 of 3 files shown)


test/roundtrip/fold-try-delegate.txt
```diff
@@ -1,28 +1,31 @@
...
+  (func
+    try
       try
         nop
-      delegate 1
-      i32.const 7
+      delegate 0
+      try
+        nop
+      delegate 0
```
aheejin, line 11: This test has changed.. Which fix is this relevant relevant to?

aheejin (commented):
Sorry for the late reply! Can we split the two fixes into two separate PRs?

takikawa:
> Sorry for the late reply! Can we split the two fixes into two separate PRs?

No problem, I've split the printing fix out as #1685 now.

test/roundtrip/fold-try-delegate.txt
takikawa: This change should now only be in the other PR #1685 for printing. The reason this test changes, btw, is just to align it with the new test `roundtrip/try-delegate.txt` in having multiple adjacent try-delegate blocks to exercise the relevant code path.

src/resolve-names.cc
```diff
@@ -412,7 +412,21 @@ Result NameResolver::OnCatchExpr(TryExpr*, Catch* catch_) {
...
+    // Invalid label (of the immediate try-delegate block) used.
+    if (delegate_target->index() == 0) {
+      PrintError(&delegate_target->loc, "invalid label variable for try-delegate");
+      return Result::Ok;
+    }
+    // Adjust label by one, because a delegate instruction cannot target its
+    // immediately enclosing try block.
+    delegate_target->set_index(delegate_target->index() - 1);
+  }
+  // Pop the label here as a try-delegate has no `end` instruction.
```
takikawa, line 428: Note: I added this line in the latest revision, as this was another bug in this resolution code. (it is sufficiently tested by the `dump` test case I added---without the pop the `$try1` label was off because it came after another `try-delegate`.)

takikawa:
In order to make the two split PRs mergeable, I changed the test addition in this PR. Instead of testing in `roundtrip`, this modifies `dump/try-delegate.txt` to ensure the label gets resolved correctly.

src/resolve-names.cc
```diff
@@ -412,7 +412,21 @@ Result NameResolver::OnCatchExpr(TryExpr*, Catch* catch_) {
...
+  PopLabel();
```
aheejin, line 429: I might be mistaken, but if we do this `PopLabel()` at the top, wouldn't we not need all these fixup routines?

The point of the current rule is to think of the whole `try`-`delegate` like a single branch instruction, and treat its immediate argument as such.

takikawa: Good point, I think that will work. I pushed a commit that makes this change. It also changes the error message on the failing test case a bit, but I think that should be fine.

aheejin (approved):
Thanks!

PR #1676, opened 2021-06-13
Allow plain `try` with no `catch` or `delegate`

takikawa:
Recent changes in the exception handling proposal spec (https://github.com/WebAssembly/exception-handling/pull/157) now allow a `try` block with no `catch` or other clauses (to reduce syntactic special cases). In this case, a `try` is identical to a `block`.

This PR adds support for this extra case.


takikawa:
Updated for compatibility with events->tag renaming.

takikawa:
@aheejin would you be a good person to review this patch? (ditto for https://github.com/WebAssembly/wabt/pull/1675) I'd use "request review" but I think I can't use that feature on this repo.

src/wast-parser.cc
```diff
@@ -2771,32 +2769,33 @@ Result WastParser::ParseExpr(ExprList* exprs) {
...
+              Consume();
+              Var var;
+              CHECK_RESULT(ParseVar(&var));
+              expr->delegate_target = var;
+              expr->kind = TryKind::Delegate;
+              EXPECT(Rpar);
+              break;
+            }
+            default:
+              break;
```
aheejin, line 2797: We enter this `switch` only when `if (PeekMatch(TokenType::Lpar))` is true, right? Then shouldn't we still expect one of `catch` or `delegate`? If this is not an error, this sounds like we don't error out in this case? (Because of mismatched parens, it will error eventually later, but not here I guess?)
```wasm
(try
  (do ... )
  ( ;; started lpar, and didn't finish ??
```
I'm not very familiar with the code here so I might be mistaken.

aheejin (commented):
Sorry for the delayed reply!

src/wast-parser.cc
takikawa: Thanks, that's a good point. I think the clause check got lost in restructuring this code. I pushed a fix and also added a new test that will exercise this case.

takikawa:
This is now rebased to fix conflicts (conflicted with `unwind` deletion PR).

aheejin (approved):
Thank you!

PR #1677, opened 2021-06-17
Add submodule init to README

ricochet:
Add submodule init before CMake commands. gtest and others will fail when running cmake without this.


kripken (approved):
Thanks!

PR #1680, opened 2021-06-26
wasm2c: Support saturating-float-to-int

SoniEx2:


SoniEx2:
(Same caveats with --disable-* flags as #1679 apply. This was fairly trivial otherwise tho, it was basically a matter of copy-paste.)

src/wasm2c.c.tmpl
```diff
@@ -170,6 +170,28 @@ DEFINE_STORE(i64_store32, u32, u64);
...
+   : (UNLIKELY(!((x)minop(min)))) ? smin                        \
+   : (UNLIKELY(!((x) < (max))))   ? smax                        \
+                                  : (ut)(st)(x))
+
+#define I32_TRUNC_SAT_S_F32(x) TRUNC_SAT_S(u32, s32, f32, (f32)INT32_MIN, INT32_MIN, >=, 2147483648.f, INT32_MAX, x)
+#define I64_TRUNC_SAT_S_F32(x) TRUNC_SAT_S(u64, s64, f32, (f32)INT64_MIN, INT64_MIN, >=, (f32)INT64_MAX, INT64_MAX, x)
+#define I32_TRUNC_SAT_S_F64(x) TRUNC_SAT_S(u32, s32, f64, -2147483649., INT32_MIN, >, 2147483648., INT32_MAX, x)
+#define I64_TRUNC_SAT_S_F64(x) TRUNC_SAT_S(u64, s64, f64, (f64)INT64_MIN, INT64_MIN, >=, (f64)INT64_MAX, INT64_MAX, x)
+
+#define TRUNC_SAT_U(ut, ft, max, smax, x) \
```
kripken, line 184: what does `s` mean in `smax`? from the code it looks like the integer / non-float maximum, but I'm not able to guess at what `s` is an abbreviation of...

SoniEx2: saturated was what we were thinking at the time.

kripken: Oh, I see.. yes, that makes sense!

kripken (approved):
Looks correct to me. Corner cases with this stuff are pretty tricky to review for, but as the spec tests pass this should be good, and it will also get fuzzed.

PR #1682, opened 2021-06-26
[EH] Remove `unwind`

aheejin:
`unwind` was removed. See WebAssembly/exception-handling#156.


src/type-checker.cc
```diff
@@ -618,15 +617,7 @@ Result TypeChecker::OnEnd() {
...
-  if (label->label_type == LabelType::Unwind) {
-    // Unwind is unusual in that it always unwinds the control stack at the end,
-    // and therefore the return type of the unwind expressions are not the same
-    // as the block return type.
-    TypeVector empty;
-    result |= OnEnd(label, empty, desc, desc);
-  } else {
-    result |= OnEnd(label, label->result_types, desc, desc);
-  }
+  result |= OnEnd(label, label->result_types, desc, desc);
```
takikawa, line 620: I believe it's possible to further simplify here, because the 2nd argument `check_type` (line 592) was added for unwind. The argument can be removed and then `label->result_types` passed in line 596 instead of `check_type`. (see commit 036a632a24679062e7fc891e7743195139bfa0a9 that added this)

test/wasm2c/spec/unwind.txt
```diff
@@ -1,5 +0,0 @@
-;;; TOOL: run-spec-wasm2c
```
takikawa, line 1: I think this file should stay actually, the `third_party/testsuite/unwind.wast` is not related to `try-unwind` AFAICT.

test/spec/unwind.txt
```diff
@@ -1,13 +0,0 @@
-;;; TOOL: run-interp-spec
```
takikawa, line 1: Ditto my other comment, I think this file should stay.

takikawa (approved):

takikawa:
LGTM other than comments above.

src/type-checker.cc
aheejin: Thanks! Done.

test/wasm2c/spec/unwind.txt
aheejin: Oh, thanks for the catch. I blindly searched for `unwind`... Restored.

test/spec/unwind.txt
aheejin: Done

sbc100 (approved):

PR #1683, opened 2021-06-27
Add clang-format check to CI

aheejin:
This adds clang-format check to the CI. This only checks the diff so it
doesn't affect the other parts of the codebase. Also generated files in
src/prebuilt/ directories are excluded from the check. A new
src/prebuilt/.clang-format is added to disable the check in the
directory.

scripts/clang-format-diff.sh is copied from the same file we are using in
Binaryen.

Two clang-format errors are added to test if CI correctly errors out;
one within a normal source file in src/, and the other in src/prebuilt.
The latter should not be shown in the error, because the directory is
excluded. The errors will be removed before merging.


aheejin:
It looks the CI correct displays the clang-format error I introduced: https://github.com/WebAssembly/wabt/pull/1683/checks?check_run_id=2923646482

sbc100 (commented):
Seems reasonable once we deal with issues in #1684/

aheejin:
@sbc100 Can you LG this?

sbc100 (approved):

PR #1684, opened 2021-06-27
Clang-format codebase

aheejin:
This applies clang-format to the whole codebase.

I noticed we have .clang-format in wabt but the codebase is not very
well formatted. This kind of mass-formatting PR has fans and skeptics
because it can mess with `git blame`, but we did a similar thing in
Binaryen a few years ago (WebAssembly/binaryen#2048, which was merged in
WebAssembly/binaryen#2059) and it was not very confusing after all.
If we are ever going to format the codebase, I think it is easier to do
it in a single big PR than dozens of smaller PRs.

This is using the existing .clang-format file in this repo, which
follows the style of Chromium. If we think this does not suit the
current formatting style, we can potentially tweak .clang-format too.
For example, I noticed the current codebase puts many `case` statements
within a single line when they are short, but the current .clang-format
does not allow that.

This does not include files in src/prebuilt, because they are generated.

This also manually fixes some comment lines, because mechanically
applying clang-format to long inline comments can look weird.

This PR is mainly to listen to opinions and it is OK we end up not
merging this. (Also even if we end up merging this, I'm not sure if I am
the right person to change this many lines of code, given that I am not
a frequent contributor in this repo.)

I also added a clang-format check hook in the Github CI in #1683, which
I think can be less controversial, given that it only checks the diff.


kripken:
This sounds good to me, but one thing we should confirm before landing (if we decide to do that) is to check if it causes large conflicts with existing or intended PRs. I'll ask around about the latter.

src/binary.cc
```diff
@@ -43,16 +43,8 @@ const char* GetSectionName(BinarySection sec) {
...
-  "module",
-  "function",
-  "local",
-  "label",
-  "type",
-  "table",
-  "memory",
-  "global",
-  "elemseg",
-  "dataseg",
```
sbc100, line 55: Can we keep this as is?    I think its usefull to be able to see the index of a given string based on its line offset (since I think the indexes here are relevant numbers in the binary format).

src/c-writer.cc
```diff
@@ -388,11 +392,16 @@ std::string CWriter::Deref(const std::string& s) {
...
+    case Type::I32:
+      return 'i';
+    case Type::I64:
+      return 'j';
+    case Type::F32:
+      return 'f';
+    case Type::F64:
+      return 'd';
+    default:
+      WABT_UNREACHABLE;
```
sbc100, line 404: I agree that these change from compact to spread out form are potentially questionable.

src/decompiler-ls.h
```diff
@@ -84,18 +102,19 @@ struct LoadStoreTracking {
   };
 
   void Track(const Node& n) {
-    for (auto& c : n.children) Track(c);
+    for (auto& c : n.children)
+      Track(c);
```
sbc100, line 106: Does chromium style not require braces around loop blocks?

src/decompiler-naming.h
```diff
@@ -173,21 +175,9 @@ void RenameAll(Module& module) {
...
-    { "void" },
-    { "int" },
-    { "unsigned" },
-    { "2" },
-    { "cxxabiv1" },
-    { "short" },
-    { "4096ul" },
+      {"const"},    {"std"},   {"allocator"}, {"char"},  {"basic"},
+      {"traits"},   {"wchar"}, {"t"},         {"void"},  {"int"},
+      {"unsigned"}, {"2"},     {"cxxabiv1"},  {"short"}, {"4096ul"},
```
sbc100, line 180: Again, I think this single-line-per-entry has some value maybe?

src/interp/interp-inl.h
```diff
@@ -351,16 +351,40 @@ bool operator!=(const RefPtr<U>& lhs, const RefPtr<V>& rhs) {
...
+  return type == ValueType::F32;
+}
+template <>
+inline bool HasType<f64>(ValueType type) {
+  return type == ValueType::F64;
+}
+template <>
+inline bool HasType<Ref>(ValueType type) {
+  return IsReference(type);
+}
```
sbc100, line 384: Can we keep the LHS here?  Seems more readable

src/interp/interp-inl.h
```diff
@@ -371,14 +395,54 @@ inline bool TypesMatch(ValueType expected, ValueType actual) {
...
+  res.v128_ = val;
+  res.SetType(ValueType::V128);
+  return res;
+}
+inline Value WABT_VECTORCALL Value::Make(Ref val) {
+  Value res;
+  res.ref_ = val;
+  res.SetType(ValueType::ExternRef);
+  return res;
+}
```
sbc100, line 445: And here.

src/interp/interp-math.h
```diff
@@ -64,32 +64,107 @@ T WABT_VECTORCALL CanonNaN(T val) {
...
+  return lhs & ~rhs;
+}
+template <typename T>
+T WABT_VECTORCALL IntAvgr(T lhs, T rhs) {
+  return (lhs + rhs + 1) / 2;
+}
+template <typename T>
+T WABT_VECTORCALL Xchg(T lhs, T rhs) {
+  return rhs;
+}
```
sbc100, line 167: Perhaps wrap these blocks of templates in `// clang-format off` `// clang-format on` ?

src/stream.cc
```diff
@@ -255,7 +255,8 @@ FileStream::~FileStream() {
 }
 
 void FileStream::Flush() {
-  if (file_) fflush(file_);
+  if (file_)
+    fflush(file_);
```
sbc100, line 259: Again, I thought the local style was to require braces around blocks.

src/test-interp.cc
```diff
@@ -140,7 +139,7 @@ TEST_F(InterpTest, Disassemble) {
   auto buf = stream.ReleaseOutputBuffer();
 
   ExpectBufferStrEq(*buf,
-R"(   0| alloca 1
+                    R"(   0| alloca 1
```
sbc100, line 142: This seems worse.   Maybe we can can `ExpectBufferStrEq(*buf, R"(\` to avoid the issue?

src/utf8.cc
```diff
@@ -23,7 +23,7 @@ namespace wabt {
 namespace {
 
 const int s_utf8_length[256] = {
- // 0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
+    // 0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
     1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,  // 0x00
```
sbc100, line 27: This is worse since the columns no longer line up

sbc100 (commented):
This seem surprisingly small actually.  

Comments in-line but it seems there are basically 3 places we might want to override/ignore/supress:

1. Switch/case statements where the oringal code is very compact and readable.
2. Templates where we use more than 80 cols to make them readble/single-line.
3. Arrays of string literals where we may prefer to keep them one-line-per-entry.

If we can address those issues I see no problem with landing this.   I don't think it really matters who actually lands it, but we should probably wait for LGTM from binji.

In terms of how to deal with those 3 types of issue above I agess we have a few options:

1. Simply revert those specific blocks in this PR (seems like the easiest option)
2. Try to craft a `.clang-format` that encompases these things
3. Use per-file, or file-local settings to locally signal, for example, our digressions from normal 80-char limit on lines in certain places.

I suggest we just do (1) and let this land as a best-effort/95% reformat.

aardappel:
It looks like we may want to enable `AllowShortCaseLabelsOnASingleLine` (and maybe `AllowShortBlocksOnASingleLine`, `AllowShortFunctionsOnASingleLine`, `AllowShortIfStatementsOnASingleLine`) to automatically solve a lot of those more readable cases @sbc100 found. Though of course that is likely to change things the other way too.

Besides that, `// clang-format off` and `on` around a block can solve individual cases.

aardappel (approved):

src/decompiler-ls.h
aheejin: Requiring braces is the area of clang-tidy, not clang-format. We [do that in .clang-tidy](https://github.com/WebAssembly/binaryen/blob/6a2d7f989065820476268a2382db2eccf72aadd7/.clang-tidy#L1) in Binaryen. clang-format's job is to format whatever code given, as long as it compiles.

We can consider adding .clang-tidy to wabt, which should be a follow-up task. In this PR I think we should decide whether we should allow one-line short `if` or `for`/`while`. The current codebase seems to be using one-line `if`s a lot, so I'll try tweaking .clang-format to allow that. Let me know if you don't prefer that way.

src/c-writer.cc
aheejin: Added `AllowShortCaseLabelsOnASingleLine: true` to fix this.

src/decompiler-ls.h
sbc100: IIUC chromium/google style (which we are basing wabt on) is OK with single line block (where the keyword and the body are on a single line), but if you have a body on a line by itself it requires braces (unlike LLVM)

See https://google.github.io/styleguide/cppguide.html#Conditionals

Basically `if (x) foo();` is ok, but not

```
if (x) 
  foo();
```

sbc100: Given that I'm a little surprised that clang-format did this here...

sbc100: Actually I think I read that wrong... for historical reasons single line block are ok... ignore me.

src/interp/interp-inl.h
aheejin: I added
```
AllowShortFunctionsOnASingleLine: true
AlwaysBreakTemplateDeclarations: MultiLine
```
to 1. Allow short functions on an single line 2. Not break after `template<..>` when the whole function declaration can fit within a single line. After that this becomes
```cpp
template <> inline bool HasType<Ref>(ValueType type) {                           
  return IsReference(type);                                                      
} 
```
Even with `AllowShortFunctionsOnASingleLine: true`, we can't still fit this within a single line like LHS because it exceeds 80 cols. I think we can exclude some code blocks from 80 cols limit if that helps readability, but I'm not sure this place warrants that exception. I actually prefer this 80 cols-complying version. What do you think?

If functions are shorter than this so that they fit in 80 cols, they will now be allowed on a isngle line.

src/decompiler-ls.h
aheejin: Sorry, I'm in the middle of fixing things, so I haven't actually committed the newly formatted code with new .clang-format. Sorry for the confusion. I added `AllowShortLoopsOnASingleLine: true`, so this will be back to `if (x) foo();`. I will commit all those changes I guess in an hour or something.

src/interp/interp-inl.h
sbc100: It seems worth preserving the existing formatting where all the types line up and we have one-line-per-type ..  I think this is just limited to this one file, so it should be possible to make an an exception.

src/stream.cc
aheejin: As explained in https://github.com/WebAssembly/wabt/pull/1684#discussion_r660172600, it is clang-tidy's job. For this PR, I allowed one-line short ifs to match the codebase.

src/test-interp.cc
aheejin: Done

aheejin: Done. Also fixed places with similar patterns in this file. By the way what is `\` at the end for?

aheejin: Hmm, I guess `\` doesn't work in raw strings...

aheejin: Hmm, I don't think there's an easy way around this. `\` doesn't work as multiline-splacer because in raw strings `\` is considered as just a literal `\`. I agree it's ugly, but [the existing code](https://github.com/WebAssembly/wabt/blob/6c9355e490a37c7a990d504f32676c9d7a2ebb13/src/test-interp.cc#L142-L161) does not completely line up either. Also we have multiple raw strings like this in this file so adding `// clang-format off` to all those places seems a little too much. How about just living with this change, even if it is a little ugly?

src/utf8.cc
aheejin: Added `// clang-format off` and `// clang-format on` around this array.

aheejin:
I pushed newly formatted code to see how it compares with the current codebase. I haven't addressed a few comments from @sbc100 on preserving the current formatting, which will be addressed soon.

src/interp/interp-inl.h
aheejin: Wrap that part with `// clang-format off` and `on`.

src/interp/interp-inl.h
aheejin: Wrapped that part with `// clang-format off` and `on`.

src/interp/interp-math.h
aheejin: Done

.clang-format
```diff
@@ -1 +1,7 @@
 BasedOnStyle: Chromium
+AllowShortBlocksOnASingleLine: true
+AllowShortCaseLabelsOnASingleLine: true
+AllowShortFunctionsOnASingleLine: true
+AllowShortIfStatementsOnASingleLine: true
+AllowShortLoopsOnASingleLine: true
+AlwaysBreakTemplateDeclarations: MultiLine
```
sbc100, line 7: Perhaps now that we have the opt-opts for specific files like the `-inl.h` we don't actually need/want these?

src/binary.cc
aheejin: Wrapped in `// clang-format off` and `on`.

src/decompiler-naming.h
aheejin: Wrapped in `// clang-format off` and `on`.

.clang-format
aheejin: Hmm, yeah, it's ambiguous because it turns out the codebase is not very consistent in one way or the other. Without these options there were many short if/for/functions that were converted to multiple lines, but with these options now we have maybe as many multiple-line if/for/functions that were converted to single lines, so I guess either way we end up with sizable lines of changes. I'll try deleting this.

aheejin: Hmm, you mean reverting all those `Allow***` lines, or only these four lines? Because there are also
```
AllowShortBlocksOnASingleLine: true                                      
AllowShortCaseLabelsOnASingleLine: true
```
above these lines.

aheejin: OK, I removed all added lines from .clang-format except for 
```
AllowShortCaseLabelsOnASingleLine: true
```

Let me know if you want this to be removed as well.

aheejin:
Also many changes are due to how we list parameters, like this:
image
We can make the current style stay by setting `BinPackParameters: true` in .clang-format. Let me know if you want that. I don't have specific preference myself.

.clang-format
```diff
@@ -1 +1,2 @@
 BasedOnStyle: Chromium
+AllowShortCaseLabelsOnASingleLine: true
```
sbc100, line 2: I think it might be best to remove this too and just exclude the few switch/case blocks where we want single line.  It doesn't seems like we want this in the general case.

src/decompiler-naming.h
```diff
@@ -172,23 +174,25 @@ void RenameAll(Module& module) {
...
-    { "wchar" },
-    { "t" },
-    { "void" },
-    { "int" },
-    { "unsigned" },
-    { "2" },
-    { "cxxabiv1" },
-    { "short" },
-    { "4096ul" },
+      {"const"},
```
sbc100, line 179: Strange that it still changed these lines despite being "off"?   Not a big deal but odd.

src/interp/interp-inl.h
```diff
@@ -351,16 +351,40 @@ bool operator!=(const RefPtr<U>& lhs, const RefPtr<V>& rhs) {
...
-template <> inline bool HasType<s32>(ValueType type) { return type == ValueType::I32; }
-template <> inline bool HasType<u32>(ValueType type) { return type == ValueType::I32; }
-template <> inline bool HasType<s64>(ValueType type) { return type == ValueType::I64; }
-template <> inline bool HasType<u64>(ValueType type) { return type == ValueType::I64; }
-template <> inline bool HasType<f32>(ValueType type) { return type == ValueType::F32; }
-template <> inline bool HasType<f64>(ValueType type) { return type == ValueType::F64; }
-template <> inline bool HasType<Ref>(ValueType type) { return IsReference(type); }
-
-template <typename T> void RequireType(ValueType type) {
+inline bool IsReference(ValueType type) {
```
sbc100: Wrap this `ValueType` block too?

src/interp/interp.cc
```diff
@@ -1174,28 +1158,28 @@ RunResult Thread::StepInternal(Trap::Ptr* out_trap) {
       break;
     }
 
-    case O::I32Load:    return DoLoad<u32>(instr, out_trap);
```
sbc100, line 1177: Maybe certain case statements like this one that use vertical alignment can be wrapped in "off" and we can leave the default to be multi-line.   Its kind of shame we can't configure it to use leave swtich/cases as single-line if and only if they are already, but I guess that isn't an option?

src/interp/interp.cc
```diff
@@ -2060,17 +2091,48 @@ RunResult Thread::DoSimdReplace(Instr instr) {
   return RunResult::Ok;
 }
 
-template <typename T> struct Simd128;
```
sbc100, line 2063: Wrap this block in off?

src/test-interp.cc
sbc100: Can we just revert these parts of the PR?  Or are you committed to having 100% of the codebase be clang-format clean?    Maybe we could punt on these chunks for now?

aheejin:
I ended up adding too many lines of `// clang-format off` and `// clang-format on` in interp-inl.h and interp-math.h, so I just wrapped the whole file with `// clang-format off` and `on` rather than trying to wrap each small region of code.

sbc100:
> Also many changes are due to how we list parameters, like this:
> image
> We can make the current style stay by setting `BinPackParameters: true` in .clang-format. Let me know if you want that. I don't have specific preference myself.

I don't feel strongly about this one because I don't feel like it worsens scanability/readability (unlike the vertical alignment stuff that we have in some places).   Happy to go either way and I guess we don't have any reason to differ from chromium on this.

.clang-format
aheejin: Done

src/decompiler-naming.h
aheejin: Chromium style is 4-space indentation for array elements, so I preserved that part before re-breaking the lines and adding `// clang-format off`. We can make this array indentation to 2-spaces with another clang-format option though.

src/interp/interp-inl.h
aheejin: I ended up wrapping the whole file (otherwise I had to add so many `// clang-format off`s) so I think this will be preserved.

src/interp/interp.cc
aheejin: I don't think we can tell clang-tidy to keep some part of the code if certain conditions are met.. 

Wrapping this part is not a problem, but this file has several similar parts:
https://github.com/WebAssembly/wabt/blob/b97dd591fad3e4b7d382d409247007cf65fc2baf/src/interp/interp.cc#L605-L610
https://github.com/WebAssembly/wabt/blob/b97dd591fad3e4b7d382d409247007cf65fc2baf/src/interp/interp.cc#L756-L762
https://github.com/WebAssembly/wabt/blob/b97dd591fad3e4b7d382d409247007cf65fc2baf/src/interp/interp.cc#L795-L801
https://github.com/WebAssembly/wabt/blob/b97dd591fad3e4b7d382d409247007cf65fc2baf/src/interp/interp.cc#L2416-L2424
https://github.com/WebAssembly/wabt/blob/b97dd591fad3e4b7d382d409247007cf65fc2baf/src/interp/interp.cc#L2427-L2439
https://github.com/WebAssembly/wabt/blob/b97dd591fad3e4b7d382d409247007cf65fc2baf/src/interp/interp.cc#L1177-L1200
...

Excluding every single one of these will not be very clean either. The switch-case you pointed is very big, so do you think wrapping only that one sounds OK?

I personally prefer fewer exceptions, especially for normal code. What I think exceptions are most warranted for is comments that don't make sense with 80-col limits, such as this one: https://github.com/WebAssembly/binaryen/blob/6a2d7f989065820476268a2382db2eccf72aadd7/src/support/safe_integer.cpp#L121-L145

sbc100: I think that sounds reasonable.

aheejin: Added `// clang-format off` and `on` to wrap this big switch-case. Left others as they are.

src/interp/interp.cc
aheejin: Wrapped... But I can't help feeling we are adding too many of this exceptions though. If we strongly want single lines to be preserved, I actually think turning on those `AllowShort***` options might be better (it also changes bunch of other existing code though)

src/test-interp.cc
aheejin: Reverted this and two more parts that use a raw string and have a similar change.

src/interp/interp.cc
sbc100: I think the rules are correct, but maybe be want exceptions (perhaps many) in this codebase?    Perhaps we not ready for rigid enforcement in wabt yet?

sbc100:
One option for us if we want to do this a little more incrementally, and with a little more flexibility, is to simply revert any part of this PR that are semi-controversial, then land it as a "format 95% of the codebase".   We would then need some way to override the clang-format CI check (I guess we already have the ability to force commit as admins in those cases).  That way we can punt the hard parts and get almost all the benefits.   If we want to revisit the final 5% at a later date we can (at which point the diff will be much more reasonably sided and focused).  WDYT?

sbc100:
> One option for us if we want to do this a little more incrementally, and with a little more flexibility, is to simply revert any part of this PR that are semi-controversial, then land it as a "format 95% of the codebase". We would then need some way to override the clang-format CI check (I guess we already have the ability to force commit as admins in those cases). That way we can punt the hard parts and get almost all the benefits. If we want to revisit the final 5% at a later date we can (at which point the diff will be much more reasonably sided and focused). WDYT?

I believe this would put is roughly in the same place as LLVM itself, which has clang-format messages as warnings in PRs but nowhere near universal conformance across the code base (we would be in way better state than that but not perfect yet).

aheejin:
> One option for us if we want to do this a little more incrementally, and with a little more flexibility, is to simply revert any part of this PR that are semi-controversial, then land it as a "format 95% of the codebase". We would then need some way to override the clang-format CI check (I guess we already have the ability to force commit as admins in those cases). That way we can punt the hard parts and get almost all the benefits. If we want to revisit the final 5% at a later date we can (at which point the diff will be much more reasonably sided and focused). WDYT?

OK, that sounds good. Do you prefer to wrap those controversial blocks with `// clang-format off` and `on` or just revert the changes? We already placed many `// clang-format off`s:
```
aheejin@aheejin:~/wabt/src$ grep clang-format * -R
binary.cc:// clang-format off
binary.cc:// clang-format on
binary.h:/* clang-format off */
binary.h:/* clang-format on */
decompiler-naming.h:  // clang-format off
decompiler-naming.h:  // clang-format on
interp/interp-inl.h:// clang-format off
interp/interp-inl.h:// clang-format on
interp/interp-math.h:// clang-format off
interp/interp-math.h:// clang-format on
interp/interp.cc:  // clang-format off
interp/interp.cc:  // clang-format on
interp/interp.cc:// clang-format off
interp/interp.cc:// clang-format on
utf8.cc:// clang-format off
utf8.cc:// clang-format on
```

And how do we figure out which parts are controversial? Do we exclude only the parts you pointed out, or, like, exclude all single-lined switch-cases from formatting, which will be harder?

sbc100:
> > One option for us if we want to do this a little more incrementally, and with a little more flexibility, is to simply revert any part of this PR that are semi-controversial, then land it as a "format 95% of the codebase". We would then need some way to override the clang-format CI check (I guess we already have the ability to force commit as admins in those cases). That way we can punt the hard parts and get almost all the benefits. If we want to revisit the final 5% at a later date we can (at which point the diff will be much more reasonably sided and focused). WDYT?
> 
> OK, that sounds good. Do you prefer to wrap those controversial blocks with `// clang-format off` and `on` or just revert the changes? We already placed many `// clang-format off`s:
> 
> ```
> aheejin@aheejin:~/wabt/src$ grep clang-format * -R
> binary.cc:// clang-format off
> binary.cc:// clang-format on
> binary.h:/* clang-format off */
> binary.h:/* clang-format on */
> decompiler-naming.h:  // clang-format off
> decompiler-naming.h:  // clang-format on
> interp/interp-inl.h:// clang-format off
> interp/interp-inl.h:// clang-format on
> interp/interp-math.h:// clang-format off
> interp/interp-math.h:// clang-format on
> interp/interp.cc:  // clang-format off
> interp/interp.cc:  // clang-format on
> interp/interp.cc:// clang-format off
> interp/interp.cc:// clang-format on
> utf8.cc:// clang-format off
> utf8.cc:// clang-format on
> ```
> 
> And how do we figure out which parts are controversial? Do we exclude only the parts you pointed out, or, like, exclude all single-lined switch-cases from formatting, which will be harder?

No harm in leaving those in there.

I'll just to one more pass though and flag anything I think we should hold off on, then we can land this I think. 

Unless @binji wants to chime in?

src/type.h
```diff
@@ -71,29 +71,47 @@ class Type {
...
+      case Type::Func:
+        return "func";
+      case Type::Void:
+        return "void";
+      case Type::Any:
+        return "any";
+      case Type::ExternRef:
+        return "externref";
+      default:
+        return "<type_index>";
```
sbc100, line 99: Revert this block

src/type.h
```diff
@@ -71,29 +71,47 @@ class Type {
...
+      case Type::FuncRef:
+        return "func";
+      case Type::ExternRef:
+        return "extern";
+      case Type::Struct:
+        return "struct";
+      case Type::Array:
+        return "array";
+      default:
+        return "<invalid>";
```
sbc100, line 114: Revert this block

src/wast-parser.cc
```diff
@@ -1076,17 +1076,28 @@ Result WastParser::ParseModuleFieldList(Module* module) {
...
+    case TokenType::Global:
+      return ParseGlobalModuleField(module);
+    case TokenType::Import:
+      return ParseImportModuleField(module);
+    case TokenType::Memory:
+      return ParseMemoryModuleField(module);
+    case TokenType::Start:
+      return ParseStartModuleField(module);
+    case TokenType::Table:
+      return ParseTableModuleField(module);
```
sbc100, line 1100: Revert this block

src/wast-parser.cc
```diff
@@ -1595,11 +1606,21 @@ Result WastParser::ParseExportDesc(Export* export_) {
...
+      break;
+    case TokenType::Memory:
+      export_->kind = ExternalKind::Memory;
+      break;
+    case TokenType::Global:
+      export_->kind = ExternalKind::Global;
+      break;
+    case TokenType::Tag:
+      export_->kind = ExternalKind::Tag;
+      break;
```
sbc100, line 1623: Revert this block

src/wast-parser.cc
```diff
@@ -2222,20 +2244,38 @@ Result WastParser::ParseSimdV128Const(Const* const_,
...
+    case TokenType::F32X4: {
+      lane_count = 4;
+      integer = false;
+      break;
+    }
+    case TokenType::F64X2: {
+      lane_count = 2;
+      integer = false;
+      break;
+    }
```
sbc100, line 2272: Revert this block

src/decompiler.cc
```diff
@@ -230,17 +238,23 @@ struct Decompiler {
                struc.empty() ? GetDecompTypeName(t) : struc);
   }
 
-  bool ConstIntVal(const Expr* e, uint64_t &dest) {
+  bool ConstIntVal(const Expr* e, uint64_t& dest) {
     dest = 0;
-    if (!e || e->type() != ExprType::Const) return false;
+    if (!e || e->type() != ExprType::Const)
+      return false;
```
sbc100, line 244: Revert or add braces.. it looks like this pattern is only common in this one file, so adding braces is probably the way to go here and elsewhere in this file.

sbc100 (approved):
lgtm with all those reverts.  

(looks like the single line if statements are limited to the decompiler so fairly localized).

sbc100 (approved):
Great!  lgmt.  Thanks for putting up with all the back and forth.

I don't see anything controversial here now.  I think we can go ahead and land.

kripken (approved):

aheejin:
I double-checked the diff and a) reverted a few more `case` patterns @sbc100 was concerned about, and b) added curlies for one-line `if`s that were merged into a single-line. 'b)' complies with the current .clang-format but due to 'a)' we have some parts of the code that don't comply the current .clang-format. The following is how much the diff between this PR and the fully-formatted codebase based on the current .clang-format:
```
$ git diff --stat
 src/binary.cc                      |   12 +-
 src/c-writer.cc                    |  157 +++-
 src/decompiler-ls.h                |   54 +-
 src/decompiler-naming.h            |   18 +-
 src/interp/binary-reader-interp.cc |   20 +-
 src/interp/interp-inl.h            |  279 +++++--
 src/interp/interp-math.h           |  289 +++++--
 src/interp/interp.cc               | 1651 +++++++++++++++++++++++++-------------
 src/ir.h                           |   21 +-
 src/opcode.cc                      |    9 +-
 src/test-interp.cc                 |    6 +-
 src/tools/spectest-interp.cc       |   22 +-
 src/type.h                         |   54 +-
 src/wast-parser.cc                 |   85 +-
 14 files changed, 1873 insertions(+), 804 deletions(-)
```

I think we can address this separately by tweaking .clang-format, or not aim for 100% compliance as @sbc100 suggested.

I'll wait for some more days in case @binji wants to chime in before landing. Thanks for the feedback!

kripken:
Let's maybe wait for the large wasm2c PR to land first, to avoid a big conflict there? (#1721)

aheejin:
I merged the current `main` into this PR and fixed a few lines that were added after this PR. clang-format CI currently fails because of one of many code parts we reverted before. This lands us in a tricky situation: I think the ideal state is the current codebase is equal to the result that we mechanically run clang-format on. If we want to exclude certain parts, we can do `// clang-format off` and `// clang-format on`. But we excluded too many parts to wrap all of them in it, mostly one-liner functions. So some parts in the code does not match the clang-format results now, which is probably OK. But that means we need to ignore the clang-format CI time to time. If people are OK with ignoring the CI failure now, I'll go ahead and land this.

aheejin:
I ended up reverting the change that caused the CI failure because I don't have the administrator access that's required to merge PRs with CI failure 😅 Will merge this now.

PR #1685, opened 2021-06-29
Fix printing of try-delegate labels

takikawa:
This is similar to a previous bug (#1609) for other kinds of block labels.

The tests check that 2+ adjacent try-delegate blocks will have labels that print correctly.

(split from #1675)

### DIFF
--- a/src/wat-writer.cc
+++ b/src/wat-writer.cc
@@ -904,6 +904,7 @@ Result WatWriter::ExprVisitorDelegate::OnUnwindExpr(TryExpr* expr) {
 
 Result WatWriter::ExprVisitorDelegate::OnDelegateExpr(TryExpr* expr) {
   writer_->Dedent();
+  writer_->EndBlock();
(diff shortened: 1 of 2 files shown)


test/roundtrip/fold-try-delegate.txt
```diff
@@ -1,28 +1,31 @@
...
 (module
-  (func (result i32)
-    try (result i32)
+  (func
+    try
       try
         nop
-      delegate 1
-      i32.const 7
+      delegate 0
```
aheejin, line 8: `WatWriter::ExprVisitorDelegate::OnDelegateExpr` doesn't seem to be used to print folded wat format, so this test isn't really affected. (This test succeeds even now.)

takikawa: Thanks I've just removed the changes to this test in the latest push and kept only the necessary one.

aheejin (approved):

PR #1686, opened 2021-07-04
wasm2c: Support multi-value

SoniEx2:


src/c-writer.cc
```diff
@@ -333,14 +335,12 @@ void CWriter::PushLabel(LabelType label_type,
                         const std::string& name,
                         const FuncSignature& sig,
                         bool used) {
-  // TODO(binji): Add multi-value support.
```
SoniEx2, line 336: cc @binji also (sorry, we should've @'ed sooner but we weren't sure)

src/prebuilt/wasm2c.include.c
```diff
@@ -22,8 +22,8 @@ const char SECTION_NAME(declarations)[] =
 "#define CALL_INDIRECT(table, t, ft, x, ...)          \\\n"
 "  (LIKELY((x) < table.size && table.data[x].func &&  \\\n"
 "          table.data[x].func_type == func_types[ft]) \\\n"
-"       ? ((t)table.data[x].func)(__VA_ARGS__)        \\\n"
-"       : TRAP(CALL_INDIRECT))\n"
+"       || TRAP(CALL_INDIRECT)                        \\\n"
+"       , ((t)table.data[x].func)(__VA_ARGS__))\n"
```
kripken, line 26: This looks like an unrelated change? Or is there a connection I am not seeing?

kripken (approved):
I'm not an expert on this code, but it looks good to me. Nice work!

src/prebuilt/wasm2c.include.c
SoniEx2: Functions now return wasm_multi_ structs in addition to the numeric types, and 0 cannot be coerced to structs. TRAP doesn't return but it still needs to typecheck, and there's not much that can be done to TRAP to make it typecheck. Changing this to use a comma operator makes CALL_INDIRECT typecheck tho, and is arguably more correct.

kripken: I see, thank you! Yes, that makes perfect sense.

I see that `wasm_rt_trap` defines the noreturn attribute as it should. It would have been nice if it were possible to assert on that in the source, as in theory a future bug could cause control flow to continue to the thing after the comma (which could be a security vulnerability). But the test suite verifies this, and that's good enough I guess.

Issue #1377, opened 2020-03-25
Implement function-references proposal


binji:
See https://github.com/webassembly/function-references and https://github.com/WebAssembly/function-references/blob/master/proposals/function-references/Overview.md.

dbezhetskov:
FYI: I'm working on it

PR #1691, opened 2021-07-12
Begin support for typed function references proposal: added the flag and supported call_ref

dbezhetskov:
The corresponding issue: https://github.com/WebAssembly/wabt/issues/1377


src/type.h
```diff
@@ -116,6 +125,10 @@ class Type {
     return static_cast<Index>(enum_);
   }
 
+  bool IsTypeIdx() const {
+    return IsIndex() && (enum_ != Type::ExternRef || enum_ == Type::FuncRef);
```
takikawa, line 129: Isn't the second condition always true if the first one is? (because `IsIndex` implies the `enum_` is zero/positive and thus not a negative value like `ExternRef`)

dbezhetskov: You are right, it was a typo, thanks for the catch

test/parse/expr/bad-callref-int32.txt
```diff
@@ -0,0 +1,15 @@
...
+;;; ERROR: 1
+(module
+  (func (export "main") (result i32)
+    (call_ref (i32.const 10)
+              (i32.const 13)
+    )
+  )
+)
+(;; STDERR ;;;
+out/test/parse/expr/bad-callref-int32.txt:6:6: error: type mismatch in call_ref, expected reference but got [i32]
```
takikawa, line 12: It looks like both of these parse failure tests (this one and `bad-callref-null.txt`) are checking type errors, so maybe they should go in `wabt/test/typecheck/` instead?

takikawa (approved):

takikawa:
LGTM other than my one comment, but I'm not an approving reviewer so I think you'll need another approval too (also apparently you need an approval to have the CI tests run for this PR https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks#approving-workflow-runs-on-a-pull-request-from-a-public-fork)

test/parse/expr/bad-callref-int32.txt
dbezhetskov: sure, I've moved them

dbezhetskov:
Hi @kripken and @sbc100, could you review these changes changes please?

src/binary-reader.cc
```diff
@@ -328,7 +328,7 @@ Result BinaryReader::ReadRefType(Type* out_value, const char* desc) {
   uint32_t type = 0;
   CHECK_RESULT(ReadS32Leb128(&type, desc));
   *out_value = static_cast<Type>(type);
-  ERROR_UNLESS(out_value->IsRef(), "%s must be a reference type", desc);
+  ERROR_UNLESS(out_value->IsFuncOrExternRef(), "%s must be a reference type", desc);
```
sbc100, line 331: Why not just leave this as `IsRef`?   Function refs are also refs, right?

src/feature.def
```diff
@@ -22,16 +22,17 @@
...
-WABT_FEATURE(annotations,      "annotations",             false,   "Custom annotation syntax")
-WABT_FEATURE(gc,               "gc",                      false,   "Garbage collection")
-WABT_FEATURE(memory64,         "memory64",                false,   "64-bit memory")
+WABT_FEATURE(exceptions,          "exceptions",              false,   "Experimental exception handling")
+WABT_FEATURE(mutable_globals,     "mutable-globals",         true,    "Import/export mutable globals")
+WABT_FEATURE(sat_float_to_int,    "saturating-float-to-int", true,    "Saturating float-to-int operators")
+WABT_FEATURE(sign_extension,      "sign-extension",          true,    "Sign-extension operators")
+WABT_FEATURE(simd,                "simd",                    false,   "SIMD support")
+WABT_FEATURE(threads,             "threads",                 false,   "Threading support")
+WABT_FEATURE(function_references, "function-references",     false,   "Typed function references")
```
sbc100, line 31: Strange that github can't do better with this diff.  I assume this is just adding the one line and the rest is indentation?

src/ir.h
```diff
@@ -578,6 +579,16 @@ class ReturnCallIndirectExpr : public ExprMixin<ExprType::ReturnCallIndirect> {
...
 };
 
+class CallRefExpr : public ExprMixin<ExprType::CallRef> {
+ public:
+  explicit CallRefExpr(const Location &loc = Location())
+      : ExprMixin<ExprType::CallRef>(loc) {}
+
+  // This field is setup only during Validate phase,
+  // so keep that in mind when you use it.
+  Var function_type_index;
```
sbc100, line 589: So there is no immediate in the binary?  Its just implied?

src/shared-validator.h
```diff
@@ -124,6 +124,7 @@ class SharedValidator {
   Result EndBrTable(const Location&);
   Result OnCall(const Location&, Var func_var);
   Result OnCallIndirect(const Location&, Var sig_var, Var table_var);
+  Result OnCallRef(const Location&, Index* function_type_index = nullptr);
```
sbc100, line 127: Why does the second arg need to have a default here?  I only see a single call site and it passes a non-null arg here?    Could we remove this and the check in OnCallRef?   In fact could we pass by ref?

src/type.h
```diff
@@ -55,18 +55,30 @@ class Type {
...
   Type() = default;  // Provided so Type can be member of a union.
-  Type(int32_t code) : enum_(static_cast<Enum>(code)) {}
-  Type(Enum e) : enum_(e) {}
+  Type(int32_t code, Nullable nullable = Nullable::Yes)
+      : enum_(static_cast<Enum>(code)), nullable(nullable) {}
+  Type(Enum e) : enum_(e), nullable(Nullable::Yes) {}
   operator Enum() const { return enum_; }
 
   bool IsRef() const {
+    return enum_ == Type::ExternRef || enum_ == Type::FuncRef || IsIndex();
```
sbc100, line 70: Why does the definition of `IsRef` need to change here?

test/dump/call_ref.txt
```diff
@@ -0,0 +1,108 @@
+;;; TOOL: run-objdump
+;;; ARGS0: -v --enable-function-references
+
+(module
+  (func (export "main") (result i32)
+    i32.const 10
+    ref.func $foo
```
sbc100, line 7: What would it look like to receive a `ref.func` as an argument?   Or is that coming in the next phase/PR?

sbc100 (commented):
Sweet!   Looking good.   Thanks for working on this.

src/feature.def
dbezhetskov: Exactly, this is just indentation + one new line

src/ir.h
dbezhetskov: Yes, according to this https://github.com/WebAssembly/function-references/blob/master/proposals/function-references/Overview.md#instructions-1 `call_ref` has no immediates, so we have to determine it's type from the context. 

I could use optional type here to express possible empty case more cleary but it looks like `Var` already has the value for indicating the invalid state.

src/type.h
dbezhetskov: According to the spec (https://github.com/WebAssembly/function-references/blob/master/proposals/function-references/Overview.md#reference-types) we have new definition of the ref type - now it is a heaptype + nullability flag, in turn heaptype ::= <typeidx> | func | extern.

src/binary-reader.cc
dbezhetskov: Thanks, in this case we indeed can use just IsRef

src/shared-validator.h
dbezhetskov: Removed default value, but preserved pointer because it is little bit readable at the call site

src/type-checker.cc
```diff
@@ -741,7 +766,7 @@ Result TypeChecker::OnRefNullExpr(Type type) {
 Result TypeChecker::OnRefIsNullExpr() {
   Type type;
   Result result = PeekType(0, &type);
-  if (!type.IsRef()) {
+  if (!type.IsFuncOrExternRef()) {
```
sbc100, line 769: Why can't OnRefIsNullExpr work for refs?   I that is to say why not use `IsRef` here?

src/type-checker.cc
```diff
@@ -790,7 +815,7 @@ Result TypeChecker::OnSelect(const TypeVector& expected) {
   result |= PeekType(1, &type1);
   result |= PeekType(2, &type2);
   if (expected.empty()) {
-    if (type1.IsRef() || type2.IsRef()) {
```
sbc100, line 793: Ditto

sbc100 (commented):
I wonder if this change can land without the addition of `IsFuncOrExternRef` and the changes to `IsNullable`?     (Those could then be added as needed and/or as followups?).

test/dump/call_ref.txt
dbezhetskov: Some new tokens need to be introduced to support high order functions so let's move that in the next PR

src/type-checker.cc
dbezhetskov: fixed

src/type-checker.cc
dbezhetskov: fixed

dbezhetskov:
> I wonder if this change can land without the addition of `IsFuncOrExternRef` and the changes to `IsNullable`? (Those could then be added as needed and/or as followups?).

Yep, let's move this functionality to the next patches because now we aren't able to test it.

sbc100 (approved):
Great!   I like how this change is now very precise.

dbezhetskov:
Hi @sbc100, binji seems unavailable, could we assign another person for review? Or could we land this?

sbc100:
I think we can can go ahead and land this.

BTW, what is the status of the tests in the funcref proposal repo?    I see they exist at  https://github.com/WebAssembly/testsuite/tree/master/proposals.   I guess we should using them here in wabt once we get a few more things working?

SinEstresthemainone:
I accidently forked this to my main pull or whatever it's called and I was trying to figure out how to close it. sorry for the trouble. I was looking into debugging my chromebook.

PR #1693, opened 2021-07-14
Branch hinting

yuri91:
This PR adds basic support for handling the custom section "branchHints" (see Branch Hinting proposal).

I added code to parse the section in the binary reader, and call new appropriate visitor methods with the parsed information.

I also Implemented the new visitor methods for the binary reader logger and the binary reader objdump.

So now it is possible to use wasm-objdump to output the contents of the section in a readable format, and wasm-validate will complain if the section is malformed.

The branchHints section parsing is done conditionally with a feature flag `--enable-branch-hinting`.

I also added some tests.

A possible further improvement would be to mark the hinted branches as such  in the code output of wasm-objdump/wasm-decompile.

### DIFF
--- a/src/binary-reader-logging.cc
+++ b/src/binary-reader-logging.cc
@@ -664,6 +664,12 @@ Result BinaryReaderLogging::OnComdatEntry(ComdatType kind, Index index) {
   return reader_->OnComdatEntry(kind, index);
 }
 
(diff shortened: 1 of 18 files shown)


yuri91:
I am closing this in favor of https://github.com/WebAssembly/wabt/pull/1724 , which add support for the more general "code annotations" ( branch hinting will soon be based on them)

PR #1695, opened 2021-08-04
Support function references in parameters and results

dbezhetskov:
This PR support usage of (ref $foo) in parameters and results of functions and blocks.

Technical notes:
- To represent symbolic types like (ref $foo) I need to extend `class Type` with `std::string` or `Var` maybe,
but `class Type` is `POD` because we use it in many places as a field of unions, so I've externally extended it with mapping from parameter/result indices to its names.
- It is not possible to use reference types in locals. I've done this locally but decided add it in the next PR not to blow this one.
- Now `(ref.func $foo)` pushes type index instead of function index, and now it is exactly matching the spec https://github.com/WebAssembly/function-references/blob/master/proposals/function-references/Overview.md#functions


dbezhetskov:
Hi Sam, could you review please? @sbc100

src/type.h
```diff
@@ -56,14 +57,20 @@ class Type {
...
+  Type(int32_t code) : enum_(static_cast<Enum>(code)), index_(UINT32_MAX) {}
+  Type(Enum e) : enum_(e), index_(UINT32_MAX)  {}
+  Type(Enum e, Index index) : enum_(e), index_(index) { assert(e == Enum::Reference); }
   operator Enum() const { return enum_; }
 
   bool IsRef() const {
     return enum_ == Type::ExternRef || enum_ == Type::FuncRef;
   }
 
+  bool IsFunctionReference() const {
```
sbc100, line 69: Its a little confusing to be me that `IsRef` would not return true for `Type::Reference` or that `IsFunctionReference` returns false for `Type::FuncRef`

Can you explain the different between the existing `Type::FuncRef` and this new `Type::Reference`.  Also aren't `Struct` and `Array` also a form of reference?    

Sorry I'm a little rusty on this stuff.

dbezhetskov: Ah, I understand your confusion.
So, in typed function references proposal `funcref` and `externref` become `(ref null func)` and `(ref null extern)` respectively.

In WABT it will look like this `Type{Type::Reference, -0x10}` and `Type{Type::Reference, -0x11}`
but without TFR proposal it should look like `Type(Type::FuncRef)` and `Type(Type::ExternRef)`.
So, to support both variant I've updated `IsRef()` so that it returns `true` for both cases.

`IsReferenceWithIndex()` needed to distinguish between `FuncRef` and `ExternRef` because sadly
`(ref null func)` and `(ref null index)` have different encodings patterns and usage.

dbezhetskov:
Gentle ping on this, @sbc100

dbezhetskov:
Gentle ping on this, @sbc100, @tlively, @aardappel

aardappel:
probably @tlively for the function reference specifics.. otherwise looks nice to me.

src/ir.h
```diff
@@ -188,6 +188,13 @@ struct FuncSignature {
   TypeVector param_types;
   TypeVector result_types;
 
+  // Some types can have names, for example (ref $foo) has type $foo.
+  // So to use this type we need to translate its name into
+  // a proper index from the module type section.
+  // This is the mapping from parameter/result index to its name.
+  std::unordered_map<uint32_t, std::string> param_type_names;
+  std::unordered_map<uint32_t, std::string> result_type_names;
```
tlively, line 196: I can't tell from this comment why the signature needs to store type names, since there is no signature involved in `(ref $foo)`. Shouldn't all type names already be stored somewhere else already?

src/shared-validator.h
```diff
@@ -180,11 +181,14 @@ class SharedValidator {
...
-    FuncType(const TypeVector& params, const TypeVector& results)
-        : params(params), results(results) {}
+    FuncType(const TypeVector& params,
+             const TypeVector& results,
+             Index type_index)
+        : params(params), results(results), type_index(type_index) {}
 
     TypeVector params;
     TypeVector results;
+    Index type_index;
```
tlively, line 191: Why does `FuncType` need to store its own index now if it didn't need to before?

tlively (commented):
I'm not familiar with WABT's code, so these might be dumb questions. But generally the syntax and semantics seem right to me from what I can tell. I will defer to @aardappel's judgement on the code.

src/ir.h
dbezhetskov: actually we should store type names because at the moment of parsing we may not know all the types.
We will resolve these type names into indices after the parsing completes and then we don't need to use these maps at all

src/shared-validator.h
dbezhetskov: previously we don't care about the function type indices because care only about function signatures.
You can get function signature via function index:
https://github.com/WebAssembly/wabt/blob/9aac0a28896a714b698ab489a473b9bc05deb7c3/src/shared-validator.h#L293

But there is no way to get function type index via function index and so, I preserve this in FuncType.

sbc100:
I think this change could be useful with updating to the latest testsuite.. would be great to get it landed.  Apologies for taking so long to review it.

src/type.h
```diff
@@ -43,27 +43,34 @@ class Type {
...
     I8U = 4,   // Not actually specified, but used internally with load/store
     I16U = 6,  // Not actually specified, but used internally with load/store
     I32U = 7,  // Not actually specified, but used internally with load/store
   };
 
   Type() = default;  // Provided so Type can be member of a union.
-  Type(int32_t code) : enum_(static_cast<Enum>(code)) {}
-  Type(Enum e) : enum_(e) {}
+  Type(int32_t code) : enum_(static_cast<Enum>(code)), index_(UINT32_MAX) {}
+  Type(Enum e) : enum_(e), index_(UINT32_MAX) {}
```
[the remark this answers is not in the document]
dbezhetskov: Huh, I thought about that, actually we can't because it will break any other code that uses `Type` as a member of union so we can't add any default initialization without breaking a lot of existing code.

src/type-checker.cc
```diff
@@ -206,6 +206,12 @@ Result TypeChecker::CheckType(Type actual, Type expected) {
   if (expected == Type::Any || actual == Type::Any) {
     return Result::Ok;
   }
+
+  if (expected == Type::Reference && actual == Type::Reference) {
+    return expected.GetFunctionIndex() == actual.GetFunctionIndex()
```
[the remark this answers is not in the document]
dbezhetskov: I think no. Actually the name of this method is misleading.
Here we have two ref types `(ref $type_1)` and `(ref $type_2)` and so, we should compare them by their indices.

`GetFunctionTypeIndex` returns index of a type in the module by a given function index.
So, for example:

```
(module
    (type $t1 (ref i32))
    (type $t2 (func (param i32) (result i32))))
    (type $t3 (func (result i32))))
    ...
)
```
Here we have 2 functions and three types, `GetFunctionTypeIndex(0)` will return `index($t2) == 1`.

BTW, I've renamed this method with `GetReferenceIndex`.

src/ir.h
dbezhetskov: > Is there a precedent for that type of thing in the codebase already?

Sure, for example names of locals handled the same way:
```
(module
  (func (param $x i32) (result i32)
    (local.get $x)
  )
)
```
In the example above we have `$x` name for the first parameter and in current design we store these mapping `$x` -> `0` here: https://github.com/WebAssembly/wabt/blob/d61a78988001cdc6d364d91cfb9b2088b9ae2798/src/ir.h#L767

We set it up inside
```
Result WastParser::ParseFuncSignature(FuncSignature* sig,
                                      BindingHash* param_bindings)
```
And this information also actually needed only for parsing, but here we are.

BTW, the ideal solution is to change Type representation and make it like `struct Var` to handle string names and indices but because of PODness of `Type` we can't do much without some refactoring.

dbezhetskov:
Hi @sbc100 , sorry for the long replay. I've rebased the code and updated the tests and provided feedback about the code, so when you have time, please review it

src/shared-validator.cc
```diff
@@ -48,15 +48,17 @@ Result SharedValidator::OnFuncType(const Location& loc,
...
     result |=
         PrintError(loc, "multiple result values not currently supported.");
   }
-  func_types_.emplace(num_types_++,
-                      FuncType{ToTypeVector(param_count, param_types),
-                               ToTypeVector(result_count, result_types)});
+  func_types_.emplace(
+      num_types_++,
+      FuncType{ToTypeVector(param_count, param_types),
+               ToTypeVector(result_count, result_types), type_index});
```
sbc100, line 61: I assume this formatting is what clang-format recommends?

sbc100 (approved):
lgtm!

I assume the plan is to integrate the function-types testsuite as a followup?

src/shared-validator.cc
dbezhetskov: yep

dbezhetskov:
> I assume the plan is to integrate the function-references testsuite as a followup?

I have a wip patch for support of references in locals https://github.com/dbezhetskov/wabt/commit/44bf83aa1fd95d0a592c3a34fd5909460a516216
with it we can integrate spec tests.

PR #1702, opened 2021-08-17
Fix a range of data type warning in gcc

clover2123:
This patch fixes a limited range of data type warning.

Warning description: 
In some part of `WABT_PRINTF_TYPE_CODE` calls, an unsigned value (uint8_t type) is passed. This value is checked if it is signed or not, but this comparison is always false due to its type. So, it triggers a range of data type warning (-Wtype-limits) in gcc compilation.

Signed-off-by: HyukWoo Park <hyukwoo.park@samsung.com>


clover2123:
Hi @sbc100 
Would you please take a look at this patch?
(its my first contribution to wabt, so I first need to run actions)

src/binary-reader.cc
```diff
@@ -2182,7 +2182,7 @@ Result BinaryReader::ReadTypeSection(Offset section_size) {
       uint8_t type;
       CHECK_RESULT(ReadU8(&type, "type form"));
       ERROR_UNLESS(type == 0x60, "unexpected type form (got " PRItypecode ")",
-                   WABT_PRINTF_TYPE_CODE(type));
+                   "", static_cast<int32_t>(type));
```
sbc100, line 2185: I think you can just replace `PRItypecode` here is `%d` and drop the cast.

clover2123: Thanks for the review. I updated it following your comment.

sbc100:
Did you mean to push more changes here?

clover2123:
@sbc100 
I've just fixed a lint error. Thanks:)

PR #1712, opened 2021-09-17
SIMD is now phase 5, enable it by default

ngzhian:


ngzhian:
Hm, so if I mark a flag as enabled, I have to remove all the flags from all the tests it looks like. Should I do that? Or can I change feature.cc to nop.

kripken:
Do we have a policy of enabling things by default once they are in phase 5? I might have missed something.

In general I think we should have a uniform policy across the toolchain space, if possible. I'm not sure what the policy should be, but it seems like there are good arguments both ways (for example, in favor: avoid users needing to enable it all the time; against: wasm files will validate that might not validate in all production VMs).

ngzhian:
Prior proposals have the flag defaulted when they are merged into the spec (see e88bc660cf89ca84dccda358cfbadc8a7c2bc935). I don't know if we want to make the distinction between phase 5 and merged into spec. There is a `--disable-simd` flag once we default to enabling simd, so users can still opt out if needed.

kripken:
Interesting, thanks @ngzhian ...

We don't do this in LLVM, emscripten, or binaryen. But perhaps those are more producers of wasm, and wabt is an inspector.

Anyhow, I don't mean to block this!

ngzhian:
> We don't do this in LLVM, emscripten, or binaryen. But perhaps those are more producers of wasm, and wabt is an inspector.

By that, you mean `-msimd128` will always exist as a flag? Yea I think keeping that around for LLVM/Emscripten/Binaryen is a good idea, give users a choice if they want to emit those instructions or not. Another data point is that V8 has SIMD enabled by default (no flag), so perhaps is a choice between producers of SIMD instructions and consumers.

No problem, thank you for your perspective! Let's see what @sbc100 thinks.

sbc100:
I think we can follow the precedent here and have wabt (like v8) enable all phase 5 proposals by default.   It think it makes sense to have a different policy for producers such as emscripten and llvm.  

If wabt was capable of generating new instructions (like binaryen does when optimizing) then we would probably re-consider.. but I don't think it is.

sbc100 (commented):
Are there any places where we can/should remove `--enable-simd` ?   It looks like at least `test/update-spec-tests.py` maybe?

sbc100:
test/wasm2c/bad-enable-feature.txt might also need updating?

ngzhian:
> Are there any places where we can/should remove `--enable-simd` ? It looks like at least `test/update-spec-tests.py` maybe?

I'll do a grep and find the relevant places. At least in all the test files we will need to remove it from ARGS.

ngzhian:
A lot of the changes are in the expected output for tests, where the line number change due to removing an entire line from `;;; ARGS:`. If you prefer I can change it to be an empty line `;;;`, then it won't change as many files.

sbc100 (approved):
No, this looks good.

PR #1713, opened 2021-09-17
Handle some SIMD opcodes in opcodecnt

ngzhian:
Most of the SIMD instructions were already transparently handled, some
(like loadlane and v128.const) don't fit in the existing callbacks, so
we need to implement (and override) the default implementation.


ngzhian:
Rebased after removing `--enable-simd`.

ngzhian:
Ping @sbc100

PR #1714, opened 2021-09-20
wasm-decompile: Avoid trailing whitespace

relrelb:


src/decompiler.cc
```diff
@@ -149,12 +149,13 @@ struct Decompiler {
     auto& right = args[1];
     BracketIfNeeded(left, precedence);
     BracketIfNeeded(right, precedence);
-    auto width = infix.size() + left.width() + right.width();
+    auto width = infix.size() + left.width() + right.width() + 2;
     if (width < target_exp_width && left.v.size() == 1 && right.v.size() == 1) {
-      return Value{{left.v[0] + infix + right.v[0]}, precedence};
+      return Value{{left.v[0] + " " + infix + " " + right.v[0]}, precedence};
```
aardappel, line 154: can use `cat` here too

src/decompiler.cc
```diff
@@ -474,7 +475,7 @@ struct Decompiler {
         auto& se = *cast<StoreExpr>(n.e);
         LoadStore(args[0], n.children[0], se.offset, se.opcode, se.align,
                   se.opcode.GetParamType2());
-        return WrapBinary(args, " = ", true, Precedence::Assign);
+        return WrapBinary(args, "=", true, Precedence::Assign);
```
aardappel, line 478: So there were no calls to `WrapBinary` that don't want the spacing?

aardappel (commented):
Thanks, looks like a nice improvement!

src/decompiler.cc
relrelb: Yup, all 3 calls had spaces before and after.

src/decompiler.cc
relrelb: Done.

aardappel (approved):

aardappel:
Thanks!

PR #1719, opened 2021-10-04
Favicons for the demo pages

pouyakary:
Hey awesome wabt team. I've been creating a webassembly assembler and found the commented binary output of the wat2wasm the best source for learning. (thanks for implementing it) and then the wat2wasm demo page became something I wanted to keep in my favorites. I just didn't like its lack of look there:

Screen Shot 1400-07-12 at 16 05 44

So I added some new stuff into it. I added favicons and colors. Now it looks like this in the upcoming safari:

Screen Shot 1400-07-12 at 16 07 00

Has favicons + iOS/Android/Chrome... 

IMG_6525

And looks much better in the favorites:

IMG_6526


keithw:
@pouyakary Thank you (belatedly) for this PR! I don't know what happened when it was submitted (I suspect either nobody had the cycles to review it, or we may also not have wanted to add all these image files to the WABT repository). It looks like we are probably not going to add these images. I think if there's a way we can do this by linking to the Wasm logos instead of including them in our own repo, that might be a smaller pill to swallow... in the meantime I'll close this for housekeeping.

pouyakary:
what a shame :(

PR #1725, opened 2021-10-11
Fix ub access of 0-size vector in data segment

ngzhian:
If the module has no memory, the we try to copy the passive data segment
to memories[0], and memories is a size 0 vector, which is UB.


sbc100:
How about a test for this?

PR #1726, opened 2021-10-11
Fix crashes in parser when f32.const is given a nan expression

ngzhian:
The parser assumes that f32.const will always be followed by a literal,
and crashes if it is followed by a nan expression when it is not
allowed.

### DIFF
--- a/src/wast-parser.cc
+++ b/src/wast-parser.cc
@@ -2363,7 +2363,13 @@ Result WastParser::ParseF32(Const* const_, ConstType const_type) {
     const_->set_f32(expected);
     return Result::Ok;
   }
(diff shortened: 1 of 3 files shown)


sbc100:
> > These tests seems like they belong upstream maybe? I wonder if we should open bugs in the spec tests when we find things like this?
> 
> These are crashes in wabt's parser, nothing to do with spec interpreter, so shouldn't need to be upstream I think.

But doesn't the upstream spec tests have test for invliad literals?   i.e. Doesn't  it have checks that certain things are un-parsable as a literal?

ngzhian:
Fyi, PR here https://github.com/WebAssembly/spec/pull/1373

Issue #1737, opened 2021-10-15
wasm2c no longer passes many spec tests


sbc100:
Many test fail because the recent spec tests now require multi-table and reference-type support.  I was forced to delete the following tests for that reason:

- test/wasm2c/spec/br_table.txt   
- test/wasm2c/spec/call_indirect.txt
- test/wasm2c/spec/elem.txt
- test/wasm2c/spec/exports.txt
- test/wasm2c/spec/imports.txt
- test/wasm2c/spec/table.txt 
- test/wasm2c/spec/linking.txt
- test/wasm2c/spec/select.txt

The new version of `test/wasm2c/spec/data.txt` fails because somehow the bounds checks don't seem to work for zero-sized memories.  This test tails to trap:

```
(assert_trap                                                                        
  (module                                                                              
    (memory 0)                                                                   
    (data (i32.const 1))                                                         
  )                                                                              
  "out of bounds memory access"                                                  
)   
```

PR #1738, opened 2021-10-15
Update testsuite

sbc100:
bulk-memory-operations and reference-types were completely
removed from the upstream testsuite becuase there were
merged into the upstream spec:

https://github.com/WebAssembly/testsuite/pull/44

In order to land this I had to disable several spec tests
under wasm2c because it lacks support for mutli-table and
reference types.  I filed #1737 to track this.


sbc100:
> Is there a way to mark a test as FAILING rather than deleting it? Seems like we might accidentally forget to add those tests back in.

Not that I know of no.

sbc100:
> > > > Is there a way to mark a test as FAILING rather than deleting it? Seems like we might accidentally forget to add those tests back in.
> > > 
> > > 
> > > Not that I know of no.
> > 
> > 
> > Thats why I open a bug and listed all the tests I was forced to delete.
> 
> Dk if this feature will be useful, but yea that tracking bug works, thanks!

Actually I forgot to run `./test/update-spec-tests.py` during this change so its not ready to land yet.

Also `./test/update-spec-tests.py`  recreates all the removed tests so its pretty clear which ones are missing and easy to re-create them once we add these features to wasm2c.

PR #1739, opened 2021-10-15
Fix crash when function counts mismatch

ngzhian:
This can happen if we don't stop on first error, and we get a malformed
module where the func counts don't match.

It's hard to write a test for this, since the kStopOnFirstError is fixed
(not set by command line), but this case is quite easy for fuzzers to
catch.


ngzhian:
> I would have thought you could write a tests in test/binary, no? (lgtm either way)

I tried, but I would never hit the crash, because `kStopOnFirstError = true`, so I hit some other error before hitting the crash.

PR #1751, opened 2021-11-04
Add support for multi-memory proposal

yhdengh:
This PR adds support for multi-memory proposal.

* Added `multi_memory` as a feature. Added memidx(s) to IR representations of memory instructions. 
* Added support for parsing and writing memidx in both binary and text format, and also wasm2c.
* Updated testsuite to https://github.com/WebAssembly/testsuite/commit/f1eaa436cc513b94c62c5ad6220e8ddc8673490f which includes testsuite for multi-memory proposal. Rebased tests. Skipped the test directory for wasm2c as suggested by #1737


yhdengh:
@sbc100 @kripken Would you be able to take a look at this?

src/binary-reader.cc
```diff
@@ -944,8 +956,22 @@ Result BinaryReader::ReadFunctionBody(Offset end_offset) {
...
-        CALLBACK(OnLoadExpr, opcode, alignment_log2, offset);
-        CALLBACK(OnOpcodeUint32Uint32, alignment_log2, offset);
+        Index memidx;
+        if (alignment_log2 >> 6) {
+          ERROR_IF(!options_.features.multi_memory_enabled(),
+              "multi_memory not allowed");
+          CHECK_RESULT(ReadMemidx(&memidx, "store memidx"));
+          alignment_log2 = alignment_log2 & ((1 << 6) - 1);
+        } else {
+          memidx = 0;
```
sbc100, line 966: Skip this else and just initialize memidx to zero above?

src/binary-reader.cc
```diff
@@ -963,28 +989,57 @@ Result BinaryReader::ReadFunctionBody(Offset end_offset) {
...
+          ERROR_UNLESS(reserved == 0, "memory.grow reserved value must be 0");
+          CALLBACK(OnMemoryGrowExpr, reserved);
+          CALLBACK(OnOpcodeUint32, reserved);
+          break; 
+        } else {
+          Index memidx;
+          CHECK_RESULT(ReadMemidx(&memidx, "memory.grow memidx"));
+          CALLBACK(OnMemoryGrowExpr, memidx);
+          CALLBACK(OnOpcodeUint32, memidx);
+          break;
```
sbc100, line 1041: Can we above this if/else here and instead just do 

```
if (!options_.features.multi_memory_enabled())  {
  ERROR_UNLESS(reserved == 0, "memory.grow memory index must be 0");
}
```

sbc100: Same for MemoryInit and all the other memory operations below.  Hopefully we can avoid duplication that way?

src/binary-writer.cc
```diff
@@ -863,27 +894,54 @@ void BinaryWriter::WriteExpr(const Func* func, const Expr* expr) {
...
-    case ExprType::MemoryFill:
+    case ExprType::MemoryFill: {
+      Index memidx =
+          module_->GetMemoryIndex(cast<MemoryFillExpr>(expr)->memidx);
       WriteOpcode(stream_, Opcode::MemoryFill);
-      WriteU32Leb128(stream_, 0, "memory.fill reserved");
+      if (options_.features.multi_memory_enabled()) {
+        WriteU32Leb128(stream_, memidx, "memory.fill memidx");
+      } else {
+        WriteU32Leb128(stream_, memidx, "memory.fill reserved");
```
sbc100, line 904: I think think we need this if/else here.  Just `WriteU32Leb128(stream_, memidx, "memory.fill memidx");` should be fine.  Same below..

src/c-writer.cc
```diff
@@ -1069,11 +1068,16 @@ void CWriter::WriteDataInitializers() {
...
 
   Write(Newline(), "static void init_memory(void) ", OpenBrace());
-  if (memory && module_->num_memory_imports == 0) {
-    uint32_t max =
-        memory->page_limits.has_max ? memory->page_limits.max : 65536;
-    Write("wasm_rt_allocate_memory(", ExternalPtr(memory->name), ", ",
-          memory->page_limits.initial, ", ", max, ");", Newline());
+  if (module_->memories.size() > module_->num_memory_imports) {
+    Index memory_idx = module_->num_memory_imports;
+    for (Index i = memory_idx; i < module_->memories.size(); i++)
```
sbc100, line 1073: Opening brace on same line as `for`.   Maybe run `git clang-format origin/main`

src/interp/binary-reader-interp.cc
```diff
@@ -1229,29 +1231,31 @@ Result BinaryReaderInterp::OnLocalTeeExpr(Index local_index) {
 
 Result BinaryReaderInterp::OnLoadExpr(Opcode opcode,
                                       Address align_log2,
-                                      Address offset) {
-  CHECK_RESULT(validator_.OnLoad(loc, opcode, GetAlignment(align_log2)));
-  istream_.Emit(opcode, kMemoryIndex0, offset);
+                                      Address offset,
+                                      Index memidx = 0) {
```
sbc100, line 1235: Is the `= 0` needed here and below?   i.e. Are there callsites that don't pass memidx?

src/ir.h
```diff
@@ -501,11 +528,20 @@ class SimdShuffleOpExpr : public ExprMixin<ExprType::SimdShuffleOp> {
 
 template <ExprType TypeEnum>
 class VarExpr : public ExprMixin<TypeEnum> {
- public:
-  VarExpr(const Var& var, const Location& loc = Location())
+  public:
+    VarExpr(const Var& var, const Location& loc = Location())
```
sbc100, line 532: I think the old indentation is what clang-format is setup to do for this project.

src/shared-validator.cc
```diff
@@ -895,11 +895,16 @@ Result SharedValidator::OnIf(const Location& loc, Type sig_type) {
...
+                               Address alignment,
+                               Var memidx) {
   Result result = Result::Ok;
   MemoryType mt;
   expr_loc_ = &loc;
-  result |= CheckMemoryIndex(Var(0, loc), &mt);
+  if (options_.features.multi_memory_enabled()) {
+    result |= CheckMemoryIndex(memidx, &mt);
+  } else {
+    result |= CheckMemoryIndex(Var(0,loc), &mt);
```
sbc100, line 906: Can't we just us `CheckMemoryIndex(memidx, &mt);` in both cases?  Won't memidx above be zero anyway for the single-memory case?

Same below.

src/wast-parser.cc
```diff
@@ -955,6 +955,27 @@ bool WastParser::ParseAlignOpt(Address* out_align) {
...
+    if (!options_->features.multi_memory_enabled()) {
+      Error(loc, "Specifying memory variable is not allowed");
+      return Result::Error;
+    }
+    EXPECT(Lpar);
+    EXPECT(Memory);
+    CHECK_RESULT(ParseVar(out_memidx));
+    EXPECT(Rpar);
+    return Result::Ok;
+  } else {
```
sbc100, line 970: Avoid else after return

Either move the `return Result::Ok;` out the top level or remove the `else` completely.

src/wast-parser.cc
```diff
@@ -1772,6 +1793,39 @@ Result WastParser::ParsePlainInstrVar(Location loc,
...
+template <typename T>
+Result WastParser::ParseMemoryInstrVar(Location loc,
+                                       std::unique_ptr<Expr>* out_expr) {
+  Var memidx;
+	Var var;
+  if (PeekMatchLpar(TokenType::Memory)) {
+    if (!options_->features.multi_memory_enabled()) {
+      Error(loc, "Specifying memory variable is not allowed");
+      return Result::Error;
+    } else {
```
sbc100, line 1805: No else after return.

src/wast-parser.cc
```diff
@@ -1772,6 +1793,39 @@ Result WastParser::ParsePlainInstrVar(Location loc,
...
+      CHECK_RESULT(ParseMemidx(loc, &memidx));
+      CHECK_RESULT(ParseVar(&var));
+      out_expr->reset(new T(var, memidx, loc));
+      return Result::Ok;
```
sbc100, line 1809: No need for this return since the it already exists at the outer level. 

Same for the blocks below.

src/wat-writer.cc
```diff
@@ -363,7 +367,7 @@ void WatWriter::WriteVar(const Var& var, NextChar next_char) {
   } else {
     WriteName(var.name(), next_char);
   }
-}
+}
```
sbc100, line 370: revert this?

sbc100 (commented):
This looks great!  Thanks for working on it.

Quite a few comments but mostly minor stuff.

src/binary-reader-ir.cc
```diff
@@ -166,17 +166,18 @@ class BinaryReaderIR : public BinaryReaderNop {
   Result OnIfExpr(Type sig_type) override;
   Result OnLoadExpr(Opcode opcode,
                     Address alignment_log2,
-                    Address offset) override;
+                    Address offset,
+                    Index memidx) override;
```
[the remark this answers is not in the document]
yhdengh: memidx appears right after the opcode in text format, but it appears after alignment and offset in binary format. I could update it such that for headers related to binary format, memidx comes last and for headers related to text format, memidx comes first, but I think it would be better if the ordering is the same for all headers.

src/binary-reader.cc
yhdengh: We kept these for similar reasons as OnOpcodeUint32Uint32Uint32. There are spec tests from WebAssembly/testsuite that require the module to be invalidated if the encoding of 0 memidx is anything other than 0x00. To remove this, we would need to require WebAssembly/testsuite to remove those tests.

src/binary-writer.cc
yhdengh: This was kept for the same reason as OnOpcodeUint32Uint32Uint32.

src/shared-validator.cc
yhdengh: This was kept for the same reason as OnOpcodeUint32Uint32Uint32. For load and store expressions, it is fine to only keep CheckMemoryIndex(memidx). However, for other memory instructions (memory.copy, memory.fill, etc.), only keeping CheckMemoryIndex(memidx) makes the program failed to pass some tests because the error messages are different.

src/binary-reader-ir.cc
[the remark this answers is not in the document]
sbc100: Agreed that they should all be the same, but I also think we should follow the text format and have memidx  come right after the opcode in all the APIs.   The fact that it comes last in the binary format is kind of an accident of history whereas the text format rightly puts it first where I think it makes more sense, no?

yhdengh:
Thanks for all the feedback! I have updated commits in the pull request to address some of those.

For the if/else duplications, we kept them so that wabt passes all the existing tests when the multi-memory feature is not enabled. Most of them could be removed if we change existing tests for multi-memory disabled cases. Is that the preferred way to approach it? 

The only exception is for reserved memidx of memory instructions in binary format, which requires a change in WebAssembly/testsuite.

src/binary-reader.cc
```diff
@@ -944,8 +956,22 @@ Result BinaryReader::ReadFunctionBody(Offset end_offset) {
...
+          alignment_log2 = alignment_log2 & ((1 << 6) - 1);
+        } else {
+          memidx = 0;
+        }
+        
+        CALLBACK(OnLoadExpr, opcode, alignment_log2, offset, memidx);
+        if (memidx == 0) {
+          CALLBACK(OnOpcodeUint32Uint32, alignment_log2, offset);
+        } else {
+          CALLBACK(OnOpcodeUint32Uint32Uint32, alignment_log2, offset, memidx);
```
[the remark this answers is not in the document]
sbc100: Oh I see.. since objdump doesn't know it can elide the memidx when dumping the instruction.     I guess this is OK for now and we can look into making objdump smart as a potential followup.

src/binary-writer.cc
sbc100: This doesn't make any sense to me since the final argument here is just the debug string which wabt prints.     I think its fine if this changes all the `memory.fill reserved` to `memory.fill memidx` in the test output if that is what you are worried about.

At the very least we should replace this with `WriteU32Leb128(stream_, memidx, options_.features.multi_memory_enabled() ? "memory.fill memidx" : "memory.fill reserved")` I think.

src/shared-validator.cc
sbc100: Interesting.  I'm curious to see what how the error message are different.  It could be that we simple accept the new error message and rebase the test expectations?  Can you given an example of the differences.

src/binary-reader-ir.cc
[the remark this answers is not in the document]
yhdengh: The binary format uses bit 6 of alignment to indicate whether the current operation has a memidx immediate. Bit 6 is 1 if and only if there is a memidx immediate. Therefore, parser has to parse alignment first to figure out whether it is going to parse memidx, and that's why memidx is placed at the last. But I agree that it makes more logical sense to put memidx right after opcode. I will update the APIs.

src/binary-reader.cc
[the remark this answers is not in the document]
yhdengh: Here is an example of the test output when we only keep OnOpcodeUint32Uint32Uint32.
>   test/opcodecnt/basic.txt                                                       
>   STDOUT MISMATCH:
>   --- expected
>   +++ actual
>   @@ -12,7 +12,7 @@
>    end: 2
>    br 0: 2
>    local.get 1: 2
>   -i32.load 2, 0: 1
>   +i32.load 2, 0, 0: 1
>    i32.const 0 (0x0): 1
>    i32.const 1 (0x1): 1
>    i32.const 2 (0x2): 1

src/binary-writer.cc
yhdengh: Here is an example of the output:

> test/dump/current-memory.txt                              
>   STDERR MISMATCH:
>   --- expected
>   +++ actual
>   @@ -32,7 +32,7 @@
>    000001b: 00                                        ; func body size (guess)
>    000001c: 00                                        ; local decl count
>    000001d: 3f                                        ; memory.size
>   -000001e: 00                                        ; memory.size reserved
>   +000001e: 00                                        ; memory.size memidx
>    000001f: 0b                                        ; end
>    000001b: 04                                        ; FIXUP func body size
>    0000019: 06                                        ; FIXUP section size

We could just rebase the test output.

src/shared-validator.cc
yhdengh: Here is an example of the test output, essentially CheckMemoryIndex(memidx) attaches a location prior to the error message: 

>  test/spec/memory.txt                                                                             
>   STDOUT MISMATCH:
>   --- expected
>   +++ actual
>   @@ -22,7 +22,7 @@
>    out/test/spec/memory.wast:37: assert_invalid passed:
>      000001d: error: load/store memory 0 out of range 0
>    out/test/spec/memory.wast:41: assert_invalid passed:
>   -  error: memory variable out of range: 0 (max 0)
>   +  0000000: error: memory variable out of range: 0 (max 0)
>      0000019: error: OnMemorySizeExpr callback failed
>    out/test/spec/memory.wast:45: assert_invalid passed:
>      error: memory variable out of range: 0 (max 0)

Again, we could also rebase the test output if the change doesn't affect any existing parts of wabt.

sbc100: Yeah I think rebasing the output for this kind of change is fine.

src/binary-reader.cc
sbc100: I see.

How about at least we do something like this to minimize duplication and make it clear precisely what the differences in the two code paths are:

```
Index memidx;
if (!options_.features.multi_memory_enabled())  {
  CHECK_RESULT(ReadU8(&reserved, "memory.grow reserved"));
  ERROR_UNLESS(memidx == 0, "memory.grow memory index must be 0");
} else {
  CHECK_RESULT(ReadMemidx(&memidx, "memory.grow memidx"));
}
CALLBACK(OnMemoryGrowExpr, memidx);
CALLBACK(OnOpcodeUint32, memidx);
```

sbc100: I think we can use this pattern here and elsewhere and it should reduce the size of this patch.

sbc100 (commented):
Otherwise looks good to me!

sbc100 (commented):
Oh, can you also update the README.md to say that we support this proposal?

src/binary-reader-logging.cc
```diff
@@ -828,16 +837,16 @@ DEFINE0(OnElseExpr)
 DEFINE0(OnEndExpr)
 DEFINE_INDEX_DESC(OnGlobalGetExpr, "index")
 DEFINE_INDEX_DESC(OnGlobalSetExpr, "index")
-DEFINE_LOAD_STORE_OPCODE(OnLoadExpr);
+DEFINE_MEMORY_LOAD_STORE_OPCODE(OnLoadExpr);
```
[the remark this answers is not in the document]
yhdengh: It is also used by other expressions with type LoadStoreExpr. I think any operations that have alignment and offset are defined as LoadStoreExpr, including operations that are not load/store related.

yhdengh:
For reserved byte related logic, I did something like: 

```
Index memidx = 0;
if (!options_.features.multi_memory_enabled())  {
  unit8_t reserved;
  CHECK_RESULT(ReadU8(&reserved, "memory.grow reserved"));
  ERROR_UNLESS(reserved == 0, "memory.grow memory index must be 0");
} else {
  CHECK_RESULT(ReadMemidx(&memidx, "memory.grow memidx"));
}
CALLBACK(OnMemoryGrowExpr, memidx);
CALLBACK(OnOpcodeUint32, memidx);
```

since I am not sure `ReadU8(&memidx)` gives the right result with different endianness. Is that correct?

Also, would you prefer to have all changes for the pull request squashed into a single commit or have them as multiple commits?

sbc100:
> For reserved byte related logic, I did something like:
> 
> ```
> Index memidx = 0;
> if (!options_.features.multi_memory_enabled())  {
>   unit8_t reserved;
>   CHECK_RESULT(ReadU8(&reserved, "memory.grow reserved"));
>   ERROR_UNLESS(reserved == 0, "memory.grow memory index must be 0");
> } else {
>   CHECK_RESULT(ReadMemidx(&memidx, "memory.grow memidx"));
> }
> CALLBACK(OnMemoryGrowExpr, memidx);
> CALLBACK(OnOpcodeUint32, memidx);
> ```
> 
> since I am not sure `ReadU8(&memidx)` gives the right result with different endianness. Is that correct?
> 
> Also, would you prefer to have all changes for the pull request squashed into a single commit or have them as multiple commits?

We always squash the PR before landing it.  Its up to you if you want to squash during the review process or not.  I tend not to look at individual commits myself.

test/dump/load.txt
```diff
@@ -174,46 +174,46 @@ Code Disassembly:
 
 00001b func[0]:
  00001c: 41 00                      | i32.const 0
- 00001e: 28 02 00                   | i32.load 2 0
```
sbc100, line 177: Actually, looks at these changes now, it does seem a little odd to show the extra zero in the disassembly.  

Can we under this part with minimal code changes?   Perhaps something like:

```
if (memidx) {
  CALLBACK(OnOpcodeUint32Uint32Uint32, alignment_log2, offset, memidx);
} else {
  CALLBACK(OnOpcodeUint32Uint32, alignment_log2, offset);
}
```

Sorry for making you go back and forth on this.

Also, should we make the memidx for the first the Uint32 arguments for the load stores so it is shown first in the disassembly?

yhdengh: I think it makes more sense to keep memidx as the last argument. The disassembly should be the same as the binary format which has memidx as the last argument, right?

sbc100: Sure, I don't feel strongly about it being the last argument or not, but lets completely elide it when its zero (or at least when multi memory is not enabled).

yhdengh: I just updated that. Tests related to disassembly/objdump are the same as how they are in the main branch now.

sbc100 (approved):
Great!

PR #1753, opened 2021-11-04
Re-enable simd in update-spec-tests.py

sbc100:
This was removed as part of #1712 but I think it still needs to be here
until simd is actually merge into the main spec repo and no longer
exists as proposal in the testsuite repo.


ngzhian (approved):
LGTM

PR #1754, opened 2021-11-04
update-spec-tests.py: Process all proposal directories.

sbc100:
Maintain an explict list of exclusions and process all other
directories.  This means that we will notice as new proposals
get added.


ngzhian (approved):
LGTM

PR #1755, opened 2021-11-05
Refactor init expressions handling in binary-reader-objdump.cc. NFC

sbc100:
This makes thinks a little more precise and is part of
paving the way for extended const/init expressions.


ngzhian (approved):
LGTM!

PR #1756, opened 2021-11-05
Remove separate OnEndFunc vs OnEndExpr.

sbc100:
We already have EndFunctionBody, and this extra distinction
doesn't seem like it is needed.


ngzhian (approved):
LGTM!

Issue #1761, opened 2021-11-16
Infinite loop on use of `return_call`


takikawa:
The following test case will cause an infinite loop (despite there being no recursion in these calls):

```
;;; TOOL: run-interp
;;; ARGS*: --enable-tail-call
(module
  (func (export "f") (result i32)
    (i64.const 1)
    (return_call $g))

  (func $g (param i64) (result i32)
    (i32.const 3)
    (return_call $h))

  (func $h (param i32) (result i32)
    (i32.const 2)
    (local.set 0)
    (local.get 0)
    return)
)
(;; STDOUT ;;;
f() => i32:2
;;; STDOUT ;;)
```

This is due to a bug in how `return_call` is implemented in the interpreter. I'm working on a patch which I'll PR when it's done.

takikawa:
This should be fixed already by #1762, closing.

PR #1762, opened 2021-11-16
interpreter: Fix infinite looping on `return_call`

takikawa:
Fixes infinite looping described in issue #1761.


sbc100:
Could you mind rebasing or merging?

takikawa:
Sure, no problem! I just pushed a rebased commit.

src/interp/binary-reader-interp.cc
```diff
@@ -1132,7 +1132,10 @@ Result BinaryReaderInterp::OnReturnCallExpr(Index func_index) {
 
   if (func_index >= num_func_imports()) {
     istream_.Emit(Opcode::InterpAdjustFrameForReturnCall, func_index);
-    istream_.Emit(Opcode::Br, GetFuncOffset(func_index));
+    istream_.Emit(Opcode::Br);
+    // We emit this separately to ensure that the index fixup location
+    // is at the right offset in the instruction stream.
+    istream_.Emit(GetFuncOffset(func_index));
```
sbc100, line 1138: I can't see how this is different to original code.  

Previously we were calling ` Istream::Emit(Opcode::Enum op, u32 val)` which looks like it does the same as `Istream::Emit(Opcode::Enum op)` + `Istream::Emit(u32 val)`..  maybe I am missing something?

takikawa: The issue is that `GetFuncOffset` has a side effect (appends a new index to `func_fixups_`) that looks up `istream_.end()`. In the original code, the appended location points to the br opcode rather than its argument `func_index` that needs fixing up.

sbc100: Oh, nasty.   How about something more like:  `// We emit this separately to ensure that the fixup generated by GetFuncOffset comes after the Br opcode`

takikawa: Sure, that's fine with me! Pushed a commit revising the comment.

sbc100 (approved):

PR #1768, opened 2021-12-01
OnRefFuncExpr takes a func index. NFC

sbc100:
The name of this argument was recently changed from `func_index`
to `type_index` as part of #1695, but I think that might have been incorrect.

The immediate that read in the binary reader is (IIRC) a function
index:

```
      case Opcode::RefFunc: {
        Index func;
        CHECK_RESULT(ReadIndex(&func, "func index"));
        CALLBACK(OnRefFuncExpr, func);
        CALLBACK(OnOpcodeUint32, func);
        break;
      }
```

and not a type index.   Indeed the interpreter seems to treat it
as a function index too:

```
Result BinaryReaderInterp::OnRefFuncExpr(Index func_index) {
  CHECK_RESULT(validator_.OnRefFunc(loc, Var(func_index)));
  istream_.Emit(Opcode::RefFunc, func_index);
  return Result::Ok;
}
```

### DIFF
--- a/src/type-checker.cc
+++ b/src/type-checker.cc
@@ -769,9 +769,9 @@ Result TypeChecker::OnTableFill(Type elem_type) {
   return PopAndCheck3Types(Type::I32, elem_type, Type::I32, "table.fill");
 }
 
-Result TypeChecker::OnRefFuncExpr(Index type_index) {
+Result TypeChecker::OnRefFuncExpr(Index func_index) {
   if (features_.function_references_enabled()) {
-    PushType(Type(Type::Reference, type_index));
+    PushType(Type(Type::Reference, func_index));
   } else {
     PushType(Type::FuncRef);
   }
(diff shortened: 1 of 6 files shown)


sbc100:
@dbezhetskov WDYT?  Was this a mistake in  #1695?

src/type-checker.cc
```diff
@@ -769,9 +769,9 @@ Result TypeChecker::OnTableFill(Type elem_type) {
   return PopAndCheck3Types(Type::I32, elem_type, Type::I32, "table.fill");
 }
 
-Result TypeChecker::OnRefFuncExpr(Index type_index) {
+Result TypeChecker::OnRefFuncExpr(Index func_index) {
   if (features_.function_references_enabled()) {
-    PushType(Type(Type::Reference, type_index));
+    PushType(Type(Type::Reference, func_index));
```
sbc100, line 774: This looks like it might be the one place we want to actually receive a type index rather than a function index.. (but as far as I now we don't).

ngzhian (approved):
Based on https://webassembly.github.io/spec/core/valid/instructions.html#valid-ref-func it is a func index, so LGTM.

sbc100:
I'm gonna go ahead and land this.  @dbezhetskov, if you disagree we can discuss more and consider reverting..

dbezhetskov:
ah, yeah, it was a typo, I've forget to return it back, thanks @sbc100 !

PR #1780, opened 2021-12-08
Add error locations to BinaryReaderInterp

sbc100:
I think it was always intended to work this way but was
left as a TODO.


sbc100:
Depends on #1779

ngzhian (approved):

PR #1782, opened 2021-12-09
Fix syntax for assert_return typecheck test

takikawa:
This test seems to be intended to test the type checking behavior, but due to a syntax error in an export statement was just testing the parsing behavior instead. This PR fixes the syntax and adjusts the test expectations.


sbc100 (approved):

sbc100:
Thanks!

sbc100:
It looks like these expectations were mistakenly updated as part of #591.  The dangers of automatic test expectation updates!  

Actually it goes way back to 63d7dbb60245136872f2ca521fbd721dc76e0b9e which was one of my changes :-/

PR #1783, opened 2021-12-10
Share validation code between constant expressions and function bodie…

sbc100:
…s. NFC

Previously we has special cases for initializer expressions (constant
expressions).  This change paves the way for adding support for
extended constant expressions that support a wider range of
instructions.

This change removes twice as many lines as it adds which shows that
this simplification is probably worthwhile even without the pending
extensions.

### DIFF
--- a/src/binary-reader-ir.cc
+++ b/src/binary-reader-ir.cc
@@ -280,6 +280,7 @@ class BinaryReaderIR : public BinaryReaderNop {
   void PushLabel(LabelType label_type,
                  ExprList* first,
                  Expr* context = nullptr);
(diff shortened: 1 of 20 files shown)


ngzhian (approved):
The failure messages are less clear i think (previously it mentions global initializer expression, now it's constant expression), but with the line and column it should be easy to pinpoint the errors.

sbc100:
> The failure messages are less clear i think (previously it mentions global initializer expression, now it's constant expression), but with the line and column it should be easy to pinpoint the errors.

I agree, it could be seen a less clear, but its also more consistent which I like.  I can take a look and see if we can be more specific perhaps.

PR #1791, opened 2021-12-19
Remove redundant `BindImports` call (NFC)

aheejin:


src/tools/wasm-interp.cc
```diff
@@ -299,8 +299,6 @@ static Result ReadAndRunModule(const char* module_filename) {
     s_stderr_stream.get()->Writef("wasi support not compiled in\n");
     return Result::Error;
 #endif
-  } else {
-    BindImports(module, imports);
   }
   BindImports(module, imports);
```
sbc100, line 303: Perhaps its this one that should be removed instead?   It looks like WASI uses `WasiBindImports` above so perhaps this  extra call is the incorrect one?     Its been a while since I looked at this code so but it seems plausible that this is why it written this way.

aheejin: Removed this one instead.

aheejin:
Ping :) Can I merge it? (Sorry I know it's early, but you seem to be OOO from tomorrow so..)

sbc100 (approved):

Issue #1796, opened 2021-12-23
Wording of error message 'multiple result values not currently supported'


jzabinski-dolios:
In `SharedValidator::OnFuncType`, there is this code:
```
if (!options_.features.multi_value_enabled() && result_count > 1) {
    result |=
        PrintError(loc, "multiple result values not currently supported.");
  }
```

WABT does indeed support multi-value, however.

Maybe the error message should say 'not currently enabled' when multi-value is disabled.

sbc100:
Yes, that sounds good.   How about `multiple result values are not supported without multi-value enabled`?  Do you have time to send a patch?

jzabinski-dolios:
Will send a patch next week when I am back from vacation

PR #1797, opened 2022-01-03
Updated multi-value message

jzabinski-dolios:
Resolves https://github.com/WebAssembly/wabt/issues/1796

To test locally:

1. Make a WAT file containing a function which returns multiple values. Example:
```
(module
    (func $mult (export "mult")
        (result f64 f64)
        f64.const 1
        f64.const 2
    )
)
```
2. Use `wat2wasm` to compile, disabling multi-value. For example:
```
sudo bin/wat2wasm --disable-multi-value examples/multi.wat -o examples/multi.wasm
```

An error should be returned by the compiler. The wording of the error should be `multiple result values are not supported without multi-value enabled.`

### DIFF
--- a/src/shared-validator.cc
+++ b/src/shared-validator.cc
@@ -53,7 +53,7 @@ Result SharedValidator::OnFuncType(const Location& loc,
   Result result = Result::Ok;
   if (!options_.features.multi_value_enabled() && result_count > 1) {
     result |=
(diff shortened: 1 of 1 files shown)


sbc100 (approved):

sbc100:
Looks like there is lint failure.. can you `git clang-format origin/main`?

PR #1804, opened 2022-01-12
Update github action used to publish releases. NFC

sbc100:
It seems that the action we were using is not longer working:
https://github.com/actions/upload-release-asset

See https://github.com/WebAssembly/binaryen/pull/4148

### DIFF
--- a/.github/workflows/build_release.yml
+++ b/.github/workflows/build_release.yml
@@ -6,6 +6,9 @@ on:
     types:
       - created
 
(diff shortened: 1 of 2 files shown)


sbc100:
You can see the new action in action here: https://github.com/WebAssembly/wabt/releases/tag/test_tag

tlively (approved):
Thanks, @sbc100!

PR #1807, opened 2022-01-14
Fix location in config.h in source list

sbc100:
The wrong name here was causing a cmake warning because
it was finding `config.h.in` instead and this behaviour is
deprecated.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,11 +50,11 @@ jobs:
     - name: mkdir
       run: mkdir -p out
     - name: cmake
-      run: cmake .. -G Ninja -DWERROR=ON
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -50,11 +50,11 @@ jobs:
     - name: mkdir
       run: mkdir -p out
     - name: cmake
-      run: cmake .. -G Ninja -DWERROR=ON
+      run: cmake .. -G Ninja -DWERROR=ON -Werror=dev
```
binji, line 53: what is -Werror=dev?

sbc100: This is cmake thing that makes cmake warnings into errors..  nothing to do with C/C++.  Helps us avoid ugly cmake warnings.

Issue #1671, opened 2021-05-26
memory management(gc) in wasm-interp


clover2123:
Hi all,
I'm trying to run the wasm-interp as WebAssembly runtime in my project.

I found a memory issue which is that the heap memory allocated through wasm-interp increases steadily without any drop.
Let me give you an example about function call,
https://github.com/WebAssembly/wabt/blob/63c350984f7dbe76f5963f4df819bfc375ea0217/src/interp/interp.cc#L344-L353
According to the above code, `Thread` object is newly created for each function call, but never deleted after calls.
In the end, these `Thread` objects will be accumulated terribly after frequent function calls.

It seems like that `Store::Collect()` is the only function that could reclaim the garbages, but this function is not explicitly called during the runtime.
I'm wondering that I should invoke `Store::Collect()` periodically or is there any other simple way to resolve this memory issue.

Thanks in advance!

clover2123:
Fixed by #1809

PR #1809, opened 2022-01-18
Change Thread in interpreter to a normal object

zherczeg:
Improves memory consumption since thread instances are freed without running garbage collector.

### DIFF
--- a/src/interp/interp.cc
+++ b/src/interp/interp.cc
@@ -256,6 +256,10 @@ void Store::Collect() {
     }
   }
 
+  for (auto thread : threads_) {
+    thread->Mark();
+  }
+
   // TODO: better GC algo.
   // Loop through all newly marked objects and mark their referents.
   std::vector<bool> all_marked(object_count, false);
@@ -366,10 +370,8 @@ Result Func::Call(Store& store,
                   Values& results,
                   Trap::Ptr* out_trap,
                   Stream* trace_stream) {
-  Thread::Options options;
-  options.trace_stream = trace_stream;
-  Thread::Ptr thread = Thread::New(store, options);
-  return DoCall(*thread, params, results, out_trap);
+  Thread thread(store, trace_stream);
+  return DoCall(thread, params, results, out_trap);
 }
 
 Result Func::Call(Thread& thread,
@@ -934,24 +936,30 @@ void Instance::Mark(Store& store) {
 }
 
 //// Thread ////
-Thread::Thread(Store& store, const Options& options)
-    : Object(skind), store_(store) {
+Thread::Thread(Store& store, Stream* trace_stream)
+    : store_(store), trace_stream_(trace_stream) {
+  store.threads().insert(this);
+
+  Thread::Options options;
   frames_.reserve(options.call_stack_size);
   values_.reserve(options.value_stack_size);
-  trace_stream_ = options.trace_stream;
-  if (options.trace_stream) {
+  if (trace_stream) {
     trace_source_ = MakeUnique<TraceSource>(this);
   }
 }
 
-void Thread::Mark(Store& store) {
+Thread::~Thread() {
+  store_.threads().erase(this);
+}
+
+void Thread::Mark() {
   for (auto&& frame : frames_) {
-    frame.Mark(store);
+    frame.Mark(store_);
   }
   for (auto index : refs_) {
-    store.Mark(values_[index].Get<Ref>());
+    store_.Mark(values_[index].Get<Ref>());
   }
-  store.Mark(exceptions_);
+  store_.Mark(exceptions_);
 }
 
 void Thread::PushValues(const ValueTypes& types, const Values& values) {
--- a/src/interp/interp.h
+++ b/src/interp/interp.h
@@ -20,6 +20,7 @@
 #include <cstdint>
 #include <functional>
 #include <memory>
+#include <set>
 #include <string>
 #include <type_traits>
 #include <vector>
@@ -90,7 +91,6 @@ enum class ObjectKind {
   Tag,
   Module,
   Instance,
-  Thread,
 };
 
 const char* GetName(Mutability);
@@ -500,11 +500,15 @@ class Store {
   const Features& features() const;
   void setFeatures(const Features& features) { features_ = features; }
 
+  std::set<Thread*>& threads();
+
  private:
   template <typename T>
   friend class RefPtr;
 
   Features features_;
+  // This set contains the currently active Thread objects.
+  std::set<Thread*> threads_;
   ObjectList objects_;
   RootList roots_;
   std::vector<bool> marks_;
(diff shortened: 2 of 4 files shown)


zherczeg:
Could somebody review this patch? Thank you!
Related issue #1671

zherczeg:
@sbc100 @dschuff @tlively Since you do approvals in this project, could you suggest somebody who can review this patch?

src/interp/interp.h
```diff
@@ -1104,13 +1098,15 @@ class Thread : public Object {
     Stream* trace_stream = nullptr;
   };
 
-  static Thread::Ptr New(Store&, const Options&);
+  explicit Thread(Store& store, Stream* trace_stream);
```
sbc100, line 1101: No need for `explicit` with multi argument constructors (IIUC)

src/interp/interp.cc
```diff
@@ -934,24 +938,33 @@ void Instance::Mark(Store& store) {
...
   values_.reserve(options.value_stack_size);
   trace_stream_ = options.trace_stream;
-  if (options.trace_stream) {
+  if (trace_stream) {
     trace_source_ = MakeUnique<TraceSource>(this);
   }
 }
 
-void Thread::Mark(Store& store) {
+Thread::~Thread()
```
sbc100, line 955: Open braces on the same line as method name

src/interp/interp.cc
```diff
@@ -934,24 +938,33 @@ void Instance::Mark(Store& store) {
...
+{
+  store_.threads_ = prev_thread_;
```
sbc100, line 957: Is it not possible for threads to be destroyed in a different order to that in which they were created?

src/interp/interp.h
```diff
@@ -503,8 +502,10 @@ class Store {
  private:
   template <typename T>
   friend class RefPtr;
+  friend class Thread;
 
   Features features_;
+  Thread *threads_ = nullptr;
```
sbc100, line 508: So this is linked list going of threads from the most recently constructed back the oldest?  Maybe add a comment here?

src/interp/interp.cc
zherczeg: It seems to me Thread objects are created on wasm function call, and so the compiler ensures they are destroyed in a reversed creation order by the compiler (unless longjmp is used, but that is discouraged to use in C++). I think the compiler would complain if a Thread is created in another way since the constructor is changed.

If an ECMAScript async like feature would be implemented in wasm in the future, that would likely need a context object, and Thread can be a member of that object. These Threads would be marked through the context object, and would not be part of this Thread stack chain.

src/interp/interp.h
zherczeg: Comment added.

src/interp/interp.cc
sbc100: I guess I'm not sure what this Thread object is then.. surely a thread (at least in the normal sense) is not created for every call but contains an entire calls stack of calls?    And there is no normally a stack relationship between running threads (I would imagine threads is more of a parent child relationship).

sbc100: Oh I see .. there are two ways to `Call` a function:

```
  Result Call(Thread& thread,                                                       
              const Values& params,                                                 
              Values& results,                                                      
              Trap::Ptr* out_trap);                                                 
                                                                                    
  // Convenience function that creates new Thread.                                  
  Result Call(Store&,                                                                                                                                                
              const Values& params,                                                 
              Values& results,                                                   
              Trap::Ptr* out_trap,                                               
              Stream* = nullptr);    
```

The second one create a brand new thread... which I guess should be destroyed when the call returns.

The first one is for when a normal function call is make within an existing thread.

sbc100: I this case I don't think we need to model threads in some kind of stack since their lifetime is independent of of each other, no?

zherczeg:
I hope I fixed the format and testing issues. CI still needs manual approval.

src/interp/interp.cc
zherczeg: Honestly, I don't know the concept behind the design. The `Result Call(Thread& thread, ...)` variant is never called directly, probably users should not create threads directly. Maybe it is a future-proof design for something (my guess is async). But this code improves the memory management of the common use case, and we can think something for the special cases later (Threads can be a field of other objects, which are inherited form GC object).

zherczeg: Regarding the stack concept, the store needs to know the active Threads for marking. We could use some high level sets to insert/delete Threads, but this stack implementation is very fast in practice. When the stack cannot be used, the Threads can be marked by the store in other ways (I described an idea for that above).

src/interp/interp.cc
```diff
@@ -934,24 +937,32 @@ void Instance::Mark(Store& store) {
...
+    : store_(store), trace_stream_(trace_stream) {
+  prev_thread_ = store.threads_;
+  store.threads_ = this;
+
+  Thread::Options options;
   frames_.reserve(options.call_stack_size);
   values_.reserve(options.value_stack_size);
-  trace_stream_ = options.trace_stream;
-  if (options.trace_stream) {
+  trace_stream_ = trace_stream;
```
zherczeg, line 948: Oops that is a nop. I will remove it.

src/interp/interp.h
```diff
@@ -503,8 +502,12 @@ class Store {
  private:
   template <typename T>
   friend class RefPtr;
+  friend class Thread;
```
zherczeg, line 505: I did this for simplicity, but I am open to other suggestions.

zherczeg (commented):
I will wait your feedback before doing changes because the CI needs manual approval.

src/interp/interp.cc
sbc100: I think just using `std::set` of active threads should be fine.

zherczeg: Ok, I have changed the patch to use std::set.

src/interp/interp.h
sbc100: It this still needed?

src/interp/interp.h
```diff
@@ -1246,6 +1241,8 @@ class Thread : public Object {
 
   RunResult StepInternal(Trap::Ptr* out_trap);
 
+  Thread* prev_thread_;
+
```
sbc100, line 1245: I think this can be removed now?

sbc100 (approved):

src/interp/interp.h
zherczeg: The `friend class Thread;` is still needed because the Thread constructor/destructor add/removes the Thread from the set in the storage. It could be removed if the `set` would be accessible for Thread somehow.

zherczeg:
Thank you for the reviews and helping for me!

src/interp/interp.h
```diff
@@ -503,8 +503,11 @@ class Store {
  private:
   template <typename T>
   friend class RefPtr;
+  friend class Thread;
 
   Features features_;
+  // This set contains the currently active Thread objects
+  std::set<Thread*> threads_;
```
sbc100, line 510: Maybe just make this public.. rather than added a friend?   Or add  `addThread` / `removeThread` methods (but that seems roughly the same as making it public).

zherczeg: I made a `thread()` inline function to get a reference to the private field.

PR #1810, opened 2022-01-20
Added support for multi-memory in apply-names

yhdengh:
This PR adds support for the multi-memory proposal in `apply-names.cc`. This change should have been part of #1751, but I didn't realize that this change is necessary.

This PR also adds name application to all exports. The current main branch only applies names if the export is of type `func`, but I think it should also apply names for `global`, `table`, `memory`, and `tag`. Is there any specific reason for only applying names to `func`s?


src/apply-names.cc
```diff
@@ -254,8 +260,30 @@ Result NameApplier::OnDataDropExpr(DataDropExpr* expr) {
   return Result::Ok;
 }
 
+Result NameApplier::OnMemoryCopyExpr(MemoryCopyExpr* expr) {
+  CHECK_RESULT(UseNameForMemoryVar(&expr->srcmemidx));
+  CHECK_RESULT(UseNameForMemoryVar(&expr->destmemidx));
```
sbc100, line 265: Could we write a test (or add to an existing test) to show that this working?

sbc100 (commented):
Nice!

lgtm % testing comment

test/roundtrip/apply-multi-memories.txt
```diff
@@ -0,0 +1,25 @@
+;;; TOOL: run-roundtrip
```
sbc100, line 1: How about calling this `apply-memory-names.txt` to match the existing `apply-global-names.txt`

sbc100 (approved):

Issue #1805, opened 2022-01-13
Thoughts on making wasm2c thread-safe


yhdengh:
Hello all,

I would like to start a conversation about the best way to make wasm2c thread-safe so that the same module can be instantiated multiple times with different memory contexts, and functions from different instances can be executed in parallel. Our current plan was to move global memory variables into a structure and then pass a pointer to that structure as a parameter to every function call (Similar changes are also proposed by @1721). Before we draft a PR, I would like to see what are people’s thoughts about this. More details are below.

Since the changes we make introduce a new layer of abstraction, we were concerned about the potential increase of memory access latency. Therefore, we investigated the assembly code generated by the output of status quo wasm2c and a toy model of the expected output of our new version of wasm2c in terms of memory access. The assembly code is compiled with `clang -O2`. Both versions require 2 memory access for a memory store.

## Status quo wasm2c
Input wat file:
```wasm
(module
  (memory $m1 1)
  (func (export "store") (param i32 i32)
     local.get 0
     local.get 1
     i32.store))
```
Assembly code compiled from wasm2c output:
```assembly
0000000000401240 <w2c_store>:
static void w2c_store(u32 w2c_p0, u32 w2c_p1) {
  401240: 50                    push   %rax
  FUNC_PROLOGUE;
  401241: 8b 05 81 2e 00 00     mov    0x2e81(%rip),%eax        # 4040c8 <wasm_rt_call_stack_depth>
  401247: 83 c0 01              add    $0x1,%eax
  40124a: 89 05 78 2e 00 00     mov    %eax,0x2e78(%rip)        # 4040c8 <wasm_rt_call_stack_depth>
  401250: 3d f5 01 00 00        cmp    $0x1f5,%eax
  401255: 73 15                 jae    40126c <w2c_store+0x2c>
  i32_store((&w2c_M0), (u64)(w2c_i0), w2c_i1);
  401257: 89 f8                 mov    %edi,%eax
DEFINE_STORE(i32_store, u32, u32);
  401259: 48 8b 0d 40 2e 00 00  mov    0x2e40(%rip),%rcx        # 4040a0 <w2c_M0>
  401260: 89 34 01              mov    %esi,(%rcx,%rax,1)
  FUNC_EPILOGUE;
  401263: 83 05 5e 2e 00 00 ff  addl   $0xffffffff,0x2e5e(%rip)        # 4040c8 <wasm_rt_call_stack_depth>
}
  40126a: 58                    pop    %rax
[...]

PR #1814, opened 2022-01-26
wasm2c: add support for module instancing

yhdengh:
This PR adds support for module instancing to wasm2c as proposed in #1805. The approach is similar to #1721. 

- Globals, memories, and tables are moved into a new structure named `module_prefix_ + _instance_t`.
- The structure is passed as the first argument of function calls. The embedder is responsible for initializing the structure.
- `run-spec-wasm2c.py` is updated such that `main.c` generated by the script reflects the changes for module instancing.


.github/workflows/build.yml
```diff
@@ -4,8 +4,6 @@ on:
   create:
     tags:
   push:
-    branches:
-      - master
```
kripken, line 8: why is this changed?

test/run-spec-wasm2c.py
```diff
@@ -163,6 +183,22 @@ def _CacheModulePrefixes(self):
...
+                module_prefix = self.module_prefix_map[idx]
+
+                self.out_file.write("static void %s_spectest_print_i32(%s *module_instance, uint32_t i) { spectest_print_i32(i); }\n" % (module_prefix, module_instance_type))
+                self.out_file.write("void (*%sZ_spectestZ_print_i32Z_vi)(%s *, uint32_t) = & %s_spectest_print_i32;\n" % (module_prefix, module_instance_type, module_prefix))
+                self.out_file.write("static void %s_spectest_print(%s *module_instance) { spectest_print(); }\n" % (module_prefix, module_instance_type))
+                self.out_file.write("void (*%sZ_spectestZ_printZ_vv)(%s *) = & %s_spectest_print;\n" % (module_prefix, module_instance_type, module_prefix))
+                self.out_file.write("\n")
+
+                idx += 1
+
```
kripken, line 201: Why are these testing changes necessary? I would hope that the tests would keep running as before, except perhaps with extra code to create an instance, but nothing more.

wasm2c/README.md
```diff
@@ -436,3 +470,95 @@ $ wat-desugar fac-flat.wat --fold -o fac-folded.wat
 
 The formatting is different and the variable and function names are gone, but
 the structure is the same.
+
+## Instantiate multiple instances of a module
+Since information about the execution context, such as memories, is encapsulated
+in `instance_module_instance`, and a pointer to the structures is being passed through 
+function calls, multiple instances of the same module can be instantiated 
+with different context.
```
kripken, line 478:
```suggestion
with different contexts. 
```

kripken (commented):
Overall this looks good to me, but I have a few questions.

@shravanrn please take a look as well. Thanks!

.github/workflows/build.yml
yhdengh: We changed it so that pushes to all branches trigger CI checks.

test/run-spec-wasm2c.py
yhdengh: `_WriteModuleImports` converts a function defined in `spec-wasm2c-prefix.c` to its corresponding version that takes a pointer to the module instance as its first argument and is name-mangled by the module name, but does not change how the function executes. Since each module defined in a `wast` file has a different instance type and name for imported functions, we need to repeat `_WriteModuleImports` for each module.

We think functions imported from the host should also take a pointer to the instance as its argument and be name-mangled by module names for the following reasons:
- Hosts should have access to instances
- Hosts should be able to choose which function they want to provide to a certain module

There are potentially other ways of adapting `spec-wasm2c-prefix.c`. What are your thoughts?

.github/workflows/build.yml
kripken: I see, sounds good.

test/run-spec-wasm2c.py
kripken: Thanks - I think I see, but I'm not sure.

Can you maybe paste the output code before and after, or a diff? I think maybe comparing them will be easier to see what you mean here. (In particular, I'm not sure what "Hosts should have access to instances" means, and how it is handled concretely in the code.)

yhdengh: The following example is from module 5 of `start.wast`. The following code snippets are taken from `start.5.h`. The module imports `spectest_print_i32`.

**Before the current PR**
```c
/* import: 'spectest' 'print_i32' */
extern void (*Z_spectestZ_print_i32Z_vi)(u32);
#ifdef __cplusplus
```

**After the current PR**
```c
/* import: 'spectest' 'print_i32' */
extern void (*WASM_RT_ADD_PREFIX(Z_spectestZ_print_i32Z_vi))(Z_startZ2E5_module_instance_t *, u32);
#ifdef __cplusplus
```

After the current PR, the imported function is name-mangled using `WASM_RT_ADD_PREFIX` and takes `Z_startZ2E5_module_instance_t *` as its first argument where `Z_startZ2E5_module_instance_t` is the type of the module instance of this module.

---
`spectest_print_i32` is defined in `spectest-wasm2c-prefix.c` as following
```c
static void spectest_print_i32(uint32_t i) {
  printf("spectest.print_i32(%d)\n", i);
}

void (*Z_spectestZ_print_i32Z_vi)(uint32_t) = &spectest_print_i32;
```
Content of `spectest-wasm2c-prefix.c` is copied into the output `main` file of `run-spec-wasm2c.py`.

---
In order to make `run-spec-wasm2c.py` output imported functions that work, we changed `run-spec-wasm2c.py` to instantiate those imported functions. And the output `start-main.c` looks like this:
```c
static void Z_start_5_spectest_print_i32(Z_startZ2E5_module_instance_t *module_instance, uint32_t i) { spectest_print_i32(i); }
void (*Z_start_5Z_spectestZ_print_i32Z_vi)(Z_startZ2E5_module_instance_t *, uint32_t) = & Z_start_5_spectest_print_i32;
```
Essentially, it defines a `Z_start_5_spectest_print_i32` that takes a pointer to the module instance and a `u32`, and calls `spectest_print_i32` on the `u32`. And then, it sets `WASM_RT_ADD_PREFIX(Z_spectestZ_print_i32Z_vi)` to point to `Z_start_5_spectest_print_i32`.

shravanrn:
Reviewing! Although its getting a bit late in my local timezone (I am currently traveling), so worst case I'll post an update in the next 12 hours or so

test/run-spec-wasm2c.py
kripken: I see, thanks!

That all makes sense to me. The name mangling by `WASM_RT_ADD_PREFIX` should have been there before, I think, and adding an instance parameter is clearly necessary.

keithw:
This now passes the current wasm2c tests, including for imports/exports/linking (these are the "old" spec tests from before bulk-memory operations and reference-types were merged).

We're still working on implementing bulk-memory operations and reference-types on top of this, which if successful will hopefully have wasm2c passing the current versions of the spec tests, as well as the remaining multi-memory proposal tests that depend on these features. I've been tracking this at https://github.com/WebAssembly/wabt/pull/1853#issuecomment-1062094906 .

shravanrn:
TLDR: we can go ahead and land this instancing support when ready. I'll deal with the incompatibilities they may introduce in our fork

@kripken @sbc100 Just wanted to briefly update everyone. I was actually away for the last few weeks hence the delayed response here. But I did play a bit more with rebasing some of these changes on our wasm2c fork. In short it should be possible with some work. That said, my current priority for Mozilla is to rewrite RLBox to add more features, so I suspect I won't be able to upstream some of our custom wasm2c features in our fork for a few more weeks. However, I don't want this to block any other ongoing work here, so feel free to land this instancing support. I am happy however to review this and any other PRs you want me to look at.

keithw:
@shravanrn Thanks, we appreciate this. Of course we'd be grateful for your review. When you do want to upstream the rlbox code, we're still happy to help rebase it at the time and preserve whatever tests are passing. I don't think the two approaches are that different; ours is mostly the same as yours + handling imports/exports/linking, cases that are probably less relevant to the RLBox use-case.

@sbc100 @kripken This is probably ready for another look. @binji gave us a review here (https://github.com/WebAssembly/wabt/pull/1877#pullrequestreview-931162689) that we've since responded to. After this one, the later-stacked PRs are #1877 (bulk memory), #1887 (reference types), and #1888 (tolerating partly invalid modules), after which wasm2c passes all the core spec tests (and all the multi-memory proposal tests) except the SIMD ones.

src/wasm2c.c.tmpl
```diff
@@ -2,6 +2,7 @@
 /* Automically generated by wasm2c */
 #include <math.h>
 #include <string.h>
+#include <stdlib.h>
```
sbc100, line 5: Why is this needed?

src/wasm2c.c.tmpl
```diff
@@ -27,7 +28,7 @@
        , ((t)table.data[x].func)(__VA_ARGS__))
 
 #define RANGE_CHECK(mem, offset, len) \
-  if (UNLIKELY(offset + (uint64_t)len > mem->size)) TRAP(OOB)
+  if (UNLIKELY(offset + (u64)len > mem->size)) TRAP(OOB)
```
sbc100, line 31: Is this change unrelated?  Maybe split into its own PR?

sbc100 (commented):
I created a separate PR to remove the signature mangling: #1896.  Perhaps we can land that first to simplify this PR?

wasm2c/wasm-rt.h
```diff
@@ -242,6 +250,11 @@ extern void wasm_rt_allocate_table(wasm_rt_table_t*,
                                    uint32_t elements,
                                    uint32_t max_elements);
 
+/**
+ * Free a Table object.
+ */
+extern void wasm_rt_free_table(wasm_rt_table_t*);
```
sbc100, line 256: Perhaps the freeing of tables and memories could also be introduced as a separate PR before this one lands?   Its seems unrelated to instancing.

src/wasm2c.c.tmpl
keithw: It's vestigial -- we'll get rid of it.

src/wasm2c.c.tmpl
keithw: This was in response to https://github.com/WebAssembly/wabt/pull/1877#discussion_r842318972 . Sure, let's split it off.

wasm2c/wasm-rt.h
keithw: Sure, we will do that. It's eventually required because otherwise asan can complain about the memory leak.

sbc100:
Can you rebase now that #1897 and #1896 have landed?  I hope it should make this change much more precise.

wasm2c/wasm-rt.h
keithw: Done in #1901.

keithw:
> Can you rebase now that #1897 and #1896 have landed? I hope it should make this change much more precise.

Here you go.

test/run-spec-wasm2c.py
```diff
@@ -145,7 +156,10 @@ def GetModulePrefix(self, idx_or_name=None):
     def GetModulePrefixUnmangled(self, idx):
         return self.unmangled_names[idx]
 
-    def _CacheModulePrefixes(self):
+    def GetModuleInstanceName(self, idx_or_name=None):
+        return self.GetModulePrefix() + '_module_instance'
+
+    def CacheModulePrefixes(self):
```
sbc100, line 162: Why not keep the leading `_` on this method name?

wasm2c/examples/fac/main.c
```diff
@@ -5,7 +5,8 @@
 
 int main(int argc, char** argv) {
   /* Make sure there is at least one command-line argument. */
-  if (argc < 2) return 1;
+  if (argc < 2)
```
sbc100, line 8: I think he wabt (aka Chromium) style is to always just curly braces if you put the body on its own line

wasm2c/examples/rot13/main.c
```diff
@@ -19,48 +19,69 @@
...
 #include "rot13.h"
 
-/* Define the imports as declared in rot13.h. */
-wasm_rt_memory_t (*Z_hostZ_mem);
-u32 (*Z_hostZ_fill_buf)(u32, u32);
-void (*Z_hostZ_buf_done)(u32, u32);
+/* Define structure to hold the imports */
+struct Z_host_module_instance_t {
+  wasm_rt_memory_t s_memory;
+  char* s_input;
```
sbc100, line 25: These members probably shouldn't have the `s_` prefix if they are not static (i think `s_` is for static here)

wasm2c/README.md
```diff
@@ -395,3 +455,104 @@ $ wat-desugar fac-flat.wat --fold -o fac-folded.wat
...
+wasm_rt_memory_t* Z_hostZ_mem(struct Z_host_module_instance_t* instance) {
+  return &instance->s_memory;
+}
+
+/* Declare the implementations of the imports. */
+static u32 fill_buf(struct Z_host_module_instance_t* instance, u32 ptr, u32 size);
+static void buf_done(struct Z_host_module_instance_t* instance, u32 ptr, u32 size);
+
+/* Define host-provided functions under the names imported by the `rot13` instance */
+u32 Z_rot13Z_hostZ_fill_buf(struct Z_host_module_instance_t* instance,
```
sbc100, line 488: I would expect this to not have the `Z_rot13` prefix..   Isn't that prefix reserved for things defined in the rot32 module?   For things imported from the `host` module I would just expect the `Z_host` prefix.

wasm2c/README.md
```diff
@@ -127,27 +142,34 @@ The generated header file looks something like this:
...
 #endif
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-extern void Z_fac_init(void);
-extern void Z_fac_free(void);
+typedef struct Z_fac_module_instance_t {
+  char dummy_member;
```
sbc100, line 157: This is a little sad that we need to do this.   I would imgine that all instances would share at least one common member but I guess not.

sbc100: Perhaps in the long term we will have a type for the module itself which each instance would have a pointer too?

src/c-writer.cc
```diff
@@ -423,11 +461,27 @@ std::string CWriter::MangleName(std::string_view name) {
   return result;
 }
 
-// static
 std::string CWriter::ExportName(std::string_view mangled_name) {
   return module_prefix_ + std::string(mangled_name);
 }
 
+// static
```
sbc100, line 468: I think this method isn't static

sbc100 (commented):
Thanks for all the work on this.   I think its in a good state, although I'm still working through the writer.cpp changes.

test/run-spec-wasm2c.py
keithw: We wanted it to be "public" so we can call it within the main function shortly after creating the CWriter (and before cwriter.GetModulePrefixUnmangled() gets called on every module).

keithw: Fair enough; latest commit keeps the leading `_` (which I think means "private"?) and adds a call from within the CWriter's `__init__`. (We need this to be called before GetModulePrefixUnmangled but we need wasm2c to run before cwriter.Write(), so it can parse the wasm2c-generated `.h` files.)

wasm2c/examples/fac/main.c
keithw: This was a clang-format autochange; restored to original.

wasm2c/examples/rot13/main.c
keithw: Ah, fixed.

wasm2c/README.md
keithw: Welll.... our general theory had been that every public symbol (import or export) gets prefixed with the module prefix to keep things isolated. E.g. (in a world without "Z"), it seems scary if a WebAssembly module could import a function named `kill` from a module named `pthread`, and now it can call the real `pthread_kill(3)` function whenever it wants.

(This also gives the host a little more power to decide what it wants to export to every module type. E.g. maybe it wants to version its API by exporting one version of the `fill_buf` function to rot13_v1, but a different version of `fill_buf` to rot13_v2.)

wasm2c/README.md
keithw: Yeah... GCC and clang are happy to allow an empty struct (except with `-pedantic` which we're not currently using), but MSVC rejects it as invalid C.

src/c-writer.cc
keithw: Thanks, fixed.

wasm2c/README.md
keithw: The downside/consequence is that the host does have to declare functions for each module it wants to be able to import something from it. If two modules (named "modulea" and "moduleb") each import a whole bunch of functions from "wasi_snapshot_preview1", then the host will have to arrange, probably via some preprocessor macro that sets this up, to declare functions named, e.g., `Z_moduleaZ_wasi_snapshot_preview1Z_fd_read` and also `Z_modulebZ_wasi_snapshot_preview1Z_fd_read`  that turn around and call the "real" WASI functions, in order for both modules to be allowed to call them.

binji (commented):
I didn't do a careful review of this PR (since I looked at the later one), but this generally looks good to me. It seems like you fixed all the notes I made in the other PR, right? If so this lgtm, though I'll defer to @sbc100 for the final approval.

wasm2c/README.md
sbc100: I think imports can be already considered prefixed/isloated because they contain the module name within them.  `Z_wasi_snapshot_preview1Z_fd_read` already contains that the `Z_wasi_snapshot_preview1` prefix.  If we want to more fine grained control, for example to be able to have two different version of `Z_wasi_snapshot_preview1Z_fd_read` in the same process, then I think we can address that as a followup and it will most likely involve using some kind of user level linking (e.g. structs of function pointers) rather than relying on the C symbol names to do the linking.

shravanrn:
> @shravanrn Thanks, we appreciate this. Of course we'd be grateful for your review. When you do want to upstream the rlbox code, we're happy to help rebase it at the time (if you want our help) and preserve whatever tests are passing. I don't think the two approaches are that different; ours is mostly the same as yours + handling imports/exports/linking, cases that are probably less relevant to the RLBox use-case.

Having a look now. Sorry for the delay

wasm2c/README.md
keithw: Hmm, this still seems pretty scary to me. Maybe I'm off-base, so let me see if I can explain why. I put a proposed solution at the bottom that hopefully will satisfy your concern and also our fears...?

I hope I'm using the right language here. We see these three things as distinct:
1. the CWriter "module_prefix," a string controlled by the caller of wasm2c (if unspecified, the module's name if it has one, or else the filename)
2. the "mod:name" of an import, a string that the module producer/author controls
3. the "nm:name" of an import or export, also a string that the module producer/author controls

In this PR, the **exported** functions turn into function definitions, with the function named "module_prefix" + "nm:name". I think we're all on the same page on that one.

What we're talking about here is about the **imported** functions, which in the current PR become extern declarations referencing functions named with the form "module_prefix" + "mod:name" + "nm:name".

What I think you're suggesting is to rename them just "mod:name" + "nm:name" (removing the "module_prefix"). There's basically two reasons we're scared of this:

1. The "mod:name" and "nm:name" are both totally under the control of the module producer/author. If the module chooses to import a function named "kill" from a mod:name "pthread", then CWriter would emit an extern declaration for a function named "pthread_kill" (assuming no "Z" here), and then if linked into an executable, the WebAssembly module would be able to call that function (or "rt_sigreturn" or "timer_delete"...) with no intervention or explicit decision to grant that capability by the host/embedder. That doesn't seem right -- and so we think the import symbols all should require a prefix that's **not** under the control of the module author/producer.
2. Even if we prefixed all import and export symbols by some distinguished but constant prefix (e.g. "w2c_" + "mod:name" + "nm:name"), it's still scary because it means that two modules would be able to call one another in a possible signature-mismatched way. E.g. if one module exports a function named w2c_modname_banana with one type signature, and another module imports a function named w2c_modname_banana with a different type signature, the user will be allowed to link the respective .o files together without any warning message and will get garbage at runtime. And that seems to break the guarantees of WebAssembly.

-------------

**How about this as a proposed solution?**
- The exported functions are defined with names like "w2c_export_" + module_prefix + nm:name
- The imported functions are declared with names like "w2c_import_" + mod:name + nm:name
- As in the current PR, it remains the host's job to define a function that matches each import declaration. But unlike the current PR, these functions would no longer be specific to each module_prefix.

This preserves the two-level hierarchy (which I think you want) but also prevents modules from "automatically" linking with random libc functions or with the exports of another function (which is our big fear).

wasm2c/examples/fac/fac.c
```diff
@@ -317,35 +314,37 @@ static u32 w2c_fac(u32 w2c_p0) {
...
-
-static void init_exports(void) {
-  /* export: 'fac' */
-  Z_facZ_fac = (&w2c_fac);
+u32 Z_facZ_fac(Z_fac_module_instance_t* module_instance, u32 w2c_p0) {
+  return w2c_fac(module_instance, w2c_p0);
 }
 
-void Z_fac_init(void) {
+void Z_fac_init_module(void) {
```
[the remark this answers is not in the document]
keithw: I think our view is that if the host wants module initialization (distinct from instance initialization) to be thread-safe, it will have to provide the appropriate locking in its runtime implementation. The default wasm-rt-impl.c isn't thread-safe because its implementation of `wasm_rt_register_func_type` isn't thread-safe -- I don't think it matters if there's a safety call to wasm_rt_init in there too. :-/

src/c-writer.cc
```diff
@@ -584,12 +648,25 @@ void CWriter::Write(const ExternalPtr& name) {
   }
 }
 
+void CWriter::Write(const ExternalInstancePtr& name) {
+  bool is_import = import_syms_.count(name.name) != 0;
+  if (is_import) {
+    Write("module_instance->", GetGlobalName(name.name));
+  } else {
+    Write("&module_instance->", GetGlobalName(name.name));
```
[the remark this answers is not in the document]
keithw: This is in the implementation of Write(ExternalInstancePtr), so the goal is to write a pointer. The imported elements are already pointers, so for those it can just write the name of the element. For something declared in the module itself, it needs to take the address of the element with the `&` operator.

wasm2c/README.md
sbc100: > Hmm, this still seems pretty scary to me. Maybe I'm off-base, so let me see if I can explain why. I put a proposed solution at the bottom that hopefully will satisfy your concern and also our fears...?
> 
> I hope I'm using the right language here. We see these three things as distinct:
> 
> 1. the CWriter "module_prefix," a string controlled by the caller of wasm2c (if unspecified, the module's name if it has one, or else the filename)
> 2. the "mod:name" of an import, a string that the module producer/author controls
> 3. the "nm:name" of an import or export, also a string that the module producer/author controls
> 
> In this PR, the **exported** functions turn into function definitions, with the function named "module_prefix" + "nm:name". I think we're all on the same page on that one.
> 
> What we're talking about here is the **imported** functions, which in the current PR become extern declarations referencing functions named with the form "module_prefix" + "mod:name" + "nm:name".
> 
> What I think you're suggesting is to rename them just "mod:name" + "nm:name" (removing the "module_prefix"). There's basically two reasons we're scared of this:
> 
> 1. The "mod:name" and "nm:name" are both totally under the control of the module producer/author. If the module chooses to import a function named "kill" from a mod:name "pthread", then CWriter would emit an extern declaration for a function named "pthread_kill" (assuming no "Z" here), and then if linked into an executable, the WebAssembly module would be able to call that function (or "rt_sigreturn" or "timer_delete"...) with no intervention or explicit decision to grant that capability by the host/embedder. That doesn't seem right -- and so we think the import symbols all should require a prefix that's **not** under the control of the module author/producer.

Given that we do have the Z_ prefix on both the module name and the field name I don't think this is (currently) a valid concern.   How about we revisit this namespace issue if/when we remove the `Z_` managing we can revisit this question.  For now the I think the name mangling avoids any collisions worries with system symbols.

> 2. Even if we prefixed all import and export symbols by some distinguished but constant prefix (e.g. "w2c_" + "mod:name" + "nm:name"), it's still scary because it means that two modules would be able to call one another in a possible signature-mismatched way. E.g. if one module exports a function named w2c_modname_banana with some type signature, and another module imports a function named w2c_modname_banana with a different type signature, the user will be allowed to link the respective .o files together without any warning message and will get garbage at runtime (since the symbols have the same name after we removed the signature-mangling). And that seems to break the guarantees of WebAssembly.

This does sounds like a problem, but maybe not one that needs to be solved as part of this PR.  The problem sounds like "how can we link wasm2c modules together in a type-safe way?"   I'm not sure the native static linker is powerfull enough be used as way to link mouldes together so maybe we need to build some higher level linking solution where imports and exports are handling at a high level (rather than relying on the static linker to link things together).   Given this issue perhaps removing the signature from the symbol name in #1896 was actually a bad idea?   

> 
> **How about this as a proposed solution?**
> 
> * The exported functions are defined with names like "w2c_export_" + module_prefix + nm:name
> * The imported functions are declared with names like "w2c_import_" + mod:name + nm:name
> * As in the current PR, it remains the host's job to define a function that matches each import declaration. But unlike the current PR, these functions would no longer be specific to each module_prefix.
> 
> This preserves the two-level hierarchy (which I think you want) but also prevents modules from "automatically" linking with random libc functions or with the exports of another function (which is our big fear).

Sure that sounds reasonable.   This is still not powerful enough to express the use case where the host night want to export two different versions of the same function to different modules, but I think to achieve that (as I say above) we would need to switch to a runtime linking system that allows the host to hook up imports and exports dynamically rather than statically.

If possible, could all of this be differed to separate PR and keep this PR focused just on instancing?

keithw: > If possible, could all of this be differed to separate PR and keep this PR focused just on instancing?

Okay -- just pushed a commit that does this. (Moves imports to just "mod:name" + "nm:name", with caveats above.) We can revisit the naming later, perhaps at the same time we remove "Z". Also updated wasm2c/README.md; the imported function names from the "host" module are now as you expected.

Re: structs of function pointers, I think for performance reasons we'd love to avoid a world where every call to an imported function has to go through a function pointer, but I admit I haven't actually tried to measure what the penalty would be there.

sbc100: Perhaps we need two different options for linking modules together:

1. Allow the host do set up imports dynamically allowing full flexibility at runtime
2. Have the static linker link everything together based on symbol names alone (i.e. matching up imports and exports symbolically).

Regarding the indirection .. don't we already use a pointer indirection for imports.   Looking at the spec test generated code for example I see:

```
extern void (*Z_spectestZ_print_i32)(u32); 
```

This is declaring a pointer to a function rather than a function so we already have that level of indirection there.  (Unless this change removes it?)

sbc100: Certainly seems like the linking issue (along with how type safety is enforced during linking) is something we should discuss further, outside of this issue.

keithw: > Perhaps we need two different options for linking modules together:
> 
>     1. Allow the host do set up imports dynamically allowing full flexibility at runtime
> 
>     2. Have the static linker link everything together based on symbol names alone (i.e. matching up imports and exports symbolically).

Maybe there's a way to enforce type safety in both settings somehow, by using the fact that we have a single global data structure of unique func types (and we know where every exported func falls in that data structure). But yes, happy to discuss later outside this issue.

> Regarding the indirection .. don't we already use a pointer indirection for imports. Looking at the spec test generated code for example I see:
> 
> ```
> extern void (*Z_spectestZ_print_i32)(u32); 
> ```
> 
> This is declaring a pointer to a function rather than a function so we already have that level of indirection there. (Unless this change removes it?)

Yes, all those pointers end up going away in this PR, at least as it currently stands -- not just the function pointers, but also the table, memory, and global pointers. The table, memory, and globals stop being pointers because now they're just members of the Z_spectest_module_instance_t structure, and any module that imports from spectest takes a Z_spectest_module_instance_t* as an argument to its init function, and stores that pointer in its own module instance structure. And the functions stop being pointers because now they're just ordinary functions that take a Z_spectest_module_instance_t* as their first argument, provided by the importing module when calling the function.

test/run-spec-wasm2c.py
```diff
@@ -105,6 +105,18 @@ def Mangle(match):
     return 'Z_' + re.sub(pattern, Mangle, s.encode('utf-8')).decode('utf-8')
 
 
+def MangleCWriterName(wasm_filename):
```
sbc100, line 108: How is this different to `MangleName` just above?   Why do we need to ways to mangle a name?

test/run-spec-wasm2c.py
```diff
@@ -122,10 +134,10 @@ def __init__(self, spec_json, prefix, out_file, out_dir):
         self.module_name_to_idx = {}
         self.module_prefix_map = {}
         self.unmangled_names = {}
-
-    def Write(self):
         self._MaybeWriteDummyModule()
```
sbc100, line 137: Why move this line from `Write` to `__init__`.. doesn't it make sense to do all the writing in `Write`?

test/run-spec-wasm2c.py
```diff
@@ -211,15 +250,19 @@ def _WriteCommand(self, command):
...
         self.module_idx += 1
-        self.out_file.write('%s_init();\n' % self.GetModulePrefix())
+        self.out_file.write('%s_init_module();\n' % self.GetModulePrefix())
+        self.out_file.write('%s_module_instance_t %s;\n' % (self.GetModulePrefix(), self.GetModuleInstanceName()))
+        self._WriteModuleInstanceImports(command, False)
 
     def _WriteModuleCleanUps(self):
-        for idx in range(1, self.module_idx):
-            self.out_file.write("%s_free();\n" % self.GetModulePrefix(idx))
+        for idx in range(0, self.module_idx):
```
sbc100, line 258: Why does this change from 1 to 0?

test/run-spec-wasm2c.py
```diff
@@ -324,9 +367,14 @@ def _Action(self, command):
         mangled_module_name = self.GetModulePrefix(action.get('module'))
         field = mangled_module_name + MangleName(action['field'])
         if type_ == 'invoke':
-            return '%s(%s)' % (field, self._ConstantList(action.get('args', [])))
+            args = self._ConstantList(action.get('args', []))
+            if len(args) == 0:
+                args = '&' + mangled_module_name + "_module_instance"
+            else:
+                args = '&' + mangled_module_name + "_module_instance" + ', ' + args
```
sbc100, line 374: Using single quotes for strings. 

We tend to avoid string addition like this and prefer `'&%s_module_instance, %s' % (mangled_module_name, args)`... but maybe its time to switch to fstrings and do `f'&{mangled_name}_module_instance, {args}'`?

test/run-spec-wasm2c.py
```diff
@@ -465,6 +508,11 @@ def main(args):
             if options.compile:
                 o_filenames.append(Compile(cc, c_filename, out_dir, includes))
 
+        cwriter.Write()
+        main_filename = utils.ChangeExt(json_file_path, '-main.c')
+        with open(main_filename, 'w') as out_main_file:
+            out_main_file.write(output.getvalue())
```
sbc100, line 514: These lines look like the just moved down?  Is that change needed?

test/run-spec-wasm2c.py
```diff
@@ -173,6 +187,31 @@ def _CacheModulePrefixes(self):
...
+        header_filename = utils.ChangeExt(command['filename'], '.h')
+        with open(os.path.join(self.out_dir, header_filename), encoding='utf-8') as f:
+            imported_modules = set()
+            for line in f:
+                if "import: " in line:
+                    line_split = line.split()
+                    import_module_name = MangleName(line_split[2][1:-1])
+                    imported_modules.add(import_module_name)
+
+            if (uninstantiable):
```
sbc100, line 200: Drop braces

test/run-spec-wasm2c.py
```diff
@@ -173,6 +187,31 @@ def _CacheModulePrefixes(self):
...
+                self.out_file.write("ASSERT_TRAP(")
+
+            if len(imported_modules) > 0:
```
sbc100, line 203: Just `if imported_modules:`

test/run-spec-wasm2c.py
```diff
@@ -173,6 +187,31 @@ def _CacheModulePrefixes(self):
...
 
+    def _WriteModuleInstanceImports(self, command, uninstantiable):
+        header_filename = utils.ChangeExt(command['filename'], '.h')
+        with open(os.path.join(self.out_dir, header_filename), encoding='utf-8') as f:
+            imported_modules = set()
+            for line in f:
+                if "import: " in line:
+                    line_split = line.split()
+                    import_module_name = MangleName(line_split[2][1:-1])
+                    imported_modules.add(import_module_name)
```
sbc100, line 198: It looks like the open file `f` is no longer needed after this line so everything below can/should be unintended perhaps?

test/run-spec-wasm2c.py
```diff
@@ -173,6 +187,31 @@ def _CacheModulePrefixes(self):
...
+
+            if (uninstantiable):
+                self.out_file.write("ASSERT_TRAP(")
+
+            if len(imported_modules) > 0:
+                self.out_file.write("%s_init(&%s_module_instance" % (self.GetModulePrefix(), self.GetModulePrefix()))
+                for imported_module in sorted(imported_modules):
+                    self.out_file.write(", &%s_module_instance" % imported_module)
+                self.out_file.write(")")
```
sbc100, line 207: Won't this code work fine even for the case where imported_modules is empty?  i.e. do you need the if/else here?

test/run-spec-wasm2c.py
```diff
@@ -173,6 +187,31 @@ def _CacheModulePrefixes(self):
                 self.module_prefix_map[name_idx] = name
                 self.unmangled_names[name_idx] = command['as']
 
+    def _WriteModuleInstanceImports(self, command, uninstantiable):
```
sbc100, line 190: The job of this function seems to be create the call to the init function?    Does it need to be renamed?  Maybe `_WriteModuleInitCall` or something like that?

test/run-spec-wasm2c.py
keithw: This was not being called -- it was apparently vestigial. I deleted it.

test/run-spec-wasm2c.py
keithw: Unfortunately not in this case -- `_MaybeWriteDummyModule` is doing a job here that's normally done by `wast2json` (writing out a wasm file). We need this to happen before `_CacheModulePrefixes` (so even the dummy module prefix is mapped) and that needs to happen before calling `GetModulePrefixUnmangled`, which is needed to run `wasm2c`.

And (and here's the change actually relevant to this PR) we need `wasm2c` to run before `cwriter.Write()`, because `_WriteModuleInstanceImports` wants to read the `.h` output of `wasm2c` to know which module types are needed in the arguments to the instance init function. These are the "imported-from" modules.

test/run-spec-wasm2c.py
keithw: This was my mistake in #1901; it wasn't caught in testing (by the asan memory leak detector) until the table stopped being a global data structure. (We were failing to free module zero...)

test/run-spec-wasm2c.py
keithw: Done.

test/run-spec-wasm2c.py
sbc100: Ah I see.  In that case maybe this can just be `range(self.module_idx)`... and maybe a separate bugfix PR if you want to land it right away?

test/run-spec-wasm2c.py
keithw: Yes (copying from above), we need wasm2c to run before cwriter.Write(), because _WriteModuleInstanceImports now wants to read the .h output of wasm2c to know which module types are needed in the arguments to the instance init function. These are the "imported-from" modules

test/run-spec-wasm2c.py
keithw: Done.

test/run-spec-wasm2c.py
keithw: Indeed -- change made.

test/run-spec-wasm2c.py
keithw: Good catch -- change made.

test/run-spec-wasm2c.py
keithw: Done.

src/c-writer.cc
```diff
@@ -848,57 +945,170 @@ void CWriter::WriteFuncTypes() {
   Write(CloseBrace(), Newline());
 }
 
-void CWriter::WriteImports() {
-  if (module_->imports.empty())
+void CWriter::ComputeUniqueImports() {
+  std::map<std::pair<std::string, std::string>, const Import*> import_map;
```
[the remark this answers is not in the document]

src/c-writer.cc
```diff
@@ -848,57 +945,170 @@ void CWriter::WriteFuncTypes() {
...
+  for (const Import* import : module_->imports) {
+    auto key = make_pair(import->module_name, import->field_name);
+    auto insertion_result = import_map.emplace(key, import);
```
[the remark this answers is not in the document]

test/run-spec-wasm2c.py
keithw: Sounds good, done here and will submit bugfix PR.

wasm2c/examples/fac/fac.c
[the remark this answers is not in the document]
shravanrn: I think the challenge here is that there seems to be three separate initializations with slightly different requirements, and unfortunately, with this change we are now tying two of these together explicitly

1. App initialization
**Goal:** Install things like the signal handler once for the application to prepare it to run wasm modules. In the future, some of the PRs we are hoping to submit will hopefully initialize any wasi runtime's one time init here too. 
**Use case and thread safety:** This doesn't have to be thread safe, and the use case of this would be to call this once in the beginning of the host app.

2. Module initialization
**Goal:** Do things like initialization of function types.
**Use case and thread safety:** In an ideal world the host app can chose to invoke this lazily during the first load of the module or once during startup (For the firefox use case, this would be done lazily). So one of the PR's i intend to submit here is to give this the same semantics of C++'s `std::call_once`. You are absolutely right in that this isn't thread_safe, but I think we should leave the door open to fix this going forward.

3. Instance intialization
**Goal:** Do things like setup the heap
**Use case and thread safety:** Called during instance creation. No thread safety needed as it should be called once per instance.

I believe we should clearly separate  App initialization and Module initialization by moving the call to `wasm_rt_init` outside the `init_module` function, so we leave the door open for the above changes. 

Please let me know your thoughts.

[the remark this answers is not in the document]
keithw: Ok, if you feel strongly, we'll change it. Our worry is that this creates a bit of a more dangerous API, where if the user doesn't remember to call `wasm_rt_init()`, then on Mac and Linux, everything proceeds seemingly happily, but a misbehaving Wasm module can now cause a real segfault that won't be caught (e.g. via stack overflow or memory or table OOB). If that's okay by all and the warning in the documentation is dire enough, fine with us.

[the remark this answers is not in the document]
sbc100: You could throw in an `assert(wasm_rt_is_initialized())` at the top of module init I guess?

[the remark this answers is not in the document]
keithw: Great, done.

wasm2c/README.md
```diff
@@ -287,25 +318,60 @@ It is only used on platforms that don't use the signal handler to detect
 exhaustion.
 
 ## Exported symbols
```
sbc100, line 320: Empty line after title here and below.

wasm2c/README.md
```diff
@@ -395,3 +461,118 @@ $ wat-desugar fac-flat.wat --fold -o fac-folded.wat
 
 The formatting is different and the variable and function names are gone, but
 the structure is the same.
+
+## Instantiate multiple instances of a module
```
sbc100, line 465: How about `Create multiple instances of a module` or `Instantiating a module multiple times`?

wasm2c/examples/fac/fac.c
```diff
@@ -317,35 +315,38 @@ static u32 w2c_fac(u32 w2c_p0) {
...
   init_func_types();
-  init_globals();
-  init_memory();
-  init_table();
-  init_exports();
 }
 
-void Z_fac_free(void) {
+void Z_fac_instantiate(Z_fac_module_instance_t* module_instance) {
+  assert(wasm_rt_is_initialized());
```
sbc100, line 345: Could we also assert that `Z_fac_init_module` has been called?

wasm2c/README.md
```diff
@@ -194,12 +216,15 @@ typedef void (*wasm_rt_funcref_t)(void);
...
-as returned by `wasm_rt_register_func_type` described below.
+as returned by `wasm_rt_register_func_type` described below. `module_instance`
+is the pointer to the module instance taht should be passed in when the func is
+called.
 
 ```c
 typedef struct {
   uint32_t func_type;
   wasm_rt_funcref_t func;
+  void* module_instance;
```
sbc100, line 227: I wonder if `wasm_rt_funcref_t` should be defined to include all these of these elements... and then `wasm_rt_elem_t` could evolve to be tagged union of `wasm_rt_funcref_t` with other possible table inhabitants?

wasm2c/README.md
keithw: Done.

wasm2c/README.md
keithw: Done ("Create multiple instances of a module")

wasm2c/examples/fac/fac.c
keithw: Done (by adding a static bool here: https://github.com/fixpointOS/wabt/blob/module_instancing/src/template/wasm2c.declarations.c#L317)

wasm2c/README.md
keithw: That is basically where we went in the later reference-types PR... happy to bring that in earlier if you want it now. (I don't know if later proposals will need the tagged union -- for reference-types, we didn't find we needed it because the code always knows statically whether it's dealing with a funcref table or an externref table.)

Here's where we were headed in the reference-types wasm-rt.h -- how do you feel about this / do you want this here or wait until later?

```cpp
/**
 * A generic function pointer type, both for Wasm functions (`code`)
 * and host functions (`hostcode`). All function pointers are stored
 * in this canonical form, but must be cast to their proper signature
 * to call.
 */
typedef void (*wasm_rt_function_ptr_t)(void);

/** A function instance (the runtime representation of a function).
 * These can be stored in tables of type funcref, or used as values. */
typedef struct {
  /** The index as returned from `wasm_rt_register_func_type`. */
  uint32_t func_type;
  /** The function. The embedder must know the actual C signature of the
   * function and cast to it before calling. */
  wasm_rt_function_ptr_t func;
  /** A function instance is a closure of the function over an instance
   * of the originating module. The module_instance element will be passed into
   * the function at runtime. */
  void* module_instance;
} wasm_rt_funcref_t;

/** An external reference. These are opaque have meaning only to the host. */
typedef void* wasm_rt_externref_t;

/** Default (null) value of an externref */
static const wasm_rt_externref_t wasm_rt_externref_null_value = NULL;
```

test/spec-wasm2c-prefix.c
```diff
@@ -202,59 +202,90 @@ static bool is_arithmetic_nan_f64(u64 x) {
...
 /*
  * spectest implementations
  */
-static void spectest_print(void) {
+void Z_spectestZ_print(Z_spectest_module_instance_t* module_instance) {
   printf("spectest.print()\n");
 }
 
-static void spectest_print_i32(uint32_t i) {
+void Z_spectestZ_print_i32(Z_spectest_module_instance_t* module_instance,
```
sbc100, line 223: I wonder if we should shorten this to `Z_spectest_instance_t* instance`?    i.e. is `modules_instance` needlessly verbose?

sbc100: (I think it might make this code more readable with less verbosity in these names .. but I could be overthinking).

wasm2c/README.md
```diff
@@ -73,23 +80,31 @@ int main(int argc, char** argv) {
...
   wasm_rt_init();
 
-  /* Initialize the fac module. */
-  Z_fac_init();
+  /* Initialize the `fac` module (this registers the module's function types in
+   * a global data structure) */
+  Z_fac_init_module();
+
+  /* Declare an instance of the `fac` module. */
+  Z_fac_module_instance_t module_instance;
```
sbc100, line 88: e.g. what if this was just `Z_fac_instance_t instance;`  ?

wasm2c/README.md
sbc100: I think we can leave this for now

test/spec-wasm2c-prefix.c
keithw: Agreed -- just pushed a rename from "module_instance" to "instance" (everywhere except the member of wasm_rt_funcref_t, where it seems useful to clarify that this is an instance of the originating module and not some other kind of instance).

wasm2c/README.md
keithw: Agreed (done).

src/c-writer.cc
```diff
@@ -1041,68 +1259,73 @@ void CWriter::WriteFuncDeclarations() {
...
   if (module_->memories.size() == module_->num_memory_imports)
     return;
 
-  Write(Newline());
-
   Index memory_index = 0;
   for (const Memory* memory : module_->memories) {
     bool is_import = memory_index < module_->num_memory_imports;
     if (!is_import) {
-      Write("static ");
```
sbc100, line 1105: Shouldn't this be static is the memory is not exported?

sbc100 (approved):
OK I think this is ready to land now.  Any final changes before we do so?

sbc100:
Ok!  Marking as auto-merge!

keithw:
> OK I think this is ready to land now. Any final changes before we do so?

Awesome! I don't think we have any final changes in this one. I just rebased it on the current tip-of-tree to make sure it applies cleanly and then rebased #1877 and #1887 on top of it.

I think after this the wasm2c roadmap from our perspective looks something like:

1. Bulk memory ops (#1877)
2. Reference types (#1887) [at this point, wasm2c will pass all the current Wasm tests (+ multi-memory and exceptions) except SIMD, and it would be cool to get it added to https://webassembly.org/roadmap/]

Followed by more speculative changes:
- Changes necessary to accommodate Firefox/RLBox (we chatted a bit with talg recently)
- SIMD [at this point, wasm2c would support all of WebAssembly 2.0 + multi-memory and exceptions]
- Adding an instance "reset" function (in addition to "initialize" and "free") to allow the runtime to to take an instance and more efficiently restore it to initial state without having to free and reallocate all the memories.
- Adding a mechanism (maybe in the IR or in the CWriter options) to tag memories for whether they should get hardware bounds-checked (with mprotect and the signal handler) or software bounds-checked (with explicit checks on every load/store) on a memory-by-memory basis. This would be helpful if the runtime has a sense of how it wants to trade off initialization latency vs. runtime speed (we are using this already in Fix).

PR #1818, opened 2022-02-01
wasm-objdump: Print local names during disassembly

### LINKED ISSUE #1815: Local names in wasm-objdump -x
So far as I can tell, wasm-objdump doesn't use local names when printing instructions such as local.get. I wanted to check if this was so (rather than, for instance, my test file specifying the names incorrectly -- wasm2wat includes the names, so I assume this isn't the case), and whether there would be any interest in changing the behaviour to include the names of locals?

mwilliamson:
Fixes #1815. Some example output:

```
module.wasm:	file format wasm 0x1

Code Disassembly:

000027 func[0] <FIRST>:
 000028: 01 7f                      | local[0] type=i32
 00002a: 20 01                      | local.get 1 <local0>
 00002c: 20 00                      | local.get 0 <arg0>
 00002e: 1a                         | drop
 00002f: 1a                         | drop
 000030: 0b                         | end
```

I haven't written a test for this yet, but I wanted to check that my approach was correct, not least because it's been a while since I've written any C++.

(Unfortunately, this doesn't actually fully solve my use case since I think wasm-ld doesn't preserve local names when linking? I haven't had a chance to investigate this yet, but mention it in case anyone knows off the top of their head.)

### DIFF
--- a/src/binary-reader-objdump.cc
+++ b/src/binary-reader-objdump.cc
@@ -54,6 +54,7 @@ class BinaryReaderObjdumpBase : public BinaryReaderNop {
  protected:
   string_view GetFunctionName(Index index) const;
   string_view GetGlobalName(Index index) const;
+  string_view GetLocalName(Index function_index, Index local_index) const;
   string_view GetSectionName(Index index) const;
   string_view GetTagName(Index index) const;
   string_view GetSymbolName(Index index) const;
@@ -143,6 +144,11 @@ string_view BinaryReaderObjdumpBase::GetFunctionName(Index index) const {
 
 string_view BinaryReaderObjdumpBase::GetGlobalName(Index index) const {
   return objdump_state_->global_names.Get(index);
+}
+
+string_view BinaryReaderObjdumpBase::GetLocalName(Index function_index,
+                                                  Index local_index) const {
+  return objdump_state_->local_names.Get(function_index, local_index);
 }
 
 string_view BinaryReaderObjdumpBase::GetSectionName(Index index) const {
@@ -275,6 +281,13 @@ class BinaryReaderObjdumpPrepass : public BinaryReaderObjdumpBase {
       default:
         break;
     }
+    return Result::Ok;
+  }
+
+  Result OnLocalName(Index function_index,
+                     Index local_index,
+                     string_view local_name) override {
+    SetLocalName(function_index, local_index, local_name);
     return Result::Ok;
   }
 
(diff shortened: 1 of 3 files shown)


src/binary-reader-objdump.cc
```diff
@@ -500,6 +520,7 @@ class BinaryReaderObjdumpDisassemble : public BinaryReaderObjdumpBase {
   Offset last_opcode_end = 0;
   int indent_level = 0;
   Index next_reloc = 0;
+  Index function_index = 0;
```
sbc100, line 523: How about `current_function_index` to match the members above?

sbc100 (commented):
Looks good to me!

I'm actually surprised none of the test output changed.  Did you run `./test/run-tests.sh -r`?

I think the test to update would be `test/dump/extended-names.txt` BTW.

sbc100:
Regarding wasm-ld, its very unlikely llvm to wasm-ld will be able to give useful names to locals.  At least not ones that comes from C++ or llvm-ir, because the wasm-level locals don't normally correspond any language level construct.  

The optimizers at various levels might choose to store many different things in given local over the course of the function, so even for things that fit in locals its probably not going to be useful to mapping higher level language names.

It still useful for hand-written stuff though.

src/binary-reader-objdump.cc
mwilliamson: Will do!

mwilliamson:
> I'm actually surprised none of the test output changed. Did you run `./test/run-tests.sh -r`?

I thought I did, I could easily be mistaken though.

> I think the test to update would be `test/dump/extended-names.txt` BTW.

Great, I'll take a look. Thanks!

> Regarding wasm-ld, its very unlikely llvm to wasm-ld will be able to give useful names to locals. At least not ones that comes from C++ or llvm-ir, because the wasm-level locals don't normally correspond any language level construct.

Just to explain my use case: I've written my own toy compiler than outputs Wasm object files, which includes local names, so having llvm-ld preserve those would be helpful for debugging.

sbc100:
> Just to explain my use case: I've written my own toy compiler than outputs Wasm object files, which includes local names, so having llvm-ld preserve those would be helpful for debugging.

Oh sure, that sounds do-able.  But yes, probably needs a wasm-ld patch to make it work.  Would be cool if we could extant that to work with `.s` format in llvm.. but that might be little more work.

mwilliamson:
I've had a go at updating `test/dump/extended-names.txt`, so I think this is ready to go, but happy to change things as needed. Thanks for the help so far.

sbc100 (approved):

PR #1819, opened 2022-02-01
objdump: Consider params when printing local decl index

mwilliamson:
Currently, objdump -d will start at index 0 when printing local declarations e.g. `local[0] type=i64`, regardless of whether the function has any params. To me, it feels like starting the index at the param count to match the index used in opcodes and elsewhere would be less confusing.


sbc100 (approved):

sbc100:
Yup that makes sense to me.

mwilliamson:
Great, thanks!

Issue #1811, opened 2022-01-21
Move to c++ 17


dbezhetskov:
Some features are really hard to implement without typed unions like std::variant and other features from c++17.

For example, with typed function references and gc proposal we have named types during parsing `(ref null $foo_func)` and `(ref null $struct_type)` but we can't just add `std::string` to struct Type because we use Type as a member of unions in `struct InitExpr` and `struct ActionResult`. So, in the end we need to maintain maps: type_name -> its index during parsing and this is awful because possibly any thing in WebAssembly from local and global to function and table can have named type.

It is 2022 and we still use c++ 2011, so I propose to move to c++17, wdyt? @sbc100

sbc100:
Sounds reasonable to me yes.   Both chromium and binaryen both recently switch to C++17.   I think we could do that same here most likely.

Would you like to send an initial PR?

PR #1825, opened 2022-02-10
Move to C++17

sbc100:
Fixes: #1811


sbc100:
The first actual change to use C++17 is the conversion to standard string_view: #1826

kripken (approved):
This sounds good to me, but perhaps we should check with users and people building the code in various places. I'm not sure how best to reach them though.

PR #1829, opened 2022-02-12
wasm2c: bounds-check active data loads

keithw:
The bulk-memory proposal changed OOB access (when loading an active data segment) from a validation failure to an initialization-time trap, but wasm2c wasn't checking OOB access in `LOAD_DATA()`.

When `WASM_RT_MEMCHECK_SIGNAL_HANDLER` is unset, this could produce memory corruption. When `WASM_RT_MEMCHECK_SIGNAL_HANDLER` is set, this led to failure in the new “data” tests that verify that modules with zero-length active data segments are uninstantiable if `offset > mem.size`.

This commit adds an unconditional bounds-check and restore the `test/wasm2c/spec/data.txt` test (one of the failing tests tracked at #1737).

(It's possible to avoid the **unconditional** bounds-check if we made `CWriter::WriteDataInitializers` more complicated to special-case the situation of a zero-length data segment. Then non-zero-length data segments could be loaded normally with `MEMCHECK` (which is a nop on platforms using `WASM_RT_MEMCHECK_SIGNAL_HANDLER`), and only zero-length segments would require the unconditional bounds-check (`RANGE_CHECK`). But given that we're talking about something that happens only once per segment on module instantiation, that extra complexity didn't seem worth it.)


sbc100 (approved):

sbc100:
Thanks for working on this.   Good to see some of those tests being re-enabled.

PR #1830, opened 2022-02-14
Improve Garbage Collection algorithm

zherczeg:
The algorithm is made partially recursive, and requires only a single pointer in the Store object.

### DIFF
--- a/src/interp/interp.cc
+++ b/src/interp/interp.cc
@@ -246,8 +246,11 @@ void Store::DeleteRoot(RootList::Index index) {
 
 void Store::Collect() {
   size_t object_count = objects_.size();
-  marks_.resize(object_count);
-  std::fill(marks_.begin(), marks_.end(), false);
+
+  assert(gc_context_.call_depth == 0);
+
+  gc_context_.marks.resize(object_count);
+  std::fill(gc_context_.marks.begin(), gc_context_.marks.end(), false);
 
   // First mark all roots.
   for (RootList::Index i = 0; i < roots_.size(); ++i) {
@@ -260,31 +263,44 @@ void Store::Collect() {
     thread->Mark();
   }
 
-  // TODO: better GC algo.
-  // Loop through all newly marked objects and mark their referents.
-  std::vector<bool> all_marked(object_count, false);
-  bool new_marked;
-  do {
-    new_marked = false;
-    for (size_t i = 0; i < object_count; ++i) {
-      if (!all_marked[i] && marks_[i]) {
-        all_marked[i] = true;
-        objects_.Get(i)->Mark(*this);
-        new_marked = true;
-      }
-    }
-  } while (new_marked);
+  // This vector is often empty since the default maximum
+  // recursion is usually enough to mark all objects.
+  while (WABT_UNLIKELY(!gc_context_.untraced_objects.empty())) {
+    size_t index = gc_context_.untraced_objects.back();
+
+    assert(gc_context_.marks[index]);
+    assert(gc_context_.call_depth == 0);
+
+    gc_context_.untraced_objects.pop_back();
+    objects_.Get(index)->Mark(*this);
+  }
+
+  assert(gc_context_.call_depth == 0);
 
   // Delete all unmarked objects.
   for (size_t i = 0; i < object_count; ++i) {
-    if (objects_.IsUsed(i) && !all_marked[i]) {
+    if (objects_.IsUsed(i) && !gc_context_.marks[i]) {
       objects_.Delete(i);
     }
   }
 }
 
 void Store::Mark(Ref ref) {
-  marks_[ref.index] = true;
+  size_t index = ref.index;
+
+  if (gc_context_.marks[index])
+    return;
+
+  gc_context_.marks[index] = true;
+
+  if (WABT_UNLIKELY(gc_context_.call_depth >= max_call_depth)) {
+    gc_context_.untraced_objects.push_back(index);
+    return;
+  }
+
+  gc_context_.call_depth++;
+  objects_.Get(index)->Mark(*this);
+  gc_context_.call_depth--;
 }
 
 void Store::Mark(const RefVec& refs) {
--- a/src/interp/interp.h
+++ b/src/interp/interp.h
@@ -506,12 +506,20 @@ class Store {
   template <typename T>
   friend class RefPtr;
 
+  struct GCContext {
+    int call_depth = 0;
+    std::vector<bool> marks;
+    std::vector<size_t> untraced_objects;
+  };
+
+  static const int max_call_depth = 10;
+
   Features features_;
+  GCContext gc_context_;
   // This set contains the currently active Thread objects.
   std::set<Thread*> threads_;
   ObjectList objects_;
   RootList roots_;
-  std::vector<bool> marks_;
 };
 
 template <typename T>
--- a/src/test-interp.cc
+++ b/src/test-interp.cc
@@ -687,6 +687,35 @@ TEST_F(InterpGCTest, Collect_InstanceExport) {
   EXPECT_EQ(after_new, store_.object_count());
 }
 
+TEST_F(InterpGCTest, Collect_DeepRecursion) {
+  const size_t table_count = 65;
+
+  TableType tt = TableType{ValueType::ExternRef, Limits{1}};
+
+  // Create a chain of tables, where each contains
+  // a single reference to the next table.
+
+  Table::Ptr prev_table = Table::New(store_, tt);
+
+  for (size_t i = 1; i < table_count; i++) {
+    Table::Ptr new_table = Table::New(store_, tt);
+
+    new_table->Set(store_, 0, prev_table->self());
+
+    prev_table.reset();
+    prev_table = std::move(new_table);
+  }
+
+  store_.Collect();
+  EXPECT_EQ(table_count + 1, store_.object_count());
+
+  // Remove the last root, now all should be removed.
+  prev_table.reset();
+
+  store_.Collect();
+  EXPECT_EQ(1u, store_.object_count());
+}
+
 // TODO: Test for Thread keeping references alive as locals/params/stack values.
 // This requires better tracking of references than currently exists in the
 // interpreter. (see TODOs in Select/LocalGet/GlobalGet)


src/interp/interp.cc
```diff
@@ -245,9 +245,17 @@ void Store::DeleteRoot(RootList::Index index) {
...
-  marks_.resize(object_count);
-  std::fill(marks_.begin(), marks_.end(), false);
+
+  GCContext context;
+
+  context.call_depth = 0;
+  context.marks.resize(object_count);
+  std::fill(context.marks.begin(), context.marks.end(), false);
+
+  gc_context_ = &context;
```
sbc100, line 258: This line sets of my danger detector, even though its been set to null at the end.    I guess passing `gc_context_` explicitly to all the `Mark` would be messier?

src/interp/interp.cc
```diff
@@ -245,9 +245,17 @@ void Store::DeleteRoot(RootList::Index index) {
...
+
   size_t object_count = objects_.size();
-  marks_.resize(object_count);
-  std::fill(marks_.begin(), marks_.end(), false);
+
+  GCContext context;
+
+  context.call_depth = 0;
+  context.marks.resize(object_count);
+  std::fill(context.marks.begin(), context.marks.end(), false);
```
sbc100, line 256: Is this line needed?  Wont the resize above take care of initializing to false?  (https://stackoverflow.com/questions/22983707/stdvectorbool-guaranteed-to-default-all-entries-to-false/22984114)

src/interp/interp.cc
```diff
@@ -260,31 +268,46 @@ void Store::Collect() {
...
-  } while (new_marked);
+  // This vector is often empty since the default maximum
+  // recursion is usually enough to mark all objects.
+  while (!context.unmarked_objects.empty()) {
+    size_t index = context.unmarked_objects.back();
+
+    assert(context.marks[index]);
+    assert(context.call_depth == 0);
+
+    context.unmarked_objects.pop_back();
```
sbc100, line 279: But this line next to the `back()` above since they are kind of part of the same operation?   (I still don't get why we get and pop the pack in single call).

src/interp/interp.cc
sbc100: Does C++17 have some way to automatically unassigned a member at the end of a scope maybe?

src/interp/interp.cc
```diff
@@ -260,31 +268,46 @@ void Store::Collect() {
...
-  marks_[ref.index] = true;
+  size_t index = ref.index;
+
+  if (gc_context_->marks[index])
+    return;
+
+  gc_context_->marks[index] = true;
+
+  if (gc_context_->call_depth >= max_call_depth) {
+    gc_context_->unmarked_objects.push_back(index);
```
sbc100, line 304: It looks like `index` is "marked" above on line 301 before being added to "unmarked_objects" .. perhaps this vector should be called something something more accurate.. obviously "marked_objects_that_have_not_yet_been_recursively_traced" won't work.   How about "untraced_objects"?

src/interp/interp.cc
```diff
@@ -260,31 +268,46 @@ void Store::Collect() {
...
-    for (size_t i = 0; i < object_count; ++i) {
-      if (!all_marked[i] && marks_[i]) {
-        all_marked[i] = true;
-        objects_.Get(i)->Mark(*this);
-        new_marked = true;
-      }
-    }
-  } while (new_marked);
+  // This vector is often empty since the default maximum
+  // recursion is usually enough to mark all objects.
```
sbc100, line 272: Do we even have test cases that trigger this condition?  If we don't I guess we should construct at least one?

sbc100 (commented):
Nice! Thanks for working on this.

src/interp/interp.cc
zherczeg: Perhaps I could pass a reference of the `gc_context_` pointer to the GCContext constructor, and its destructor resets its to nullptr. Or creating a special class for this purpose. Which one do you prefer?

src/interp/interp.cc
zherczeg: It was in the original code. I can remove it of course.

src/interp/interp.cc
zherczeg: I also find it strange, but it seems `pop_back` is a void function:
https://www.cplusplus.com/reference/vector/vector/pop_back/

I would prefer to do the pop and return the popped item in one step if there is an operation for that.

src/interp/interp.cc
zherczeg: Good idea!

src/interp/interp.cc
sbc100: Right, but in the original code the `marks_` object persisted between calls to Collect.   In this new code `context.marks` is new on each call. 

Speaking of which would it make sense to keep the `GCContext` in the Store rather than creating it anew each `Collect`?

src/interp/interp.cc
zherczeg: Good suggestion. What is the easiest way to construct such a chain? Tables, which contain a single element to another Table? If I need to do it in C++ level, where should I put the test?

Thank you for the review!

sbc100: I'm not sure I'm afraid, I'm not to familiar with the the GC/interpreter yet.

I think `InterpGCTest` in `src/test-interp.cc` is probably a good place to start.

src/interp/interp.cc
zherczeg: Could be, but that increases memory consumption.

I have another suggestion: what about making this class public as a member of `Store`. The name could be `MarkContext` or something else. Then move the `Mark(object)` function from `Store` to this new class, and change the `Mark()` virtual function of Object to get this class as an argument rather than the `Store`. The problem is, you can call the `Mark()` function anytime, which is incorrect if you are not in the `Collect()` function. This way the compiler protects you from misusing `Mark()`, because it simply does not exist outside of `Collect()` function.

sbc100: What do you mean "increases memory consumption." ?  With the old code the "marks_" object was always allocated as part of the store and always as large as the number of objects from the most recent collection.     Storing GCContext in directly in the store seems pretty much exactly the same size.

I also think its a reasonable to have the `GCContext` persist otherwise it will just need to get relocated on the heap constantly.

zherczeg: I know the old code always used some extra memory, and I wanted to reduce it. But we can keep the current approach of course, and I think it is not worth to introduce the `GCContext` in this case. The advantage of the current approach is that if the object count does not change, the reallocation is basically free. Ok, I keep the current approach.

src/interp/interp.cc
zherczeg: You are right. I have added a test. But before other changes, I will wait your reply what should we do with GCContext.

sbc100: I recommend defining GCContext as you have, but including an instance in the Store rather than a pointer.  That way you get (a) fewer allocations and (b) no lifetime issues with pointers to stack data.

src/interp/interp.cc
```diff
@@ -260,31 +264,44 @@ void Store::Collect() {
...
+  while (!gc_context_.untraced_objects.empty()) {
```
clover2123, line 269: May I add some review?
It seems that wabt adopts UNLIKELY macro (`WABT_UNLIKELY`), so how about using it for this uncommon condition?

src/interp/interp.cc
```diff
@@ -260,31 +264,44 @@ void Store::Collect() {
...
+
+  if (gc_context_.marks[index])
+    return;
+
+  gc_context_.marks[index] = true;
+
+  if (gc_context_.call_depth >= max_call_depth) {
+    gc_context_.untraced_objects.push_back(index);
+    return;
+  }
```
clover2123, line 300: This if statement is also uncommon, so it looks suitable to use UNLIKELY here too.

src/interp/interp.h
```diff
@@ -506,12 +506,20 @@ class Store {
   template <typename T>
   friend class RefPtr;
 
+  struct GCContext {
+    int call_depth;
+    std::vector<bool> marks;
+    std::vector<size_t> untraced_objects;
```
clover2123, line 512: Just my suggestion.
It seems that these two vectors are allocated for calculation in GC(`Collect`) only, and never reclaimed.
If you concern about memory consumption, how about release these two vectors after `Collect`  calling like `shrink_to_fit()` vector function?

sbc100: But if collection happens frequently is there is advantage to grown in a shrinking these between collections?  Wouldn't that just add to fragmentation and cause the allocator do do extra work?

clover2123: I think that it depends on the situation.
If you put a high priority on memory usage, shrinking after GC would be a better choice.

To relieve fragmentation in shrinking, we can consider another option which just releases the whole space allocated for these vectors like below.
```c++
std::vector<bool>().swap(marks); // convert `marks` to an empty vector
```

AKAIK, `Collect` is not called inside wabt engine. So, user should call it explicitly to reclaim unused memory.
It would be great if GC is triggered automatically inside wabt engine (future work?)

src/interp/interp.cc
zherczeg: I have added these macros.

src/interp/interp.h
zherczeg: I have also wondering when calling GC automatically would be appropriate. It seems majority of GC objects are created by a user before validation, and should be released after a byte code is released. It seems the engine creates the trap object, but that should be very rare. Is there any object type which is created frequently by the engine? Maybe the frame object?

zherczeg:
Please let me know if any more changes are needed. Thank you!

src/interp/interp.cc
```diff
@@ -203,6 +203,7 @@ Store::Store(const Features& features) : features_(features) {
   Ref ref{objects_.New(new Object(ObjectKind::Null))};
   assert(ref == Ref::Null);
   roots_.New(ref);
+  gc_context_.call_depth = 0;
```
sbc100, line 206: This should be the definition of GCContext I think.  (can you just do `int call_depth = 0;` there?)

src/test-interp.cc
```diff
@@ -687,6 +687,35 @@ TEST_F(InterpGCTest, Collect_InstanceExport) {
   EXPECT_EQ(after_new, store_.object_count());
 }
 
+TEST_F(InterpGCTest, Collect_DeepRecursion) {
+  const size_t table_count = 65;
+
+  TableType tt = TableType{ValueType::ExternRef, Limits{1}};
+
+  // Create a chain of tables, where each contains
+  // a single reference to the next table.
```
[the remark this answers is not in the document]
zherczeg: I just exploited the fact that a table is an array of references :) Is there a better structure for this purpose?

sbc100 (approved):

PR #1834, opened 2022-02-17
wasm2c: run multi-memory tests

keithw:
~~This PR is sequenced behind #1810 and #1829.~~

PR #1751 added support for multi-memory and ran the multi-memory tests with `run-interp-spec`. This PR also runs (some of) the multi-memory tests under wasm2c, with `run-spec-wasm2c`. Only 5 of the 9 multi-memory tests are supported by wasm2c right now; the others require reference types, multi-table, or bulk memory.

### DIFF
--- a/src/tools/wasm2c.cc
+++ b/src/tools/wasm2c.cc
@@ -56,6 +56,13 @@ examples:
   $ wasm2c test.wasm --no-debug-names -o test.c
 )";
 
+static const std::string supported_features[] = {"multi-memory"};
+
+static bool IsFeatureSupported(const std::string& feature) {
+  return std::find(std::begin(supported_features), std::end(supported_features),
+                   feature) != std::end(supported_features);
+};
+
 static void ParseOptions(int argc, char** argv) {
   OptionParser parser("wasm2c", s_description);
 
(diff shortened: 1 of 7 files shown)


keithw:
(Rebased after #1810 and #1829 were merged.)

src/tools/wasm2c.cc
```diff
@@ -94,6 +94,7 @@ static void ParseOptions(int argc, char** argv) {
     exit(1);
   }
   s_features.disable_bulk_memory();
+  s_features.enable_multi_memory();
```
sbc100, line 97: I'm not sure we want to enable this in add cases.  The tests that need it can/should probably enable it on the command line  (see  test/spec/multi-memory/data.txt for example)

test/run-spec-wasm2c.py
```diff
@@ -386,6 +386,7 @@ def main(args):
             error_cmdline=options.error_cmdline)
         wast2json.verbose = options.print_cmd
         wast2json.AppendOptionalArgs({'-v': options.verbose})
+        wast2json.AppendArg('--enable-multi-memory')
```
sbc100, line 389: Again, this should probably go in the test file

sbc100 (commented):
Nice!

lgtm % comments

test/run-spec-wasm2c.py
keithw: Can do (if I understood your comment above correctly) -- we'll also have to plumb through the selectable features in run-spec-wasm2c.py. It's not as sophisticated as the interpreter runner and I don't think it supports command-line enable/disable arguments in the test file yet.

sbc100: I think it probably makes sense to pass features through to wasm2c, yes.   Mostly because all the other wabt tools do that.

src/tools/wasm2c.cc
```diff
@@ -82,13 +89,15 @@ static void ParseOptions(int argc, char** argv) {
 
   // TODO(binji): currently wasm2c doesn't support any non-default feature
   // flags.
```
sbc100, line 91: This comment is now stale

sbc100 (approved):

PR #1835, opened 2022-02-21
Implement a separate free list for Refs

zherczeg:
Create a memory optimized free list for Refs in the interpreter.

This patch is based on a comment in the original Free List. The highest bit cannot be set for valid Ref indices, so it can be used to represent free items. If the patch is accepted, I will specialize the original free list for objects, where the lowest bit will represent unused slots.

### DIFF
--- a/src/interp/interp-inl.h
+++ b/src/interp/interp-inl.h
@@ -190,6 +190,44 @@ auto FreeList<T>::count() const -> Index {
   return list_.size() - free_.size();
 }
 
+//// FreeRefList ////
+inline FreeRefList::Index FreeRefList::New(Ref ref) {
+  if (free_head_ == 0) {
+    list_.push_back(ref);
+    return list_.size() - 1;
+  }
+
+  Index index = free_head_ - 1;
+  assert(!IsUsed(index));
+  free_head_ = list_[index].index & (freeBit - 1);
+  list_[index] = ref;
+  return index;
+}
+
+inline void FreeRefList::Delete(Index index) {
+  assert(IsUsed(index));
+  list_[index].index = free_head_ | freeBit;
+  free_head_ = index + 1;
+}
+
+inline bool FreeRefList::IsUsed(Index index) const {
+  return (list_[index].index & freeBit) == 0;
+}
+
+inline const Ref& FreeRefList::Get(Index index) const {
+  assert(IsUsed(index));
+  return list_[index];
+}
+
+inline Ref& FreeRefList::Get(Index index) {
+  assert(IsUsed(index));
+  return list_[index];
+}
+
+inline FreeRefList::Index FreeRefList::size() const {
+  return list_.size();
+}
+
 //// RefPtr ////
 template <typename T>
 RefPtr<T>::RefPtr() : obj_(nullptr), store_(nullptr), root_index_(0) {}
--- a/src/interp/interp.h
+++ b/src/interp/interp.h
@@ -435,18 +435,42 @@ class FreeList {
   //                 1 when the payload is the index of the next free object
   //   n: the payload
   //
-  // When T is a Ref (see Store::RootList below), we'd need to store the
-  // "is_free" bit in most-significant bit instead.
-  //
   std::vector<T> list_;
   std::vector<size_t> free_;
   std::vector<bool> is_free_;
 };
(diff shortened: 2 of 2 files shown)


src/interp/interp.h
```diff
@@ -443,10 +443,33 @@ class FreeList {
   std::vector<bool> is_free_;
 };
 
+class FreeRefList {
```
sbc100, line 446: I guess the comment in the TODO above can't should probably be updated?

src/interp/interp-inl.h
```diff
@@ -190,6 +190,44 @@ auto FreeList<T>::count() const -> Index {
...
+}
+
+inline Ref& FreeRefList::Get(Index index) {
+  assert(IsUsed(index));
+  return list_[index];
+}
+
+inline FreeRefList::Index FreeRefList::size() const {
+  return list_.size();
+}
```
sbc100, line 229: Some of these methods look identical to `FreeList<T>`..  can we avoid the duplication somehow?    Partial template specialization maybe?

src/interp/interp.h
```diff
@@ -443,10 +443,33 @@ class FreeList {
...
+
+  bool IsUsed(Index) const;
+
+  const Ref& Get(Index) const;
+  Ref& Get(Index);
+
+  Index size() const;  // 1 greater than the maximum index.
+
+ private:
+  static const size_t unusedStart = (SIZE_MAX >> 1) + 1;
```
sbc100, line 461: This probably warrants a comment.    Also, why not give this type `Index` ?  

Also, would it be more clear to the reader if we instead defined freeBit and use bitwise logic elsewhere rather than greater than / less then operations on this start number?   e.g. something like `bool isFree = myindex & FREE_BIT`.   I don't free strongly but it seems like when using magic bit packing it makes sense to use bitwise logic.

sbc100:
Can I ask what motivated this change?  Are you just trying to address that TODO?  Or maybe you are seeing real world performance issues with the interpreter?

zherczeg:
Thank you for the review! The motivation is reducing the memory consumption, not performance, although this version should be faster. However, considering the whole runtime this is probably just a small fragment of it. Kind of a low-hanging fruit: you gain a little, loose nothing, why not. Gradual code improvement.

src/interp/interp.h
zherczeg: Ok, from cpu perspective, there is no difference.

src/interp/interp-inl.h
zherczeg: I am a c/assembly dev and c++ magic is a bit confusing for me. FreeList<T> has fields which are not used for Refs and vice versa, and I don't know how can I make unused fields disappear. If FreeList<T> will be also reworked, specializing will be easier. Shall I do the whole rework in one step?

sbc100:
> The motivation is reducing the memory consumption, not performance, although this version should be faster. However, considering the whole runtime this is probably just a small fragment of it. Kind of a low-hanging fruit: you gain a little, loose nothing, why not. Gradual code improvement.

Fair enough.   I'm a little concerned that  doing these kind of optimizations (for performance or memory consumption) is somewhat new to wabt and we don't currently have any way to measure the gain of prevent regressions.  Perhaps if we are going to continue this route we should add some improvements in the testing/measuring area as well.

zherczeg:
I like the idea of putting a benchmark, maybe I can set it up. However putting a good benchmark is not easy:
- Should cover real world use cases
- Licensing issues
- Runtime should not be too long / too short
- Minimal external dependencies

Usually data encryption, simulation, recursive computation, sorting algorithms are good for benchmarking.

sbc100:
> I like the idea of putting a benchmark, maybe I can set it up. However putting a good benchmark is not easy:
> 
> * Should cover real world use cases
> * Licensing issues
> * Runtime should not be too long / too short
> * Minimal external dependencies
> 
> Usually data encryption, simulation, recursive computation, sorting algorithms are good for benchmarking.

I agree a full benchmark would be complicated.   I'm hoping we could could start with basic measurement of simple instantiations.. although even that too is perhaps harder than it sounds.     Could we do something really simple like use `mallinfo` to measure high watermarks when we run the test suite?

PR #1836, opened 2022-02-22
Add missing wasm2c feature flags

SoniEx2:
See #1679 #1680 #1686


SoniEx2:
It would probably be a good idea to add tests for these at some point...

sbc100 (approved):
Yes please to tests!

PR #1838, opened 2022-02-23
Fix CI on main branch

SoniEx2:


sbc100 (approved):

keithw:
If possible, I think we'd love to just remove the "branches" section (as in #1814) so that CI will be run for people developing on other branches in forks. But I'm not sure if there is benefit to specifying just one branch...

SoniEx2:
oh hm, looks like CI is broken https://github.com/WebAssembly/wabt/runs/5311574587?check_suite_focus=true

at least it's *running* now... but hmm...

SoniEx2:
oh okay `lint` is just broken on the main branch we guess, but it never ran before so it's only being caught now. cc #1683 / @aheejin

sbc100:
> If possible, I think we'd love to just remove the "branches" section (as in [#1814 (comment)](https://github.com/WebAssembly/wabt/pull/1814#discussion_r796055373)) so that CI will be run for people developing on other branches in forks. But I'm not sure if there is benefit to specifying just one branch...

I imagine the reason for that being there is that we maybe wanted folks to be able to great branches but not run CI until the branch graduates to PR status.

We could remove the restriction probably.

But as a workaround I think can also create a PR on your fork `fork/branch -> fork/main` and that would trigger the CI.

sbc100:
The `./scripts/clang-format-diff.sh` script is designed to be run on PRs, and not on the full codebase.  The full codebase is not yet full lint clean by design (We have some cases where we still want to override the linter).

sbc100:
So I guess the lint job should move to a different workflow that only runs on PRs?

SoniEx2:
the clang-format-diff script runs on diffs, and pushes on the main branch are arguably diffs. we're not sure why it breaks when run like that.

we think it should still run, particularly because the lint job also includes flake8 - which works fine. you could run it only on PRs as a workaround, but also it also [breaks for anyone running a fork](https://github.com/SoniEx2/wasm2kotlin/runs/5312185205?check_suite_focus=true) so we really think it should be fixed somehow.

sbc100:
clang-format-diff tries to compare the current HEAD with its upstream branch so the logic would need to be updated somehow to handle the case when its just run on `main`:

```
if [ -n "$1" ]; then                                                                
  BRANCH="$1"                                                                       
elif [ -n "$GITHUB_BASE_REF" ]; then                                                
  BRANCH="origin/$GITHUB_BASE_REF"                                                  
else                                                                                
  BRANCH="@{upstream}"                                                              
fi                                                                                  
                                                                                    
FORMAT_MSG=$(git clang-format $MERGE_BASE -q --diff -- src/) 
```

The specific problem is trying to figure out that the upstream branch to compare to would be (called `$BRANCH in the script).   If you want to try and make it work thats fine.. but also we could just have it do nothing on pushes to a branch with no upstream?

SoniEx2:
when git runs a post-receive hook, it passes a list of refs and the "old" and "new" commit IDs. we're not sure if github actions has something similar...

aheejin:
I noticed we run lint only on PRs in Binaryen: WebAssembly/binaryen#3708
But I don't remember why it didn't make sense on commits, even though it says I reviewed it.. @sbc100 Do you remember?

SoniEx2:
we could always hardcode a "known-good" base revision and lint diffs based on that? :stuck_out_tongue_closed_eyes:

sbc100:
Since we accept PRs that fail this particular test (its not supposed to be strict in that sense, at least not yet), I think it might not be worth the effort to run it on ever push to a branch (that isn't a PR).

SoniEx2:
we think `github.event.before` should have the original rev, so if we could grab that somehow and inject it into the script...?

sbc100:
If you want to try and make that work sounds reasonable.

PR #1839, opened 2022-02-24
Fix lint CI on push

SoniEx2:
Because clang-format-diff is designed to check only the diff, this effectively only checks the relevant commit instead of the whole repo. So it should be fine to accept PRs that don't pass lint - the lint job will fail for that commit, but following pushes should pass.

Granted this isn't so relevant for the main repo where everyone goes through PRs, but for forks/etc it makes a difference. An alternative would be to just skip clang-format-diff on push (with something like `if: github.event_name != 'push'` instead of `env: ...`) but, again: forks/etc.

### DIFF
--- a/scripts/clang-format-diff.sh
+++ b/scripts/clang-format-diff.sh
@@ -3,10 +3,10 @@
 set -o errexit
 set -o pipefail
 
-if [ -n "$GITHUB_EVENT_BEFORE" ] && [ "push" = "$GITHUB_EVENT_NAME" ]; then
-  BRANCH="$GITHUB_EVENT_BEFORE"
-elif [ -n "$1" ]; then
+if [ -n "$1" ]; then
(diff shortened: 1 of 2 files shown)


scripts/clang-format-diff.sh
```diff
@@ -3,7 +3,9 @@
 set -o errexit
 set -o pipefail
 
-if [ -n "$1" ]; then
+if [ -n "$GITHUB_EVENT_BEFORE" ] && [ "push" = "$GITHUB_EVENT_NAME" ]; then
+  BRANCH="$GITHUB_EVENT_BEFORE"
+elif [ -n "$1" ]; then
```
sbc100, line 8: Can you put this second so that $1 always wins, if specified.

sbc100 (approved):

sbc100 (approved):

PR #1840, opened 2022-02-24
Code Metadata

yuri91:
This PR adds support for Code Metadata.

See https://github.com/WebAssembly/tool-conventions/blob/main/CodeMetadata.md  for the specification.

In particular this pr implements the following:

- Parsing code metadata sections in BinaryReader, providing appropriate callbacks that a BinaryReaderDelegate can implement:
    - BinaryReaderObjdump: show the sections in a human-readable form
    - BinaryReaderIr: add code metadata in the IR as expressions
- Parsing code metadata annotations in text format, adding them in the IR like the BinaryReaderIR does
- Writing the code metadata present in the IR in the proper sections when converting IR to binary
- Support in wasm-decompiler for showing code metadata as comments in the pseudo-code

All the features have corresponding tests.

Support for code metadata is gated through the --enable-code-metadata feature. For reading/writing in the text format, --enable-annotations is also required.

Missing features:

    Support for function-level code metadata (offset 0)
    Extensive validation in validator.cc (like making sure that all metadata instances are at the same code offset of an instruction)

### DIFF
--- a/src/binary-reader-ir.cc
+++ b/src/binary-reader-ir.cc
@@ -329,7 +329,6 @@ class BinaryReaderIR : public BinaryReaderNop {
   Result OnCodeMetadataFuncCount(Index count) override;
   Result OnCodeMetadataCount(Index function_index, Index count) override;
   Result OnCodeMetadata(Offset offset, const void* data, Address size) override;
-  Result EndCodeMetadataSection() override;
 
   Result OnTagSymbol(Index index,
                      uint32_t flags,
@@ -1539,10 +1538,6 @@ Result BinaryReaderIR::OnCodeMetadata(Offset offset,
       MakeUnique<CodeMetadataExpr>(current_metadata_name_, std::move(data_));
   meta->loc.offset = offset;
   code_metadata_queue_.push_metadata(std::move(meta));
-  return Result::Ok;
-}
-
-Result BinaryReaderIR::EndCodeMetadataSection() {
   return Result::Ok;
 }
 
--- a/src/binary-reader-ir.cc
+++ b/src/binary-reader-ir.cc
@@ -324,13 +324,10 @@ class BinaryReaderIR : public BinaryReaderNop {
                          uint32_t flags,
                          Index section_index) override;
   /* Code Metadata sections */
-  Result BeginCodeMetadataSection(std::string_view name,
-                                    Offset size) override;
+  Result BeginCodeMetadataSection(std::string_view name, Offset size) override;
   Result OnCodeMetadataFuncCount(Index count) override;
   Result OnCodeMetadataCount(Index function_index, Index count) override;
-  Result OnCodeMetadata(Offset offset,
-                          const void* data,
-                          Address size) override;
+  Result OnCodeMetadata(Offset offset, const void* data, Address size) override;
   Result EndCodeMetadataSection() override;
 
   Result OnTagSymbol(Index index,
@@ -1518,7 +1515,7 @@ Result BinaryReaderIR::OnLocalNameLocalCount(Index index, Index count) {
 }
 
 Result BinaryReaderIR::BeginCodeMetadataSection(std::string_view name,
-                                                  Offset size) {
+                                                Offset size) {
   current_metadata_name_ = name;
   return Result::Ok;
 }
@@ -1527,19 +1524,18 @@ Result BinaryReaderIR::OnCodeMetadataFuncCount(Index count) {
   return Result::Ok;
 }
 
-Result BinaryReaderIR::OnCodeMetadataCount(Index function_index,
-                                             Index count) {
+Result BinaryReaderIR::OnCodeMetadataCount(Index function_index, Index count) {
   code_metadata_queue_.push_func(module_->funcs[function_index]);
   return Result::Ok;
 }
 
 Result BinaryReaderIR::OnCodeMetadata(Offset offset,
-                                        const void* data,
-                                        Address size) {
+                                      const void* data,
+                                      Address size) {
   std::vector<uint8_t> data_(static_cast<const uint8_t*>(data),
                              static_cast<const uint8_t*>(data) + size);
-  auto meta = MakeUnique<CodeMetadataExpr>(current_metadata_name_,
-                                            std::move(data_));
+  auto meta =
+      MakeUnique<CodeMetadataExpr>(current_metadata_name_, std::move(data_));
   meta->loc.offset = offset;
   code_metadata_queue_.push_metadata(std::move(meta));
   return Result::Ok;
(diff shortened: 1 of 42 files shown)


src/binary-reader-ir.cc
```diff
@@ -43,6 +44,51 @@ struct LabelNode {
...
+  void push_func(Func* f) { entries.emplace_back(f); }
+  void push_metadata(std::unique_ptr<CodeMetadataExpr> meta) {
+    assert(!entries.empty());
+    entries.back().func_queue.push_back(std::move(meta));
+  }
+
+  std::unique_ptr<CodeMetadataExpr> pop_match(Func* f, Offset offset) {
+    std::unique_ptr<CodeMetadataExpr> ret;
+    if (entries.empty())
+      return ret;
```
sbc100, line 67: IIRC, Wabt coding style always uses { } , even for single line block.

src/binary-reader-ir.cc
```diff
@@ -660,6 +720,7 @@ Result BinaryReaderIR::OnFunctionBodyCount(Index count) {
 
 Result BinaryReaderIR::BeginFunctionBody(Index index, Offset size) {
   current_func_ = module_->funcs[index];
+  current_func_->loc = GetLocation();
```
sbc100, line 723: Seems odd that this was not needed before.. do you know why this is needed now?

src/binary-reader-ir.cc
```diff
@@ -1447,6 +1517,38 @@ Result BinaryReaderIR::OnLocalNameLocalCount(Index index, Index count) {
...
+                             static_cast<const uint8_t*>(data) + size);
+  auto meta = MakeUnique<CodeMetadataExpr>(current_metadata_name_,
+                                            std::move(data_));
+  meta->loc.offset = offset;
+  code_metadata_queue_.push_metadata(std::move(meta));
+  return Result::Ok;
+}
+
+Result BinaryReaderIR::EndCodeMetadataSection() {
+  return Result::Ok;
```
sbc100, line 1549: Maybe just remove this ?

src/binary-reader-logging.cc
```diff
@@ -638,6 +638,22 @@ Result BinaryReaderLogging::OnComdatEntry(ComdatType kind, Index index) {
...
+Result BinaryReaderLogging::BeginCodeMetadataSection(std::string_view name,
+                                                       Offset size) {
+  LOGF("BeginCodeMetadataSection('" PRIstringview "', size:%" PRIzd ")\n",
+       WABT_PRINTF_STRING_VIEW_ARG(name), size);
+  Indent();
+  return reader_->BeginCodeMetadataSection(name, size);
+}
+Result BinaryReaderLogging::OnCodeMetadata(Offset code_offset,
+                                             const void* data,
+                                             Address size) {
```
sbc100, line 650: indentation looks wrong here.   Run `git clang-format origin/main` to fix this stuff?

src/binary-reader.cc
```diff
@@ -2258,6 +2260,57 @@ Result BinaryReader::ReadTagSection(Offset section_size) {
...
+  CHECK_RESULT(ReadCount(&num_funcions, "function count"));
+  CALLBACK(OnCodeMetadataFuncCount, num_funcions);
+
+  Index last_function_index = kInvalidIndex;
+  for (Index i = 0; i < num_funcions; ++i) {
+    Index function_index;
+    CHECK_RESULT(ReadCount(&function_index, "function index"));
+    ERROR_UNLESS(function_index >= num_func_imports_,
+                 "function import can't have metadata (got %" PRIindex ")",
+                 function_index);
```
sbc100, line 2277: Since this is a custom section I guess we should not completely fail if its malformed?   

(Update: I see tests below where it looks like there are just warnings... I guess that downgrade from error to warning ins somewhere in the custom section handling code?)

src/binary-reader.cc
```diff
@@ -2280,6 +2333,12 @@ Result BinaryReader::ReadCustomSection(Index section_index,
     CHECK_RESULT(ReadTargetFeaturesSections(section_size));
   } else if (section_name == WABT_BINARY_SECTION_LINKING) {
     CHECK_RESULT(ReadLinkingSection(section_size));
+  } else if (options_.features.code_metadata_enabled() &&
+             section_name.find(WABT_BINARY_SECTION_CODE_METADATA) == 0) {
```
sbc100, line 2337: Is there no `startswith` for C++ strings?

src/common.h
```diff
@@ -354,6 +354,12 @@ enum class SymbolBinding {
   Local = 2,
 };
 
+/* matches binary format, do not change */
+enum class BranchHintKind : uint32_t {
```
sbc100, line 358: Maybe elaborate on what this class represents?   Should we maybe create a new header called `code_annotations.h`?   We don't want folks reading this to think it relates to branches in the wabt codebase itself.

sbc100 (commented):
Looking good!

src/binary-reader.cc
yuri91: it seems that they added it in c++20. But wabt uses c++17 currently.

src/common.h
yuri91: good catch. It is a remnant of a previous implementation, but it is not used anymore. I will remove it.

src/binary-reader.cc
yuri91: Yes, the PrintError method automatically converts to warning if it is invoked when parsing a custom section.

src/binary-reader-logging.cc
yuri91: I think I forgot to run clang-format after the last rebase. I will do it now.

src/binary-reader-ir.cc
yuri91: ok. I added it for completeness but it is not necessary, at least for now

src/binary-reader-ir.cc
yuri91: I am not sure why it wasn't used before, and maybe I am using it as a hack?

Anyway, It is now needed by CodeMetadataExprQueue: the offsets are always absolute but code metadata need relative offsets from the beginning of the function. So it seemed natural to fill this existing field here and use it to get relative offsets later

sbc100: Oh, actually you are adding this new `loc` field to `struct Func`.   I guess that makes sense for this use case and we already track this for other things such as expressions, and I guess the location a function is different from the location of its first expression due to param declaration?

sbc100 (approved):

src/binary-reader-ir.cc
yuri91: Sorry, I made this change some time ago and I didn't remember that I added the field!

But yes, it felt natural, since like you say it is there for similar things.

And also yes, the difference with the location of the first expression is that there are the locals in between.

Issue #1044, opened 2019-03-15
6 test failures on x86 32bit (i686)


rathann:
Running the 1.0.10 testsuite (`test/run-tests.py`) on x86 32bit (i686) yields the following failed tests:
```
- test/interp/unary.txt
  STDOUT MISMATCH:
  --- expected
  +++ actual
  @@ -21,7 +21,7 @@
   f64_abs() => f64:100.000000
   f64_sqrt_neg_is_nan() => i32:1
   f64_sqrt_100() => f64:10.000000
  -f64_ceil() => f64:-0.000000
  +f64_ceil() => f64:0.000000
   f64_floor() => f64:-1.000000
   f64_trunc() => f64:-0.000000
   f64_nearest_lo() => f64:1.000000
...
- test/wasm2c/spec/conversions.txt
  expected error code 0, got 1.
  STDERR MISMATCH:
  --- expected
  +++ actual
  @@ -0,0 +1,17 @@
  +Traceback (most recent call last):
  +  File "/builddir/build/BUILD/wabt-1.0.10/test/run-spec-wasm2c.py", line 407, in <module>
  +    sys.exit(main(sys.argv[1:]))
  +  File "/builddir/build/BUILD/wabt-1.0.10/test/run-spec-wasm2c.py", line 400, in main
  +    forward_stdout=True).RunWithArgs()
  +  File "/builddir/build/BUILD/wabt-1.0.10/test/utils.py", line 104, in RunWithArgs
  +    raise error
  +utils.Error: Error running "out/test/wasm2c/spec/conversions/conversions":
  +conversions-main.c:1394: assertion failed: in Z_conversions_0Z_i32Z2Ereinterpret_f32Z_if(make_nan_f32(0x200000)): expected 2141192192, got 2145386496.
  +conversions-main.c:1397: assertion failed: in Z_conversions_0Z_i32Z2Ereinterpret_f32Z_if(-make_nan_f32(0x200000)): expected 4288675840, got 4292870144.
  +conversions-main.c:1439: assertion failed: in Z_conversions_0Z_i64Z2Ereinterpret_f64Z_jd(make_nan_f64(0x4000000000000)): expected 9219994337134247936, got 9222246136947933184.
  +conversions-main.c:1442: assertion failed: in Z_conversions_0Z_i64Z2Ereinterpret_f64Z_jd(-make_nan_f64(0x4000000000000)): expected 18443366373989023744, got 18445618173802708992.
[...]

PR #1843, opened 2022-03-02
Add windows implementation of wasm2c runtime

sbc100:
All tests are now passing with cl.exe under x64.   With x86 there are some test failure that I believe relate
the use of the x87 registers to pass floating point numbers.   I suggest we look into fixing those as a followup.

Split out from #1833


wasm2c/wasm-rt-os-unix.c
```diff
@@ -0,0 +1,190 @@
...
+}
+
+void* os_mmap(void* hint, size_t size, int prot, int flags) {
+  int map_prot = PROT_NONE;
+  int map_flags = MAP_ANONYMOUS | MAP_PRIVATE;
+  uint64_t request_size, page_size;
+  uint8_t* addr;
+
+  page_size = (uint64_t)os_getpagesize();
+  request_size = (size + page_size - 1) & ~(page_size - 1);
```
binji, line 43: This is common enough in the code below it probably should be a macro (`ALIGN_UP`?)

wasm2c/wasm-rt-os-win.c
```diff
@@ -0,0 +1,256 @@
...
+  void* addr;
+
+  page_size = os_getpagesize();
+  request_size = (size + page_size - 1) & ~(page_size - 1);
+
+  if (request_size < size)
+    /* integer overflow */
+    return NULL;
+
+  if (request_size == 0)
```
binji, line 88: No similar check for macos/unix?

wasm2c/wasm-rt-os-win.c
```diff
@@ -0,0 +1,256 @@
...
+  DWORD alloc_flag = MEM_RESERVE | MEM_COMMIT;
+  return win_mmap(hint, size, prot, flags, alloc_flag);
+}
+
+#ifndef DONT_USE_VIRTUAL_ALLOC2
+static void* win_mmap_aligned(void* hint,
+                              size_t size,
+                              int prot,
+                              int flags,
+                              size_t pow2alignment) {
```
binji, line 75: seems weird to call this specifically `pow2alignment`, since all of the functions require power-of-two alignment

wasm2c/wasm-rt-os-unix.c
```diff
@@ -0,0 +1,190 @@
...
+    map_prot |= PROT_EXEC;
+
+  return mprotect(addr, request_size, map_prot);
+}
+
+void* os_mmap_aligned(void* addr,
+                      size_t requested_length,
+                      int prot,
+                      int flags,
+                      size_t alignment,
```
binji, line 116: probably should check that alignment is a power of two, since it's required below

wasm2c/wasm-rt-impl.c
```diff
@@ -130,19 +134,25 @@ void wasm_rt_allocate_memory(wasm_rt_memory_t* memory,
 
   /* Reserve 8GiB. */
   void* addr =
-      mmap(NULL, 0x200000000ul, PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+      os_mmap_aligned(NULL, 0x200000000ul, MMAP_PROT_NONE, MMAP_MAP_NONE,
+                      WASM_HEAP_ALIGNMENT, /*alignment_offset=*/0);
```
binji, line 138: is `os_mmap_aligned` ever called with a non-zero `alignment_offset`? If not, then maybe that extra code should be removed.

wasm2c/wasm-rt-os-win.c
```diff
@@ -0,0 +1,256 @@
...
+#include <windows.h>
+
+#ifdef VERBOSE_LOGGING
+#define VERBOSE_LOG(...) \
+  { printf(__VA_ARGS__); }
+#else
+#define VERBOSE_LOG(...)
+#endif
+
+#define DONT_USE_VIRTUAL_ALLOC2
```
binji, line 24: Seems odd to have this always defined. Why not remove the other option below then?

binji (approved):
Generally looks good, though it seems like the code could use some cleaning up

src/wasm2c.c.tmpl
```diff
@@ -22,8 +22,8 @@
        || TRAP(CALL_INDIRECT)                        \
        , ((t)table.data[x].func)(__VA_ARGS__))
 
-#define RANGE_CHECK(mem, a, t) \
-  if (UNLIKELY((a) + sizeof(t) > mem->size)) TRAP(OOB)
+#define RANGE_CHECK(mem, addr, len) \
```
keithw, line 25: The MEMCHECK macro (line 31) uses RANGE_CHECK with a typename when doing software bounds-checks. L31 could be updated to something like `#define MEMCHECK(mem, a, t) RANGE_CHECK(mem, a, sizeof(t))` to keep this working on the non-signal-handler platforms. (Also I think this needs a `make update-wasm2c` for these changes to be synced to the prebuilt headers and tested.)

src/wasm2c.c.tmpl
```diff
@@ -22,8 +22,8 @@
...
+  if (UNLIKELY((addr) + len > mem ->size)) TRAP(OOB)
```
keithw, line 26: The MEMCHECK macro (line 31) uses RANGE_CHECK with a typename when doing software bounds-checks. L31 could be updated to something like #define MEMCHECK(mem, a, t) RANGE_CHECK(mem, a, sizeof(t)) to keep this working on the non-signal-handler platforms. (Also I think this needs a make update-wasm2c for these changes to be synced to the prebuilt headers and tested.)

sbc100:
The failures on windows are no reduces to what looks like fp/nan related failures:

spec/float_exprs:

- float_exprs.wast:2337
- float_exprs.wast:2338
- float_exprs.wast:2347
- float_exprs.wast:2348

spec/float_memory:

- float_memory.wast:21
- float_memory.wast:46
- float_memory.wast:73
- float_memory.wast:98

spec/float_literals:

- float_literals.wast:109
- float_literals.wast:111
- float_literals.wast:112
- float_literals.wast:113
- float_literals.wast:141
- float_literals.wast:143
- float_literals.wast:144
- float_literals.wast:145

spec/float_literals:

- conversions.wast:112
-  ...

sbc100:
It looks like that these tests maybe are relying on the nan bits in f64s.. does that sound right @binji ?  Could this be due to the fact that microsoft x64 ABI use the XMM registers for to f64 pass params and results?   Its seems that the nan bits are maybe not preserved when f64 (double) is returned..

binji:
@sbc100 In my incomplete windows patch, I ended up having to do a few more tweaks to handle floating-point values on Windows, see the description of this commit: https://github.com/WebAssembly/wabt/commit/3d0989635b4d02000981c4f805be8e989a651b87

sbc100:
> @sbc100 In my incomplete windows patch, I ended up having to do a few more tweaks to handle floating-point values on Windows, see the description of this commit: [3d09896](https://github.com/WebAssembly/wabt/commit/3d0989635b4d02000981c4f805be8e989a651b87)

Thanks!  I didn't know about that.

A lot of the issue I was seeing went away when I switch from i386 to x64, but now I see new issues.  I will see if I can get any insight from your work.

sbc100:
I removed a lot of the extra complexity from this change.  It now doesn't add any new `.c` files to wasm2c.. and just includes the very minimum needed to get windows to pass all the tests.

sbc100:
This is now a much smaller PR and does not include a bunch of the stuff from #1833 such as `mmap` alignment, or splitting the rt library into several C files.     All that can come later once its needed/justified I think.

sbc100:
@binji PTAL

wasm2c/wasm-rt-impl.c
```diff
@@ -108,6 +114,54 @@ static void signal_handler(int sig, siginfo_t* si, void* unused) {
...
+  if (errorMessageID != 0) {
+    LPSTR messageBuffer = 0;
+    // The api creates the buffer that holds the message
+    size_t size = FormatMessageA(
+        FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM |
+            FORMAT_MESSAGE_IGNORE_INSERTS,
+        NULL, errorMessageID, MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
+        (LPSTR)&messageBuffer, 0, NULL);
+    (void)size;
+    // Copy the error message into a std::string.
```
binji, line 139: comment is wrong

wasm2c/wasm-rt-impl.c
```diff
@@ -179,6 +240,78 @@ uint32_t wasm_rt_grow_memory(wasm_rt_memory_t* memory, uint32_t delta) {
...
+    return quiet_nan(x);
+  }
+  return nearbyint(x);
+}
+
+float wasm_rt_fabsf(float x) {
+  if (isnan(x)) {
+    uint32_t tmp;
+    memcpy(&tmp, &x, 4);
+    if (tmp & (1 << 31)) {
```
binji, line 292: why not always mask off the top bit?

wasm2c/wasm-rt.h
```diff
@@ -199,6 +225,22 @@ extern void wasm_rt_allocate_table(wasm_rt_table_t*,
...
+float wasm_rt_nearbyintf(float x);
+double wasm_rt_nearbyint(double x);
+float wasm_rt_fabsf(float x);
+double wasm_rt_fabs(double x);
+#else
+#define wasm_rt_truncf(x) truncf(x)
+#define wasm_rt_trunc(x) trunc(x)
+#define wasm_rt_nearbyintf(x) nearbyintf(x)
+#define wasm_rt_nearbyint(x) nearbyint(x)
+#define wasm_rt_fabsf(x) fabsf(x)
```
binji, line 240: interesting, this wasn't an issue for copysign or neg?

binji (approved):
Nice, looks much better

wasm2c/wasm-rt.h
sbc100: I was curious why it was only these functions that were effected.. I have no idea why.  I simply added wrappers as needed.

wasm2c/wasm-rt-impl.c
sbc100: Good point .. I'll try that.

sbc100: Done

rathann:
> [...] With x86 there are some test failure that I believe relate the use of the x87 registers to pass floating point numbers. I suggest we look into fixing those as a followup.

Maybe compare with #1044 as a number of tests are still failing on Linux/x86, too.

PR #1844, opened 2022-03-02
Fix compiler warnings in wasm2c output

sbc100:
Use stricter compiler settings and fix all the resulting warnings.
This is precursor to landing windows compiler support for wasm2c.
See #1843.


src/c-writer.cc
```diff
@@ -1017,8 +1023,15 @@ void CWriter::WriteMemory(const std::string& name) {
 }
 
 void CWriter::WriteTables() {
-  if (module_->tables.size() == module_->num_table_imports)
+  if (module_->tables.size() == module_->num_table_imports) {
     return;
+  }
+
+  if (!module_->types.size()) {
+    // If not types are defined then there is no way to use the table
```
kripken, line 1031:
```suggestion
    // If no types are defined then there is no way to use the table
```

src/c-writer.cc
```diff
@@ -1082,17 +1100,27 @@ void CWriter::WriteDataInitializers() {
...
   }
 
   Write(CloseBrace(), Newline());
 }
 
 void CWriter::WriteElemInitializers() {
+  Write(Newline(), "static void init_table(void) ", OpenBrace());
+
+  if (!module_->types.size()) {
+    // If there are no function types there cannot be any table entries either.
```
kripken, line 1115: I don't think that's true in general. You can have a table that contains non-function types with wasm GC at least.

kripken (approved):
lgtm % comments

src/c-writer.cc
sbc100: Right, but wouldn't the mean that `module_->types` would have to be non-empty?  I'll update the comment.

kripken: Ah, yes, if these are all types and not just function types, which I realize now must be the case, then just updating the comment sgtm.

PR #1846, opened 2022-03-03
fix an unused variable warning in opt

walkingeyerobot:


sbc100 (approved):

sbc100 (commented):
Do you know why we don't see this in CI?

PR #1848, opened 2022-03-04
wasm2c: Add wasm_rt_strerror to covert trap code to string

sbc100:


sbc100:
Sadly we don't currently report expected traps in the stdout when running wasm2c spec tests.. so this change doesn't actually show up in the test expectations.

binji (approved):

PR #1851, opened 2022-03-06
Update wasm2c example output. NFC

sbc100:
Do we really want to have this output checked in like this?
If so, we should probably automate its update ot at least add
a check to avoid it getting out-of-date.


binji (approved):
Agree that it's a bit odd to have checked-in files like this. OTOH, I kinda like that you can quickly see what the output looks like. In either case, lgtm

PR #1852, opened 2022-03-07
Improve the maintenace of ObjectKind types.

zherczeg:
I have realized the Thread has not been removed and Exception has not been added to the ObjectKind as string list. This patch makes the maintenance automated. It seems this macro is used for debugging so I am not sure how to test this.

### DIFF
--- a/src/interp/interp.cc
+++ b/src/interp/interp.cc
@@ -41,9 +41,12 @@ const char* GetName(ExternKind kind) {
 
 const char* GetName(ObjectKind kind) {
   static const char* kNames[] = {
-      "Null",   "Foreign", "Trap", "DefinedFunc", "HostFunc", "Table",
-      "Memory", "Global",  "Tag",  "Module",      "Instance", "Thread",
+      "Null",  "Foreign", "Trap",   "Exception", "DefinedFunc", "HostFunc",
+      "Table", "Memory",  "Global", "Tag",       "Module",      "Instance",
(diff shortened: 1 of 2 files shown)


src/interp/interp.cc
```diff
@@ -41,8 +41,9 @@ const char* GetName(ExternKind kind) {
 
 const char* GetName(ObjectKind kind) {
   static const char* kNames[] = {
-      "Null",   "Foreign", "Trap", "DefinedFunc", "HostFunc", "Table",
-      "Memory", "Global",  "Tag",  "Module",      "Instance", "Thread",
+#define WABT_INTERP_OBJECT_KIND_AS_STRING(Kind) #Kind,
+      WABT_INTERP_OBJECT_KIND(WABT_INTERP_OBJECT_KIND_AS_STRING)
+#undef WABT_INTERP_OBJECT_KIND_AS_STRING
```
sbc100, line 46: I'm not sure this pattern is really worth the added obfuscation that comes with it.

Elsewhere in the codebase we do this instead, which is probably good enough here too?:

```WABT_STATIC_ASSERT(WABT_ARRAY_SIZE(g_reloc_type_name) == kRelocTypeCount);``

zherczeg: Good idea. I have reworked the check to follow the other enums in the code.

sbc100 (approved):

PR #1853, opened 2022-03-08
wasm2c: restore older versions of some spec tests

keithw:
Test wasm2c against versions of the "missing" spec tests (tracked at #1737) from before the bulk-memory and reference-types proposals were merged (as of commit https://github.com/WebAssembly/testsuite/commit/a8bcbafe6d2fb191ce0188de0e18fdc107fa2598). It seems helpful to make sure these tests don't break through various refactors (including our #1814 which doesn't currently pass all of these). These can be replaced with the current versions of the tests once wasm2c supports bulk-memory and reference-types.

~I had to make two changes to get these tests to pass today:~

~1. https://github.com/WebAssembly/testsuite/blob/a8bcbafe6d2fb191ce0188de0e18fdc107fa2598/elem.wast#L15-L18 (removed duplicate $t ids or else wast2json errors)
2. https://github.com/WebAssembly/testsuite/blob/a8bcbafe6d2fb191ce0188de0e18fdc107fa2598/exports.wast#L119-L137 (removed table export section or else wasm2c triggers an assertion failure)~


sbc100:
In general I like the idea of have more test coverage. 

The fact that you have to modify these tests makes me a little sad.   Could we perhaps fix the code to support the old tests without modifying them?   Or would that end up moving the code backwards in some way?

How far are we from supporting bulk-memory and reference-types and thus not needing this PR?   Could you make it clear in the README.txt that these files should be delete as soon as we do support support bulk-memory and reference-types.  Perhaps with a link to `https://github.com/WebAssembly/wabt/issues/1737`?

keithw:
> The fact that you have to modify these tests makes me a little sad. Could we perhaps fix the code to support the old tests without modifying them? Or would that end up moving the code backwards in some way?

I'm guessing it would be pretty doable since wast2json and wasm2c passed these tests in the pass. Let me take a look.

> How far are we from supporting bulk-memory and reference-types and thus not needing this PR?

We've started working on this because of our own needs (you can see WIP at https://github.com/fixpointOS/wabt/tree/bulk-memory-on-instances), but if possible I'd rather not have to implement these features once for the un-instanced code and then rebase/reimplement for instanced modules (or vice versa). So if it were up to us and based on your comment in #1833, the sequence would go:

- [ ] Get this PR merged (after changes necessary to include the un-modified old tests)
- [ ] Get our module_instancing branch passing these newly added (but old) tests
- [ ] Finish implementing bulk-memory, and start and finish implementing reference-types, on top of our module_instancing branch
- [ ] Confirm that all the current versions of the spec tests pass, restore them to the spec directory, and remove old-spec
- [ ] If rlbox passes the testsuite by this point, rebase it on top of module_instancing; otherwise, worry about it later.
- [ ] Merge module_instancing

Thoughts?

> Could you make it clear in the README.txt that these files should be delete as soon as we do support support bulk-memory and reference-types. Perhaps with a link to `https://github.com/WebAssembly/wabt/issues/1737`?

Sure.

sbc100:
Thats order of work sounds right to me!  Thanks again for working on this.

keithw:
Here you go -- the changes necessary to get the old tests passing unmodified were pretty small.

src/c-writer.cc
```diff
@@ -1026,12 +1026,6 @@ void CWriter::WriteTables() {
     return;
   }
 
-  if (!module_->types.size()) {
-    // If no types are defined then there is no way to use the table
-    // for anything.
-    return;
-  }
```
sbc100, line 1033: Hmm.. why was this needed?

test/run-spec-wasm2c.py
```diff
@@ -392,7 +394,9 @@ def main(args):
         wast2json.verbose = options.print_cmd
         wast2json.AppendOptionalArgs({
             '-v': options.verbose,
-            '--enable-multi-memory': options.enable_multi_memory})
+            '--enable-multi-memory': options.enable_multi_memory,
+            '--disable-bulk-memory': options.disable_bulk_memory,
+            '--disable-reference-types': options.disable_reference_types})
```
sbc100, line 399: Are these needed?  Why?

sbc100 (approved):
Thanks!  LGTM.   Can you explain why you needed those changes under src?

src/c-writer.cc
keithw: Here's a test case that fails without this change:
```wat
(module (table 0 funcref) (export "a" (table 0)))
```

The module doesn't have any types, so, this code causes `CWriter::WriteTables` to early-out and omit the call to `WriteTable(DefineGlobalScopeName(table->name));`.

Later, when `CWriter::WriteExports(WriteExportsKind::Initializers)` is called, it's trying to write the body of the `init_exports` function, and it tries to execute `Write(mangled_name, " = ", ExternalPtr(internal_name), ";");` (https://github.com/WebAssembly/wabt/blob/main/src/c-writer.cc#L1234), and the call to `Write(const ExternalPtr&)` tries to call `GetGlobalName` which hits the assertion failure (this one: https://github.com/WebAssembly/wabt/blob/main/src/c-writer.cc#L595) because the table's name never made it to `DefineGlobalScopeName` back when `WriteTables` was called.

I guess the big-picture subtlety here is that these `CWriter` methods are designed to be invoked in a particular order, in which some of them call `DefineGlobalScopeName` (the first time a name is seen) and then other methods call `GetGlobalName` on the same name and assume that the name has already been defined. And these ordering requirements/dependencies between the methods are not completely obvious or explicit.

sbc100: Oh, yes good point.  That looks like a fix for a real issue.

test/run-spec-wasm2c.py
keithw: The offending test looks something like this:
```wat
(module
  (table $t 10 funcref)
  (elem $t (i32.const 0))
  (elem $t (i32.const 0))
)
```
(simplified from here: https://github.com/WebAssembly/testsuite/blob/a8bcbafe6d2fb191ce0188de0e18fdc107fa2598/elem.wast#L15-L16)

Before bulk-memory, the `elem` segments were parsed so that the `$t` was the `tableidx` of the table that the element segment would be initializing. So it's perfectly legal to have a duplicate idx here. After bulk-memory, my understanding is that the syntax changed so that the `$t` is now parsed to be the `id` of the elem segment itself. (In WebAssembly 1.0, elem and data segments did not have ids.) So with bulk-memory on, `wat2wasm` and `wast2json` refuse to parse this old spec test -- hence the need to disable it for this one.

PR #1854, opened 2022-03-08
Fix Store object assignment.

zherczeg:
When a new store object is assigned to a value, its FreeList is copied without duplicating the referenced objects, so these objects are destroyed twice.

### DIFF
--- a/src/interp/interp.h
+++ b/src/interp/interp.h
@@ -453,6 +453,10 @@ class Store {
 
   explicit Store(const Features& = Features{});
 
+  Store(const Store&) = delete;
+  Store& operator=(const Store&) = delete;
+  Store& operator=(const Store&&) = delete;
+
   bool IsValid(Ref) const;
   bool HasValueType(Ref, ValueType) const;
   template <typename T>


src/interp/interp.h
```diff
@@ -414,6 +414,8 @@ class FreeList {
   Index New(Args&&...);
   void Delete(Index);
 
+  FreeList& operator=(FreeList&&);
```
sbc100, line 417: That about the copy assignment operator?   Is that no longer implicitly defined?

i.e. what happens if you do the following:

```  
wabt::interp::Store store1;
wabt::interp::Store store2;
store1 = store2;
```

Does that now fail to compile?  Or do you have to explictly add:

```
operator= (const FreeList&) = delete;
```

?

sbc100: It sounds like it should not be needed since:

```
Deleted implicitly-declared copy assignment operator
An implicitly-declared copy assignment operator for class T is defined as deleted if any of the following is true:

T has a user-declared move constructor;
T has a user-declared move assignment operator.
```

src/test-interp.cc
```diff
@@ -718,6 +718,13 @@ TEST_F(InterpGCTest, Collect_DeepRecursion) {
   EXPECT_EQ(1u, store_.object_count());
 }
 
+TEST_F(InterpGCTest, Store_Assignemnt) {
+  wabt::interp::Store store;
+
+  store = wabt::interp::Store();
```
sbc100, line 724: I think you can do this all one line no?

sbc100 (approved):

src/test-interp.cc
sbc100: Should this just be `store_ = wabt::interp::Store();`  (note the trailing underscore).   Since the assertion below relates to the member `store_`?

src/test-interp.cc
```diff
@@ -718,6 +718,12 @@ TEST_F(InterpGCTest, Collect_DeepRecursion) {
   EXPECT_EQ(1u, store_.object_count());
 }
 
+TEST_F(InterpGCTest, Store_Assignment) {
```
sbc100, line 721: Maybe add a comment here.  Something like:

"Verify that move assignment operator works.  We previously has issues with double free that this test would fail under with ASAN".

src/interp/interp-inl.h
```diff
@@ -218,6 +218,17 @@ void FreeList<T>::Delete(Index index) {
   free_items_++;
 }
 
+template <typename T>
+FreeList<T>& FreeList<T>::operator=(FreeList<T>&& other) {
+  list_ = other.list_;
```
sbc100, line 223: I think you would need to free all the items you are currently holding before moving the new items into place.

Given the complexity here maybe its best to just delete the copy constructor?

@steven-johnson WDYT?

sbc100: Looking at the halide use case it looks like the copy could easily be avoided by initializing `store` in the initlizer list part of `WasmModuleContents::WasmModuleContents`  rather than using `store = wabt::interp::Store(calc_features(halide_module.target()));` in the body.   So I think deleting the operator makes sense.

zherczeg: You are right about deleting the items. Ok, I will wait your decision before doing further changes. Anyway, these operations needed to be done before even if the compiler did it, so reassignment was always a costly operation.

sbc100: Please go ahead and remove the assignment operations.. I'm pretty its the right approach for these classes.

We should also delete the copy constructor too.

sbc100 (approved):

steven-johnson:
Why do we delete the move-assignment operator but not the move-constructor? Normally you'd want to implement (or delete) these in tandem.

sbc100:
No reason.. other than C++11 ignorance on my part I think.  I can never remember the rules for when implicit constructors and deleted...

steven-johnson:
https://en.cppreference.com/w/cpp/language/rule_of_three

I don't know if it's an *error* per se, but it makes my spidey-sense tingle; if the move-assign needs special-casing, the move-ctor almost always does too (and vice versa).

sbc100:
Thanks!  Yes we should have delete the move constructor too.

PR #1856, opened 2022-03-09
Build and run wasm2c-generated code under sanitizers in CI. NFC

sbc100:
Split out from #1843

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -84,6 +84,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       USE_NINJA: "1"
+      WASM2C_CFLAGS: "-fsanitize=address"
     steps:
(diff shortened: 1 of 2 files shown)


sbc100:
@keithw @shravanrn.  This is needed if the wasm2c tests are going to use the `libwam-rt` that is build via cmake, but should also be good for keep undefined behaviour out the wasm2c-generated code I hope.

keithw:
Sounds great to me, although if there's a worry about adversarial WebAssembly triggering UB in wasm2c-generated code, I'm not sure if the spec tests are "adversarial enough." It would be great to have a fuzzer in oss-fuzz that tests wasm2c and the wasm2c-generated code. Right now the only fuzz target is for wabt::ReadBinaryIr.

shravanrn:
sgtm

Issue #1257, opened 2019-11-30
GC proposal


skuzmich:
Is there a roadmap of typed function references and GC proposal support?

Tools like **wasm-interp** or **wasm2c** would be useful to test compilers while waiting for browser support.

binji:
Not yet. As with other proposals, I'd like to add this functionality. I probably won't have time to work on it, but I'm happy to review changes if anyone else wants to work on it. That said, since it's a bigger feature, I'd like to discuss a plan first if someone wants to take it on.

binji:
I've started to work on this: https://github.com/WebAssembly/wabt/pull/1345

PR #1860, opened 2022-03-10
Use a shared null object for all Stores.

zherczeg:


sbc100:
What is the goal of this change?  Is it an optimization or something about being able to share this object between stores?

If its an optimization then I would ask why such micro-optimizations are needed?   Are you huge number of stores?

zherczeg:
It is a simple optimization. If you think this is unnecessary, I can close the PR. I would like to help improving the interpreter, is there any task which is useful for you? I would focus on garbage collection and memory, but I can help with other things.

sbc100:
> It is a simple optimization. If you think this is unnecessary, I can close the PR. I would like to help improving the interpreter, is there any task which is useful for you? I would focus on garbage collection and memory, but I can help with other things.

If the optimization is to avoid the bytes needed to represent NULL in each store I do think it might be premature, yes.   I imagine this is a couple of words of memory per-store?  But I don't know if anyone who uses the wabt interpreter on a scale where that would matter.

If you are interested in helping out wabt in general what would be great!   Let me go through the open issues and prioritize some of them and get back to you.

sbc100:
If are you interested in helping out with GC in general we could use some help workon and merging the work that @binji started in to get the GC proposal implemented: https://github.com/WebAssembly/wabt/issues/1257

zherczeg:
Thank you for the suggestion. At first sight it looks like it adds a new struct / array related code. The work looks quite a big, and does a lot of different things. Is it possible to split this into well defined subtasks? I see references to draft documents. Is there a final version available somewhere?

sbc100:
> Thank you for the suggestion. At first sight it looks like it adds a new struct / array related code. The work looks quite a big, and does a lot of different things. Is it possible to split this into well defined subtasks? I see references to draft documents. Is there a final version available somewhere?

The GC proposal is still in flight but recently reached stage 2 so is less likely to have huge changes going forward.

But yes, I think it pretty large amount of work.    There are certainly many smaller issues in wabt that we could have you look at.

sbc100:
Another useful task that could be more reasonable in size would be to update the `testsuite` repo (using this script: https://github.com/WebAssembly/testsuite/blob/main/update-testsuite.sh) and then update the submodule dependency in wabt, and then run `./test/update-spec-tests.py` here in wabt, and fix any issues that come out of that.

That process bascially ensure that wabt stays to-to-date with any specs that it purports to support.

keithw:
Closing as dormant (it looks like we decided not to do this).

PR #1861, opened 2022-03-10
Rebase the rlbox repo

shravanrn:
Rebasing PR per our discussion
@sbc100 Would it be possible to get write access to the rlbox branch in the wabt repo?


sbc100:
Done. I gave you write access the whole repo (couldn't figure out how to do it on just one branch).

sbc100:
Also, I'm guessing we don't expect the tests to pass on this branch yet?  (I think that OK for now since IIUC they have not be passing previously, and our near term goal is to get them passing?)

BTW do you know up to which revision in main this gets you?  Maybe worth nothing that in the description?

PR #1870, opened 2022-03-18
Use CMAKE_CXX_STANDARD and bump cmake minimum version to 3.1

sbc100:
See https://cmake.org/cmake/help/latest/variable/CMAKE_CXX_STANDARD.html


binji (approved):
nice!

PR #1875, opened 2022-03-31
wasm2c: use signal handler to detect stack exhaustion

keithw:
Use the signal-handler method to detect and recover from stack exhaustion, in addition to OOB accesses. This eliminates the global `wasm_rt_call_stack_depth` and the need to increment/decrement it on function entries and exits. Eliminating this global variable is helpful for module instancing and for running parallel invocations in a thread-safe manner.

(The catch is that MacOS doesn't seem to make it easy to distinguish OOB vs. exhaustion, so these trap codes have to be merged on MacOS and perhaps other BSDs. On Linux, it's possible to tell them apart based on the si_code of the SIGSEGV.)


wasm2c/wasm-rt-impl.c
```diff
@@ -161,15 +168,29 @@ static void os_print_last_error(const char* msg) {
 }
 #endif
 
-void wasm_rt_allocate_memory(wasm_rt_memory_t* memory,
-                             uint32_t initial_pages,
-                             uint32_t max_pages) {
-  uint32_t byte_length = initial_pages * PAGE_SIZE;
+void wasm_rt_init() {
```
binji, line 171: Should be `void wasm_rt_init(void)`

wasm2c/wasm-rt-impl.c
```diff
@@ -180,7 +201,20 @@ void wasm_rt_allocate_memory(wasm_rt_memory_t* memory,
       abort();
     }
   }
+#endif
+}
 
+void wasm_rt_free() {
```
binji, line 207: same here `void wasm_rt_free(void)`

binji (approved):
lgtm w/ a few comments

PR #1876, opened 2022-03-31
Allow Code Metadata Annotations to have arbitrary hex strings

jacob-abraham:
This PR adds support for arbitrary hex strings in code metadata annotations. Previously, the decoded strings would be checked as utf-8 strings. However, this prohibits something like `(@metadata.code.trace_inst "\FF\FF")` since 0xffff is not a valid utf-8 string.

Also modified existing roundtrip test to have a string that includes a non-utf8 character

### DIFF
--- a/src/wast-parser.cc
+++ b/src/wast-parser.cc
@@ -977,7 +977,7 @@ bool WastParser::ParseRefTypeOpt(Type* out_type) {
   return true;
 }
 
-Result WastParser::ParseQuotedText(std::string* text) {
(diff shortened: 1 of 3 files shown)


binji (approved):
lgtm, this seems to match the expected behavior described here: https://github.com/WebAssembly/annotations/blob/main/proposals/annotations/Overview.md

keithw:
Thank you -- merging as this was approved and also lgtm.

PR #1877, opened 2022-03-31
wasm2c: implement the bulk memory operations proposal

keithw:
~~(This PR is sequenced behind #1875 and #1814.)~~

Support bulk memory operations in wasm2c (the next step in the checklist at https://github.com/WebAssembly/wabt/pull/1853#issuecomment-1062094906).

This PR also adds test coverage:

- The current testsuite's `elem.wast` test (we had previously been using a pre-bulk-memory version in old-spec. These "old" tests are tracked at #1737.)
- The current testsuite's `bulk.wast`, `memory_copy.wast`, `memory_fill.wast`, and `memory_init.wast` (these were added in the bulk-memory proposal)
- The current multi-memory proposal's `load.wast` and `store.wast` tests (we had previously been omitting these because they require bulk-memory support, discussed at #1834. Another multi-memory tests requires reference-types support, and the final multi-memory test requires SIMD.)
- Bulk-memory proposal versions of the `imports.wast`, `linking.wast`, `table_copy.wast`, and `table_init.wast` tests. (The current versions of these are waiting on reference-types support.)


src/c-writer.cc
```diff
@@ -1177,70 +1458,209 @@ void CWriter::WriteExports(WriteExportsKind kind) {
...
+  Write(CloseBrace(), Newline());
+}
+
+void CWriter::WriteInitInstanceImport() {
+  if (module_->imports.empty())
+    return;
+
+  Write(Newline(), "static void init_instance_import(",
+        MangleModuleInstanceTypeName(module_name_), "* module_instance");
+  for (auto import_module_name : import_module_set_) {
```
binji, line 1634: this loop seems to show up a lot too, might be nice for a helper function

src/wasm2c.c.tmpl
```diff
@@ -288,3 +296,78 @@ DEFINE_REINTERPRET(i32_reinterpret_f32, f32, u32)
...
+  memmove(dest->data + dest_addr, src->data + src_addr, n);
+}
+
+static inline void memory_init(wasm_rt_memory_t* dest,
+                               const u8* src,
+                               u32 src_size,
+                               u32 dest_addr,
+                               u32 src_addr,
+                               u32 n) {
+  if (UNLIKELY(src_addr + (uint64_t)n > src_size)) TRAP(OOB);
```
binji, line 320: nit: u64  here and below

src/wasm2c.c.tmpl
```diff
@@ -288,3 +296,78 @@ DEFINE_REINTERPRET(i32_reinterpret_f32, f32, u32)
...
+                              u32 src_addr,
+                              u32 n,
+                              void* module_instance,
+                              const u32* func_types) {
+  if (UNLIKELY(src_addr + (uint64_t)n > src_size)) TRAP(OOB);
+  if (UNLIKELY(dest_addr + (uint64_t)n > dest->size)) TRAP(OOB);
+  for (u32 i = 0; i < n; i++) {
+    const wasm_elem_segment_expr_t* src_expr = &src[src_addr + i];
+    dest->data[dest_addr + i] =
+        (wasm_rt_elem_t){func_types[src_expr->func_type_index], src_expr->func,
```
binji, line 343: I wonder if we should use c99 here? Not sure if the previous code already did

test/run-spec-wasm2c.py
```diff
@@ -165,7 +187,73 @@ def _CacheModulePrefixes(self):
                 else:
                     name_idx = idx - 1
 
+                if name_idx in self.idx_to_module_name.keys():
```
binji, line 190: I think you can just do `if name_idx in self.idx_to_module_name:`

test/run-spec-wasm2c.py
```diff
@@ -165,7 +187,73 @@ def _CacheModulePrefixes(self):
...
+                c_filename = command['as'] + '.c'
+                self.idx_to_c_name[name_idx] = c_filename
+
+    def _WriteModuleInstanceImports(self, command, uninstantiable):
+        c_filename = os.path.join(self.out_dir, self.idx_to_c_name[self.module_idx - 1])
+        header_filename = utils.ChangeExt(c_filename, '.h')
+        with open(header_filename, encoding='utf-8') as f:
+            headerfile = f.readlines()
+            imported_modules = set()
+            for i in range(0, len(headerfile)):
```
binji, line 202: nit: `for line in f:` to read iterate over all lines in the file

test/run-spec-wasm2c.py
```diff
@@ -165,7 +187,73 @@ def _CacheModulePrefixes(self):
...
+            self.out_file.write(";\n")
+
+    def _WriteModuleImports(self):
+        for idx, c_filename in self.idx_to_c_name.items():
+            header_filename = utils.ChangeExt(os.path.join(self.out_dir, c_filename), '.h')
+
+            with open(header_filename, encoding='utf-8') as f:
+                headerfile = f.readlines()
+                imported_functions = {}
+                for i in range(0, len(headerfile)):
```
binji, line 231: same here

test/run-spec-wasm2c.py
```diff
@@ -165,7 +187,73 @@ def _CacheModulePrefixes(self):
...
+                    line = headerfile[i]
+                    if "import: " in line:
+                        next_line = headerfile[i + 1]
+                        if "extern" in next_line:
+                            line_split = line.split()
+                            import_module_name = MangleName(line_split[2][1:-1])
+                            import_field_name = MangleName(line_split[3][1:-1])
+                            imported_functions[import_module_name + "_" + import_field_name] = next_line[7:-2]
```
binji, line 239: `[7:-2]` looks a little odd, maybe a comment here explaining the format being parsed?

test/run-spec-wasm2c.py
```diff
@@ -165,7 +187,73 @@ def _CacheModulePrefixes(self):
...
+                line = headerfile[i]
+                if "import: " in line:
+                    line_split = line.split()
+                    import_module_name = MangleName(line_split[2][1:-1])
+                    imported_modules.add(import_module_name)
+
+            if (uninstantiable):
+                self.out_file.write("ASSERT_TRAP(")
+
+            if (len(imported_modules) > 0):
```
binji, line 212: nit: no braces around if

test/run-spec-wasm2c.py
```diff
@@ -324,15 +413,18 @@ def _ActionSig(self, action, expected):
...
-        expected = command['expected']
         type_ = action['type']
         mangled_module_name = self.GetModulePrefix(action.get('module'))
-        field = (mangled_module_name + MangleName(action['field']) +
-                 MangleName(self._ActionSig(action, expected)))
+        field = mangled_module_name + "_" + MangleName(action['field'])
         if type_ == 'invoke':
-            return '%s(%s)' % (field, self._ConstantList(action.get('args', [])))
+            args = self._ConstantList(action.get('args', []))
+            if (len(args) == 0):
```
binji, line 421: nit: no braces

wasm2c/README.md
```diff
@@ -443,3 +487,104 @@ $ wat-desugar fac-flat.wat --fold -o fac-folded.wat
 
 The formatting is different and the variable and function names are gone, but
 the structure is the same.
+
```
binji, line 490: awesome! Thanks for writing this up

wasm2c/examples/fac/fac.wat
```diff
@@ -1,3 +1,4 @@
+(memory $mem 1)
```
binji, line 1: is this necessary to add now?

binji (approved):
lgtm! Since I'm not maintaining this anymore you may want to wait for further review, but I wanted to give you some feedback

keithw:
@binji Thank you muchly for the reviews! I wish GitHub had a way for us to mark the stacked PRs more clearly -- if possible, could we take these in order (so #1814 [module instancing] would be next, before we get to this one [bulk memory], reference types [#1887], and tolerating partly invalid modules [#1888])? Some of your comments here are things we'll need to fix in #1814 where the relevant commits originate, and then our plan was to rebase the later PRs on top of whatever ultimately lands.

src/c-writer.cc
```diff
@@ -127,7 +134,8 @@ class CWriter {
       : options_(options),
         c_stream_(c_stream),
         h_stream_(h_stream),
-        header_name_(header_name) {}
+        header_name_(header_name),
+        module_name_(header_name_.substr(0, header_name_.size() - 2)) {}
```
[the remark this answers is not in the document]
keithw: Thanks, added a more verbose version at https://github.com/fixpointOS/wabt/blob/module_instancing/src/c-writer.cc#L139-L142

src/c-writer.cc
```diff
@@ -875,62 +946,162 @@ void CWriter::WriteFuncTypes() {
   Write(CloseBrace(), Newline());
 }
 
-void CWriter::WriteImports() {
-  if (module_->imports.empty())
+void CWriter::ComputeUniqueImports() {
+  std::map<std::pair<std::string, std::string>, const Import*> import_map;
+  for (const Import* import : module_->imports) {
+    import_map.try_emplace({import->module_name, import->field_name}, import);
```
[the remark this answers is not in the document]
keithw: It's fine for a module to import the same external field under multiple ids (if the local fields are all the same type), but not so fine otherwise. Added a warning here on duplicate imports and an abort on clearly contradictory imports (https://github.com/fixpointOS/wabt/blob/module_instancing/src/c-writer.cc#L953-L962). I don't think this will bomb out on every contradictory import, though (e.g. if the same field is imported as two functions with different signatures).

src/c-writer.cc
```diff
@@ -875,62 +946,162 @@ void CWriter::WriteFuncTypes() {
...
+
+  for (const auto& node : import_map) {
+    unique_imports_.push_back(node.second);
+  }
+}
+
+void CWriter::WriteInstanceImports() {
+  if (module_->imports.empty()) {
+    Write("typedef struct ", MangleModuleInstanceTypeName(module_name_), " ",
+          OpenBrace());
```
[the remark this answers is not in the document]
keithw: Changed to BeginInstance

src/c-writer.cc
```diff
@@ -952,64 +1123,76 @@ void CWriter::WriteFuncDeclarations() {
...
-  if (decl.GetNumParams() == 0) {
-    Write("void");
-  } else {
-    for (Index i = 0; i < decl.GetNumParams(); ++i) {
-      if (i != 0)
-        Write(", ");
-      Write(decl.GetParamType(i));
-    }
+  Write(MangleModuleInstanceTypeName(module_name_), "*");
+  WriteParamTypes(decl);
```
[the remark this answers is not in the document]
keithw: Moved the close paren back to the callsites.

test/run-spec-wasm2c.py
keithw: This one seems trickier because we want to look ahead one line to detect a function import. Added a comment to clarify (https://github.com/fixpointOS/wabt/blob/module_instancing/test/run-spec-wasm2c.py#L223-L224).

binji:
Sorry about reviewing out of order, I should have looked more closely at the description!

sbc100:
Does this change really depend on instancing ( #1814)?  I wonder if it can land first?

keithw:
> Does this change really depend on instancing ( #1814)? I wonder if it can land first?

It's possible, but it would be a bit of work to backport this to a pre-instanced implementation. I guess I thought we had coalesced on this ordering when we talked at https://github.com/WebAssembly/wabt/pull/1853#issuecomment-1062094906.

The major "instance-related" design decisions/consequences here are:

- as before, we store each segment as a static const array at file scope (belonging to the "module" itself), but now each module instance has a boolean (in a bitfield) to track whether the corresponding segment has been dropped. (A drop instruction doesn't actually free memory here; even if all extent instances have dropped the segment, the caller might later create another instance of the same module...)

- before bulk memory, it was possible to use element segments only at "wasm2c-time", to write the code that initializes the table as part of `init_table`, and then not actually store the element segment itself. But after bulk memory, the element segment has to be stored so that it can be used at runtime in a `table.init`. To do that, we have to be a bit careful in choosing the representation of the init expression, because at runtime the members of a funcref-type table are function **instances** (which close over the originating module), but when writing down the element segment itself (before any module has been instantiated with its imports), you don't know what the originating module will be. So, we represent the elem segment init expr as a local type index, a funcref, and an **offset** within the module instance structure to the originating module pointer. At runtime, the implementation of `table_init` converts this to a global type index, the funcref, and the actual pointer to the originating module (looked up from the module instance structure).

sbc100:
> > Does this change really depend on instancing ( #1814)? I wonder if it can land first?
> 
> It's possible, but it would be a bit of work to backport this to a pre-instanced implementation. I guess I thought we had coalesced on this ordering when we talked at [#1853 (comment)](https://github.com/WebAssembly/wabt/pull/1853#issuecomment-1062094906).
> 
> The major "instance-related" design decisions/consequences here are:
> 
> * as before, we store each segment as a static const array at file scope (belonging to the "module" itself), but now each module instance has a boolean (in a bitfield) to track whether the corresponding segment has been dropped. (A drop instruction doesn't actually free memory here; even if all extent instances have dropped the segment, the caller might later create another instance of the same module...)
> * before bulk memory, it was possible to use element segments only at "wasm2c-time", to write the code that initializes the table as part of `init_table`, and then not actually store the element segment itself. But after bulk memory, the element segment has to be stored so that it can be used at runtime in a `table.init`. To do that, we have to be a bit careful in choosing the representation of the init expression, because at runtime the members of a funcref-type table are function **instances** (which close over the originating module), but when writing down the element segment itself (before any module has been instantiated with its imports), you don't know what the originating module will be. So, we represent the elem segment init expr as a local type index, a funcref, and an **offset** within the module instance structure to the originating module pointer. At runtime, the implementation of `table_init` converts this to a global type index, the funcref, and the actual pointer to the originating module (looked up from the module instance structure).

No problem.  If its not a simple rebase we can keep the sequence as is

remko:
@keithw I'm getting an "unimplemented: ..." error when running `wasm2c`. It seems it's in a function using `memory.copy`, so I'm assuming it's related to the patch.

Here's my WASM file:
[waforth.zip](https://github.com/WebAssembly/wabt/files/9079551/waforth.zip)

keithw:
@sbc100 Do you want to take a look at this one? (It will be helpful for updating the testsuite as more of the spec tests have started to require bulk-memory.)

scripts/gen-wasm2c-templates.cmake
```diff
@@ -1,4 +1,5 @@
 # https://stackoverflow.com/a/47801116
 file(READ ${in} content)
-set(content "const char* ${symbol} = R\"w2c_template(${content})w2c_template\";")
+string(REGEX REPLACE "(.[^\n]*\n)" "R\"w2c_template(\\1)w2c_template\"\n" content "${content}")
+set(content "const char* ${symbol} = ${content};")
```
sbc100, line 4: Why this change?

src/template/wasm2c.declarations.c
```diff
@@ -444,4 +444,85 @@ static float wasm_sqrtf(float x) {
   return sqrtf(x);
 }
 
+static inline void memory_fill(wasm_rt_memory_t* mem, u32 d, u32 val, u32 n) {
```
sbc100, line 447: Would it make sense to somehow only emit this code when bulk memory is enabled?  Perhaps we could define `_WASM2C_HAS_BULK_MEMORY` and make this conditional?  I guess it doesn't make much difference?  Maybe tiny modules would compile marginally faster? 

Maybe not worth it?

wasm2c/examples/fac/fac.c
```diff
@@ -468,6 +551,20 @@ static void init_tags(void) {
 
 static u32 w2c_fac(Z_fac_instance_t*, u32);
 
+static void init_globals(Z_fac_instance_t* instance) {
+}
+static void init_memory(Z_fac_instance_t* instance) {
```
sbc100, line 556: Missing newline here before `init_memory`

wasm2c/examples/fac/fac.c
```diff
@@ -468,6 +551,20 @@ static void init_tags(void) {
...
+}
+
+static void init_data_instances(Z_fac_instance_t *instance) {
+}
+
+static void init_table(Z_fac_instance_t* instance) {
+}
+
+static void init_elem_instances(Z_fac_instance_t *instance) {
+}
```
sbc100, line 566: It it worth eliding all these function when they are empty (to make the output more readable?)

src/c-writer.cc
```diff
@@ -1385,21 +1390,35 @@ void CWriter::WriteGlobalInitializers() {
...
+}
+
+void CWriter::WriteDataInstances() {
+  for (const DataSegment* data_segment : module_->data_segments) {
+    DefineGlobalScopeName(data_segment->name);
+    if (is_droppable(data_segment)) {
+      Write("bool ", "data_segment_dropped_", GetGlobalName(data_segment->name),
+            " : 1;", Newline());
+    }
+  }
```
sbc100, line 1410: I wonder if we should have some more examples of wasm2c output actual checked in so that when reviewing these changes we can see the effect they have on the output?  Not sure what the best way to do that is.

src/c-writer.cc
```diff
@@ -1410,95 +1429,148 @@ void CWriter::WriteDataInitializers() {
...
   }
-  data_segment_index = 0;
   for (const DataSegment* data_segment : module_->data_segments) {
-    Write("LOAD_DATA(", ExternalInstanceRef(memory->name), ", ");
-    WriteInitExpr(data_segment->offset);
-    Write(", data_segment_data_", data_segment_index, ", ",
-          data_segment->data.size());
-    Write(");", Newline());
-    ++data_segment_index;
+    if (data_segment->kind == SegmentKind::Active) {
```
sbc100, line 1449: Below you do:

```
    if (elem_segment->kind != SegmentKind::Active) {
      continue;
    }
 ``` 

perhaps its worth being consistent?  (I don't feel strongly about which you should prefer).

sbc100 (commented):
Nice! 

lgtm with comments

scripts/gen-wasm2c-templates.cmake
keithw: MSVC has a limit on the length of string literals that was getting exceeded (I think because wasm2c.declarations.c is now longer than 16KiB). This switches it to being one string literal per line of source.

src/template/wasm2c.declarations.c
keithw: We can do this if you want, but given that bulk memory is enabled by default (and has been for almost a year), it seems like a niche case... and it's not like these are particularly complicated functions to compile. (Some of the math functions are arguably more complicated and we emit those unconditionally...)

wasm2c/examples/fac/fac.c
keithw: Fixed.

sbc100 (approved):

wasm2c/examples/fac/fac.c
keithw: Sounds good, done in #1999 which hopefully we can do as a follow-on to this one.

src/c-writer.cc
keithw: I also like this idea but not sure how far to take it... some options are probably:
- check in rot13.c like we do fac.c?
- check in some representative outputs of running wasm2c on the spec testsuite?
- check in every single output of running wasm2c on the spec testsuite?

src/c-writer.cc

src/c-writer.cc
sbc100: How about a few hand-crafted tests under `test/wasm2c` that include the input wat and the output c code in the same same file.

keithw: Did a first stab at this in #2006.

Issue #1612, opened 2021-02-25
wat2wasm doesn't typecheck (elem) against table types


ghost:
This snippet is invalid,
```wat
(module
	(import "" "" (func $f))
	(table 1 1 externref)
	(elem (i32.const 0) $f)
)
```

as `func` is not a super-type of `externref`.

wat2wasm currently doesn't catch it.

7ombie:
I personally found it tricky to type-check active table segments, as you need to refer back to the table it populates to know the reftype, which in turn, requires resolving the number literal or identifier that expresses the table index. This should happen during parsing, but the indices are not worked out until the compilation stage.

You cannot just track how many tables have been imported or defined in the parser, as a later import will be enumerated in the binary ahead of any definitions, even if the definition appears first in the source, so the indices can be invalidated.

The compiler can check the types as it compiles everything, and has the information it needs, but this creates a subtle issue: If your example code was at line 50, and another error was at line 100, the compiler would complain about line 100, and fixing that bug would reveal the error at line 50. This is a small thing, but you really don't want debugging tools to occasionally do something subtly different from what they normally do.

It's really a consequence of the design of WAT. It could have required that active segments are directly attached (lexically) to the memory or table they populate, which would have made validation easy, but would have been inflexible linguistically.

PR #1887, opened 2022-04-04
wasm2c: implement the reference-types proposal

### LINKED ISSUE #1737: wasm2c no longer passes many spec tests

### LINKED ISSUE #1985: Aborted in CWriter::Write at wasm2c
### Environment
```
OS      : Linux ubuntu 5.15.0-46-generic #49~20.04.1-Ubuntu SMP Thu Aug 4 19:15:44 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
Commit  : 3054d61f703d609995798f872fc86b462617c294
Version : 1.0.29
Build   : make clang-debug-asan
```

### Proof of concept
poc_wasm2c-1.wasm
[poc_wasm2c-1.wasm.zip](https://github.com/WebAssembly/wabt/files/9489964/poc_wasm2c-1.wasm.zip)

### Stack dump
```
gdb /wabt/out/clang/Debug/asan/wasm2c
pwndbg> r  --enable-multi-memory ./poc_wasm2c-1.wasm
context:
LEGEND: STACK | HEAP | CODE | DATA | RWX | RODATA
─────────────────────────────────[ REGISTERS ]──────────────────────────────────
 RAX  0x0
*RBX  0x7ffff7a357c0 ◂— 0x7ffff7a357c0
*RCX  0x7ffff7a7b00b (raise+203) —▸ 0x10824848b48 ◂— 0x0
 RDX  0x0
*RDI  0x2
*RSI  0x7fffffff6990 ◂— 0x0
 R8   0x0
*R9   0x7fffffff6990 ◂— 0x0
*R10  0x8
*R11  0x246
*R12  0x43e420 (_start) ◂— endbr64
*R13  0x7fffffffe070 ◂— 0x3
 R14  0x0
 R15  0x0
*RBP  0x7fffffff88f0 —▸ 0x7fffffff8930 —▸ 0x7fffffffa650 —▸ 0x7fffffffa690 —▸ 0x7fffffffc3c0 ◂— ...
*RSP  0x7fffffff6990 ◂— 0x0
*RIP  0x7ffff7a7b00b (raise+203) —▸ 0x10824848b48 ◂— 0x0
───────────────────────────────────[ DISASM ]───────────────────────────────────
 ► 0x7ffff7a7b00b <raise+203>    mov    rax, qword ptr [rsp + 0x108]
   0x7ffff7a7b013 <raise+211>    xor    rax, qword ptr fs:[0x28]
   0x7ffff7a7b01c <raise+220>    jne    raise+260                <raise+260>
    ↓
   0x7ffff7a7b044 <raise+260>    call   __stack_chk_fail                <__stack_chk_fail>

   0x7ffff7a7b049                nop    dword ptr [rax]
   0x7ffff7a7b050 <killpg>       endbr64
   0x7ffff7a7b054 <killpg+4>     test   edi, edi
   0x7ffff7a7b056 <killpg+6>     js     killpg+16                <killpg+16>

   0x7ffff7a7b058 <killpg+8>     neg    edi
   0x7ffff7a7b05a <killpg+10>    jmp    kill                <kill>

   0x7ffff7a7b05f <killpg+15>    nop
───────────────────────────────────[ STACK ]────────────────────────────────────
00:0000│ rsi r9 rsp 0x7fffffff6990 ◂— 0x0
01:0008│            0x7fffffff6998 ◂— '(ut)(x))unimplemented: %'
02:0010│            0x7fffffff69a0 ◂— 'unimplemented: %'
03:0018│            0x7fffffff69a8 ◂— 'ented: %'
04:0020│            0x7fffffff69b0 ◂— 0x0
... ↓               3 skipped
─────────────────────────────────[ BACKTRACE ]──────────────────────────────────
 ► f 0   0x7ffff7a7b00b raise+203
   f 1   0x7ffff7a5a859 abort+299
   f 2         0x52769e
   f 3         0x5315c3
   f 4         0x52760d
   f 5         0x5315c3
   f 6         0x52760d
   f 7         0x51dd51
────────────────────────────────────────────────────────────────────────────────

backtrace_msg:
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7a5a859 in __GI_abort () at abort.c:79
#2  0x000000000052769e in wabt::(anonymous namespace)::CWriter::Write (this=0x7fffffffce30, exprs=...) at ../../../../src/c-writer.cc:1969
#3  0x00000000005315c3 in wabt::(anonymous namespace)::CWriter::Write<wabt::(anonymous namespace)::Newline, wabt::intrusive_list<wabt::Expr> const&> (this=0x7fffffffce30, t=..., u=...) at ../../../../src/c-writer.cc:205
#4  0x000000000052760d in wabt::(anonymous namespace)::CWriter::Write (this=0x7fffffffce30, exprs=...) at ../../../../src/c-writer.cc:1945
#5  0x00000000005315c3 in wabt::(anonymous namespace)::CWriter::Write<wabt::(anonymous namespace)::Newline, wabt::intrusive_list<wabt::Expr> const&> (this=0x7fffffffce30, t=..., u=...) at ../../../../src/c-writer.cc:205
#6  0x000000000052760d in wabt::(anonymous namespace)::CWriter::Write (this=0x7fffffffce30, exprs=...) at ../../../../src/c-writer.cc:1945
#7  0x000000000051dd51 in wabt::(anonymous namespace)::CWriter::Write<wabt::intrusive_list<wabt::Expr> const&, wabt::(anonymous namespace)::LabelDecl> (this=0x7fffffffce30, t=..., u=...) at ../../../../src/c-writer.cc:204
#8  0x000000000051bd15 in wabt::(anonymous namespace)::CWriter::Write (this=0x7fffffffce30, func=...) at ../../../../src/c-writer.cc:1423
#9  0x000000000051b647 in wabt::(anonymous namespace)::CWriter::Write<wabt::(anonymous namespace)::Newline, wabt::Func const&, wabt::(anonymous namespace)::Newline> (this=0x7fffffffce30, t=..., u=..., args=...) at ../../../../src/c-writer.cc:205
#10 0x000000000051182d in wabt::(anonymous namespace)::CWriter::WriteFuncs (this=0x7fffffffce30) at ../../../../src/c-writer.cc:1393
#11 0x0000000000500bf4 in wabt::(anonymous namespace)::CWriter::WriteCSource (this=0x7fffffffce30) at ../../../../src/c-writer.cc:2794
#12 0x00000000004ffcd7 in wabt::(anonymous namespace)::CWriter::WriteModule (this=0x7fffffffce30, module=...) at ../../../../src/c-writer.cc:2807
#13 0x00000000004ff48d in wabt::WriteC (c_stream=0x7fffffffdaa0, h_stream=0x7fffffffdaa0, header_name=0x7ccce0 <str> "wasm.h", module=0x7fffffffd2b0, options=...) at ../../../../src/c-writer.cc:2819
#14 0x00000000004f11b4 in ProgramMain (argc=3, argv=0x7fffffffe078) at ../../../../src/tools/wasm2c.cc:179
#15 0x00000000004f37f2 in main (argc=3, argv=0x7fffffffe078) at ../../../../src/tools/wasm2c.cc:190
#16 0x00007ffff7a5c083 in __libc_start_main (main=0x4f37d0 <main(int, char**)>, argc=3, argv=0x7fffffffe078, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffe068) at ../csu/libc-start.c:308
#17 0x000000000043e44e in _start ()
```

keithw:
~~(This PR is sequenced behind #1875, #1814, #1877 and #1999.)~~

This PR adds support for the reference-types proposal (per https://github.com/WebAssembly/wabt/pull/1853#issuecomment-1062094906). This makes it possible to run all of the non-SIMD spec tests in the current testsuite.

Closes #1737. Closes #1985.


src/c-writer.cc
```diff
@@ -796,6 +803,12 @@ void CWriter::Write(Type type) {
     case Type::I64: Write("u64"); break;
     case Type::F32: Write("f32"); break;
     case Type::F64: Write("f64"); break;
+    case Type::FuncRef:
+      Write("wasm_rt_funcref_t");
+      break;
+    case Type::ExternRef:
+      Write("wasm_rt_externref_t");
+      break;
```
sbc100, line 811: Ditto single line cases.

src/c-writer.cc
```diff
@@ -807,6 +820,12 @@ void CWriter::Write(TypeEnum type) {
     case Type::I64: Write("WASM_RT_I64"); break;
     case Type::F32: Write("WASM_RT_F32"); break;
     case Type::F64: Write("WASM_RT_F64"); break;
+    case Type::FuncRef:
+      Write("WASM_RT_FUNCREF");
+      break;
+    case Type::ExternRef:
+      Write("WASM_RT_EXTERNREF");
+      break;
```
sbc100, line 828: Ditto single line cases.

src/c-writer.cc
```diff
@@ -1158,10 +1202,11 @@ void CWriter::BeginInstance() {
         break;
       }
 
-      case ExternalKind::Table: {
-        Write("wasm_rt_table_t");
+      case ExternalKind::Table:
+        Write("wasm_rt_",
+              GetReferenceTypeName(cast<TableImport>(import)->table.elem_type),
+              "_table_t");
```
sbc100, line 1208: Can we add a helper called `WriteTableType` that takes a type name? Then we could use it in `WriteTable` and `WriteTablePtr` below too?

src/c-writer.cc
```diff
@@ -1537,20 +1584,30 @@ void CWriter::WriteElemInitializers() {
...
   for (const ElemSegment* elem_segment : module_->elem_segments) {
     if (elem_segment->kind != SegmentKind::Active) {
       continue;
     }
 
-    Write("table_init(", ExternalInstancePtr(table->name), ", ");
+    const Table* table = module_->GetTable(elem_segment->table_var);
+
+    if (table->elem_type != Type::FuncRef) {
+      WABT_UNREACHABLE;
```
sbc100, line 1607: Don't we have tests for elem segments that contain extern refs?

src/c-writer.cc
```diff
@@ -1638,7 +1695,7 @@ void CWriter::WriteExports(WriteExportsKind kind) {
         const Table* table = module_->GetTable(export_->var);
         mangled_name = ExportName(MangleName(export_->name));
         internal_name = table->name;
-        WriteTablePtr(mangled_name);
+        WriteTablePtr(mangled_name, GetReferenceTypeName(table->elem_type));
```
sbc100, line 1698: Can we have WriteTablePtr take the table itself as arg2 and have that function to the `GetReferenceTypeName` call?

src/c-writer.cc
```diff
@@ -1967,7 +2023,8 @@ void CWriter::WriteParamTypes(const FuncDeclaration& decl) {
 
 void CWriter::WriteLocals(const std::vector<std::string>& index_to_name) {
   Index num_params = func_->GetNumParams();
-  for (Type type : {Type::I32, Type::I64, Type::F32, Type::F64}) {
+  for (Type type : {Type::I32, Type::I64, Type::F32, Type::F64, Type::FuncRef,
+                    Type::ExternRef}) {
```
sbc100, line 2027: Is the some way to C++ to say `for (Type type : AllTypeEnumValues)` to avoid spelling this out here and below?

src/c-writer.cc
```diff
@@ -2489,7 +2552,12 @@ void CWriter::Write(const ExprList& exprs) {
             module_->tables[module_->GetTableIndex(inst->table_index)];
         const ElemSegment* src_segment =
             module_->GetElemSegment(inst->segment_index);
-        Write("table_init(", ExternalInstancePtr(dest_table->name), ", ");
+        if (dest_table->elem_type != src_segment->elem_type) {
+          WABT_UNREACHABLE;
+        }
```
sbc100, line 2557: Should this be an assert instead?   (I guess it would be validation error if this were true.. so it should be impossible for that to happen here?)

src/c-writer.cc
```diff
@@ -2531,20 +2599,107 @@ void CWriter::Write(const ExprList& exprs) {
         Table* dest_table =
             module_->tables[module_->GetTableIndex(inst->dst_table)];
         const Table* src_table = module_->GetTable(inst->src_table);
-        Write("table_copy(", ExternalInstancePtr(dest_table->name), ", ",
+        if (dest_table->elem_type != src_table->elem_type) {
+          WABT_UNREACHABLE;
+        }
```
sbc100, line 2604: assert instead?

src/c-writer.cc
```diff
@@ -3387,6 +3542,30 @@ Result CWriter::WriteModule(const Module& module) {
...
+    default:
+      WABT_UNREACHABLE;
+  }
+}
+
+// static
+const char* CWriter::GetReferenceNullValue(const Type& type) {
+  switch (type) {
+    case Type::FuncRef:
+      return "(wasm_rt_funcref_t){0, NULL, 0}";
```
sbc100, line 3561: Can we define `wasm_rt_funcref_null_value` too?   Also, can we avoid having `wasm_rt_externref_null_value` in the wasm-rt.h header file if its only  used here in the generated code?

sbc100 (commented):
Awesome!  Just a few more nits and we good to land this I think.

wasm2c/wasm-rt.h
```diff
@@ -308,23 +329,46 @@ uint32_t wasm_rt_grow_memory(wasm_rt_memory_t*, uint32_t pages);
...
+void wasm_rt_allocate_funcref_table(wasm_rt_funcref_table_t*,
+                                    uint32_t elements,
+                                    uint32_t max_elements);
+
+/**
+ * Free a funcref Table object.
+ */
+void wasm_rt_free_funcref_table(wasm_rt_funcref_table_t*);
+
+/** Initialize an externref Table object with an element count of `elements` and
```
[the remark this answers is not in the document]
keithw: Done.

wasm2c/wasm-rt-impl.c
```diff
@@ -350,27 +350,54 @@ void wasm_rt_free_memory(wasm_rt_memory_t* memory) {
...
 
-void wasm_rt_allocate_table(wasm_rt_table_t* table,
-                            uint32_t elements,
-                            uint32_t max_elements) {
-  table->size = elements;
-  table->max_size = max_elements;
-  table->data = calloc(table->size, sizeof(wasm_rt_elem_t));
-}
+#define DEFINE_TABLE_OPS(type)                                          \
+  void wasm_rt_allocate_##type##_table(wasm_rt_##type##_table_t* table, \
```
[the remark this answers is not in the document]
keithw: Afraid so...

wasm2c/examples/fac/fac.c
```diff
@@ -485,58 +485,96 @@ static inline void memory_init(wasm_rt_memory_t* dest,
 
 typedef struct {
   uint32_t func_type_index;
-  wasm_rt_funcref_t func;
+  wasm_rt_function_ptr_t func;
   size_t module_offset;
 } wasm_elem_segment_expr_t;
```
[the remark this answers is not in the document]
keithw: It's definitely possible! The `module_offset` field here is what handles that situation (which I believe is tested in the testsuite, e.g. in linking.wast). FWIW this code was added already in handling bulk memory ops.

The reason these are separate types is that a `wasm_rt_funcref_t` includes a pointer to an instance of the originating module, which is what you want when calling a funcref. But when CWriter is representing the elem expr, we don't have the instance of the originating module because the module hasn't been instantiated yet. So instead the elem segment expr contains basically a pointer-to-member; the offset within the module instance of where the function's originating module instance will be found, so that `funcref_table_init` (previously just `table_init`) can properly initialize the table at runtime. Also, the `wasm_elem_segment_expr_t` stores the function's type index within the module, whereas a `wasm_rt_funcref_t` stores the global opaque type ID returned by `wasm_rt_register_func_type`.

wasm2c/examples/fac/fac.c
```diff
@@ -485,58 +485,96 @@ static inline void memory_init(wasm_rt_memory_t* dest,
...
+    }                                                                        \
+                                                                             \
+    if (dest->data + dest_addr < src->data + src_addr) {                     \
+      for (u32 i = 0; i < n; i++) {                                          \
+        dest->data[dest_addr + i] = src->data[src_addr + i];                 \
+      }                                                                      \
+    } else {                                                                 \
+      for (u32 i = n; i > 0; i--) {                                          \
+        dest->data[dest_addr + i - 1] = src->data[src_addr + i - 1];         \
+      }                                                                      \
```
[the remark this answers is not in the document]
keithw: Hmm, indeed (as memmove, as it has to handle overlapping copies). Split out in #2012.

test/wasm2c/old-spec/README.md
```diff
@@ -1,7 +0,0 @@
-WebAssembly spec tests from before the bulk-memory and reference-types
```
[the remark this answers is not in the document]
keithw: Indeed! We only have SIMD left before wasm2c passes the whole testsuite again (+ multi-memory and exception-handling).

test/spec-wasm2c-prefix.c
```diff
@@ -169,6 +189,29 @@ static bool is_equal_u64(u64 x, u64 y) {
...
+  return (wasm_rt_externref_t)x;
+}
+
+wasm_rt_funcref_t funcref(u64 x) {
+  if (x == 0) {
+    return (wasm_rt_funcref_t){0, NULL, 0};
+  } else {
+    abort();
+  }
+}
```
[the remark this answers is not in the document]
keithw: Done.

src/c-writer.cc
keithw: The linter seems to force this format for newly added sections. :-/

src/c-writer.cc
keithw: Same -- the linter seems to force this formatting for newly added lines of code. :-/

src/c-writer.cc
keithw: Sure, done.

src/c-writer.cc
keithw: No... I don't think it's (currently) possible to represent an elem segment of externrefs. I think they are generally opaque to the Wasm module; only the host knows how to create one.

src/c-writer.cc
keithw: Sure, done.

src/c-writer.cc
keithw: Unfortunately I don't know of a _great_ way, and it's made more complicated by the fact that we're skipping a bunch of Type enum values that aren't meaningful or supported here (e.g. I8, Func, Void, Reference, Any, V128...).

src/c-writer.cc
keithw: I agree this should be a validation error, but my understanding is that the validator doesn't currently catch this one. :-( Per https://github.com/WebAssembly/wabt/issues/1612

src/c-writer.cc
keithw: (same as previous)

src/c-writer.cc
keithw: Sure, done (wasm_rt_funcref_null_value). We wanted wasm_rt_externref_null_value to be in the wasm-rt.h header so that if the embedder wants to choose a different type for wasm_rt_externref_t (something other than a void*), then the embedder can also specify the appropriate null value. (In Fix we're using externref = m256 instead of void*.)

src/c-writer.cc
sbc100: Presumably a module could declare an element segment containing null externrefs?  Not needed for PR either way... since all tests pass.

src/c-writer.cc
sbc100: Can you wrap the whole case statement in `// clang-format off  ... // clang-format on` to avoid having two different styles.  We can re-visit the style question later perhaps.

keithw: Done. I didn't know about that!

src/c-writer.cc
keithw: Done (same as above)

src/c-writer.cc
keithw: This is... hmm. This is not supposed to be possible (https://webassembly.github.io/spec/core/syntax/modules.html#element-segments: "In the current version of WebAssembly, only tables of element type can be initialized with an element segment. This limitation may be lifted in the future."), and in the binary format, there is currently no way to represent an element segment of any type other than funcref (https://webassembly.github.io/spec/core/binary/modules.html#element-section defines elemkind 0x00 as funcref and no other elemkinds are defined).

However, I can't find where this same formal limitation is imposed on the text format. wat2wasm currently accepts an active element segment of type externref, but then wasm-validate rejects the resulting module. I think this needs further study -- I will look into it.

src/c-writer.cc
sbc100: Still, why not use assert here?

wasm2c/wasm-rt.h
```diff
@@ -308,23 +332,48 @@ uint32_t wasm_rt_grow_memory(wasm_rt_memory_t*, uint32_t pages);
...
+                                      uint32_t max_elements);
 
 /**
- * Free a Table object.
+ * Free an externref Table object.
  */
-void wasm_rt_free_table(wasm_rt_table_t*);
+void wasm_rt_free_externref_table(wasm_rt_externref_table_t*);
+
+/** Grow a Table object by `delta` elements (giving the new elements the value
```
sbc100, line 367: Can you put opening and closing comment marks on lines by themsevles.

sbc100 (approved):
lgtm with a final comments

wasm2c/wasm-rt.h
keithw: Done.

src/c-writer.cc
keithw: My thinking was that an assert isn't strong enough here, because cmake disables asserts on the Release builds. :-/ If we made this an assert, then a user could take an invalid module like this:
```wat
(module
 (table 1 externref)
 (func $f)
 (elem (i32.const 0) func $f)
)
```
and successfully pass it through wat2wasm and wasm2c (release build) and all they would get is a warning when compiling the generated code (`warning: passing argument 1 of ‘funcref_table_init’ from incompatible pointer type; note: expected ‘wasm_rt_funcref_table_t *’ but argument is of type ‘wasm_rt_externref_table_t *’`), and unless they compile everything with `-Werror`, the module would end up with the ability to create a garbage externref that the host never gave it. This feels kind of icky to have as a known bug...

sbc100: True. but UNREACHABLE isn't much better as an end user experience is it?  Wouldn't it manifest as a crash, which then would require debugger to figure out what happens?  IIRC it doesn't even given a line number.

I guess the real fix is to update the validator.    How we add a TODO that says something like "This should really be and assert since the validator should catch this, but currently it doesn't?"

keithw: Done.

src/c-writer.cc
keithw: Submitted https://github.com/WebAssembly/spec/issues/1543

PR #1888, opened 2022-04-04
wasm2c: support final (non-SIMD) spec tests by adding an unsafe "ignore validation failure" flag

keithw:
(This PR is sequenced behind ~~#1875,~~ #1814, #1877, and #1887.)

To pass the ref_func.wasm and unreached-valid.wast tests (the final missing non-SIMD spec tests in the core testsuite), wasm2c needs to be willing to continue translating a module that has failed to validate. (The interpreter doesn't seem to enforce validity as strictly.)

Expose this behavior in wasm2c behind a `--unsafe-ignore-validation-failure` flag, and add the final two missing tests.


keithw:
I'm going to close this -- per #1894, the better approach is probably to fix the underlying issues that are causing these modules to fail validation. After d08b0d9f, the only remaining case involves validating `br_table` in the `unreached-valid.wast` test (https://github.com/WebAssembly/wabt/pull/1894#issuecomment-1097489677).

PR #1889, opened 2022-04-04
Fix parsing for typed function reference types

### LINKED ISSUE #1881: Typed funcref support should support subtyping and "concrete type" equality
A test like the following currently crashes in type-checking:

```
;;; TOOL: wat2wasm
;;; ARGS: --enable-function-references
;;; ERROR: 1
(module
  (type $f32-f32-1 (func (param f32) (result f32)))
  (type $f32-f32-2 (func (param f32) (result f32)))

  (func $foo (param $f (ref $f32-f32-1)) (result f32)
    (call_ref (f32.const 42.0) (local.get $f))
  )

  (func $bar (type $f32-f32-2)
    (f32.const 1.0)
  )

  (func (export "main") (result f32)
    ;; $f32-f32-1 and $f32-f32-2 should be equal
    (call $foo (ref.func $bar))
  )

  (elem declare funcref (ref.func $bar))
)

(;; STDERR ;;;
;;; STDERR ;;)
```

Error:

```
- test/typecheck/funcref-equality.txt (roundtrip)                                   
  b'Signal raised running "wat2wasm": SIGABRT\nwat2wasm: ../../../src/type.h:132: wabt::Index wabt::Type::GetReferenceIndex() const: Assertion `enum_ == Enum::Reference\' failed.\n'
```

I think the immediate reason is that type checking should account for subtyping, and make sure to compare heap types instead of assuming all reference types have an index (i.e., are a "concrete reference type" as described in the GC spec).

More generally, I think the type-checker needs to also be extended to look at the type section (or a more abstract "type store") to look up type indices when checking equality of concrete types. Alternatively, concrete ref types should be stored pre-canonicalized as a pointer (this would depend on https://github.com/WebAssembly/wabt/pull/1828) to a representative type in the canonicalized type store.

takikawa:
Previously an indexed ref type was parsed as a `Var`, assuming that regular types would be represented as index vars and indexed ref types would be named vars. The names would be resolved to actual indices later in a separate pass.

Unfortunately, it's also possible for a ref type to be an indexed var that overlaps with a type (e.g., the type "any" which is 0x0 and the index 0 for the type `(ref 0)`).

One way to fix this would be to return a `Type` *and* a `Var` from `ParseValueType` and then to handle this everywhere that `ParseValueType` is used. This would require handling the additional `Var` and resolving the type index everywhere that type parsing is done, such as for globals and so on as well.

Instead, it's also possible to avoid threading through all this state by just doing the lookups inside the `Module` that we are constructing. Since the `Module` is not always available during parsing, and there can be nested modules set up, this needs to be done a bit carefully so the patch also adds an RAII class to help track this.

This PR will close #1881.


takikawa:
Pinging @dbezhetskov in case you have any thoughts on this approach too. :)

src/wast-parser.cc
```diff
@@ -866,8 +866,17 @@ Result WastParser::ParseValueType(Var* out_type) {
...
+    Var name;
+    CHECK_RESULT(ParseVar(&name));
+    if (out_var) {
+      *out_var = name;
+    }
     EXPECT(Rpar);
+    if (name.is_index()) {
+      *out_type = Type(Type::Reference, name.index());
+    } else {
+      *out_type = Type(Type::Reference, kInvalidIndex);
```
sbc100, line 923: Should this not be an error?

sbc100:
I think the idea is that builtin types should all have negative values .. so we know that thing >= to zero is a user type.   The fact that we have this internal "Any" think value of zero sounds like a bug.

src/wast-parser.cc
```diff
@@ -3147,15 +3148,11 @@ Result WastParser::ParseGlobalType(Global* global) {
   WABT_TRACE(ParseGlobalType);
   if (MatchLpar(TokenType::Mut)) {
     global->mutable_ = true;
-    Var type;
-    CHECK_RESULT(ParseValueType(&type));
-    global->type = Type(type.index());
+    CHECK_RESULT(ParseValueType(&global->type));
```
sbc100, line 3151: I like the simplification we get from have ParseValueType return the Type directly.

keithw:
@takikawa Do you have cycles to respond to this review? Would be nice to get this merged (especially as #1661 is blocked on a bug that this fixes).

Issue #1890, opened 2022-04-05
Typed funcref: handle unbound type names in validation


takikawa:

zherczeg:
I am new here, and may misunderstand something but the `IsModuleField(PeekPair())` while loop completes in `WastParser::ParseModuleFieldList` before calling `ResolveFuncTypes(module, errors_)`, so all types should be available.

https://github.com/WebAssembly/wabt/blob/main/src/wast-parser.cc#L1165

I try to figure out what is happening here.

zherczeg:
#1892 eliminates the assert issue

PR #1892, opened 2022-04-11
Throw error when a reference is not found.

zherczeg:


zherczeg:
This is an attempt to fix #1890 
The patch enhances `ResolveTypeName` to return with an error. However location info is missing, and tests are also missing. If this is the right way to fix this please let me know.

Issue #1893, opened 2022-04-12
[reference types]: `ref.func` declaration check is overly strict


takikawa:
When function indices are used in `ref.func`, there is a check to make sure the index is declared in the element section before the code section.

However, this check is too strong in wabt as uses in the global section should be ok. Example:

```
;; the reference in the global is ok
(module (func $f1) (global funcref (ref.func $f1)))
```

It fails to validate in wabt (validates in the reference interpreter though):

```
func.wat:1:46: error: function 0 is not declared in any elem sections
(module (func $f1) (global funcref (ref.func $f1)))
```

takikawa:
Closed by #1894

PR #1894, opened 2022-04-12
Fix checking of ref.func index declarations

takikawa:
The validation was overly strict for ref.func index uses. In the spec, the ref index just needs to appear in

> the set of function indices occurring in the module, except in its functions or start function.

(https://webassembly.github.io/spec/core/valid/modules.html#valid-module)

which includes uses in the global and export sections. There were some test failures in the spec tests that were fixed by this, so the PR also changes test expectations.

Fixes issue #1893

### DIFF
--- a/src/shared-validator.cc
+++ b/src/shared-validator.cc
@@ -952,12 +952,12 @@ Result SharedValidator::OnRefFunc(const Location& loc, Var func_var) {
   Result result = CheckInstr(Opcode::RefFunc, loc);
   result |= CheckFuncIndex(func_var);
   if (Succeeded(result)) {
-    // Reference declarations aren't needed for uses in global sections, and
-    // the use in a global itself counts as a declaration.
-    if (!in_init_expr_) {
-      check_declared_funcs_.push_back(func_var);
-    } else {
+    // References in initializer expressions are considered declarations, as
+    // opposed to references in function bodies that are considered usages.
+    if (in_init_expr_) {
       declared_funcs_.insert(func_var.index());
+    } else {
+      check_declared_funcs_.push_back(func_var);
     }
     Index func_type = GetFunctionTypeIndex(func_var.index());
     result |= typechecker_.OnRefFuncExpr(func_type);
--- a/src/shared-validator.cc
+++ b/src/shared-validator.cc
@@ -211,6 +211,7 @@ Result SharedValidator::OnExport(const Location& loc,
   switch (kind) {
     case ExternalKind::Func:
       result |= CheckFuncIndex(item_var);
+      declared_funcs_.insert(item_var.index());
       break;
 
     case ExternalKind::Table:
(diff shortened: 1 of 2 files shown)


keithw:
This looks right to me (you may have seen #1888 where I tried to work around this).

What do you think about https://github.com/WebAssembly/wabt/blob/main/test/spec/unreached-valid.txt , which seems to be the other test with failed validation?

src/shared-validator.cc
```diff
@@ -951,7 +952,13 @@ Result SharedValidator::OnRefFunc(const Location& loc, Var func_var) {
   Result result = CheckInstr(Opcode::RefFunc, loc);
   result |= CheckFuncIndex(func_var);
   if (Succeeded(result)) {
-    check_declared_funcs_.push_back(func_var);
+    // Reference declarations aren't needed for uses in global sections, and
+    // the use in a global itself counts as a declaration.
+    if (!in_init_expr_) {
```
sbc100, line 957: Maybe flip the arms and invert the condition so its in the positive?

src/shared-validator.cc
```diff
@@ -951,7 +952,13 @@ Result SharedValidator::OnRefFunc(const Location& loc, Var func_var) {
   Result result = CheckInstr(Opcode::RefFunc, loc);
   result |= CheckFuncIndex(func_var);
   if (Succeeded(result)) {
-    check_declared_funcs_.push_back(func_var);
+    // Reference declarations aren't needed for uses in global sections, and
+    // the use in a global itself counts as a declaration.
```
sbc100, line 956: Hoe about "References in initializer expressions are considered declarations, as opposed to references in function bodies that are considered usages".   (or something like that)

sbc100 (approved):
Thanks for working on this.  I do wonder if we are lacking in test coverage (in the spec repo) given that the overall number of passing/failing tests stayed the same here (only the specific outout text changed).

takikawa:
> What do you think about https://github.com/WebAssembly/wabt/blob/main/test/spec/unreached-valid.txt , which seems to be the other test with failed validation?

The problem there seems to be that the validation semantics of `br_table` was not totally clear (see https://github.com/WebAssembly/relaxed-dead-code-validation/issues/4) and had been revised in 2021. I think in any case, wabt should change to implement the current algorithm in https://webassembly.github.io/spec/core/appendix/algorithm.html#validation-of-opcode-sequences for `br_table`, which only checks that the current stack matches each target, rather than that all targets have equal type.

takikawa:
> Thanks for working on this. I do wonder if we are lacking in test coverage (in the spec repo) given that the overall number of passing/failing tests stayed the same here (only the specific outout text changed).

Thanks for the feedback, I think I've addressed the comments.

Re: tests, I think the issue here is that a spec test command like `(module ...)` that doesn't assert anything and just checks that the module parses/validates doesn't currently count as a pass/fail test case for the wabt spec test runner.

sbc100:
> > Thanks for working on this. I do wonder if we are lacking in test coverage (in the spec repo) given that the overall number of passing/failing tests stayed the same here (only the specific outout text changed).
> 
> Thanks for the feedback, I think I've addressed the comments.
> 
> Re: tests, I think the issue here is that a spec test command like `(module ...)` that doesn't assert anything and just checks that the module parses/validates doesn't currently count as a pass/fail test case for the wabt spec test runner.

OK, that sounds like maybe something we should fix but is not related to this change.

Are you saying that if wabt can fail to validate such a `(module ...)` and that won't case a test failure?   That sounds like something we should fix.

takikawa:
> Are you saying that if wabt can fail to validate such a `(module ...)` and that won't case a test failure? That sounds like something we should fix.

Right, because just a `(module ...)` on its own won't count as a test case. It will just have the stderr printout.

Just adding a `TallyCommand` here might solve this: https://github.com/WebAssembly/wabt/blob/main/src/tools/spectest-interp.cc#L1277 (and then many spec test expecations would need to be updated)

keithw:
#1895 is my attempt to address this with the smallest possible change (bombing out on error instead of adding TallyCommand and updating all tests).

PR #1895, opened 2022-04-13
update br_table validation rule + enforce module validity in spectest-interp

keithw:
- https://github.com/WebAssembly/wabt/pull/1888 and https://github.com/WebAssembly/wabt/pull/1894#issuecomment-1098434603 observed that spectest-interp wasn't checking for errors on a "module" command. (Failure of a module to validate wasn't, by itself, a test failure.)
- and https://github.com/WebAssembly/wabt/pull/1894#issuecomment-1097489677 observed that the semantics of `br_table` were relaxed after reference-types (leading to a test error message when wabt tried to enforce the old rule, but not a test failure because of the above)

This PR updates the `br_table` validation and then modifies `spectest-interp` to require success on "module" and "register" commands. (The only other test with this problem was just fixed in #1894.)

(Disclaimer: I'm not 100% positive that this correctly captures the new `br_table` validation algorithm -- it seems somehow too simple. But it does seem to be what the appendix (https://webassembly.github.io/spec/core/appendix/algorithm.html#validation-of-opcode-sequences) is asking for, and it does pass all of the core spec tests.)


src/tools/spectest-interp.cc
```diff
@@ -1274,15 +1274,21 @@ wabt::Result CommandRunner::Run(const Script& script) {
   for (const CommandPtr& command : script.commands) {
     switch (command->type) {
       case CommandType::Module:
-        OnModuleCommand(cast<ModuleCommand>(command.get()));
+        if (Failed(OnModuleCommand(cast<ModuleCommand>(command.get())))) {
+          PrintError(command->line, "invalid module");
+          return wabt::Result::Error;
```
sbc100, line 1279: Why break out of the loop here?  Wouldn't you want to continue to try to run other tests?

src/tools/spectest-interp.cc
```diff
@@ -1274,15 +1274,21 @@ wabt::Result CommandRunner::Run(const Script& script) {
...
 
       case CommandType::Action:
         TallyCommand(OnActionCommand(cast<ActionCommand>(command.get())));
         break;
 
       case CommandType::Register:
-        OnRegisterCommand(cast<RegisterCommand>(command.get()));
+        if (Failed(OnRegisterCommand(cast<RegisterCommand>(command.get())))) {
+          PrintError(command->line, "invalid register command");
+          return wabt::Result::Error;
```
sbc100, line 1290: I guess register is different because it can/will effect all future commands so hard here makes more sense.

sbc100 (commented):
Regarding the br_table validation rules I think we can but fairly sure its good if the spec tests all path.

lgtm % comment.

BTW, why not include modules in the tally.. then I would expect the see the overall number of tests go up in most files.. which would tell us just how much coverage we've been missing out on.

src/tools/spectest-interp.cc
keithw: Sure, happy to do it that way -- it just requires rebasing all the test expectations. Done in latest commit.

keithw:
Done. If it's any solace, the corresponding wasm2c spec tests (for the same wast files) have been enforcing module validity all along, and that goes through basically the same codepath, so it hasn't exactly been missing out on this test coverage completely. (And of course the absence of an error message in the .txt file was also test coverage -- there were only two tests that had this issue, and now it's going to be zero.)

sbc100 (approved):
Wow.. we have quite a lack coverage there it seems.

PR #1896, opened 2022-04-13
Remove signature mangling from wasm2c output

### LINKED ISSUE #1858: Should we remove signature mangling from wasm2c output?
;tldr; I think we should.

It has been proposed in #1833 that we don't need to be mangling the names at all (except to turn them into valid C symbols).

In particular there is no need to append the mangled signature to the symbol names.  e.g. foo_vi (for a void function `foo` that takes an i32).

For exports there should be no issue at all since exports already need to have unique names.   However, wasm has historically allowed the same symbol to imported with various different signatures.  In particular the test suite it self used to do this when it imported the same `spectest.print` symbol with various signatures.    However, this was (conveniently) removed in https://github.com/WebAssembly/spec/pull/652.    In the design repo there was some discussion of removing duplicate imports: https://github.com/WebAssembly/design/issues/1402.  However that doesn't seem like its doing to happen.   In fact the discussion seems to be moving towards giving the host more power to link module together based on both names and positions (and signatures) of the imports.

Based the above it seems like the naive linking scheme we currently have is already insufficient and not general enough to provide all the things that a wasm host might want to do.. so I propose that we remove name mangling and add a more powerful linking mechanism (that is not just based on the symbol resolution in the linker) if we even need one.

sbc100:
This effectively means that we no longer support imports that are
overloaded by signature only, which is not something that we need to
support in order to support the core wasm spec.

This feature is available in the JS embedding but there is no good
reason (AFAICT) to support it in wasm2c, and this simplifies the
generated code.

Fixes #1858

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -161,13 +161,8 @@ class CWriter {
   static std::string Deref(const std::string&);
 
   static char MangleType(Type);
-  static std::string MangleTypes(const TypeVector&);
   static std::string MangleMultivalueTypes(const TypeVector&);
   static std::string MangleName(std::string_view);
-  static std::string MangleFuncName(std::string_view,
-                                    const TypeVector& param_types,
-                                    const TypeVector& result_types);
-  static std::string MangleGlobalName(std::string_view, Type);
   static std::string LegalizeName(std::string_view);
   static std::string ExportName(std::string_view mangled_name);
   std::string DefineName(SymbolSet*, std::string_view);
@@ -396,18 +391,6 @@ char CWriter::MangleType(Type type) {
 }
 
 // static
-std::string CWriter::MangleTypes(const TypeVector& types) {
-  if (types.empty())
-    return std::string("v");
-
-  std::string result;
-  for (auto type : types) {
-    result += MangleType(type);
-  }
-  return result;
-}
-
-// static
 std::string CWriter::MangleMultivalueTypes(const TypeVector& types) {
   assert(types.size() >= 2);
   std::string result = "wasm_multi_";
@@ -434,20 +417,6 @@ std::string CWriter::MangleName(std::string_view name) {
   }
 
   return result;
-}
-
-// static
-std::string CWriter::MangleFuncName(std::string_view name,
-                                    const TypeVector& param_types,
-                                    const TypeVector& result_types) {
-  std::string sig = MangleTypes(result_types) + MangleTypes(param_types);
-  return MangleName(name) + MangleName(sig);
-}
-
-// static
-std::string CWriter::MangleGlobalName(std::string_view name, Type type) {
-  std::string sig(1, MangleType(type));
-  return MangleName(name) + MangleName(sig);
 }
 
 // static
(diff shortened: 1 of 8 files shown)


keithw:
In principle this is good by us, and we'll conform to whatever you want, but, @angelamontemayor and @yhdengh had already done a version of this in #1814 (including the name of the module in the name of an exported function -- e.g. "Z_caesar_Z_rot13" for a module named "caesar" that exports a function named "rot13"), with #1877 and #1887 stacked behind it.

We added some checking for clearly erroneous duplicate imports (https://github.com/fixpointOS/wabt/blob/module_instancing/src/c-writer.cc#L953-L962) in response to @binji's review at https://github.com/WebAssembly/wabt/pull/1877#discussion_r842297572 , which might be a good idea in any case.

Bottom line, if it were up to us, our preference would be to do this as part of #1814 (after working through a review), rather than trying to rebase our PRs later. In the end, though, obviously it's your call and we'll work with whatever you prefer.

sbc100:
> In principle this is good by us, and we'll conform to whatever you want, but, @angelamontemayor and @yhdengh had already done a version of this in #1814 (including the name of the module in the name of an exported function -- e.g. "Z_caesar_Z_rot13" for a module named "caesar" that exports a function named "rot13"), with #1877 and #1887 stacked behind it.
> 
> We added some checking for clearly erroneous duplicate imports (https://github.com/fixpointOS/wabt/blob/module_instancing/src/c-writer.cc#L953-L962) in response to @binji's review at [#1877 (comment)](https://github.com/WebAssembly/wabt/pull/1877#discussion_r842297572) , which might be a good idea in any case.
> 
> Bottom line, if it were up to us, our preference would be to do this as part of #1814 (after working through a review), rather than trying to rebase our PRs later. In the end, though, obviously it's your call and we'll work with whatever you prefer.

I like the idea to doing it stages:

1. remove signature mangling
2. Remove optional prefixing (make it mandatory)
3. Add instancing (#1814)

These all seems like logically separate things and should make #1814 a lot smaller and more precise.

I'm also looking to ways to remove the `Z_` where possible.   It would be nicer do just have `caesar_rot13` as the symbol name... but its complicated because module names and function names can also contain "_".

keithw:
Ok, fine with us.

On #2 (remove optional prefixing) after #3, we think it's simplest if instance naming dictates the prefixing:
- the names of exported functions should be determined by the name of the module (e.g. "caesar_rot13")
- the exported globals, memories, and tables will be members of the corresponding instance structure, which in turn is named something like "caesar_module_instance_t".

But we can cross that bridge when we come to it.

We'd love to get rid of "Z"; I think what we'd have to do at that point would be to escape "_" if it appears in the name of a module or function export name. If every exported function ("modname_funcname") has an "_" in it somewhere, then I don't think you have to worry about colliding with a C reserved word (like "void").

sbc100:
I also have PR for removal of optional prefixing: #1897.   I made these a few weeks ago but went on vacation before uploading them.   Sorry I didn't realize that both of them were already basically contains in  #1814

PR #1897, opened 2022-04-13
wasm2c: Always use a module prefix and set one by default

sbc100:
The module prefix is no longer optional and is now set by default to the
name of the module as given in the name section, or taken from the name
of the input file.

A new `-n/--module-name` command line flag can also be used to override.

The prefix used by the generated code is now fixed at wasm2c time which
makes the output code easier to read and also avoid the symbol pasting
in the C pre-processor which makes the source hard to understand.  For
example, it makes symbols hard to `grep` for.


sbc100:
For now this change leaves the `Z_` mangling in place.  We can consider ways to remove that as a followup.

src/wasm2c.h.tmpl
```diff
@@ -1,20 +1,10 @@
...
-
 #include <stdint.h>
 
 #include "wasm-rt.h"
 
-#ifndef WASM_RT_MODULE_PREFIX
-#define WASM_RT_MODULE_PREFIX
-#endif
-
 #define WASM_RT_PASTE_(x, y) x ## y
```
keithw, line 6: `WASM_RT_PASTE_` and `WASM_RT_PASTE` can also be eliminated now.

src/c-writer.cc
```diff
@@ -2376,11 +2379,14 @@ void CWriter::Write(const LoadZeroExpr& expr) {
 void CWriter::WriteCHeader() {
   stream_ = h_stream_;
   std::string guard = GenerateHeaderGuard();
+  Write("/* Automically generated by wasm2c */", Newline());
```
keithw, line 2382: should probably say "Automatically"

src/tools/wasm2c.cc
```diff
@@ -78,6 +78,14 @@ static void ParseOptions(int argc, char** argv) {
         s_outfile = argument;
         ConvertBackslashToSlash(&s_outfile);
       });
+  parser.AddOption(
+      'n', "module-name", "MODNAME",
+      "Unique name for the module being generated. This name is prefixes to "
```
keithw, line 83: This help text needs more newlines to wrap nicely in the `wasm2c --help` output. Should say "prefixed" (instead of "prefixes").

src/tools/wasm2c.cc
```diff
@@ -78,6 +78,14 @@ static void ParseOptions(int argc, char** argv) {
...
+      "each"
```
keithw, line 84: Needs a space after "each" (before "of")

src/tools/wasm2c.cc
```diff
@@ -78,6 +78,14 @@ static void ParseOptions(int argc, char** argv) {
...
+      "of the global C symbols that are generated. By default the module\n"
+      "module name from the names section is used. If that is not present the\n"
```
keithw, line 86: "module" is repeated here from the previous line

src/tools/wast2json.cc
```diff
@@ -139,9 +139,8 @@ int ProgramMain(int argc, char** argv) {
     }
 
     if (Succeeded(result)) {
-      for (auto iter = module_streams.begin(); iter != module_streams.end();
-           ++iter) {
-        result = iter->stream->WriteToFile(iter->filename);
+      for (auto& pair : module_streams) {
```
keithw, line 142: `const auto&`?

wasm2c/README.md
```diff
@@ -187,20 +182,6 @@ from stack-size exhaustion). This defaults to 500:
 #endif
 ```
 
-Next we can specify a module prefix. This is useful if you are using multiple
-modules that may use the same name as an export. Since we only have one module
-here, it's fine to use the default which is an empty prefix:
-
-```c
-#ifndef WASM_RT_MODULE_PREFIX
```
keithw, line 195: The "Exported symbols" section of wasm2c/README.md (lines 318-337 in the revised file) also needs to be revised now. (I guess GitHub doesn't let me leave a comment down there because there are no changes...)

keithw (approved):
lgtm % comments

Issue #1692, opened 2021-07-13
accommodate outputting the binary wasm to stdout


sfranzyshen:
## update: jump to the bottom of all this ...

Pipes are used to give the output (stdout) of a command as the input (stdin) to another command ... command line piping has always been a power part of the Unix command line. This feature is absent in wabt. Is this something to consider ... as the concept is already being discussed and implemented within the webassembly.sh and wasi environments for example ...

### example:
`cat some.wasm | wasm2wat | wat2wasm > copyofsome.wasm`

sfranzyshen:
from what I can see there was at least the attempt to support this by using a input file name of "-" but it doesn't work. Somewhere during the parsing of the command line the "-" filename thing gets lost ... I tried to follow along the process flow of the wasm2wat utility ... but I cannot figure out where ... Here's a example of how it fails ...
```
$cat some.wasm | wasm2wat -
stat failed: No such file or directory

```
* [wabt/src/tools/wasm2wat.cc](https://github.com/WebAssembly/wabt/blob/main/src/tools/wasm2wat.cc)
  *  [static std::string s_infile;](https://github.com/WebAssembly/wabt/blob/main/src/tools/wasm2wat.cc#L38) 
  * [int main(int argc, char** argv)](https://github.com/WebAssembly/wabt/blob/main/src/tools/wasm2wat.cc#L145)
    * [int ProgramMain(int argc, char** argv)](https://github.com/WebAssembly/wabt/blob/main/src/tools/wasm2wat.cc#L100)
      * [static void ParseOptions(int argc, char** argv)](https://github.com/WebAssembly/wabt/blob/main/src/tools/wasm2wat.cc#L60)
        * [parser.AddArgument("filename", OptionParser::ArgumentCount::One,](https://github.com/WebAssembly/wabt/blob/main/src/tools/wasm2wat.cc#L92)
        * [parser.Parse(argc, argv);](https://github.com/WebAssembly/wabt/blob/main/src/tools/wasm2wat.cc#L97) =>  [wabt/src/option-parser.cc](https://github.com/WebAssembly/wabt/blob/main/src/option-parser.cc)
      * [result = ReadFile(s_infile.c_str(), &file_data);](https://github.com/WebAssembly/wabt/blob/main/src/tools/wasm2wat.cc#L107) => [wabt/src/common.cc](https://github.com/WebAssembly/wabt/blob/main/src/common.cc)
[...]

PR #1902, opened 2022-04-15
wat2wasm: support "-" for outputting to stdout (#1692)

remko:


binji (approved):
lgtm, but maybe worth mentioning that "-" works in the wat2wasm.cc help too

remko:
> lgtm, but maybe worth mentioning that "-" works in the wat2wasm.cc help too

Done

src/stream.cc
```diff
@@ -157,6 +157,18 @@ Result OutputBuffer::WriteToFile(std::string_view filename) const {
   return Result::Ok;
 }
 
+Result OutputBuffer::WriteToStdout() const {
```
sbc100, line 160: Why not call this `WriteToFile(FILE* file)` then it can be used both with `stdout_fileno` or with the result of fopen, and we can avoid duplicating the code between WriteToFile and WriteToStdout which do mostly the same thing.

remko: @sbc100 I considered doing that, but wasn't sure what to do with the error message in the shared part (which contains the filename): emitting the error message at the call site and doing an if error check there too (more code), or passing a filename into the shared part (sounds bad)

sbc100: OK sounds reasonable.

PR #1909, opened 2022-04-28
Implement \u{hex} escape parsing

zherczeg:
Support \u{hex} escape sequences for string literals. These literals are converted to UTF-8 characters.

### DIFF
--- a/src/wast-lexer.cc
+++ b/src/wast-lexer.cc
@@ -354,6 +354,39 @@ Token WastLexer::GetStringToken(WastParser* parser) {
             }
             break;
 
+          case 'u': {
+            token_start_ = cursor_ - 2;
+            if (ReadChar() != '{') {
+              goto error;
+            }
+
+            // Value must be a valid unicode scalar value.
+            uint32_t digit;
+            uint32_t scalar_value = 0;
+
+            while (IsHexDigit(PeekChar())) {
+              ParseHexdigit(*cursor_++, &digit);
+
+              scalar_value = (scalar_value << 4) | digit;
+              // Maximum value of a unicode code point.
+              if (scalar_value >= 0x110000) {
+                goto error;
+              }
+            }
+
+            if (PeekChar() != '}') {
+              goto error;
+            }
+
+            // Scalars between 0xd800 and 0xdfff are not allowed.
+            if ((scalar_value >= 0xd800 && scalar_value < 0xe000) ||
+                token_start_ == cursor_ - 3) {
+              ReadChar();
+              goto error;
+            }
+            break;
+          }
+
           default:
             token_start_ = cursor_ - 2;
             goto error;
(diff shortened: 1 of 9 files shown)


test/parse/string-utf-escape.txt
```diff
@@ -0,0 +1,2 @@
+;;; TOOL: wat2wasm
+(module (func) (export "a:\u{0}b:\u{00001}c:\u{7f}d:\u{80}e:\u{7ff}f:\u{800}g:\u{ffff}h:\u{10000}i:\u{10ffff}" (func 0)))
```
zherczeg, line 2: This is a similar file to string-escape.txt, but I couldn't find its corresponding binary expected output file. It would be good to prove the encoding is correct.

test/parse/bad-string-utf-escape-large.txt
```diff
@@ -0,0 +1,14 @@
+;;; TOOL: wat2wasm
```
sbc100, line 1: Should these be called `unicode-escape`?   I'm not actually sure.. I guess the string is utf8 but the thing being escaped is a unicode scalar value right?

test/parse/string-utf-escape.txt
zherczeg: I am not familiar with the term round tripping. Is that a command?

src/wast-parser.cc
```diff
@@ -76,6 +76,44 @@ void RemoveEscapes(std::string_view text, OutputIter dest) {
...
+          do {
+            if (Succeeded(ParseHexdigit(src[0], &digit))) {
+              value = (value << 4) | digit;
+            } else {
+              assert(0);
+            }
+            src++;
+          } while (src[0] != '}');
+
+          assert(value < 0x110000);
```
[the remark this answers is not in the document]
zherczeg: Looks like this is a correct term for these values. I only heard code points before. I can add extra comments of course.

src/wast-lexer.cc
```diff
@@ -354,6 +354,37 @@ Token WastLexer::GetStringToken(WastParser* parser) {
             }
             break;
 
+          case 'u': {
+            token_start_ = cursor_ - 2;
+            if (ReadChar() != '{') {
+              goto error;
```
zherczeg, line 360: Currently every error displays the same simple error message. If you prefer, I can add more sophisticated error messages as well.

sbc100: I think this is fine for now.

test/parse/string-utf-escape.txt
sbc100: Take a look at `test/roundtrip/`..  I think this test should probably live there.

zherczeg:
Thank you for the review, I have updated the patch.

rossberg:
That would be great, yes. Nobody has bothered contributing lexical tests before, it seems.

sbc100 (approved):

PR #1910, opened 2022-04-29
wasm2c: Avoid unnecessary `extern` in function decls. NFC

sbc100:


sbc100:
> lgtm.
> 
> If you want to get rid of all of these, I think here are a few more locations:
> 
> * most of the wasm-rt.h function declarations have `extern`, e.g. `extern uint32_t wasm_rt_register_func_type()`, `extern void wasm_rt_allocate_memory()`, etc.

Done.

> * src/c-writer.cc currently writes `extern` for all types of imports (https://github.com/WebAssembly/wabt/blob/main/src/c-writer.cc#L861) and exports (https://github.com/WebAssembly/wabt/blob/main/src/c-writer.cc#L1139); I think if we really don't want to use `extern` where it's already the default (on a function declaration), these sites could also be modified.

At first I thought that same thing, but because they are declaring pointers to functions, and not functions, the export is actually needed in these cases.

keithw:
Ah, yes, you're right.

PR #1911, opened 2022-04-29
[WIP] wasm2c: group imports for a given module into a struct type.

sbc100:
Rather than given each import a magic name we can group imports into a
single struct and pass that struct into the init function.

This means that in the generated code all access to imported elements
now goes via the `g_imports` struct.  Note that this is not a pointer
but a copy of the passed in struct, to avoid adding any extra
indirection.

This can be considered one step towards module instancing.

It also avoids the need for name mangling since individual imports no
longer live in the global namespace.


sbc100:
@keithw .. this is just a WIP .. but perhaps it makes a good stepping stone to #1814.

PR #1928, opened 2022-05-27
List funcref as one of the expected value types in WastParser::ParseValueType

asb:
Previously, externref was listed but not funcref.


asb:
(Force pushed a version fixing the clang-format issue).

I think to be comprehensive, we'd make v128 conditional on simd being enabled, and externref/funcref conditional on reference types being enabled. Doable, but as you say it's not obvious it's worth it.

Issue #1929, opened 2022-05-27
Out-of-bound Read at `BinaryReaderInterp::OnReturnCallIndirectExpr`


Mem2019:
Commit: `83a226bcd1816c86fb1fcb0963c6a11e2c373aaf`

## Reproduce

Sample: [interp_sig_index_oob.zip](https://github.com/WebAssembly/wabt/files/8789638/interp_sig_index_oob.zip)

```bash
./wasm-objdump -d interp_sig_index_oob.wasm
./wasm-interp --enable-tail-call interp_sig_index_oob.wasm
```

Output:

```
interp_sig_index_oob.wasm:  file format wasm 0x1

Code Disassembly:

000063 func[0] <_start>:
 000064: 01 7c                      | local[0] type=f64
 000066: 01 7d                      | local[1] type=f32
 000068: 01 7e                      | local[2] type=i64
 00006a: 01 7f                      | local[3] type=i32
 00006c: 13 13 13                   | return_call_indirect 19 19
 
 
=================================================================
==4201==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6060000008b8 at pc 0x559d6f7da883 bp 0x7ffd3d189f90 sp 0x7ffd3d189f80
READ of size 8 at 0x6060000008b8 thread T0
    #0 0x559d6f7da882 in std::vector<wabt::Type, std::allocator<wabt::Type> >::size() const /usr/include/c++/9/bits/stl_vector.h:916
    #1 0x559d6f823a4a in GetReturnCallDropKeepCount /wabt/src/interp/binary-reader-interp.cc:438
    #2 0x559d6f8360be in OnReturnCallIndirectExpr /wabt/src/interp/binary-reader-interp.cc:1146
    #3 0x559d6f935ab9 in ReadInstructions /wabt/src/binary-reader.cc:879
    #4 0x559d6f92de6d in ReadFunctionBody /wabt/src/binary-reader.cc:609
    #5 0x559d6f95d8d3 in ReadCodeSection /wabt/src/binary-reader.cc:2731
    #6 0x559d6f960d23 in ReadSections /wabt/src/binary-reader.cc:2885
    #7 0x559d6f961b6e in ReadModule /wabt/src/binary-reader.cc:2946
    #8 0x559d6f9620fd in wabt::ReadBinary(void const*, unsigned long, wabt::BinaryReaderDelegate*, wabt::ReadBinaryOptions const&) /wabt/src/binary-reader.cc:2964
[...]

Issue #1924, opened 2022-05-22
Assertion `is_name()` failed at wasm-decompiler caused by empty global name


Mem2019:
Commit: `d8517aa922ddd3ffd6d7f28f00d2bb0f9853cf88`

## Reproduce

Fuzzed Sample: [wabt_decomp_is_name_assert.zip](https://github.com/WebAssembly/wabt/files/8750089/wabt_decomp_is_name_assert.zip)

Simplified Sample:

```
$ cat simplified.wat 
(module
  (type (;0;) (func (result i64)))
  (func (;0;) (type 0) (result i64)
    global.get 0)
  (global (;0;) i64 (i64.const 8))
  (export "" (global 0))
)
$ ./wat2wasm ./simplified.wat && ./wasm-decompile ./simplified.wasm 
wasm-decompile: /wabt/src/ir.h:61: const string& wabt::Var::name() const: Assertion `is_name()' failed.
Aborted (core dumped)
```

## Analysis

`struct Var` has a `union` field, and it has a `type_` field to indicate which union entry should be accessed.

```c++

// Definition of VarType
enum class VarType {
  Index,
  Name,
};

// Private fields of `struct Var` 
  VarType type_;
  union {
    Index index_;
    std::string name_;
  };
```

If `type_` is `Index`, `index_` should be accessed; if `type_` is `Name`, `name_` should be accessed. Current assertion error is caused by we are accessing `name_` of a `struct Var` whose `type_` is `Index`, and in release build in which `assert` is removed, the type confusion can occur (e.i. an unsigned integer will be used as a pointer).

### Root Cause

Here is the intended behavior first:

1. `ReadBinaryIr` is called in `ProgramMain` of `wasm-decompile` to parse the binary, in which `Var var` field of `GlobalGetExpr` is [initialized to](https://github.com/WebAssembly/wabt/blob/d8517aa922ddd3ffd6d7f28f00d2bb0f9853cf88/src/binary-reader-ir.cc#L924) `type_ == Index` first.
2. Then `ApplyNames` in `ProgramMain` is called to [convert](https://github.com/WebAssembly/wabt/blob/d8517aa922ddd3ffd6d7f28f00d2bb0f9853cf88/src/apply-names.cc#L426) `var` field of every `GlobalGetExpr` to `type_ == Name` version.
[...]

Issue #1922, opened 2022-05-21
OOB Read at wasm-decompile caused by lack of BrTable handling


Mem2019:
Commit: `d8517aa922ddd3ffd6d7f28f00d2bb0f9853cf88`

## Reproduce

Fuzzed Sample: 
[decomp_size_ge_nargs.zip](https://github.com/WebAssembly/wabt/files/8748269/decomp_size_ge_nargs.zip)

Simplified Sample:

```bash
$ cat simplified.wat
(module
  (type (;0;) (func (param i32 i32) (result i32)))
  (func (;0;) (type 0) (param i32 i32) (result i32)
    local.get 0
    loop (param i32) (result i32)  ;; label = @1
      unreachable
      br_table 0 (;@1;) 0 (;@1;)
      i32.add
    end))
$ ./wat2wasm simplified.wat && ./wasm-decompile simplified.wasm
wasm-decompile: wabt/src/decompiler-ast.h:98: wabt::Node& wabt::AST::InsertNode(wabt::NodeType, wabt::ExprType, const wabt::Expr*, wabt::Index): Assertion `exp_stack.size() >= nargs' failed.
Aborted
```

## Analysis

In debug mode, assertion failure is triggered; in release mode, since there is no such assertion, `std::move` after such length check can cause an OOB read.

```c++
std::move(exp_stack.end() - nargs, exp_stack.end(),
          std::back_inserter(n.children));
// Since size of `exp_stack` is less than `nargs`,
// `exp_stack.end() - nargs` points to a negative index
```

### Root Cause

[...]

PR #1931, opened 2022-06-01
Fix several issues found by fuzzing

Mem2019:
Fixes https://github.com/WebAssembly/wabt/issues/1922
Fixes https://github.com/WebAssembly/wabt/issues/1924
Fixes https://github.com/WebAssembly/wabt/issues/1929

### DIFF
--- a/src/decompiler-ast.h
+++ b/src/decompiler-ast.h
@@ -226,7 +226,6 @@ struct AST {
             mc.GetLabel(cast<BrIfExpr>(&e)->var)->label_type;
         return;
       }
-      // https://github.com/WebAssembly/wabt/issues/1922
       case ExprType::BrTable: {
         InsertNode(NodeType::Expr, ExprType::BrTable, &e, 1).u.lt =
             mc.GetLabel(cast<BrTableExpr>(&e)->default_target)->label_type;
--- a/src/decompiler-ast.h
+++ b/src/decompiler-ast.h
@@ -226,6 +226,12 @@ struct AST {
             mc.GetLabel(cast<BrIfExpr>(&e)->var)->label_type;
         return;
       }
+      // https://github.com/WebAssembly/wabt/issues/1922
+      case ExprType::BrTable: {
+        InsertNode(NodeType::Expr, ExprType::BrTable, &e, 1).u.lt =
+            mc.GetLabel(cast<BrTableExpr>(&e)->default_target)->label_type;
+        return;
+      }
       default: {
         InsertNode(NodeType::Expr, e.type(), &e, arity.nargs);
         return;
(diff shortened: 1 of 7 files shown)


src/decompiler-ast.h
```diff
@@ -226,6 +226,12 @@ struct AST {
             mc.GetLabel(cast<BrIfExpr>(&e)->var)->label_type;
         return;
       }
+      // https://github.com/WebAssembly/wabt/issues/1922
```
sbc100, line 229: While its useful for the review, I don't think we need to annotate all these locations with bug urls.

It would be great to see tests added though.

Normally for santizer failures one need to construct a test under `test/binary/` containing the module the triggers the bad code path.   Since the tests run under santizers it should be possible to verify that the tests fail before this change and pass afterwards.

sbc100 (commented):
Could you update the title so be a little more specific.   Perhaps "Fix several issues found by fuzzing", and the you can do "Fixes #xxx, #yyy" in the description so have github automatically close those issue.

keithw:
@Mem2019 Thank you for fixing these! Do you have cycles to make the changes requested in the review and get this merged? Would be great to have this in...

keithw (approved):
@sbc100 In an effort to get these issues closed, I made the requested changes and confirmed that the four regression tests fail before the patch is applied (and pass afterwards). Will merge early next week unless you want to take another look.

src/interp/binary-reader-interp.cc
```diff
@@ -1169,6 +1172,9 @@ Result BinaryReaderInterp::OnReturnCallExpr(Index func_index) {
 
 Result BinaryReaderInterp::OnReturnCallIndirectExpr(Index sig_index,
                                                     Index table_index) {
+  if (sig_index >= module_.func_types.size()) {
+    return Result::Error;
+  }
```
sbc100, line 1177: I wonder if these checks should be in the validator instead?

sbc100: Actually it looks like we already have this code above, can we call this instead?

 ```
CHECK_RESULT(
      validator_.OnCall(GetLocation(), Var(func_index, GetLocation())));		    
```

keithw: Ah, good point. I changed them to use validator_.OnReturnCall and validator_.OnReturnCallIndirect which looks like what it should have been doing all along.

sbc100 (approved):
I wonder if that fact that we were missing those validation checks means there are some tests that are missing from the spec tests?  i.e. why didn't the spec tests find these?  I'm they are missing from the spec tests we should followup by filing bugs upstream and eventually remove these downstream tests.

keithw:
> I wonder if that fact that we were missing those validation checks means there are some tests that are missing from the spec tests? i.e. why didn't the spec tests find these?

I looked into this. In this case, the tail-call proposal tests would have caught these binary-reader-interp bugs (causing a segfault or asan failure). But we're not running the tail-call proposal tests because the interpreter doesn't support tail calls yet. :-(

Issue #1927, opened 2022-05-26
wast2json changes element section contents in `(module binary ...)`


osa1:
Tried with d8517aa9.

test.wast:

```wast
(module binary
  "\00asm" "\01\00\00\00"
  "\04\04\01"                          ;; Table section with 1 entry
  "\70\00\00"                          ;; no max, minimum 0, funcref
  "\09\09\01"                          ;; Element section with 1 entry
  "\02"                                ;; Element with explicit table index
  "\80\00"                             ;; Table index 0, encoded with 2 bytes
  "\41\00\0b\00\00"                    ;; (i32.const 0) with no elements
)
```

(Copied from [spec test suite][1])

Generate Wasm module with wast2json:

```
$ wast2json test.wast 
```

Generated Wasm module is different than the input module:

```
$ hd test.0.wasm
00000000  00 61 73 6d 01 00 00 00  04 04 01 70 00 00 09 06  |.asm.......p....|
00000010  01 00 41 00 0b 00                                 |..A...|
00000016
```

Note that after `0x09` (element section), section size is 6 in the generated Wasm module. In the input the section size is 9. Section contents (element segments) are also different.

[1]: https://github.com/WebAssembly/testsuite/blob/e25ae159357c055b3a6fac99043644e208d26d2a/binary-leb128.wast#L32-L40

sbc100:
Hmm.. that does seems like a serious issue.  wast2json should simply be writing those bytes out verbatim.   Any idea whats going on?

osa1:
As far as I understand, when wast2json sees a `(module binary ...)` it still parses the module and then converts the parsed module back to binary. So it doesn't directly emit the bytes it sees.

I just checked and the module contents are as expected after `WastParser::ParseScriptModule`.

However for binary modules `WastParser::ParseModuleCommand` still parses the module contents to a `Module`: https://github.com/WebAssembly/wabt/blob/d8517aa922ddd3ffd6d7f28f00d2bb0f9853cf88/src/wast-parser.cc#L3319-L3343

In the code above, `bsm->data` is correct, so the problem should be when serializing the parsed module back to binary. It seems like the module deserializer and serializer do not have roundtrip property: if you deserialize a binary module and serialize it again you sometimes get a different serialization.

This may or may not be a problem, depending on what wast2json wants to do. I'm not a Wasm expert but I'm guessing it's possible to have two different binary encodings of a single module (e.g. we can probably add empty sections without changing module semantics?).

[...]

PR #1932, opened 2022-06-02
wast2json: write binary modules verbatim

binji:
Previously any top-level module defined in a .wast file would be parsed
and converted to a `Module` and stored in a `ModuleCommand`, since it
may be used later by the script (e.g. via an `invoke` command). This
means that when it was written out later it may have been changed, which
is undesirable for binary and quoted modules.

This change adds a new `ScriptModuleCommand`, which stores both a
`Module` and a `ScriptModule`. The `Module` contains the parsed
information, and the `ScriptModule` contains the original contents of
the binary or quoted data.

This fixes issue #1927.

### DIFF
--- a/test/wast2json/module-binary.txt
+++ b/test/wast2json/module-binary.txt
@@ -1,0 +1,15 @@
+;;; TOOL: wast2json
+;;; ARGS: -v
+(module binary
+  "\00asm" "\01\00\00\00"
+  "\04\04\01"                          ;; Table section with 1 entry
+  "\70\00\00"                          ;; no max, minimum 0, funcref
+  "\09\09\01"                          ;; Element section with 1 entry
+  "\02"                                ;; Element with explicit table index
+  "\80\00"                             ;; Table index 0, encoded with 2 bytes
+  "\41\00\0b\00\00"                    ;; (i32.const 0) with no elements
+)
+(;; STDERR ;;;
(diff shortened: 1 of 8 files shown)


test/wast2json/module-binary.txt
```diff
@@ -0,0 +1,15 @@
+;;; TOOL: wast2json
```
sbc100, line 1: How does this test work?  I thought wast2json write a json file with binaries on the side.

sbc100 (commented):
When wast2json spits out .wat files are that byte for byte verbatim or are they also round-tripped trough IR.

It seems like all the output files from wast2json should be verbatim chunks of the input, right?

test/wast2json/module-binary.txt
binji: It does, but it also outputs the contents of the module if you pass `-v` too

binji:
> When wast2json spits out .wat files are that byte for byte verbatim or are they also round-tripped trough IR.
> 
> It seems like all the output files from wast2json should be verbatim chunks of the input, right?

The wat files are also written out directly verbatim too, at least when used in `assert_malformed` constructs. The issue here was specifically with top-level module definitions. AFAIK there are no examples of using a "quoted" module at the top level, because there's no benefit to doing so. This currently will error out, but could be supported later if needed.

PR #1939, opened 2022-06-30
wasm2c: run tests with -O2 on non-Windows

### LINKED ISSUE #1925: Several spec tests fail if wasm2c output compiled with `-O` or `-O2`
wasm2c is failing several spec tests if we add `-O` or `-O2` to the compiler command line.

- Example commit here: https://github.com/fixpointOS/wabt/commit/551b3dba69f9d35ddf0271e0e0383e3525d2e50f
- Corresponding CI output on Ubuntu: https://github.com/fixpointOS/wabt/runs/6547538885?check_suite_focus=true

There seem to be at least a few buckets:

1. The WebAssembly spec requires that "There is no observable difference between quiet and signalling NaNs," and the spec tests enforce that some operations transform signalling NaNs into quiet NaNs (e.g. here: https://github.com/WebAssembly/spec/blob/main/test/core/float_exprs.wast#L2361).

   But, gcc and clang often make optimizations that preserve signalling NaNs, even when the same code, compiled without optimization, will turn a signalling NaN into a quiet NaN. E.g. in this code which is essentially what wasm2c generates for a promote/demote:

   ```c
   float func(float x)
   {
     double y = x;
     return y;
   }
   ```

   On x86-64 and compiling without optimization, gcc and clang will turn this into a cvtss2sd/cvtsd2ss, which turns a signalling NaN into a quiet NaN (passing the tests). If we compiled with `-O`, the function turns into a NOP and the signalling NaN is preserved (breaking the tests).  I couldn't find a way to change this behavior with fenv(3) or with a `-fno-...` optimization flag. One possible solution might be to mark all floating-point stack variables as `volatile`, but this may be a heavy price in terms of slowdown. Maybe there's a path forward that involves explicitly clearing a signalling NaN only when (a) storing into linear memory, (b) reinterpreting as integer, or (c) returning to a host function? It might be hard to guarantee this catches all the edge cases...
   
 2. The spec requires OOB loads/stores to trap deterministically, but when using mprotect & a signal handler to detect OOB memory access, gcc/clang `-O` omits loads/stores if they don't affect the output or a visible side-effect (which the segfault doesn't count as). E.g. even this test fails: https://github.com/WebAssembly/wabt/blob/main/test/wasm2c/address-overflow.txt
    
    I'm curious how this is done in other Wasm engines that also use the "signal handler trick" to detect OOB memory access (while, hopefully, also optimizing to remove dead code while still passing all the spec tests)?
   
3. The spec tests say, "Implementations are required to have every call consume some abstract resource towards exhausting some abstract finite limit, such that infinitely recursive test cases reliably trap in finite time," and enforce this, e.g. here: https://github.com/WebAssembly/spec/blob/master/test/core/call.wast#L337 But, I think gcc/clang `-O` are optimizing these kinds of runaway recursive functions into infinite iteration/proper tail recursion, so the exhaustion segfault never arrives. (I fear this issue was recently introduced by #1875.)

---
For our purposes, we'd like to compile wasm2c output with optimization, and we also want the "almost-deterministic"/spec-conforming behavior of WebAssembly, so it seems worthwhile to get the tests passing even with `-O`. I'd love to hear if this has been discussed before and what others think about the best route (@binji @sbc100 @kripken ?).

keithw:
~~(This PR is stacked behind #1814, #1877, #1887, and #1930, but if there is interest in landing it sooner, it could be rebased on top of the current ToT.)~~

Enable optimization when compiling the wasm2c output on non-Windows platforms (effectively GCC and clang). This required:
- Preventing load instructions from being optimized away if their value is unused (using inline assembly with an input operand and empty code). This is necessary to force an OOB trap on platforms that use mprotect and the signal handler to detect OOB.
- Disabling tail-call optimization in the compiler, to make sure that infinite recursion traps. (This required bumping the version of macOS in GitHub Actions to get a new-enough AppleClang. We should revert this back to 'macos-latest' as soon as that becomes the default.)
- Using NaN-quieting versions of a bunch of FP ops that were previously only used on Windows, and adding floor/ceil and promotion/demotion.
- Using the '-frounding-math' and '-fsignaling-nans' compiler flags to tell GCC and clang not to fold certain FP ops (e.g. subtracting zero, multiplying by 1).

Fixes #1925.


sbc100:
I think it would be great if this and all the other changes in the stack could be rebased to avoid #1814.. then we could land them without being blocked on the instancing discussions.

Could you arrange to have the instancing at the other end of the chain, or do that others somehow depend deeply on it?

sbc100:
> #1877 (bulk memory) and #1887 (reference types) unfortunately depend on #1814 in a way that would be a bit annoying to rebase (more details here about why: [#1877 (comment)](https://github.com/WebAssembly/wabt/pull/1877#issuecomment-1112504465)).
> 
> The other two would be much easier to rebase (#1930 (exceptions) and #1939 (tests pass with optimization)), and I'm happy to do it if it would help get things moving.
> 
> But, instancing is important to us (and I imagine other users) and it would be our hope to actually have that discussion and reach alignment (especially now that we've been building on top of this code for five months in our local fork and have ticked the boxes at [#1853 (comment)](https://github.com/WebAssembly/wabt/pull/1853#issuecomment-1062094906)). How can we help get things moving again there?

My apologies for dragging my feet on that.   Perhaps we could have a meeting discuss it next week?

keithw:
Happy to -- I'll send email, but in general July 6 looks pretty good on our end and we're happy to come up to SF.

wasm2c/wasm-rt.h
```diff
@@ -264,22 +264,6 @@ void wasm_rt_allocate_table(wasm_rt_table_t*,
...
-float wasm_rt_fabsf(float x);
-double wasm_rt_fabs(double x);
-#else
-#define wasm_rt_truncf(x) truncf(x)
-#define wasm_rt_trunc(x) trunc(x)
-#define wasm_rt_nearbyintf(x) nearbyintf(x)
-#define wasm_rt_nearbyint(x) nearbyint(x)
-#define wasm_rt_fabsf(x) fabsf(x)
-#define wasm_rt_fabs(x) fabs(x)
-#endif
```
sbc100, line 281: Is there some way we could keep these here in the shared header rather than making them part of the generated code?     Having them live in generator makes them a little hard to work on and make the diffs harder to read.

Also, I wonder if we could have a define to enable and disable them?  For example `#ifdef WASM2C_NON_CONFORMING_FLOAT_BITS` so something like that.  Then folks who don't care about the determinism of the nan bits can make a faster/slimmer build?

test/run-spec-wasm2c.py
```diff
@@ -341,10 +341,13 @@ def Compile(cc, c_filename, out_dir, *args):
     if IS_WINDOWS:
         args += ['/nologo', '/MDd', '/c', c_filename, '/Fo' + o_filename]
     else:
-        args += ['-c', c_filename, '-o', o_filename,
+        args += ['-c', c_filename, '-o', o_filename, '-O2',
```
sbc100, line 344: Lets add some comment here maybe?  And perhaps move reference these flags in the wasm2c README file?

What about `-O3`, does that break anything?

wasm2c/wasm-rt.h
keithw: Sure, they could live anywhere, but it defeats inlining if they are part of the runtime -- my working assumption was that we wanted to reserve runtime hooks for things where (a) different embedders might want to implement different behavior at runtime, or (b) there's some necessary whole-runtime state, e.g. `wasm_rt_register_func_type`. Whereas these functions seem about as universal as `popcount`, `ctz`, `trunc_sat`, `divrem`, etc., and I think will benefit from inlining or at least the compiler being able to see the implementation.

How would you feel about doing this as a command-line argument to wasm2c, e.g., `--non-conforming-float-bits` ? I could imagine a similar flag for `--non-conforming-memory-loads` (which would remove the load liveness stuff at the cost of sometimes omitting a load that would have produced a deterministic trap).

(BTW if you want the template to be easier to work on, we _could_ split it into separate files for the `%%includes` and `%%declarations` sections, and then both of those would become fully-valid C files. And we could omit the `src/prebuilt/wasm2c.include.*` files (or even fac.c) from the repository and just always build them as part of the build system, which would make the diffs cleaner in Git.)

test/run-spec-wasm2c.py
keithw: Just checked and it still works iwth -O3 (at least with gcc on Linux).

wasm2c/wasm-rt.h
sbc100: If we are going to have flags for controlling spec compliance do you think they should be wasm2c arguments that control codegen, or should we emit code with `#ifdef` and make these compile time options?   I'm leaning towards that latter but I don't feel strongly and you have done more work on wasm2c recently, so I think we should pick the one that makes maintenance of wasm2c easier.

keithw: I don't have a strong view -- honestly the easiest is probably just to not do it until somebody asks for it, but second easiest is probably the `#ifdef` option. Just pushed this (with `WASM2C_NON_CONFORMING_NAN_BITS` as the symbol).

test/run-spec-wasm2c.py
keithw: Added a discussion of the flags to the wasm2c README.

sbc100 (approved):
lgtm.. did we decide that we don't need to give folks an opt-out for these these things are don't relate directly to sandbox security? (but are there for spec compliance).

I think I'm OK wither way..

keithw:
I think that was our tentative conclusion, but we're okay either way too. I guess my mild preference is to wait until somebody wants it and avoid the added complexity if we can.

(Are you able to change the repo configuration to accept a "macos-12" status check instead of "macos-latest"? GitHub has suggested they will update "macos-latest" by August so hopefully it's temporary...)

sbc100:
> I think that was our tentative conclusion, but we're okay either way too. I guess my mild preference is to wait until somebody wants it and avoid the added complexity if we can, but no strong preference.
> 
> (Are you able to change the repo configuration to accept a "macos-12" status check instead of "macos-latest"? GitHub has suggested they will update "macos-latest" by August so hopefully it's temporary...)

Do I have to do something to enable that?   I thought it was all driven by the yaml file?

keithw:
> Do I have to do something to enable that? I thought it was all driven by the yaml file?

I think the list of "required" status checks is only in the GitHub "Settings" > "Branches" > Branch Protection Rules for main > "Edit" > Status Checks that Are Required. I think it needs to have "build (macos-12)" added and remove "build (macos-latest)" before this can be merged afaik, at least without an override.

PR #1945, opened 2022-07-13
wasm2c: Use `static` for wasm-rt globals. NFC

sbc100:
And use the `wasm_rt_` prefix for the one non-static global.

While reviewing #1930 I noticed these globals could/should be static.

### DIFF
--- a/wasm2c/wasm-rt-impl.c
+++ b/wasm2c/wasm-rt-impl.c
@@ -46,23 +46,24 @@ typedef struct FuncType {
 } FuncType;
 
 #if WASM_RT_MEMCHECK_SIGNAL_HANDLER
-bool g_signal_handler_installed = false;
-char* g_alt_stack;
+static bool g_signal_handler_installed = false;
+static char* g_alt_stack;
 #else
 uint32_t wasm_rt_call_stack_depth;
-uint32_t g_saved_call_stack_depth;
+uint32_t wasm_rt_saved_call_stack_depth;
 #endif
 
-jmp_buf g_jmp_buf;
-FuncType* g_func_types;
-uint32_t g_func_type_count;
+static FuncType* g_func_types;
+static uint32_t g_func_type_count;
+
+jmp_buf wasm_rt_jmp_buf;
 
 void wasm_rt_trap(wasm_rt_trap_t code) {
   assert(code != WASM_RT_TRAP_NONE);
 #if !WASM_RT_MEMCHECK_SIGNAL_HANDLER
-  wasm_rt_call_stack_depth = g_saved_call_stack_depth;
+  wasm_rt_call_stack_depth = wasm_rt_saved_call_stack_depth;
 #endif
-  WASM_RT_LONGJMP(g_jmp_buf, code);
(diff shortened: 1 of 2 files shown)


wasm2c/wasm-rt-impl.c
```diff
@@ -46,23 +46,24 @@ typedef struct FuncType {
...
 
 #if WASM_RT_MEMCHECK_SIGNAL_HANDLER
-bool g_signal_handler_installed = false;
-char* g_alt_stack;
+static bool g_signal_handler_installed = false;
+static char* g_alt_stack;
 #else
-uint32_t wasm_rt_call_stack_depth;
-uint32_t g_saved_call_stack_depth;
+static uint32_t wasm_rt_call_stack_depth;
```
keithw, line 52: I think this one needs to keep external linkage because it's used both in the generated output (in the function prologue/epilogue) and here in the runtime implementation. (wasm-rt.h declares it extern.)

wasm2c/wasm-rt-impl.c
```diff
@@ -46,23 +46,24 @@ typedef struct FuncType {
 } FuncType;
 
 #if WASM_RT_MEMCHECK_SIGNAL_HANDLER
-bool g_signal_handler_installed = false;
-char* g_alt_stack;
+static bool g_signal_handler_installed = false;
```
keithw, line 49: Given the way the signal handler works it could be more meaningful for this to remain an extern global -- it's not like somebody can link two wasm-rt-impl.o's together in the same process and have the different signal handlers play nicely together.

keithw (approved):
lgtm % comments

wasm2c/wasm-rt-impl.c
sbc100: Hmm I guess you are correct.. i wonder how the tests passed though?

wasm2c/wasm-rt-impl.c
sbc100: Perhaps.. although I'm not sure that a symbol collision is the best way to prevent that... not sure what is though.

wasm2c/wasm-rt-impl.c
keithw: We only use wasm_rt_call_stack_depth on platforms without the signal handler to catch OOB/exhaustion (so basically just Windows and emscripten).

sbc100: OK leaving these as public.

So we not test wasm2c under windows in CI?

wasm2c/wasm-rt-impl.c
sbc100: I think I'll leave this one as static for now.   Double linking of wasm-rt-impl.o will already fail because of all the public symbols would be duplicately defined.

sbc100:
ptal

wasm2c/wasm-rt-impl.c
keithw: :+1: The CI does test wasm2c under Windows (and Emscripten) -- those two tests failed with the first version of the commit (https://github.com/WebAssembly/wabt/actions/runs/2660225161) but are passing now.

keithw:
lgtm

PR #1947, opened 2022-07-18
Fix interpreter to run opcodes consecutively

clover2123:
For better performance, I fixed the interpreter to run opcodes consecutively rather than single execution for each opcode.

### DIFF
--- a/src/interp/interp.cc
+++ b/src/interp/interp.cc
@@ -1071,31 +1071,6 @@ void Thread::PopValues(const ValueTypes& types, Values* out_values) {
   values_.resize(values_.size() - types.size());
 }
 
-RunResult Thread::Run(Trap::Ptr* out_trap) {
-  const int kDefaultInstructionCount = 1000;
-  RunResult result;
-  do {
-    result = Run(kDefaultInstructionCount, out_trap);
-  } while (result == RunResult::Ok);
-  return result;
-}
-
-RunResult Thread::Run(int num_instructions, Trap::Ptr* out_trap) {
-  DefinedFunc::Ptr func{store_, frames_.back().func};
-  for (; num_instructions > 0; --num_instructions) {
-    auto result = StepInternal(out_trap);
-    if (result != RunResult::Ok) {
-      return result;
-    }
-  }
-  return RunResult::Ok;
-}
-
-RunResult Thread::Step(Trap::Ptr* out_trap) {
-  DefinedFunc::Ptr func{store_, frames_.back().func};
-  return StepInternal(out_trap);
-}
-
 Value& Thread::Pick(Index index) {
   assert(index > 0 && index <= values_.size());
   return values_[values_.size() - index];
@@ -1138,9 +1113,13 @@ void Thread::Push(Ref ref) {
   values_.push_back(Value::Make(ref));
 }
 
-RunResult Thread::StepInternal(Trap::Ptr* out_trap) {
+RunResult Thread::Run(Trap::Ptr* out_trap) {
   using O = Opcode;
 
+  DefinedFunc::Ptr func{store_, frames_.back().func};
+  RunResult result = RunResult::Ok;
+
+NextOpcode:
   u32& pc = frames_.back().offset;
   auto& istream = mod_->desc().istream;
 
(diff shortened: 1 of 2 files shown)


clover2123:
From a simple evaluation below, this patch showed a significant performance improvement.
```c
// fibonacci calculation
#include <stdio.h>

int fib(int n) {
  if (n == 1) return 1;
  if (n == 2) return 1;
  return fib(n-1) + fib(n-2);
}

int main()
{
    fib(30);
    return 0;
}
```

Performance Result
- This patch : 22.1 sec
- Origin : 25.1 sec

src/interp/interp.cc
```diff
@@ -1138,775 +1122,1263 @@ void Thread::Push(Ref ref) {
...
-  if (trace_stream_) {
-    istream.Trace(trace_stream_, pc, trace_source_.get());
-  }
+  DefinedFunc::Ptr func{store_, frames_.back().func};
+  for (; num_instructions > 0; --num_instructions) {
+    u32& pc = frames_.back().offset;
+    auto& istream = mod_->desc().istream;
 
-  auto instr = istream.Read(&pc);
-  switch (instr.op) {
```
sbc100, line 1152: To avoid re-indenting this huge function (and to make this diff easier to read) pehaps keep the existing function as `RunOneInstruction()` and call it from within this loop?  if inclining is important it can be marked as such.

Also perhaps make a separate PR with the `using O = Opcode;` change and land that first?  (I thin that would also help make this diff more readable).

clover2123: Originally, interpreter was composed of one loop (`Thread::Run()`) that invokes a single opcode execution (`Thread::StepInternal()`) internally same as your suggestion. But this approach incurs exessive call overhead. So I just merge the loop and single execution call into one huge loop to reduce this call overhead (inlining the single opcode execution call into the loop). To do this merging, It seems that re-indenting is inevitable 😢 

`using O = Opcode;` is not newly added, but also included in the original single execution function (`Thread::StepInternal()`)
Shall I separate another PR for this?

sbc100: I see.

How about try marking the Step method as `__attribute__((always_inline))` ?

clover2123: I've updated the patch to avoid re-indenting as much as possible by adding a label (`NextOpcode`).
What do you think about this?

Actually, after this patch landed, I'm going to accelerate the interpreter performance more by adopting `threaded interpretation` method. This approach would greatly reduce switch-case overhead by directly jumping to the next opcode instead of decode-dispatch process. To apply this optimization, I first need to change the interpreter function as to one huge loop-switch code.

clover2123: I merged two `Thread::Run` functions and this showed a better performance for the same fibonacci code above.

* Update : 21.6 sec
* Origin : 25.1 sec

sbc100 (approved):
lgtm

Can I ask what you interest is in making the wabt interpreter fast?    IIUC the goal the wabt interpreter is not to be ask fast as possible, but to be correct and easy to maintain.   If you want to run wabt in production or at speed the wabt interpreter is probably not the right tool.

clover2123:
@sbc100 
I'm using `wasm-c-api` to run wasm codes in my project. And AFAIK `wabt` is the only wasm engine built by c/c++ that fully supports the `wasm-c-api` standard.

I'm also happy to contribute to this project and I'll upgrade the interpreter as much as possible not harming the overall sustainability or readability. Thanks :)

sbc100:
> @sbc100 I'm using `wasm-c-api` to run wasm codes in my project. And AFAIK `wabt` is the only wasm engine built by c/c++ that fully supports the `wasm-c-api` standard.
> 
> I'm also happy to contribute to this project and I'll upgrade the interpreter as much as possible not harming the overall sustainability or readability. Thanks :)

I though that v8 supported that API?   Does it now?  What about wasmtime? 

In any case, its great that you are using wabt and we will happily accept your patches, but please bear in mind that the wabt is probably not the best choice for a production runtime.. at least not in the long term.

clover2123:
@sbc100 
I need a lightweight wasm engine for several reasons now, so I choose `wabt`.

`wasmtime` is also one of good alternatives, but unfortunately it is based on `rust` language 😢 
I'll keep in mind your advice Thanks :)

SoniEx2:
it might be possible to simply add an appropriate inline attribute instead?

keithw:
Hi -- is there still interest in merging this (in which case I think the diff should probably be cleaned up to avoid re-formatting those big sections), or should we close as dormant?

clover2123:
This has been inactive for a long time, so I close it.

PR #1954, opened 2022-08-02
add wasm2wat offset in verbose mode

wilsonwang371:
in wasmtime, we sometimes only get the offset information. to be able to collect more information, we need to the offset information.

### DIFF
--- a/src/binary-reader-logging.cc
+++ b/src/binary-reader-logging.cc
@@ -28,6 +28,7 @@ namespace wabt {
 
 #define LOGF(...)               \
   do {                          \
+    WriteOffset();              \
     WriteIndent();              \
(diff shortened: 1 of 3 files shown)


sbc100:
> in wasmtime, we sometimes only get the offset information. to be able to collect more information, we need to the offset information.

Does `wasm-objdump -d` not give you the offsets you need?

sbc100:
> > in wasmtime, we sometimes only get the offset information. to be able to collect more information, we need to the offset information.
> 
> Does `wasm-objdump -d` not give you the offsets you need?

The reason I ask is because the logging is really a side effect of was2wasm, and not really designed to be an end user feature.

PR #1957, opened 2022-08-12
Generate SLSA signatures for GitHub release artifacts

laurentsimon:
Hi,

I'm reaching out on behalf of the [Open Source Security Foundation (openssf.org)](https://openssf.org/). We work on improving the security of critical open source projects like yours. 

Together with [GitHub](https://github.blog/2022-04-07-slsa-3-compliance-with-github-actions/), we designed a free, easy-to-use method of code signing. It will help your users verify that your release binaries were built from your repository’s [workflow](https://github.com/WebAssembly/wabt/blob/main/.github/workflows/build_release.yml) and not altered by anyone. It’s just a few lines of code, but it will make your project more secure against third-party tampering and attacks like [Codecov](https://about.codecov.io/apr-2021-post-mortem/) and [CTX](https://sockpuppets.medium.com/how-i-hacked-ctx-and-phpass-modules-656638c6ec5e).

This PR shows how to add this code signing to your workflow. You don’t have to be a cryptography expert or learn complicated tools and [verification](https://github.com/slsa-framework/slsa-verifier/blob/main/README.md#verification-of-provenance) is simple for your users.  

You can read more on the [SLSA blog](https://slsa.dev/blog/2022/06/slsa-github-workflows). Please reach out if you have any questions!


keithw:
Thanks for the pointers, especially the link to https://github.com/grpc-ecosystem/grpc-gateway/pull/2847#issuecomment-1212137854 -- that's very helpful. I have to admit this seems... super-complicated. :-(

- The PR seems to be asking us to 'contents: write' and 'id-token: write' permissions to the slsa-github-generator workflow that you maintain (and... the workflows that this workflow depends on?). Isn't that adding some exposure to our project, e.g. if you (or a workflow you depend on) get owned or have an exploitable bug? How should we think about this risk?

- You write, "This is not something GitHub can do by default," but... I don't follow why not. GitHub is executing our workflow on their GitHub-hosted runners, based on checking out a Git commit (pointed to by a tag), all of which GitHub hosts. The workflow produces release artifacts that they are uploading to their webserver and serving. If GitHub wants to sign a certificate stating that they built these artifacts on their runners from the such-and-such Git commit and workflow, why don't they just do that (signing the certificate with a well-known GitHub key)? I don't understand why we have to be involved in this. GitHub is the ones doing all the work. :-)

laurentsimon:
> Thanks for the pointers, especially the link to [grpc-ecosystem/grpc-gateway#2847 (comment)](https://github.com/grpc-ecosystem/grpc-gateway/pull/2847#issuecomment-1212137854) -- that's very helpful. I have to admit this seems... super-complicated. :-(

It may seems complicated, especially the first time you look at it. It builds upon standard blocks like transparency logs (for auditability; the same principle used for TLS https://certificate.transparency.dev/), OIDC (a way to authenticate a workflow).
You're correct that it still is a lot of work, and that's why it'a cross-company effort within the https://openssf.org/. We are working with ecosystem like npm (https://github.blog/2022-08-08-new-request-for-comments-on-improving-npm-security-with-sigstore-is-now-open/), and Maven (https://blog.sonatype.com/maven-central-and-sigstore), for example.

> * The PR seems to be asking us to add 'contents: write' and 'id-token: write' permissions to the slsa-github-generator workflow that you maintain (and... the workflows that this workflow depends on?). Isn't that adding some exposure to our project, e.g. if you (or a workflow you depend on) get owned or have an exploitable bug? How should we think about this risk?

Great question. `contents:write` has little consequence on your project per se. You have branch protection enabled. If a dependency of the builder was compromised, it could not push code to the repo. It could create release assets though. This is not really different from your existing workflow, since it does require `contents: write` https://github.com/WebAssembly/wabt/blob/main/.github/workflows/build_release.yml#L9-L10. However, because the provenance contains the version of the builder, we could trace exactly which build might have been compromised. This is one of the key features of having provenance. The certificate is attested to by GitHub, and a builder compromise cannot change that.

`id-token: write` allows the builder to sign, but the signature is bound to the builder's identity. This is why we could trace back and remove artifacts that could have been compromised. The identity of the builder is listed as a oid in the x509 cert.

> * You write, "This is not something GitHub can do by default," but... I don't follow why not. GitHub is executing our workflow on their GitHub-hosted runners, based on checking out a Git commit (pointed to by a tag), all of which GitHub hosts. The workflow produces release artifacts that they are uploading to their webserver and serving. If GitHub wants to sign a certificate stating that they built these artifacts on their runners from the such-and-such Git commit and workflow, why don't they just do that (signing the certificate with a well-known GitHub key)? I don't understand why we have to be involved in this. GitHub is the ones doing all the work. :-)

Technically, I think what you're saying would be possible. But that's not the direction they are taking. They prefer an opt-in solution for users. It has a few benefits:
- they don't need to sign everything, which means it scales better for them
- they don't need to maintain backward compatibility in the way they sign / format / etc. Having a different project is easier for users to follow semver, upgrade, use the version they want, etc.
- they want the solution to be community-driven
- even if they signed automatically, they would be limited in what they can attest to. It's hard for them to attest to the build steps that were taken, for example, without tracing every syscall. We have Go builders that do this, and we're working on extending it other ecosystems, later this year. All the builders generate the same attestation format, so there is an easy upgrade path without breaking consumers who verify the provenance. (Note: This PR does not populate build steps). Build steps let consumers re-build using the same steps that were taken, and try to reproduce builds. This adds further guarantees if they re-build on a different platform, because it's harder for an attacker to compromise many systems at once. Note that having build steps reported in the provenance is one reason we're building an npm trusted builder in collaboration with them

Hope this answers some of your concerns.

Please don't hesitate if you have more questions.

keithw:
I don't want to string you along -- maybe somebody else is more enthusiastic about this than me, but, I don't think the risk-benefit calculation supports merging this at this time.

If I understand correctly (which I may not), on the "risk" side,
- We'd be giving the power to create WABT release artifacts to you, your co-maintainers, your future co-maintainers, and the current and future co-maintainers of every action that your action depends on, now or in the future
- Even if no changes are made in the future, we're signing up for a network of interconnected actions that seem quite difficult to follow or audit. E.g., we'd use "generator_generic_slsa3.yml@v1.2.0", and that action uses "detect-workflow@1d646d70aeba1516af69fb0ef48206580122449b" and "upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8", and those actions use other versions of your workflow (e.g. the first one requires "slsa-github-generator v1.1.1", etc.).
- Because this is something "we"'re doing (instead of GitHub just going ahead and attesting to the artifacts of GitHub-hosted workflows), we seem more responsible for this process; i.e. we have to review your PR, probably future PRs when things change or need updating, and we're just more on the hook. Which suggests we should understand what we're doing, but:
- It seems pretty difficult to trace the logic from the human description. E.g.:
   - "The running workflow [...] generate an ECDSA private/public key pair." Where does this happen (making sure to get the version that is actually pointed-to by the currently referenced tag and its transitive dependencies)? What is the source of randomness used to generate the private/public key pair? Where is it documented that GitHub runners have an adequate source of high-quality randomness to generate a keypair like this?
   - "Then it proves its identity to a remote CA called Fulcio using OIDC." When you write "proves its identity," do you just mean it proves possession of the private half of the private/public key pair that it just created in the previous step? Or do you mean it proves that it's an authentic GitHub-hosted runner executing the authentic "build_release.yml" workflow from the WebAssembly/wabt repository? If the former, why is this interesting/valuable? (Can't anybody create a new keypair and then demonstrate possession of the private half...?) If the latter, well, how -- what secret does it possess that only the true GitHub-hosted runner of our workflow would have?
   - "standard blocks like transparency logs (for auditability; the same principle used for TLS...)". Transparency logs are useful for TLS server certificates because domain name owners want to make sure that no CA is issuing certificates claiming that the domain name in question belongs to a cryptographic identity that isn't theirs. I'm not sure I follow the analogy for this use case, though. Who's going to be reading these transparency logs to make sure nobody else is putting out WABT releases that didn't come from "our" GitHub-hosted runners running the workflow from one of "our" Git tags? Are we supposed to be scanning the log every day? How will we recognize an unauthorized entry -- how do we know it wasn't a valid release? Just based on timing and whether we were expecting a release that day...?
   
On the "benefit" side, as I understand it, people who want to know the provenance of WABT binaries today have a few options:
- Build from source in their own trusted environment (optionally: verifying that the binaries match the reproducible WABT build from somebody else, e.g. https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/wabt.html)
- Get the binaries from somebody like Debian who signs the binary package and effectively attests that it came from running the corresponding source package in a hermetic environment. (It seems to me that this is the low-drama way to achieve the stated goals here... basically just copy what Debian and similar distributors/builders have been doing. But it would more of an active role from GitHub.)
- Get the binaries from the GitHub release, compute the SHA-256 checksum, and then live-download the ".tar.gz.sha256" artifact from GitHub via HTTPS (doing an online cryptographic handshake with GitHub's public key to make sure you're really talking to GitHub) and compare the checksums.

As I understand it, this PR adds a fourth option, that is:
- Get the binaries from the GitHub release and also get the signed provenance, which you can largely verify offline without an online connection (but to truly verify the provenance requires a live connection to the transparency log?).

I see the appeal and I do support reproducible builds and known provenance, but in terms of the risk-benefit and the complexity burden we'd be taking on, I can only speak for myself, but I think my own vote would be for "wait-and-see" rather than merging this now. If this ends up being the direction that most of FLOSS goes in, I assume we'd go with the flow...

laurentsimon:
> I don't want to string you along -- maybe somebody else is more enthusiastic about this than me, but, I don't think the risk-benefit calculation supports merging this at this time.

np, I've added some answer to your comments. Feel free to close.

> 
> If I understand correctly (which I may not), on the "risk" side,
> 
> * We'd be giving the power to create WABT release artifacts to you, your co-maintainers, your future co-maintainers, and the current and future co-maintainers of every action that your action depends on, now or in the future

correct. This is not different from trusting Debian maintainers. Our action does not depend on external actions that have privileged permissions. It's also not very different from "trusting GitHub" only: like any other companies, they have engineer turnover; services they run rely on dependencies, etc. You don't see it, but it's there. For our builders you can see the code in an open-source repository, and who works on it if you want.

> * Even if no changes are made in the future, we're signing up for a network of interconnected actions that seem quite difficult to follow or audit. E.g., we'd use "generator_generic_slsa3.yml@v1.2.0", and that uses "detect-workflow@1d646d70aeba1516af69fb0ef48206580122449b" and "upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8", and those actions use other versions of your workflow (e.g. the first one requires "slsa-github-generator v1.1.1", etc.). Just trying to follow the transitive closure of what we're signing up for would require a lot of manual link-following to navigate all these different tags and commit hashes -- I'm not aware of a tool that just gathers everything.

These actions were created as modular components, the same way you'd write C++ using classes. The equivalent of classes in GitHub workflows are actions. There is one action that needs to be pinned by hash: `detect-env` because it detects the version / hash the builder is called at (this is due to a limitation of GitHub APIs but will be removed later). The others need not be pinned. Point taken, I'll see if we can remove the pinning for internal actions to improve auditability.

> * Because this is something "we"'re doing (instead of GitHub just going ahead and attesting to the artifacts of GitHub-hosted workflows), we seem more responsible for this process; i.e. we have to review your PR, probably future PRs when things change or need updating, and we're just more on the hook. 

You don't have to review our PR., but you need to update the version via dependabot.

> Which suggests we should understand what we're doing, but:
> * It seems pretty difficult to trace the logic from the human description. E.g.:
>   
>   * "The running workflow [...] generate an ECDSA private/public key pair." Where does this happen (making sure to get the version that is actually pointed-to by the currently referenced tag and its transitive dependencies)? 

It's a GitHub API to fetch an OIDC token. The token contains the ref / tag / workflow name. It happens inside the builder.

> What is the source of randomness used to generate the private/public key pair? Where is it documented that GitHub runners have an adequate source of high-quality randomness to generate a keypair like this?

GitHub runners run Ubuntu on Azure machine; and "support" TLS, so they have a source of randomness. Key pair uses the Linux PRNG (/dev/urandom), seeded from the hardware generator. FWIW, you have only 20mn to crack a keypair because validation qill check the certificate validity period (20mn). 

>   * "Then it proves its identity to a remote CA called Fulcio using OIDC." When you write "proves its identity," do you just mean it proves possession of the private half of the private/public key pair that it just created in the previous step? Or do you mean it proves that it's an authentic GitHub-hosted runner that was invoked to execute the "build_release.yml" 
workflow from the WebAssembly/wabt repository at some particular tag? 

both the above.


> If the former, why is this interesting/valuable? (Can't anybody create a new keypair and then demonstrate possession of the private half...?) If the latter, well, how -- what secret does it possess that only the true GitHub-hosted runner of our workflow would have?

It's a GitHub API to retrieve an OIDC token. It's a GitHub feature. (https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect). GitHub attests to the identity of the run / runner. 

>   * "standard blocks like transparency logs (for auditability; the same principle used for TLS...)". Transparency logs are useful for TLS server certificates because domain name owners want to make sure that no CA is issuing certificates claiming that the domain name in question belongs to a cryptographic identity that isn't theirs. I'm not sure I follow the analogy for this use case, though. 

The same way you want to be sure a certificate was not issued for your domain in the TLS PKI, here you want to be sure nobody signed something with your identity. You can monitor the log, e.g. for each release you have, and be sure there's a single signature for a certificate with your identity, and it has signed the binary you have released. In practice you'd sign up for a monitor service (like a CVE alert service), or you'd run the monitor yourself. 

> Who's going to be reading these transparency logs to make sure nobody else is putting out WABT releases that didn't come from "our" GitHub-hosted runners running the workflow from one of "our" Git tags? Are we supposed to be scanning the log every day? How will we recognize an unauthorized entry -- how do we know it wasn't a valid release? Just based on timing and whether we were expecting a release that day...?

This is WIP and monitors are being worked on. It would work by looking at something signed for your repositories, and you'd verify that the hash signed is one of the binaries you released. I think there's no free lunch here: if we want an append-only log of key / identity uses, the log cannot be owned by a maintainer. It does come with some additional complexity, but we're getting there.

> 
> On the "benefit" side, as I understand it, people who want to know the provenance of WABT binaries today have a few options:
> 
> * Build from source in their own trusted environment (optionally: verifying that the binaries match the reproducible WABT build from somebody else, e.g. https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/wabt.html)
> * Get the binaries from somebody like Debian who signs the binary package and effectively attests that it came from running the corresponding source package in a hermetic environment. (It seems to me that this is the low-drama way to achieve the stated goals here... basically just copy what Debian and similar distributors/builders have been doing. But it would more of an active role from GitHub.)
> * Get the binaries from the GitHub release, compute the SHA-256 checksum, and then live-download the ".tar.gz.sha256" artifact from GitHub via HTTPS (doing an online cryptographic handshake with GitHub's public key to make sure you're really talking to GitHub) and compare the checksums.
> 
> As I understand it, this PR adds a fourth option, that is:
> 
> * Get the binaries from the GitHub release and also get the signed provenance, which you can largely verify offline without an online connection (but to truly verify the provenance requires a live connection to the transparency log?).

There is an important difference between "Get the binaries from the GitHub release, compute the SHA-256 checksum, and then live-download the ".tar.gz.sha256" artifact from GitHub via HTTPS" and the solution proposed in this PR. The GitHub release and their checksum do not go thru code reviews, and can be changed / uploaded by any project maintainer. That is, there is no guarantee the binaries are the results of compiling the source code. This is exactly the problem SLSA signatures address: they guarantee the downloaded binary was built from the workflow in this repository, and maintainer cannot change that. It adds this layer of transparency and resilience.

> I see the appeal and I do support reproducible builds and known provenance, but in terms of the risk-benefit and the complexity burden we'd be taking on, I can only speak for myself, but I think my own vote would be for "wait-and-see" rather than merging this now. If this ends up being the direction that most of FLOSS goes in, I assume we'd go with the flow...

keithw:
Thanks, appreciate your reply and clarifications. I'm hardly the boss of this project, but insofar as anything is my call, I think we'll wait-and-see on this and if this becomes the prevailing way to generate provenances, we'll go with the flow. My own preference would be for either (a) GitHub to just attest to these artifacts in basically the same way Debian/Ubuntu/Fedora do, and leave us out of it, or (b) your organization to implement this in a simpler fashion, e.g., a single script that we can drop in to our repository (or ideally a tool accessible via apt, but I guess you want this to be cross-platform). But that's just my 2c; I don't think we expect to blaze new trails on this. Closing for now and will monitor.

PR #1961, opened 2022-08-15
wasm2c: simplify handling of exception tags

keithw:
Previously, the wasm2c generated code used an array to store the tag for each index, and looked up tags by index in the array at runtime. This PR changes the handling of tags to match everything else (one named symbol per tag). This is a stepping-stone on the way to #1814.

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -922,32 +922,37 @@ void CWriter::WriteFuncTypes() {
 }
 
 void CWriter::WriteTags() {
-  for (auto it = module_->tags.cbegin() + module_->num_tag_imports;
-       it != module_->tags.cend(); ++it) {
-    const Tag* tag = *it;
-    Write("static u32 ", DefineGlobalScopeName(tag->name), ";", Newline());
+  Index tag_index = 0;
+  for (const Tag* tag : module_->tags) {
+    bool is_import = tag_index < module_->num_tag_imports;
+    if (!is_import) {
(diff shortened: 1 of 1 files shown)


src/c-writer.cc
```diff
@@ -922,33 +922,22 @@ void CWriter::WriteFuncTypes() {
 }
 
 void CWriter::WriteTags() {
-  if (module_->tags.empty()) {
-    Write("static void init_tags(void) ", OpenBrace(), CloseBrace(), Newline());
-    return;
+  for (auto it = module_->tags.cbegin() + module_->num_tag_imports;
+       it != module_->tags.cend(); ++it) {
```
sbc100, line 926: The pattern used in other functions of this type (e.g. WriteFuncDeclarations) is to use:

```
for (const Func* func : module_->funcs) {                                         
    bool is_import = func_index < module_->num_func_imports;                        
    if (!is_import) {                                                               
       .. do something                                            
    }                                                                               
    ++func_index;                                                                   
}
```   

Even if `func_index` isn't actually used.   Maybe use this same pattern and then we can consider a separate patch to change to this iterator pattern instead everywhere if you think its better.

sbc100 (commented):
lgtm, with couple of comments.

Looks at this change makes me think it would be nice if we had more tests showed the wasm2c output itself .. so the patch could what effect it has on the generated code.

sbc100 (approved):

src/c-writer.cc

PR #1962, opened 2022-08-15
Support multi-memory in all memory ops and in apply/resolve-names

### LINKED ISSUE #1958: Wrong encoding WAT format with the multiple memories proposal
Hi,

I found that when I encode the following WAT into WASM by the `wat2wasm` binary tool in both `1.0.29` or `master`:
(The WAT is in the [test case of the multiple memories proposal](https://github.com/WebAssembly/multi-memory/blob/main/test/core/memory-multi.wast#L47-L84))

```
(module
  (memory 1)
  (memory $m 1)

  (func
    (local $v v128)

    (drop (v128.load8_lane 1 (i32.const 0) (local.get $v)))
    (drop (v128.load8_lane 1 offset=0 1 (i32.const 0) (local.get $v)))
    (drop (v128.load8_lane 1 offset=0 align=1 1 (i32.const 0) (local.get $v)))
    (drop (v128.load8_lane 1 align=1 1 (i32.const 0) (local.get $v)))

    (drop (v128.load8_lane $m 1 (i32.const 0) (local.get $v)))
    (drop (v128.load8_lane $m offset=0 1 (i32.const 0) (local.get $v)))
    (drop (v128.load8_lane $m offset=0 align=1 1 (i32.const 0) (local.get $v)))
    (drop (v128.load8_lane $m align=1 1 (i32.const 0) (local.get $v)))

    (drop (v128.load8_lane 1 1 (i32.const 0) (local.get $v)))
    (drop (v128.load8_lane 1 offset=0 1 (i32.const 0) (local.get $v)))
    (drop (v128.load8_lane 1 offset=0 align=1 1 (i32.const 0) (local.get $v)))
    (drop (v128.load8_lane 1 align=1 1 (i32.const 0) (local.get $v)))

    (v128.store8_lane 1 (i32.const 0) (local.get $v))
    (v128.store8_lane offset=0 1 (i32.const 0) (local.get $v))
    (v128.store8_lane offset=0 align=1 1 (i32.const 0) (local.get $v))
    (v128.store8_lane align=1 1 (i32.const 0) (local.get $v))

    (v128.store8_lane $m 1 (i32.const 0) (local.get $v))
    (v128.store8_lane $m offset=0 1 (i32.const 0) (local.get $v))
    (v128.store8_lane $m offset=0 align=1 1 (i32.const 0) (local.get $v))
    (v128.store8_lane $m align=1 1 (i32.const 0) (local.get $v))

    (v128.store8_lane 1 1 (i32.const 0) (local.get $v))
    (v128.store8_lane 1 offset=0 1 (i32.const 0) (local.get $v))
    (v128.store8_lane 1 offset=0 align=1 1 (i32.const 0) (local.get $v))
    (v128.store8_lane 1 align=1 1 (i32.const 0) (local.get $v))
  )
)
```

Here, I take the first line of function body `(drop (v128.load8_lane 1 (i32.const 0) (local.get $v)))` as `Line 1`, and so on.

The command is:

```bash
$ ./wat2wasm --output=test.wasm --enable-multi-memory test.wat
```

The output `test.wasm` in binary is as following:
(The tar file here: [test.tar.gz](https://github.com/WebAssembly/wabt/files/9332424/test.tar.gz))

```
;; WASM magic and version
00 61 73 6D 01 00 00 00
;; Type section
01 04 01 60 00 00
;; Function section
03 02 01 00
;; Memory section
05 05 02 00 01 00 01
;; Code section
0A EB 01
;; Vector of code segment
01
;; Code segment: size
E8 01
;; Code segment: local (local $v v128)
01 01 7B
;; Expression start
;; Line 1: (drop (v128.load8_lane 1 (i32.const 0) (local.get $v)))
41 00 20 00 FD 54 00 00 01 1A
;; Line 2: (drop (v128.load8_lane 1 offset=0 1 (i32.const 0) (local.get $v)))
41 00 20 00 FD 54 00 00 01 1A
;; Line 3: (drop (v128.load8_lane 1 offset=0 align=1 1 (i32.const 0) (local.get $v)))
41 00 20 00 FD 54 00 00 01 1A
;; Line 4: (drop (v128.load8_lane 1 align=1 1 (i32.const 0) (local.get $v)))
41 00 20 00 FD 54 00 00 01 1A
;; Line 5: (drop (v128.load8_lane $m 1 (i32.const 0) (local.get $v)))
41 00 20 00 FD 54 00 00 01 1A
;; Line 6: (drop (v128.load8_lane $m offset=0 1 (i32.const 0) (local.get $v)))
41 00 20 00 FD 54 00 00 01 1A
;; Line 7: (drop (v128.load8_lane $m offset=0 align=1 1 (i32.const 0) (local.get $v)))
41 00 20 00 FD 54 00 00 01 1A
;; Line 8: (drop (v128.load8_lane $m align=1 1 (i32.const 0) (local.get $v)))
41 00 20 00 FD 54 00 00 01 1A
;; Line 9: (drop (v128.load8_lane 1 1 (i32.const 0) (local.get $v)))
41 00 20 00 FD 54 00 00 01 1A
;; Line 10: (drop (v128.load8_lane 1 offset=0 1 (i32.const 0) (local.get $v)))
41 00 20 00 FD 54 00 00 01 1A
;; Line 11: (drop (v128.load8_lane 1 offset=0 align=1 1 (i32.const 0) (local.get $v)))
41 00 20 00 FD 54 00 00 01 1A
;; Line 12: (drop (v128.load8_lane 1 align=1 1 (i32.const 0) (local.get $v)))
41 00 20 00 FD 54 00 00 01 1A
;; Line 13: (v128.store8_lane 1 (i32.const 0) (local.get $v))
41 00 20 00 FD 58 00 00 01
;; Line 14: (v128.store8_lane offset=0 1 (i32.const 0) (local.get $v))
41 00 20 00 FD 58 00 00 01
;; Line 15: (v128.store8_lane offset=0 align=1 1 (i32.const 0) (local.get $v))
41 00 20 00 FD 58 00 00 01
;; Line 16: (v128.store8_lane align=1 1 (i32.const 0) (local.get $v))
41 00 20 00 FD 58 00 00 01
;; Line 17: (v128.store8_lane $m 1 (i32.const 0) (local.get $v))
41 00 20 00 FD 58 00 00 01
;; Line 18: (v128.store8_lane $m offset=0 1 (i32.const 0) (local.get $v))
41 00 20 00 FD 58 00 00 01
;; Line 19: (v128.store8_lane $m offset=0 align=1 1 (i32.const 0) (local.get $v))
41 00 20 00 FD 58 00 00 01
;; Line 20: (v128.store8_lane $m align=1 1 (i32.const 0) (local.get $v))
41 00 20 00 FD 58 00 00 01
;; Line 21: (v128.store8_lane 1 1 (i32.const 0) (local.get $v))
41 00 20 00 FD 58 00 00 01
;; Line 22: (v128.store8_lane 1 offset=0 1 (i32.const 0) (local.get $v))
41 00 20 00 FD 58 00 00 01
;; Line 23: (v128.store8_lane 1 offset=0 align=1 1 (i32.const 0) (local.get $v))
41 00 20 00 FD 58 00 00 01
;; Line 24: (v128.store8_lane 1 align=1 1 (i32.const 0) (local.get $v))
41 00 20 00 FD 58 00 00 01
;; Expression end
0B
```

As above, the instructions of `Line 1` to `Line 12` are not totally the same, but the encode result binaries are the same.
The same situation occurs at `Line 13` to `Line 24`.

Take the `Line 7: (drop (v128.load8_lane $m offset=0 align=1 1 (i32.const 0) (local.get $v)))` as example.
The `wat2wasm` encoding is:

```
;; local.get $v
41 00
;; i32.const 0
20 00
;; v128.load8_lane $m offset=0 align=1 1
FD 54 00 00 01
;; drop
1A
```

We can concentrate on the `v128.load8_lane` instruction encoding: `FD 54 00 00 01`.

With the [binary format in multiple memories proposal](https://webassembly.github.io/multi-memory/core/binary/instructions.html#memory-instructions), the memory `$m` will be index `1`, so the `memarg` immdiate of the instruction should not be `00 00`.

I think the encoding should be:

```
;; Instruction v128.load8_lane
FD 54
;; memarg: align (bit 6 of the flags field containing alignment is set)
40
;; memarg: offset
00
;; memarg: memory index
01
;; lane index
01
```

And so on with the other instructions which contains the memory index > 0 with the multiple memory proposal.

Can you help to check if this is en error?

Thanks.

keithw:
Previously, wabt could parse multi-memory SIMD loads/stores but not write them. (The testsuite for the multi-memory proposal doesn't currently test writing or execution of multi-memory SIMD ops, so this is untested. However, it does appear to roundtrip its own output on the "memory-multi" test.)

Closes #1958.

### DIFF
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -119,9 +119,18 @@ class BinaryReader {
   Result ReadOffset(Offset* offset, const char* desc) WABT_WARN_UNUSED;
   Result ReadAlignment(Address* align_log2, const char* desc) WABT_WARN_UNUSED;
   Result ReadMemidx(Index* memidx, const char* desc) WABT_WARN_UNUSED;
-  Result ReadAlignMem(Address* align_log2,
-                      Index* memidx,
-                      const char* desc) WABT_WARN_UNUSED;
+  Result ReadMemLocation(Address* alignment_log2,
+                         Index* memidx,
+                         Address* offset,
+                         const char* desc_align,
+                         const char* desc_memidx,
+                         const char* desc_offset,
+                         uint8_t* lane_val = nullptr) WABT_WARN_UNUSED;
+  Result CallbackMemLocation(const Address* alignment_log2,
+                             const Index* memidx,
+                             const Address* offset,
+                             const uint8_t* lane_val = nullptr)
+      WABT_WARN_UNUSED;
   Result ReadCount(Index* index, const char* desc) WABT_WARN_UNUSED;
   Result ReadField(TypeMut* out_value) WABT_WARN_UNUSED;
 
@@ -170,14 +179,6 @@ class BinaryReader {
   Result ReadTagSection(Offset section_size) WABT_WARN_UNUSED;
   Result ReadSections(const ReadSectionsOptions& options) WABT_WARN_UNUSED;
   Result ReportUnexpectedOpcode(Opcode opcode, const char* message = nullptr);
-
-  Result CallbackAlignMemOffset(Index memidx,
-                                Address alignment_log2,
-                                Address offset);
-  Result CallbackAlignMemOffsetLane(Index memidx,
-                                    Address alignment_log2,
-                                    Address offset,
-                                    uint8_t lane_val);
 
   size_t read_end_ = 0;  // Either the section end or data_size.
   BinaryReaderDelegate::State state_;
@@ -441,16 +442,47 @@ Result BinaryReader::ReadMemidx(Index* memidx, const char* desc) {
   return Result::Ok;
 }
 
-Result BinaryReader::ReadAlignMem(Address* alignment_log2,
-                                  Index* memidx,
-                                  const char* desc) {
-  CHECK_RESULT(ReadAlignment(alignment_log2, desc));
+Result BinaryReader::ReadMemLocation(Address* alignment_log2,
+                                     Index* memidx,
+                                     Address* offset,
+                                     const char* desc_align,
+                                     const char* desc_memidx,
+                                     const char* desc_offset,
+                                     uint8_t* lane_val) {
+  CHECK_RESULT(ReadAlignment(alignment_log2, desc_align));
   *memidx = 0;
   if (*alignment_log2 >> 6) {
     ERROR_IF(!options_.features.multi_memory_enabled(),
              "multi_memory not allowed");
     *alignment_log2 = *alignment_log2 & ((1 << 6) - 1);
-    CHECK_RESULT(ReadMemidx(memidx, desc));
+    CHECK_RESULT(ReadMemidx(memidx, desc_memidx));
+  }
+  CHECK_RESULT(ReadAddress(offset, 0, desc_offset));
+
+  if (lane_val) {
+    CHECK_RESULT(ReadU8(lane_val, "Lane idx"));
+  }
+
+  return Result::Ok;
+}
+
+Result BinaryReader::CallbackMemLocation(const Address* alignment_log2,
+                                         const Index* memidx,
+                                         const Address* offset,
+                                         const uint8_t* lane_val) {
+  if (lane_val) {
+    if (*memidx) {
+      CALLBACK(OnOpcodeUint32Uint32Uint32Uint32, *alignment_log2, *memidx,
+               *offset, *lane_val);
+    } else {
+      CALLBACK(OnOpcodeUint32Uint32Uint32, *alignment_log2, *offset, *lane_val);
+    }
+  } else {
+    if (*memidx) {
+      CALLBACK(OnOpcodeUint32Uint32Uint32, *alignment_log2, *memidx, *offset);
+    } else {
+      CALLBACK(OnOpcodeUint32Uint32, *alignment_log2, *offset);
+    }
   }
 
   return Result::Ok;
@@ -938,12 +970,12 @@ Result BinaryReader::ReadInstructions(bool stop_on_end,
       case Opcode::V128Load32X2U: {
         Address alignment_log2;
         Index memidx;
-        CHECK_RESULT(ReadAlignMem(&alignment_log2, &memidx, "load alignment"));
         Address offset;
-        CHECK_RESULT(ReadAddress(&offset, 0, "load offset"));
-
+        CHECK_RESULT(ReadMemLocation(&alignment_log2, &memidx, &offset,
+                                     "load alignment", "load memidx",
+                                     "load offset"));
         CALLBACK(OnLoadExpr, opcode, memidx, alignment_log2, offset);
-        CHECK_RESULT(CallbackAlignMemOffset(memidx, alignment_log2, offset));
+        CHECK_RESULT(CallbackMemLocation(&alignment_log2, &memidx, &offset));
         break;
       }
 
@@ -959,11 +991,12 @@ Result BinaryReader::ReadInstructions(bool stop_on_end,
       case Opcode::V128Store: {
         Address alignment_log2;
         Index memidx;
-        CHECK_RESULT(ReadAlignMem(&alignment_log2, &memidx, "store alignment"));
         Address offset;
-        CHECK_RESULT(ReadAddress(&offset, 0, "store offset"));
+        CHECK_RESULT(ReadMemLocation(&alignment_log2, &memidx, &offset,
+                                     "store alignment", "store memidx",
+                                     "store offset"));
         CALLBACK(OnStoreExpr, opcode, memidx, alignment_log2, offset);
-        CHECK_RESULT(CallbackAlignMemOffset(memidx, alignment_log2, offset));
+        CHECK_RESULT(CallbackMemLocation(&alignment_log2, &memidx, &offset));
         break;
       }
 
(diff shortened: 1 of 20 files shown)


src/binary-writer.cc
```diff
@@ -720,8 +720,15 @@ void BinaryWriter::WriteSimdLoadStoreLaneExpr(const Func* func,
...
-  stream_->WriteU8(log2_u32(align), "alignment");
-  WriteU32Leb128(stream_, typed_expr->offset, desc);
+  Index memidx = module_->GetMemoryIndex(typed_expr->memidx);
+  if (memidx != 0) {
+    stream_->WriteU8(log2_u32(align) | (1 << 6), "alignment");
+    WriteU32Leb128(stream_, memidx, "memidx");
+    WriteU32Leb128(stream_, typed_expr->offset, desc);
+  } else {
+    stream_->WriteU8(log2_u32(align), "alignment");
+    WriteU32Leb128(stream_, typed_expr->offset, desc);
```
sbc100, line 730: Wove this last line outside the condition?

sbc100 (approved):

src/binary-writer.cc
keithw: Done.

keithw:
Further inspection revealed a large number of additional memory ops (atomics, SIMD splat and zero, etc.) that weren't parsing memidx when the memarg flag was set. The validator also had memidx=0 hardcoded for several of these (including memory.fill).

Just pushed a second commit that tries to handle this comprehensively by always parsing alignment and memidx in the same place, and carrying them together through the binary readers. This should get reviewed again before merging because it's a much bigger change now.

src/binary-reader.h
```diff
@@ -59,6 +59,11 @@ struct TypeMut {
 };
 using TypeMutVector = std::vector<TypeMut>;
 
+struct AlignMem {
+  Address alignment_log2;
+  Index memidx;
```
sbc100, line 64: Is there some reason in particular to wrap these two into a struct?   Do we do this for other parts of the binary reader API?   If we are going to do this why not also include the offset?  

Since this is makes the change so large maybe it could be split out into an NFC and landed first before any simd changes/fixes?

test/spec/memory_fill.txt
```diff
@@ -10,7 +10,7 @@ out/test/spec/memory_fill.wast:118: assert_trap passed: out of bounds memory acc
 test() =>
 test() =>
 out/test/spec/memory_fill.wast:175: assert_invalid passed:
-  out/test/spec/memory_fill/memory_fill.8.wasm:000002c: error: memory variable out of range: 0 (max 0)
+  0000000: error: memory variable out of range: 0 (max 0)
```
sbc100, line 13: Why is this zero?

src/binary-reader.h
keithw: Now that I think I've found all the locations where this is needed, I think we could split it back out and have every callsite go back to being `Index memidx, Address alignment_log2, Address offset`. Let me try that...

test/spec/memory_fill.txt
keithw: It's because of the change here: https://github.com/fix-project/wabt/blob/simd-mm-fix/src/shared-validator.cc#L922

Previously the SharedValidator was hardcoding `Var(0, loc)` as the memidx, so it had the current location in the binary module. Now it's just using the memidx it gets from the IR, which came from an Index so didn't have a location attached to it.

sbc100: This seems like a regression, no?   It looks like the `Location` is passed to OnMemoryFill so why not use that?   ...  can't CheckMemoryIndex take advantage of it?  Or alternatively, why doesn't the incoming `Var memidx` have location associated with it?

src/binary-reader.h
keithw: Done. Should make for a somewhat cleaner diff.

src/binary-reader.cc
```diff
@@ -1557,12 +1558,13 @@ Result BinaryReader::ReadInstructions(bool stop_on_end,
...
-        CHECK_RESULT(ReadAlignment(&alignment_log2, "store alignment"));
+        Index memidx;
+        CHECK_RESULT(ReadAlignMem(&alignment_log2, &memidx, "store alignment"));
         Address offset;
         CHECK_RESULT(ReadAddress(&offset, 0, "store offset"));
 
-        CALLBACK(OnAtomicStoreExpr, opcode, alignment_log2, offset);
-        CALLBACK(OnOpcodeUint32Uint32, alignment_log2, offset);
+        CALLBACK(OnAtomicStoreExpr, opcode, memidx, alignment_log2, offset);
+        CHECK_RESULT(CallbackAlignMemOffset(memidx, alignment_log2, offset));
```
sbc100, line 1567: Nice!  This change certainly looks like a step in the right direction.

What do you think about combining `ReadAlignMem` with `ReadAddress` into a single function `ReadMemroyLocation` which would replace `ReadAlignMem` and include the reading of the offset?  Could that function also include `CallbackAlignMemOffset` (removing the need for that new function)?

test/spec/memory_fill.txt
keithw: Basically the BinaryReaderIR and BinaryReaderInterp were converting a lot of Index to Vars (including memidx's) without including the location. Fixed in #1963 which I split out from this one because it includes a lot of changing test output.

src/binary-reader.cc
keithw: Done. I couldn't quite combine it with the CallbackAlignMemOffset because that needs to happen after the callback for the actual expression (otherwise all the opcnts and dumps are off), but I was able to clean it up a little.

sbc100 (approved):
Thanks for working on this!

PR #1963, opened 2022-08-15
Track locations of Vars in BinaryReaderIR and BinaryReaderInterp

keithw:
The BinaryReaderIR and BinaryReaderInterp were converting a lot of Indexes to Vars without tagging them with the source location, as discovered/discussed at https://github.com/WebAssembly/wabt/pull/1962#discussion_r946069037 .

The PR has the reader delegates tag each Index with the current location when converting to Var. Split out from #1962.


sbc100 (approved):
Nice!

(cam we remove the single argument constructor to avoid this happening in the future?)

keithw:
Done, which caught a few more locations.

PR #1967, opened 2022-08-23
Update declaration of wasm_rt_get/set_unwind_target. NFC

sbc100:
This matches how they are declared in the header file.


keithw (approved):
lgtm

PR #1968, opened 2022-08-25
Raise minimum CMake version to 3.16

### LINKED ISSUE #1568: Statically link Visual C++ runtime when releasing on Windows
Currently we need to install Visual C++ Redistributable on Windows in order to run WABT, which is a slight inconvenience.

Could you link it statically?

I believe it is controlled by `-MT` and `-MTd` compiler fags and which are supported in CMake build via [MSVC_RUNTIME_LIBRARY](https://cmake.org/cmake/help/v3.19/prop_tgt/MSVC_RUNTIME_LIBRARY.html) property, but I'm don't have a Windows dev box to try it myself.

alexreinking:
The stated minimum, 3.1, was released on Dec 17, 2014 and did not support a value of 17 for `CMAKE_CXX_STANDARD`. The first version to do so was 3.8. In fact, attempting to build with CMake 3.7 fails with errors like:

```
CMake Error in CMakeLists.txt:
  Target "wasm-strip" requires the language dialect "CXX17" , but CMake does
  not know the compile flags to use to enable it.
```

So we might as well take this chance to raise the minimum to something more recent. I propose 3.16 here because of the following features:

* CMP0077 (v3.13) makes the lives of `FetchContent` users easier by allowing `option()` to be overridden.
* CMP0082 (v3.14) corrects the semantics of `install()` rules, which is relevant for `FetchContent` users.
* CMake 3.12 allows the namelink of a shared library on Linux to be installed separately (i.e. in a dev package)
* CMake 3.14 integrates the standard install destination variables from `GNUInstallDirs` with the `install()` command.
* CMake 3.15 introduced the `CMAKE_MSVC_RUNTIME_LIBRARY` variable for controlling the selection of the runtime library from the outside.
* CMake 3.16 gained support for generator expressions in RPATH properties, which will make it easier to ship shared libraries.

This version is quite conservative and I would even suggest upgrading further. Windows and macOS users enjoy frequent updates from Homebrew and Microsoft Visual Studio. Ubuntu Linux 20.04 LTS ships 3.16 and the newest LTS ships 3.22. Linux users can always install the latest version through PIP, even on ARM, PowerPC, and s390x, and even without sudo.

I have also added a GHA job to the CI workflow to test that WABT builds under its stated minimum version. This should prevent unintentional compatibility issues from creeping in. The other workflows can use newer CMake versions to catch issues there.

Fixes #1568

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -116,3 +116,35 @@ jobs:
     - run: make clang-release-ubsan
     - run: make test-clang-debug-ubsan
     - run: make test-clang-release-ubsan
+
+  build-min-cmake:
+    name: min-cmake
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+      with:
+        submodules: true
+    - uses: actions/setup-python@v1
+      with:
+        python-version: '3.x'
+    - name: Install Ninja
+      run: sudo apt-get install ninja-build
+    - name: Detect minimum CMake version
+      run: >
+        awk 'match($0, /cmake_minimum_required\(VERSION *([0-9]+\.[0-9]+)\)/, a)
+        { print "WABT_CMAKE_VER=" a[1] }' CMakeLists.txt | tee $GITHUB_ENV
+    - name: Install minimum CMake
+      run: |
+        python -m pip install -U setuptools wheel pip
+        python -m pip install "cmake==${WABT_CMAKE_VER}.*"
+        cmake --version
+    - name: Configure WABT
+      run: cmake -G Ninja -S . -B out -DCMAKE_BUILD_TYPE=Release
+    - name: build
+      run: cmake --build out
+    - name: unittests
+      run: cmake --build out --target run-unittests
+    - name: c-api-tests
+      run: cmake --build out --target run-c-api-tests
+    - name: tests
+      run: cmake --build out --target run-tests
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,16 +14,13 @@
 # limitations under the License.
 #
 
-cmake_minimum_required(VERSION 3.1.0)
+cmake_minimum_required(VERSION 3.16)
 project(WABT LANGUAGES C CXX VERSION 1.0.29)
+
 include(GNUInstallDirs)
 
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
-
-if (POLICY CMP0077)
-  cmake_policy(SET CMP0077 NEW)
-endif (POLICY CMP0077)
 
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
@@ -631,10 +628,6 @@ if (BUILD_TESTS)
     LIBS gtest_main gtest ${CMAKE_THREAD_LIBS_INIT}
   )
 
-  if (NOT CMAKE_VERSION VERSION_LESS "3.2")
-    set(USES_TERMINAL USES_TERMINAL)
-  endif ()
-
   # test running
   set(RUN_TESTS_PY ${WABT_SOURCE_DIR}/test/run-tests.py)
 
@@ -642,20 +635,20 @@ if (BUILD_TESTS)
     COMMAND ${PYTHON_EXECUTABLE} ${RUN_TESTS_PY} --bindir $<TARGET_FILE_DIR:wat2wasm>
     DEPENDS ${WABT_EXECUTABLES}
     WORKING_DIRECTORY ${WABT_SOURCE_DIR}
-    ${USES_TERMINAL}
+    USES_TERMINAL
   )
 
   add_custom_target(run-unittests
     COMMAND $<TARGET_FILE:wabt-unittests>
     DEPENDS wabt-unittests
     WORKING_DIRECTORY ${WABT_SOURCE_DIR}
-    ${USES_TERMINAL}
+    USES_TERMINAL
   )
 
   add_custom_target(run-c-api-tests
     COMMAND ${PYTHON_EXECUTABLE} ${WABT_SOURCE_DIR}/test/run-c-api-examples.py --bindir $<TARGET_FILE_DIR:wat2wasm>
     WORKING_DIRECTORY ${WABT_SOURCE_DIR}
-    ${USES_TERMINAL}
+    USES_TERMINAL
   )
 
   add_custom_target(check DEPENDS run-unittests run-tests run-c-api-tests)


alexreinking:
Bumping up to 3.15 would:

1. Provide an immediate solution to https://github.com/WebAssembly/wabt/issues/1568 in the form of [`CMAKE_MSVC_RUNTIME_LIBRARY`](https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html)
2. Let us write installation instructions that use the nicer `cmake --install ...` command than `cmake ... -P build/cmake_install.cmake`.

steven-johnson:
I have to LGTM power here, but I endorse this PR :-)

keithw:
Thanks! If you think 3.15 would be helpful for these various issues, maybe we should just bump to 3.15? Debian packages 3.16 in stretch-backports, and CentOS 8 "AppStream" seems to have 3.20, so given that this is only a build-dep, it doesn't seem that aggressive if it would help.

Would this also help us with #1602?

.github/workflows/build.yml
```diff
@@ -116,3 +116,35 @@ jobs:
...
+      run: >
+        awk 'match($0, /cmake_minimum_required\(VERSION *([0-9]+\.[0-9]+)\)/, a)
+        { print "WABT_CMAKE_VER=" a[1] }' CMakeLists.txt | tee $GITHUB_ENV
+    - name: Install minimum CMake
+      run: |
+        python -m pip install -U setuptools wheel pip
+        python -m pip install "cmake==${WABT_CMAKE_VER}.*"
+        cmake --version
+    - name: Configure WABT
+      run: cmake -G Ninja -S . -B out -DCMAKE_BUILD_TYPE=Release
```
keithw, line 142: How do we know that the cmake that pip installs will end up first in the PATH? Could we add a check on the output of `cmake --version` and have it bomb out if it doesn't match (or, install that version to a particular location and then execute cmake with a fully qualified pathname)?

alexreinking:
@keithw - absolutely it would help. I strongly favor being aggressive with CMake version updates. Halide just upgraded to 3.22. Why not just bump to 3.16 since that's the lower of the two baselines you mentioned?

.github/workflows/build.yml
alexreinking: I'll check the former. I'm not sure the path is stable.

keithw:
3.16 sgtm

alexreinking:
@keithw - done! I think I need you to re-start the workflows.

.github/workflows/build.yml
alexreinking: Done

alexreinking:
@keithw - and we're green! Anything left for me to do here?

.github/workflows/build.yml
```diff
@@ -116,3 +116,36 @@ jobs:
...
+    - name: Detect minimum CMake version
+      run: >
+        awk 'match($0, /cmake_minimum_required\(VERSION *([0-9]+\.[0-9]+)\)/, a)
+        { print "WABT_CMAKE_VER=" a[1] }' CMakeLists.txt | tee $GITHUB_ENV
+    - name: Install minimum CMake
+      run: |
+        python -m pip install -U setuptools wheel pip
+        python -m pip install "cmake==${WABT_CMAKE_VER}.*"
+        cmake --version
+        [[ "$(cmake --version)" =~ ${WABT_CMAKE_VER}. ]]
```
keithw, line 141: I think this test may not quite capture it. E.g. if WABT_CMAKE_VER is "3.2", this will pass if `cmake --version` prints "cmake version 3.22.1". (I think the trailing "." is still part of the extended regex and matches any character.) On the other hand, if `cmake --version` actually does print "cmake version 3.2", this will fail (but should pass).

My shell scripting is not great, but what would you think about something like
```sh
cmake --version | head -n1 | grep -P "^cmake version \Q${REGEX}\E(\.\d+)?$"
```

alexreinking: Yeah, that seems fine too. Shell scripting is so finicky.

alexreinking: Fwiw my understanding was that setup-python put Python's bin directory early in the path. Reading the code suggests that is the case, though documentation is sparse.

alexreinking: https://github.com/actions/toolkit/blob/bc4be505973a6a7344bfd71e1b32f77e1755310c/packages/core/src/core.ts#L124

https://github.com/actions/setup-python/blob/f6795b163d3422199fb74e169e9dc477edf891b1/src/find-python.ts#L125

keithw: Ok, it does seem a bit fragile to rely on an undocumented PATH order, but the current behavior of that code looks good, and this isn't launching missiles so I think we can live with it. Up to you if you want to use the exact test or just remove the test. lgtm either way.

alexreinking: I think I'll cut it. It might be undocumented, but path-prepending is so central to GHA's workflow ecosystem, I'm quite sure Hyrum's Law prevents them from ever changing it.

alexreinking: Done!

keithw (approved):

sbc100:
In the past we have tried to avoid bumping this to a version higher than is needed by the two peer projects binaryen (currently on 3.10.2) and llvm (currently on 3.13.4).   However, less relevant today since we don't build wabt on emscripten-releases waterfall (normally the limiting factor) anymore.

I do like the new test, but i would have rather we did a minimal bump to 3.8 (or 3.10 to match binaryen).

alexreinking:
In our experience on Halide, telling users that `pip install --user cmake` will install the latest version on every Linux imaginable (s390x with musl libc? yep.) without root permissions adequately addresses their concerns. Yet, I doubt users will complain about this change in the first place since even CentOS ships 3.20 now.

LLVM is a massive project and taking advantage of newer CMake features is an endeavor in itself. I'm not sure why binaryen doesn't upgrade more frequently. A comment in their code says they based their version off of the one in Ubuntu bionic. Maybe now that there are two LTS releases ahead of it, they would consider upgrading, too.

And then there are the several clear benefits to upgrading that I outlined above...

keithw:
FWIW, I guess I wouldn't be thrilled for us to bump the required cmake version beyond what can be built for a recent version of Debian or other major Linux distribution of recent vintage. I recognize that Halide is a different story, but I don't think the pip availability of binaries for later cmake versions really helps when we're talking about something we want packaged by the major distributors -- they want to be building packages in their hermetic/reproducible build environments, with build-deps that are available in the build environment.

alexreinking:
> I wouldn't be thrilled for us to bump the required cmake version beyond what can be built for a recent version of Debian or other major Linux distribution of recent vintage.

Well, I don't think that's an issue either since, as you previously observed, all the distros of recent vintage have at _least_ CMake 3.16. The older distros aren't going to upgrade WABT beyond what they already packaged. Debian stable (bullseye) has 3.18.

penzn:
> > I wouldn't be thrilled for us to bump the required cmake version beyond what can be built for a recent version of Debian or other major Linux distribution of recent vintage.
> 
> @keithw - Well, I don't think that's an issue either since, as you previously observed, all the distros of recent vintage have at _least_ CMake 3.16. The older distros aren't going to upgrade WABT beyond what they already packaged. Debian stable (bullseye) has 3.18. We got Halide packaged in Debian when it had a 3.16 requirement.

This PR disabled building on latest LTS Ubuntu (22.04), it only has 3.14 at the moment.

alexreinking:
> This PR disabled building on latest LTS Ubuntu (22.04), it only has 3.14 at the moment.

@penzn - That is incorrect. The latest Ubuntu LTS (jammy) ships with CMake 3.22. Your environment is broken, somehow.

penzn:
Yeah, I found a rogue CMake on my system...

PR #1970, opened 2022-08-25
Fix test dependencies

### LINKED ISSUE #1385: Don't require Python in build (if not found, skip those targets)

alexreinking:
This PR consists of two tightly related changes. It depends on #1968

### 1. Fix usage of Threads
    
`find_package(Threads)` was called redundantly and a bad check was used to disable a pthreads-specific test. Rather than checking `WIN32`, one should check `CMAKE_USE_PTHREADS_INIT`.

### 2. Only search for Python when `BUILD_TESTS=ON`
    
Rather than disable tests with a warning when Python cannot be found, require it when tests are enabled. Avoid the search altogether when tests are disabled.

This check additionally causes issues with FetchContent users that use the newer FindPython(3) module.

Fixes #1385

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -570,8 +570,10 @@ if (BUILD_TOOLS)
   endif ()
 endif ()
 
-find_package(Threads)
 if (BUILD_TESTS)
+  set(THREADS_PREFER_PTHREAD_FLAG ON)
+  find_package(Threads REQUIRED)
+
   # Python 3.5 is the version shipped in Ubuntu Xenial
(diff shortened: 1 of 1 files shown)


keithw (approved):
lgtm

alexreinking:
Thanks @keithw so much for all your time reviewing my PRs today!

PR #1971, opened 2022-08-27
Add GHA workflows for testing s390x and i386

alexreinking:
The failures in these workflows on CI match what I observed locally while testing #1969.

* Opened #1972 for an s390x failure discovered by this testing.
* Opened #1973 for i386 failures discovered by this testing.

Hopefully once the failures are fixed, this will prevent future regressions.


alexreinking:
Opened #1972 for an s390x failure discovered by this testing

PR #1974, opened 2022-08-28
Build fixes for FetchContent users

alexreinking:
This PR makes two important fixes for FetchContent users.

### 1. Add WABT_INSTALL_RULES option

This option is enabled by default when WABT is the top-level project and disabled by default when it is not. It is used (and should continue to be used) to guard uses of the install() command.

This fixes an issue where FetchContent users were forced to install parts of WABT. This most notably included the wasm-rt-impl library and its associated headers.

### 2. Add components to install() rules

By default, CMake assigns every install() rule to an "Unspecified" component. By adding COMPONENT annotations, users become empowered to install only the parts of WABT that they need. WABT now provides three components:

1. `wabt-runtime` - the executables in WABT_EXECUTABLES
2. `wabt-development` - the wasm-rt static library and headers
3. `wabt-documentation` - the man pages, on UNIX only

Users can access this functionality using, e.g.:
```
cmake --install build --component wabt-development
```
**When a user specifies no component, all are installed.**

FetchContent users benefit as well. If they enable WABT's install rules, they can build accurate packaging dependencies around these components.

---

There are also a handful of small drive-by fixes: (1) rename `WABT_MASTER_PROJECT` to [`PROJECT_IS_TOP_LEVEL`](https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html?)for easier CMake upgrades down the line. (2) drop arguments to `install()` rules that match defaults as of CMake 3.14+. (3) delete the top-level `cmake/` folder that doesn't have anything useful in it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,10 +25,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 # Check if wabt is being used directly or via add_subdirectory, FetchContent, etc
-set(WABT_MASTER_PROJECT OFF)
-if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
-  set(WABT_MASTER_PROJECT ON)
-endif()
+string(COMPARE EQUAL "${CMAKE_CURRENT_SOURCE_DIR}" "${CMAKE_SOURCE_DIR}" PROJECT_IS_TOP_LEVEL)
 
 # For git users, attempt to generate a more useful version string
 if (EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/.git)
@@ -43,9 +40,9 @@ if (EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/.git)
               GIT_VERSION
           OUTPUT_STRIP_TRAILING_WHITESPACE)
   if (${GIT_VERSION_RESULT})
-    # Don't issue warning if we aren't the master project;
+    # Don't issue warning if we aren't the top-level project;
     # just assume that whoever included us knows the version they are getting
-    if (WABT_MASTER_PROJECT)
+    if (PROJECT_IS_TOP_LEVEL)
       message(WARNING "${GIT_VERSION_ERROR} Error running git describe to determine version")
     endif()
   else ()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -359,10 +359,14 @@ add_library(wabt STATIC ${WABT_LIBRARY_SRC})
 IF (NOT WIN32)
   add_library(wasm-rt-impl STATIC wasm2c/wasm-rt-impl.c wasm2c/wasm-rt-impl.h)
   if (WABT_INSTALL_RULES)
-    install(TARGETS wasm-rt-impl)
     install(
-      FILES wasm2c/wasm-rt.h wasm2c/wasm-rt-impl.h
+      TARGETS wasm-rt-impl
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -360,8 +358,18 @@ add_library(wabt STATIC ${WABT_LIBRARY_SRC})
...
+  if (WABT_INSTALL_RULES)
+    install(
+      TARGETS wasm-rt-impl
+      COMPONENT WABT_Development
+    )
+    install(
+      DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/wasm2c/"
+      TYPE INCLUDE
+      COMPONENT WABT_Development
+      FILES_MATCHING PATTERN "wasm-rt*.h"
```
sbc100, line 370: Can we keep this as a list of two files for now, just to be explicit, and also the keep this change more focused?

We can change it a wildcard later perhaps if the list of headers ever grows?  (I doubt it will).

CMakeLists.txt
```diff
@@ -360,8 +358,18 @@ add_library(wabt STATIC ${WABT_LIBRARY_SRC})
 
 IF (NOT WIN32)
   add_library(wasm-rt-impl STATIC wasm2c/wasm-rt-impl.c wasm2c/wasm-rt-impl.h)
-  install(TARGETS wasm-rt-impl DESTINATION ${CMAKE_INSTALL_LIBDIR})
-  install(FILES wasm2c/wasm-rt.h wasm2c/wasm-rt-impl.h DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+  if (WABT_INSTALL_RULES)
+    install(
+      TARGETS wasm-rt-impl
+      COMPONENT WABT_Development
+    )
```
sbc100, line 365: Does this not fit on one line?

sbc100: Is there a naming convention that cmake tests to use for components?  My inclination would be use all lowercase here (e.g. wabt_development) because we don't use the uppercase name wabt in source code or filenames.

sbc100: Why don't we need `DESTINATION` here anymore?  Previously we had it set to `CMAKE_INSTALL_LIBDIR`... is that the default?

cmake/README.md
```diff
@@ -1,3 +0,0 @@
-FindRE2C.cmake is copied from the [CMakeXFind
```
sbc100, line 1: Ah, looks like this should have been removed back in #1058

sbc100 (commented):
lgtm % a few comments

CMakeLists.txt
alexreinking: Sure, happy to back that out a bit.

CMakeLists.txt
alexreinking: It does fit on one line. Happy to do that, was just trying to be consistent. There is not an established convention, I was using the `<ProjectName>_Component` pattern that Craig Scott, one of the maintainers, [advocated in his CppCon talk](https://www.youtube.com/watch?v=m0DwB4OvDXk). Happy to make it `wabt-development` if you prefer.

alexreinking: `DESTINATION` is no longer needed because it takes the defaults from `GNUInstallDirs`, `CMAKE_INSTALL_LIBDIR` included.

alexreinking: Done.

CMakeLists.txt
alexreinking: Done.

alexreinking:
@sbc100 - thanks! I just updated my PR to address your comments. PTAL 🙂

sbc100 (approved):

PR #1980, opened 2022-09-02
Make wabt find_package compatible

alexreinking:
Allows users to install wabt to a local directory and use `find_package` to load the `wabt` and `wasm-rt-impl` libraries into their build. These users, as well as FetchContent users, may link to these libraries using, for example:

    find_package(wabt REQUIRED)
    target_link_libraries(myApp PRIVATE wabt::wabt)

Note the `wabt::` namespace. The installed package requires only CMake 3.8 (on account of C++17).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -351,13 +351,41 @@ set(WABT_LIBRARY_SRC
 )
 
 add_library(wabt STATIC ${WABT_LIBRARY_SRC})
+add_library(wabt::wabt ALIAS wabt)
+
+target_compile_features(wabt PUBLIC cxx_std_17)
+
+if (WABT_INSTALL_RULES)
+  install(
+    TARGETS wabt EXPORT wabt-targets
+    COMPONENT wabt-development
+    INCLUDES DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/wabt"
+  )
+  install(
+    DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/src"
+    DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/wabt"
+    COMPONENT wabt-development
+    FILES_MATCHING
+    PATTERN "*.h"
+    PATTERN "*.def"
+  )
+  install(
+    FILES "${CMAKE_CURRENT_BINARY_DIR}/config.h"
+    DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/wabt/"
+    COMPONENT wabt-development
+  )
+endif ()
 
 IF (NOT WIN32)
   add_library(wasm-rt-impl STATIC wasm2c/wasm-rt-impl.c wasm2c/wasm-rt-impl.h)
+  add_library(wabt::wasm-rt-impl ALIAS wasm-rt-impl)
+
   if (WABT_INSTALL_RULES)
     install(
       TARGETS wasm-rt-impl
+      EXPORT wabt-targets
       COMPONENT wabt-development
+      INCLUDES DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
     )
     install(
       FILES "wasm2c/wasm-rt-impl.h" "wasm2c/wasm-rt.h"
@@ -743,5 +771,40 @@ if (EMSCRIPTEN)
     PROPERTIES
     LINK_FLAGS "${LIBWABT_LINK_FLAGS_STR}"
     LINK_DEPENDS "${WABT_POST_JS};${EMSCRIPTEN_EXPORTED_JSON}"
+  )
+endif ()
+
+# Create find_package configuration files
+if (WABT_INSTALL_RULES)
+  include(CMakePackageConfigHelpers)
+
+  set(WABT_INSTALL_CMAKEDIR "${CMAKE_INSTALL_LIBDIR}/cmake/wabt"
+      CACHE STRING "Path to wabt CMake files")
+
+  install(
+    EXPORT wabt-targets
+    DESTINATION "${WABT_INSTALL_CMAKEDIR}"
+    NAMESPACE wabt::
+    FILE wabt-targets.cmake
+    COMPONENT wabt-development
+  )
+
+  configure_package_config_file(
+    scripts/wabt-config.cmake.in
+    wabt-config.cmake
+    INSTALL_DESTINATION "${WABT_INSTALL_CMAKEDIR}"
+    NO_SET_AND_CHECK_MACRO
+  )
+
+  write_basic_package_version_file(
+    wabt-config-version.cmake COMPATIBILITY ExactVersion
+  )
+
+  install(
+    FILES
+    "${CMAKE_CURRENT_BINARY_DIR}/wabt-config.cmake"
+    "${CMAKE_CURRENT_BINARY_DIR}/wabt-config-version.cmake"
+    DESTINATION "${WABT_INSTALL_CMAKEDIR}"
+    COMPONENT wabt-development
   )
 endif ()
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -132,7 +132,7 @@ jobs:
     - name: Detect minimum CMake version
       run: >
         awk 'match($0, /cmake_minimum_required\(VERSION *([0-9]+\.[0-9]+)\)/, a)
-        { print "WABT_CMAKE_VER=" a[1] }' CMakeLists.txt | tee $GITHUB_ENV
+        { print "WABT_CMAKE_VER=" a[1]; exit; }' CMakeLists.txt | tee $GITHUB_ENV
     - name: Install minimum CMake
       run: |
         python -m pip install -U setuptools wheel pip
--- a/scripts/wabt-config.cmake.in
+++ b/scripts/wabt-config.cmake.in
@@ -1,0 +1,5 @@
+cmake_minimum_required(VERSION 3.8)
+@PACKAGE_INIT@
+
+include("${CMAKE_CURRENT_LIST_DIR}/wabt-targets.cmake")
+check_required_components(wabt)


alexreinking:
For the sake of review, here is the output of `tree _local` where `_local` is the install prefix I chose for this demonstration:

```
$ cmake -G Ninja -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo
$ cmake --build build
$ cmake --install build --prefix _local
$ tree _local
_local/
├── bin
│   ├── spectest-interp
│   ├── wasm2c
│   ├── wasm2wat
│   ├── wasm-decompile
│   ├── wasm-interp
│   ├── wasm-objdump
│   ├── wasm-opcodecnt
│   ├── wasm-strip
│   ├── wasm-validate
│   ├── wast2json
│   ├── wat2wasm
│   └── wat-desugar
├── include
│   ├── wabt
│   │   ├── config.h
│   │   └── src
│   │       ├── apply-names.h
│   │       ├── base-types.h
│   │       ├── binary.h
│   │       ├── binary-reader.h
│   │       ├── binary-reader-ir.h
│   │       ├── binary-reader-logging.h
│   │       ├── binary-reader-nop.h
│   │       ├── binary-reader-objdump.h
│   │       ├── binary-reader-opcnt.h
│   │       ├── binary-writer.h
│   │       ├── binary-writer-spec.h
│   │       ├── binding-hash.h
│   │       ├── cast.h
│   │       ├── circular-array.h
│   │       ├── color.h
│   │       ├── common.h
│   │       ├── c-writer.h
│   │       ├── decompiler-ast.h
│   │       ├── decompiler.h
│   │       ├── decompiler-ls.h
│   │       ├── decompiler-naming.h
│   │       ├── error-formatter.h
│   │       ├── error.h
│   │       ├── expr-visitor.h
│   │       ├── feature.def
│   │       ├── feature.h
│   │       ├── filenames.h
│   │       ├── generate-names.h
│   │       ├── interp
│   │       │   ├── binary-reader-interp.h
│   │       │   ├── interp.h
│   │       │   ├── interp-inl.h
│   │       │   ├── interp-math.h
│   │       │   ├── interp-util.h
│   │       │   ├── interp-wasi.h
│   │       │   ├── istream.h
│   │       │   └── wasi_api.def
│   │       ├── intrusive-list.h
│   │       ├── ir.h
│   │       ├── ir-util.h
│   │       ├── leb128.h
│   │       ├── lexer-source.h
│   │       ├── lexer-source-line-finder.h
│   │       ├── literal.h
│   │       ├── make-unique.h
│   │       ├── opcode-code-table.h
│   │       ├── opcode.def
│   │       ├── opcode.h
│   │       ├── option-parser.h
│   │       ├── prebuilt
│   │       ├── range.h
│   │       ├── resolve-names.h
│   │       ├── result.h
│   │       ├── shared-validator.h
│   │       ├── stream.h
│   │       ├── string-format.h
│   │       ├── string-util.h
│   │       ├── template
│   │       │   ├── wasm2c.bottom.h
│   │       │   └── wasm2c.top.h
│   │       ├── token.def
│   │       ├── token.h
│   │       ├── tools
│   │       ├── tracing.h
│   │       ├── type-checker.h
│   │       ├── type.h
│   │       ├── utf8.h
│   │       ├── validator.h
│   │       ├── wast-lexer.h
│   │       ├── wast-parser.h
│   │       └── wat-writer.h
│   ├── wasm-rt.h
│   └── wasm-rt-impl.h
├── lib
│   ├── cmake
│   │   └── wabt
│   │       ├── wabt-config.cmake
│   │       ├── wabt-config-version.cmake
│   │       ├── wabt-targets.cmake
│   │       └── wabt-targets-relwithdebinfo.cmake
│   ├── libwabt.a
│   └── libwasm-rt-impl.a
└── share
    └── man
        └── man1
            ├── spectest-interp.1
            ├── wasm2c.1
            ├── wasm2wat.1
            ├── wasm-decompile.1
            ├── wasm-interp.1
            ├── wasm-objdump.1
            ├── wasm-opcodecnt.1
            ├── wasm-strip.1
            ├── wasm-validate.1
            ├── wast2json.1
            ├── wat2wasm.1
            └── wat-desugar.1

14 directories, 100 files
```

alexreinking:
There's a naming issue with using the wabt library outside the project tree: all of the headers are prefixed with `src/` and have very common names like `ir.h` or `config.h`.

It would be much better if the headers lived inside a `wabt/` folder and were included as, e.g. `wabt/ir.h`.

sbc100:
> There's a naming issue with using the wabt library outside the project tree: all of the headers are prefixed with `src/` and have very common names like `ir.h` or `config.h`. It would be much better if the headers lived inside a `wabt/` folder and were included as, e.g. `wabt/ir.h`.
> 
> Such a change is clearly outside the scope of this PR, but I wonder if it would be acceptable in the first place?

I guess it depends if there are users of libwabt out there who what to be able to use the library in this way.  Almost all wabt users that I know of are users of the tools (wasm-rt-impl being some what separate). 

We don't really make any attempt to define a public library API, but I assume if we did it would not include all of the headers under src/?  Or would it? 

Do you have a use case in mind?   Is that use case currently blocked on this change?

alexreinking:
> I guess it depends if there are users of libwabt out there who what to be able to use the library in this way. Almost all wabt users that I know of are users of the tools (wasm-rt-impl being some what separate). [...] Do you have a use case in mind? Is that use case currently blocked on this change?

Halide is such a user. We use WABT (or, experimentally, V8) to test our WebAssembly backend in-process, which we found is much more efficient than going through the command line interface. We hack around the includes problem by adding a directory layer, i.e. our includes look like so:

```
#include "wabt-src/src/binary-reader.h"
#include "wabt-src/src/cast.h"
#include "wabt-src/src/common.h"
#include "wabt-src/src/error-formatter.h"
#include "wabt-src/src/error.h"
#include "wabt-src/src/feature.h"
#include "wabt-src/src/interp/binary-reader-interp.h"
#include "wabt-src/src/interp/interp-util.h"
#include "wabt-src/src/interp/interp.h"
#include "wabt-src/src/interp/istream.h"
#include "wabt-src/src/result.h"
#include "wabt-src/src/stream.h"
```

We aren't _blocked_, per se, but maintaining custom build rules to use libwabt in our build and to rewrite header paths has long been a maintenance burden.

> We don't really make any attempt to define a public library API, but I assume if we did it would not include all of the headers under src/? Or would it?

It's okay (for us) if you don't. We are comfortable requiring an exact version and upgrading slowly. So no need to define a public API for our sake. The headers I listed above are the ones we need, plus the set of transitive includes.

alexreinking:
By inspecting the output of `g++ -H`, these are the headers we need:

```
build/config.h
src/base-types.h
src/binary.h
src/binary-reader.h
src/cast.h
src/color.h
src/common.h
src/error-formatter.h
src/error.h
src/feature.def
src/feature.h
src/interp/binary-reader-interp.h
src/interp/interp.h
src/interp/interp-inl.h
src/interp/interp-util.h
src/interp/istream.h
src/leb128.h
src/lexer-source.h
src/lexer-source-line-finder.h
src/make-unique.h
src/opcode-code-table.h
src/opcode.def
src/opcode.h
src/range.h
src/result.h
src/stream.h
src/string-format.h
src/type.h
```

CMakeLists.txt
```diff
@@ -745,3 +773,46 @@ if (EMSCRIPTEN)
...
+    FILE wabt-targets.cmake
+    COMPONENT wabt-development
+  )
+
+  file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/wabt-config.cmake.in" [[
+cmake_minimum_required(VERSION 3.8)
+@PACKAGE_INIT@
+
+include("${CMAKE_CURRENT_LIST_DIR}/wabt-targets.cmake")
+check_required_components(wabt)
```
sbc100, line 797: Why inline this file here rather than checking it in?

CMakeLists.txt
```diff
@@ -745,3 +773,46 @@ if (EMSCRIPTEN)
...
+    NAMESPACE wabt::
+    FILE wabt-targets.cmake
+    COMPONENT wabt-development
+  )
+
+  file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/wabt-config.cmake.in" [[
+cmake_minimum_required(VERSION 3.8)
+@PACKAGE_INIT@
+
+include("${CMAKE_CURRENT_LIST_DIR}/wabt-targets.cmake")
```
sbc100, line 796: Hhere does wabt-targets.cmake come file?

CMakeLists.txt
```diff
@@ -351,13 +351,41 @@ set(WABT_LIBRARY_SRC
 )
 
 add_library(wabt STATIC ${WABT_LIBRARY_SRC})
+add_library(wabt::wabt ALIAS wabt)
+
+target_compile_features(wabt PUBLIC cxx_std_17)
```
sbc100, line 356: How is this different to the existing `set(CMAKE_CXX_STANDARD 17)`?

CMakeLists.txt
```diff
@@ -745,3 +773,46 @@ if (EMSCRIPTEN)
...
+if (WABT_INSTALL_RULES)
+  include(CMakePackageConfigHelpers)
+
+  set(WABT_INSTALL_CMAKEDIR "${CMAKE_INSTALL_LIBDIR}/cmake/wabt"
+      CACHE STRING "Path to wabt CMake files")
+
+  install(
+    EXPORT wabt-targets
+    DESTINATION "${WABT_INSTALL_CMAKEDIR}"
+    NAMESPACE wabt::
```
sbc100, line 787: Is this reason we require the aliases above?   Are the aliases actually required?

CMakeLists.txt
alexreinking: It was expedient, but I'm happy to check it in under `scripts/` or to resurrect the `cmake/` folder.

CMakeLists.txt
alexreinking: That comes from here: 

https://github.com/WebAssembly/wabt/blob/52ec70d09a6ef86eda4995168a5572092c6ba706/CMakeLists.txt#L784-L790

CMakeLists.txt
alexreinking: It's a requirement that propagates to consumers... CMake will pick the highest standard version among the transitively linked set of targets and issue an error if the active toolchain is incapable of compiling it.

CMakeLists.txt
alexreinking: Linking in CMake is... fraught. 

Normally, when you link to a library `A` via `target_link_libraries(tgt PRIVATE A)`, CMake first checks to see if `A` is the name of a CMake target. If so, it can be smarter about how it links to `tgt`. However, if it can't find the target `A` then it shrugs its shoulders and says "eh, this is a link flag". In the best case, this defers what should be a configure-time error to a build-time linker failure. In the worst case, this _succeeds_ and links to an unexpected copy of `A` laying around in the user's system. Nasal demons, ahoy!

To fix this, CMake adopted the convention that namespaced (with `::`) targets are _always_ considered CMake targets and will never be silently treated as linker flags. If the target doesn't exist, you get an error at configure time. Unfortunately, there's a caveat: only `ALIAS` and `IMPORTED` targets can have a `:` in the name. Why? Because the `:` would conflict with drive letter syntax on Windows. So:

1. We _have_ to declare the build target `wabt` without a namespace.
2. But linking to non-namespaced targets invites bad behavior.
3. So we export it with a namespace as `wabt::wabt`. Good.
4. But now the `FetchContent` users have to link to `wabt` and if they want to switch to `find_package`, they have to change `wabt` to `wabt::wabt` elsewhere in their code. Bad.
5. So we create an alias in the build to match the exported target.

Now, both `FetchContent` and `find_package` users can be confident linking to `wabt::wabt`.

CMakeLists.txt
alexreinking: I moved this to `scripts/`.

sbc100 (approved):

PR #1994, opened 2022-09-11
Implement Relaxed SIMD proposal

marcusb:
This adds support for the new opcodes from the Relaxed SIMD proposal (https://github.com/WebAssembly/relaxed-simd) behind the "--enable-relaxed-simd" flag.


keithw:
Thank you for doing this! I'm probably not the most qualified person to review this, but, as a starter, I don't think we want to vendor the relaxed-simd test scripts into wabt unless we really have to. My suggestion would be to submit a PR to add relaxed-simd to the testsuite repo list (https://github.com/WebAssembly/testsuite/blob/main/update-testsuite.sh#L9) and then add them there, and then coming back here with a PR that updates the testsuite submodule, and then we go from there.

marcusb:
Ok, added to testsuite in https://github.com/WebAssembly/testsuite/pull/62

keithw:
Those test changes don't look like a problem, but we should probably update the testsuite in a separate PR... For the non-wasm2c tests, it looks like they just need to be rebased. For the wasm2c tests, either this will need to happen after #1877/#1887 (in which case updating the test suite is just a rebase too) or more tests will need to be moved to old-spec.

marcusb:
The Relaxed SIMD proposal is still in flux so there will be some changes down the line:
* The FMS instructions have new names and operand order, but the spec tests haven't been adjusted to match so I'm holding off with that change.
* There is a new instruction `f32x4.relaxed_dot_bf16x8_add_f32x4`, not implemented in this PR.

src/interp/interp.cc
```diff
@@ -1691,36 +1698,66 @@ RunResult Thread::StepInternal(Trap::Ptr* out_trap) {
     case O::F32X4Add:          return DoSimdBinop(Add<f32>);
     case O::F32X4Sub:          return DoSimdBinop(Sub<f32>);
     case O::F32X4Mul:          return DoSimdBinop(Mul<f32>);
-    case O::F32X4Div:          return DoSimdBinop(FloatDiv<f32>);
-    case O::F32X4Min:          return DoSimdBinop(FloatMin<f32>);
-    case O::F32X4Max:          return DoSimdBinop(FloatMax<f32>);
+    case O::F32X4Div:
+      return DoSimdBinop(FloatDiv<f32>);
```
sbc100, line 1702: Why not keep this on a single line like the others?  It doesn't looks like we are near the line limit.

src/opcode-code-table.h
```diff
@@ -24,7 +24,12 @@
 extern "C" {
 #endif
 
-#define WABT_OPCODE_CODE_TABLE_SIZE 65536
+#define WABT_OPCODE_CODE_TABLE_SIZE 131072
+
+/*
+ * Number of bits required to store an opcode
+ */
+#define MAX_OPCODE_BITS 9
```
sbc100, line 32: Is this an intentional part of the relaxed simd proposal, that the opcode space is now larger and a byte?

src/interp/interp.cc
marcusb: This was done by clang-format and enforced by the linting, if I'm not mistaken.

sbc100: This seems unfortunate.. I guess we should exclude this whole block from clang-formatting.

src/opcode-code-table.h
marcusb: The vector instructions were always using variable-length integer opcodes, but these are the only instructions in this implementation with opcodes above 0xff. This necessitates a change in the structure of the internal lookup table (and also unfortunately doubles its size to 128 kB).

sbc100: Just to be clear, adding relaxed SIMD introduces opcode whose encoding is > 0xff for the first time?

marcusb: Yes, to be precise we're [adding](https://github.com/WebAssembly/wabt/pull/1994/files#diff-61d33dba45369870124687df0c499690e95ce7336a06f3e2df45226e25034d2fR505) prefixed opcodes with suffixes over 0xff.

keithw:
@marcusb Do you have cycles to respond to the review (which might just be about formatting at this point) and rebase? This seems close and would be great to have it in...

(I should point out that nominally the relaxed-simd spec is supposed to enforce that an implementation always chooses the same option for an "either" [it's not an independent free choice each time], but they don't seem to require this in the reference implementation so I think we can defer worrying about it until they decide how/if to test that.)

marcusb:
@keithw I'll bring the PR up to date with the latest changes in the spec. Here is the [testsuite bump PR](https://github.com/WebAssembly/testsuite/pull/63) for a start.

marcusb:
@keithw Seems there are a few unrelated test failures after pulling in the latest testsuite, eg some elem-type mismatches.

sbc100:
> @keithw Seems there are a few unrelated test failures after pulling in the latest testsuite, eg some elem-type mismatches.

I've been working on a PR to update the testsuite .. along with the corresponding fixes.  Uploading now.

sbc100:
https://github.com/WebAssembly/wabt/pull/2054

marcusb:
Rebased and brought up to date with spec.

sbc100 (commented):
Looks good to me, but maybe lets hear from @ngzhian first

ngzhian (approved):
Neat!

keithw:
Ok to merge this now?

marcusb:
Ok to merge as far as I'm concerned.

keithw:
Ok, merging. Hope there's no objection, but I'd like to copy the note from the commit message (about the missing f32x4.relaxed_dot_bf16x8_add_f32x4 instruction) to a footnote in the README so we don't forget about it and nobody gets surprised and files more issues if they try to use that instruction and get an error.

keithw:
Sigh, I thought I was going to be able to merge this but I guess we have to wait until after #2082.

PR #2003, opened 2022-09-22
Update spec tests

marcusb:


marcusb:
@keithw

marcusb:
Not sure if I should also run `./test/update-spec-tests.py` to add the new test cases, but those do not generally pass.

sbc100 (approved):
lgtm, I think we can worry about any news tests in separate PRs.

sbc100:
(Will wait for @keithw approval before landing)

keithw:
lgtm. An update that runs update-spec-tests.py is at #2001; this brings in the new tokens.wast test, which requires a change to the Wast lexer to get everything passing.

keithw (approved):

PR #2004, opened 2022-09-22
Updated man pages

tDwtp:
update the man pages based on the output of `--help` of the individual commands

### DIFF
--- a/help2man.lua
+++ b/help2man.lua
@@ -1,0 +1,154 @@
+
+
+local format_br = '\n.br\n'
+local tools = {
+	"wasm-decompile",	"wasm-interp",	"wasm-objdump",	"wasm-opcodecnt",
+	"wasm-strip",	"wasm-validate",	"wasm2c",	"wasm2wat",
+	"wast2json",	"wat-desugar",	"wat2wasm",	"spectest-interp"
+}
+
+local name, mainarg, argvs
+local usage
+local short = ""
+local full = {}
+local examples = {}
+local options = {}
+
+local last = { }
+
+do -- parse usage
+	local usage = io.read('*line');
+	name, mainarg, argvs = usage:match("usage: ([^%s%c]+) %[options%] ([^%s%c]+) %[([^%s%c]+)%]%.%.%.")
+	if not argvs then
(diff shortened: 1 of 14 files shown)


sbc100 (approved):
Is there some way to automate this?

tDwtp:
You can definetly parse the output of the `--help` for this.
Actually, I wrote a little script to help me, but its in Lua. (I was lazy and did it in like 10 minutes flat with `CTRL+C`).

Also: I want to expand the `OptionParser parser("wabt-tool", s_description);` stuff, to include proper description so it is distinguished from the short description. Would that be ok?

##### Format:
```
usage: <usage>
<short-description> one liner; no indentation
\n
  full description
  indented by 2 spaces!
\n
examples:
  # sample-description
  # may span multiple lines
  $ sample input
\n
options:
  -x, --xyz     two (2) spaces indention for "multi-option"
      --abx     six (6) spaces indention for "single-option"
```

##### The Lua-script:
```lua
local line = ""
repeat
	line = io.read('*line');
until line == "options:"

local FULL = "^      %-%-"
local SHORT = "^  %-(.), %-%-"
local R_FULL = ".It Fl Fl "
local R_SHORT = ".It Fl %1 , Fl Fl "

for line in io.stdin:lines() do
	if line:match("^ +%-") then
		local parsed = line:match(FULL) and
			line:gsub(FULL, R_FULL) or
			line:gsub(SHORT, R_SHORT)
		print((parsed:gsub("   +", '\n')))
	else
		print(line)
	end
end
```

P.S.: I would like to add the option `-m, --module=[NAME:]PATH   use PATH as a wasm module and provide all exports under the name NAME (filename if not present) for imports`
Later maybe dlopen/GetModuleHandle could be supported? you could simply write your own wasi and wasi will just export wasi.[dll|so] and start the function, done. easily replacable wasi runtime. good for experimentation.

help2man.lua
```diff
@@ -0,0 +1,154 @@
+
```
sbc100, line 1: Wow.. I guess there is no reason not to include a lua script here.. most folks won't ever need to run it.

Can you put this in the scripts/ directory?   And maybe add a `#!` to the top?

tDwtp: aw yeah... i should really use my linux laptop more ...
... done

Issue #2005, opened 2022-09-22
Load other wasm (and possibly native) modules for exposing exports/imports


tDwtp:
I would like to make imports simpler for wasm testing with the `wasm-interp` tool.
Nothing fancy just something like  
`-m, --module=[NAME:]PATH`  
`use PATH as a wasm module and provide all exports under the name NAME (filename if not present) for imports`

This way you could load any wasm-modules or possibly native libraries and expose all names to others. testing wasm modules would be a lot simpler.

Adding the argument to the parser is fine, literally a oneliner.  
However, I am confused on how to link exports from one module to the imports of others. I checked `BindImports` from `src/tools/wasm-interp.cc` and `WasiBindImports` from `src/interp/interp-wasm.cc` and I think I get how to add them to the imports-list, but how do I *get* the function for pushing onto the imports?

keithw:
> However, I am confused on how to link exports from one module to the imports of others.

Here's how the spectest-interp runner does it: https://github.com/WebAssembly/wabt/blob/main/src/tools/spectest-interp.cc#L1455-L1473

sbc100:
I can see an argument for wasm-interp itself being able to run more than one module at a time though?    Should we leave this issue open in case somebody wants to add this feature?

tDwtp:
I am working on it... I just had to work on weekend and have free today.

Also: As there can (potentially) be multiple memory entries. I would like to add the behaviour, that if a memory is exported using the name of the module it is importing, that is linked to the "memory" export of the imported module. or shoudl I just join by name? (I am still figuring out things..)

sbc100:
Linking imports and exports should always be by name, and should work the same way for all types of exports.

keithw:
Happy to reopen. :-)

tDwtp:
Is there a chat, IRC or Discord or whatever... I am stuck.

tDwtp:
To be more specific with my problem:
It compiles fine. It even sets the exports, but they dont seem to be loaded properly.
I basically do the same as described by these 3 functions:
https://github.com/WebAssembly/wabt/blob/main/src/tools/spectest-interp.cc#L1455-L1483
I just added the host.print exception.

Here is the loop over the module-arguments:
```cpp

  std::vector<Module::Ptr> modules_loaded(s_modules.size());
  ExportMap load_map;
  for (auto& import_module : s_modules) {
    std::string module_path = GetPathName(import_module);

    Module::Ptr load_module;
[...]

PR #2009, opened 2022-09-28
Module loading is now implemented

tDwtp:
#2005 feature is finally implemented.
Modules can be loaded via `-m --module=NAME:PATH`
Maybe I will expand it in the future to accept dll/so/dylib/bundle as well. I am open for suggestions.
`-N --native=mod:lib.so` would be nice, or make `--module` more dynamic? \*
'@' will override the module name
'$' is reserved to prefer prefer the debug name if provided in the name section.
providing no NAME use the file-name. again reserved for the future is to use the debug name given in the name section

\* note that native libraries either need some witx glue, or ffi or an API to bind against.

### DIFF
--- a/src/tools/wasm-interp.cc
+++ b/src/tools/wasm-interp.cc
@@ -119,7 +119,9 @@ static void ParseOptions(int argc, char** argv) {
       [](const std::string& argument) { s_wasi_dirs.push_back(argument); });
   parser.AddOption(
       'm', "module", "[(@|$)NAME]:PATH",
-      "load wasm-module PATH and provide all exports via module-name NAME for imports of following modules",
+      "load wasm-module PATH and provide all exports via module-name NAME for imports of following modules. "
+      "@ will always use NAME"/*", $ will prefer a given debug-name in the namesection"*/ ". "
+      "if NAME isn't provided, the file-name without the extension will be used.",
       [](const std::string& argument) { s_modules.push_back(argument); });
   parser.AddOption(
       "run-all-exports",
@@ -222,49 +224,6 @@ static void PopulateExports(const Instance::Ptr& instance,
   }
 }
 
-/*
-void CommandRunner::PopulateExports(const Instance::Ptr& instance,
-                                    ExportMap* map) {
-  map->clear();
-  interp::Module::Ptr module{store_, instance->module()};
-  for (size_t i = 0; i < module->export_types().size(); ++i) {
-    const ExportType& export_type = module->export_types()[i];
-    (*map)[export_type.name] = store_.UnsafeGet<Extern>(instance->exports()[i]);
-  }
-}
-
-wabt::Result CommandRunner::OnModuleCommand(const ModuleCommand* command) {
-  Errors errors;
-  auto module = ReadModule(command->filename, &errors);
-  FormatErrorsToFile(errors, Location::Type::Binary);
-
-  if (!module) {
-    PrintError(command->line, "error reading module: \"%s\"",
-               command->filename.c_str());
-    return wabt::Result::Error;
-  }
-
-  RefVec imports;
-  PopulateImports(module, &imports);
-
-  Trap::Ptr trap;
-  auto instance = Instance::Instantiate(store_, module.ref(), imports, &trap);
-  if (trap) {
-    assert(!instance);
-    PrintError(command->line, "error instantiating module: \"%s\"",
-               trap->message().c_str());
-    return wabt::Result::Error;
-  }
-
-  PopulateExports(instance, &last_instance_);
-  if (!command->name.empty()) {
-    instances_[command->name] = last_instance_;
-  }
-
-  return wabt::Result::Ok;
-}
-//*/
-
 static std::string GetPathName(std::string module_arg) {
   size_t path_at = module_arg.find_first_of(':');
 
@@ -276,9 +235,12 @@ static std::string GetPathName(std::string module_arg) {
 static std::string GetRegistryName(std::string module_arg,
                                    const Module::Ptr& module) {
   size_t split_pos = module_arg.find_first_of(':');
-  std::string override_name = module_arg.substr(0, split_pos);
-  std::string path_name = module_arg.substr(split_pos + 1);
-  std::string debug_name = ""; //GetDebugName(module);
+  if (split_pos == std::string::npos) {
+    split_pos = 0;
+  }
+  std::string override_name = module_arg.substr(0, split_pos ? split_pos : 0);
+  std::string path_name = module_arg.substr(split_pos);
+  std::string debug_name = ""; // GetDebugName(module);
 
   // check if override_name starts with @ and return override_name
   if (override_name[0] == '@') {
@@ -481,11 +443,6 @@ static Result ReadAndRunModule(const char* module_filename) {
   }
 #endif
 
-  for (auto &pair : s_registry) {
-    pair.second.clear();
-  }
-  s_registry.clear();
-
   return Result::Ok;
 }
 
@@ -498,6 +455,12 @@ int ProgramMain(int argc, char** argv) {
   s_store.setFeatures(s_features);
 
   wabt::Result result = ReadAndRunModule(s_infile);
+
+  for (auto &pair : s_registry) {
+    pair.second.clear();
+  }
+  s_registry.clear();
+
   return result != wabt::Result::Ok;
 }
 
--- a/src/tools/wasm-interp.cc
+++ b/src/tools/wasm-interp.cc
@@ -210,18 +210,6 @@ static Extern::Ptr GetImport(const std::string& module,
   return {};
 }
 
-/*static void PopulateImports(const Module::Ptr& module,
-                            RefVec* imports) {
-  for (auto&& import : module->desc().imports) {
-    if (IsHostPrint(import)) {
-      imports->push_back(GenerateHostPrint(import));
-      continue;
-    }
-    auto extern_ = GetImport(import.type.module, import.type.name);
-    imports->push_back(extern_ ? extern_.ref() : Ref::Null);
-  }
-}//*/
-
 static void PopulateExports(const Instance::Ptr& instance,
                             Module::Ptr &module,
                             ExportMap &map) {
@@ -386,9 +374,11 @@ static Result ReadAndRunModule(const char* module_filename) {
 
   RefVec imports;
 
+  // if we only have dummy imports, we wont need to reghister anything
+  // but we still want to load them.
   std::vector<Module::Ptr> modules_loaded(s_modules.size());
-  ExportMap load_map;
   for (auto import_module : s_modules) {
+    ExportMap load_map;
     std::string module_path = GetPathName(import_module);
 
     Module::Ptr load_module;
@@ -399,26 +389,18 @@ static Result ReadAndRunModule(const char* module_filename) {
     }
 
     RefVec load_imports;
-    // PopulateImports(module, &imports);
     BindImports(load_module, load_imports);
 
-    /* // this is how wasm-interp.exe does it
+    // this is how wasm-interp.exe does it
     Instance::Ptr load_instance;
     CHECK_RESULT(InstantiateModule(load_imports, load_module, &load_instance));
-    /*/ // this is how spectest-interp.exe does it
-    Trap::Ptr trap;
-    auto load_instance = Instance::Instantiate(s_store, load_module.ref(), load_imports, &trap);
-    if (trap) {
-      assert(!load_instance);
-      s_stderr_stream.get()->Writef("error instantiating module: \"%s\"",
-                                    trap->message().c_str());
-      return wabt::Result::Error;
-    }
-    //*/
 
-    std::string reg_name = GetRegistryName(import_module, load_module);
-    PopulateExports(load_instance, load_module, load_map);
-    s_registry[reg_name] = std::move(load_map);
+    // we wont need to register anything, if we only have dummy imports anyway
+    if (!s_dummy_import_func) {
+      std::string reg_name = GetRegistryName(import_module, load_module);
+      PopulateExports(load_instance, load_module, load_map);
+      s_registry[reg_name] = std::move(load_map);
+    }
 
     modules_loaded.push_back(load_module);
   }
--- a/src/tools/wasm-interp.cc
+++ b/src/tools/wasm-interp.cc
@@ -465,7 +465,6 @@ static Result ReadAndRunModule(const char* module_filename) {
     return Result::Error;
 #endif
   } else {
-    // PopulateImports(module, &imports);
     BindImports(module, imports);
   }
 
--- a/src/tools/wasm-interp.cc
+++ b/src/tools/wasm-interp.cc
@@ -21,6 +21,7 @@
 #include <memory>
 #include <string>
 #include <vector>
+#include <map>
 
 #include "src/binary-reader.h"
 #include "src/error-formatter.h"
@@ -39,6 +40,9 @@
 using namespace wabt;
 using namespace wabt::interp;
 
+using ExportMap = std::map<std::string, Extern::Ptr>;
+using Registry = std::map<std::string, ExportMap>;
+
 static int s_verbose;
 static const char* s_infile;
 static Thread::Options s_thread_options;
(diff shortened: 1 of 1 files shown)


tDwtp:
How can I write a test, to check for it to work porperly.
I found how to run `app.wasm`, but how can I provide a second file?

src/tools/wasm-interp.cc
```diff
@@ -110,6 +117,13 @@ static void ParseOptions(int argc, char** argv) {
   parser.AddOption(
       'd', "dir", "DIR", "Pass the given directory the the WASI runtime",
       [](const std::string& argument) { s_wasi_dirs.push_back(argument); });
+  parser.AddOption(
+      'm', "module", "[(@|$)NAME:]FILE",
```
sbc100, line 121: I wonder if we can come up with better name for this parameter?     How about something like `--preload`?

Also perhaps we could just get away without adding a new option here at all and just treat these module like main binary we are running. i.e could we just allow wasm-interp to take any number of binaries as arguments rather than the current limit of just one?

src/tools/wasm-interp.cc
```diff
@@ -39,6 +40,9 @@
 using namespace wabt;
 using namespace wabt::interp;
 
+using ExportMap = std::map<std::string, Extern::Ptr>;
+using Registry = std::map<std::string, ExportMap>;
```
sbc100, line 44: Can/should we use unordered_map here?

src/tools/wasm-interp.cc
```diff
@@ -239,6 +346,37 @@ static Result ReadAndRunModule(const char* module_filename) {
 
   RefVec imports;
 
+  // if we only have dummy imports, we wont need to reghister anything
+  // but we still want to load them.
+  std::vector<Module::Ptr> modules_loaded(s_modules.size());
```
sbc100, line 351: Can we unify the loading of the main `module_filename` above with the loading of `s_modules` here?

src/tools/wasm-interp.cc
```diff
@@ -161,33 +175,126 @@ Result RunAllExports(const Instance::Ptr& instance, Errors* errors) {
...
+  }
+
+  // prefer debug_name if override_name starts with '$'
+  if (override_name[0] == '$' && !debug_name.empty()) {
+    return debug_name;
+  }
+
+  // return the override_name (remove leading '$' if present)
+  bool leading_dollar = override_name[0] == '$';
+  return override_name.substr(leading_dollar ? 1 : 0);
```
sbc100, line 283: Do we need this dollar sign syntax?    What about this simpler scheme:

1. Override name, if present
2. Debug name, if present
3. Fall back to file name

src/tools/wasm-interp.cc
tDwtp: i wanted module to be used for the import/exprot because thats what its called in the docs of WebAssembly. `--preload` would not be to far fetched, but I still think `--module` (or just `--mod`) would still be best. (You can choose, just wanted to give my reasoning).

Sadly if we want it to interoperate with `--wasi` (and as long as we don't have a glue for native libraries) a new option is a must.
I was thinking about multiple different ways, but the OptionParser does not really like them. I cannot specify multiple modules, then a main one, and then optional parameters without *some* new option to mark the main module. I guessed `-m` for conveniance á la '-L"path" -llibname' would be best and in the end just stuck to the `-m`/`--module` to be more like the other options available.

I was thinking about something to glue native libraries and just have `--wasi` be an indicator which native library to give the optional arguments to.


references:
https://webassembly.github.io/spec/core/valid/modules.html#imports
https://webassembly.github.io/spec/core/text/modules.html#imports
https://webassembly.github.io/spec/core/syntax/modules.html#imports
https://webassembly.github.io/spec/core/binary/modules.html#import-section

src/tools/wasm-interp.cc
tDwtp: I am new to the game (last time I programmed C++ was 5-8 years ago), but I guess yes, does not make any difference.
I am going to update this.

src/tools/wasm-interp.cc
tDwtp: Still a new option, but if I waht you want, this would be better suited:
we could have an `-a, --arg` parameter right before the arguments. I just don't know how to make it optional and still have a list of arguments following it.

src/tools/wasm-interp.cc
tDwtp: I was thinking about it early, but dropped it, because I did not understand how it would make it work.
Probably why I have that vector in the first place.
Now on my TODO. After I figure out how to do the module option thingy.

src/tools/wasm-interp.cc
tDwtp: We need something to explicitly go for the debug name. I dont want something unexpected. Like downloading a module from the web, and not have it use the filename, because you might not check for the debug name. (just for convenience)

I came up with
1. `override_name` if present
2. if `override_name == "$"` debug_name if present
3. Fall back to file_name

```cpp
  if (!override_name.empty()) {
    // prefer override_name unless its just "$"
    if (override_name != "$") {
      return override_name;
    }
    
    // prefer debug_name if override_name == "$"
    if (!debug_name.empty()) {
      return debug_name;
    }
  }
  
  // fall back to file-name
  return FileNameOfPath(path_name);
```

src/tools/wasm-interp.cc
tDwtp: I know now why I did this, only the main module has access to `--wasi` the other ones don't.
I first need to figure out how to provide wasi support to all of them, then how to do the `--arg` thingy and then I don't mind.

src/tools/wasm-interp.cc
tDwtp: or we go for this:
1. if `override_name == "@"` use `file_name`
2. `overriide_name` if present
3. `debug_name` if present
4. Fall back to `file_name`

Binary compatibility woutl then be prefered

src/tools/wasm-interp.cc
tDwtp: Should I add a Divider to Option Parser?
```cpp
  parser.AddArgument("filename", OptionParser::ArgumentCount::OneOrMore, [...]);
  // parser.AddDivider("--", OptionParser::Divider::Optional);
  parser.AddArgument("arg", OptionParser::ArgumentCount::ZeroOrMore, [...]);
```
This way I could distinguish between modules and wasi arguments.

sbc100: Yes, I think using `--` to distinguish between arguments passed to wasi and argument to wasm-interp itself is good idea.

In general I'm a fan of the direction of this PR, and I think we could probably even land with without too much more iteration, but I would also like to be able to iterate on the ergonomics of it.  But perhaps we can do a lot of that iteration after first landing it.

src/tools/wasm-interp.cc
sbc100: I just don't think we need a way to force the file name to be used over the debug name.   If you really want to first the filename why not just use the override mechism.  e.g -m libfoo:libfoo.wasm.

In other words, always prefer the debug name, but give users an opt out to override if they want to.  That would avoid the magic `@` thing.

src/tools/wasm-interp.cc
sbc100: Speaking of `-L`.. I wonder if we could have wasi-interp automatically looks for wasm files in certain locations and load them on demand to satisfy imports?   

We could even do something fancy like JS import maps: https://github.com/WICG/import-maps

But all that is farther down the line after we land this initial support.

src/tools/wasm-interp.cc
```diff
@@ -327,6 +465,12 @@ int ProgramMain(int argc, char** argv) {
   s_store.setFeatures(s_features);
 
   wabt::Result result = ReadAndRunModule(s_infile);
+
+  for (auto& pair : s_registry) {
+    pair.second.clear();
+  }
+  s_registry.clear();
```
sbc100, line 472: I think it should be enough to just call `.clear()` on the `s_registry` .. this will automatically destroy the `ExportMap` object which are the values in the map, no?

sbc100 (approved):
I think the main thing we need to do before landing this is figure out a way to start testing this. 

We can bikeshed the name of the argument as a followup.   I would like to remove the magic `@` syntax though, since that seems unnecessary for v1.

src/tools/wasm-interp.cc
tDwtp: was just for conveniance... But tbh I don't care to much.
Its done already. I removed the magic thingy.

tDwtp: Oh I also can't find a way to get the DebugName, my best guess was it's hidden somewhere in `Instance`, but I cant find it.

src/tools/wasm-interp.cc
tDwtp: Well: Off I go to find out how to implement that... Possibly wihtout implementing a whole Divider thing.

tDwtp:
wasi is working now. and i change unordered_map and the file fallback.

I really tried to make the arguments work, but I would need another argument to `AddArgument`.
```cpp
parser.AddArgument("arg-name", OptionParser, [callback], std::string stopper);
```

I could also just add it to `OneOrMore` and `ZeroOrMore` and generally assume its `.` as i assumed for my case.
It cannot be `--` as this is handled by the parser itself! It does not seem worth the hassle.
I am *not* fluent enough in *all* shells to attempt it. I'd pick between `+`, `.` or `,` but I really dont know.
I picked `.` in my hack because it seemed most sensible.

sbc100:
What do you think about renaming this PR to "wasm-interp: Add multi-module support"?

sbc100:
> wasi is working now. and i change unordered_map and the file fallback.
> 
> I really tried to make the arguments work, but I would need another argument to `AddArgument`.
> 
> ```c++
> parser.AddArgument("arg-name", OptionParser, [callback], std::string stopper);
> ```
> 
> I could also just add it to `OneOrMore` and `ZeroOrMore` and generally assume its `.` as i assumed for my case. It cannot be `--` as this is handled by the parser itself! It does not seem worth the hassle. I am _not_ fluent enough in _all_ shells to attempt it. I'd pick between `+`, `.` or `,` but I really dont know. I picked `.` in my hack because it seemed most sensible.

Awesome, I don't think we need to worry about getting the option separator thing in as part of this change.

src/tools/wasm-interp.cc
```diff
@@ -228,92 +372,129 @@ static Result InstantiateModule(RefVec& imports,
...
-    uvwasi_options_t init_options;
-
-    std::vector<const char*> argv;
-    argv.push_back(module_filename);
-    for (auto& s : s_wasi_argv) {
-      if (s_trace_stream) {
-        s_trace_stream->Writef("wasi: arg: \"%s\"\n", s.c_str());
-      }
-      argv.push_back(s.c_str());
+    result = InitWasi(&uvwasi);
```
sbc100, line 435: What do you think aInitWasibout making a separate NFC (non-functional change) which would just refactor all this wasi stuff.  i.e. InitWasi, RegisterWasiInstance, etc?  

That would make this change much simpler to understand I think.

src/interp/interp-wasi.h
```diff
@@ -28,10 +28,14 @@ struct uvwasi_s;
...
-Result WasiBindImports(const Module::Ptr& module,
-                       RefVec& imports,
-                       Stream* err_stream,
-                       Stream* trace_stream);
+Result RegisterWasiInstance(const Instance::Ptr& instance,
+                            uvwasi_s* uvwasi,
+                            Stream* stream,
+                            Stream* trace_stream);
+
+void UnregisterWasiInstance(const Instance::Ptr& instance);
```
sbc100, line 36: Can we make all these function symbols start with `Wasi`?

tDwtp:
Don't have much time for today,

> We can bikeshed the name of the argument as a followup. I would like to remove the magic `@` syntax though, since that seems unnecessary for v1.

Will revert the hack later.

> I think the main thing we need to do before landing this is figure out a way to start testing this.

I have a simple test ready. will add 4 more simple ones at least. Already have them in mind.
I convert a wat file to wasm and use it for module imports. I just don't know where to put them and how the testing works exactly.

CU tomorrow.

tDwtp:
I... do not remember how I had planned the tests... I need help again.

src/interp/interp-wasi.cc
```diff
@@ -35,6 +35,7 @@
 
 #include "uvwasi.h"
 
+#include <cstdio>
```
sbc100, line 38: Is this needed?

src/tools/wasm-interp.cc
```diff
@@ -161,33 +175,147 @@ Result RunAllExports(const Instance::Ptr& instance, Errors* errors) {
...
+
+static std::string GetPathName(std::string module_arg) {
+  size_t path_at = module_arg.find_first_of(':');
+
+  path_at = path_at == std::string::npos ? 0 : path_at + 1;
+
+  return module_arg.substr(path_at);
+}
+
+static std::string FileNameOfPath(std::string path_name) {
```
sbc100, line 252: I think you can use `GetBasename` + `StripExtension` from `wabt/filenames.h`?

src/interp/interp-wasi.cc
```diff
@@ -652,85 +653,90 @@ std::unordered_map<Instance*, WasiInstance*> wasiInstances;
...
-                     import.type.module.c_str());
-      return Result::Error;
-    }
+  if (!memory) {
+    err_stream->Writef("wasi error: memory export not found\n");
+    return Result::Error;
+  }
+
+  WasiInstance* wasi = new WasiInstance(instance, uvwasi, std::move(memory).get(), trace_stream);
+  wasiInstances[instance.get()] = std::move(wasi);
```
sbc100, line 682: I don't think `std::move` is meaningful on raw pointer.

Maybe just avoid the local and do `wasiInstances[instance.get()] = new WasiInstance...`

src/interp/interp-wasi.cc
```diff
@@ -652,85 +653,90 @@ std::unordered_map<Instance*, WasiInstance*> wasiInstances;
...
+}
 
-    // TODO(sbc): Validate signatures of imports.
-#define WASI_FUNC(NAME)                                 \
-  if (import.type.name == #NAME) {                      \
-    host_func = HostFunc::New(*store, func_type, NAME); \
-    goto found;                                         \
+void WasiUnregisterInstance(const Instance::Ptr& instance) {
+  WasiInstance* wasi = wasiInstances[instance.get()];
+  if (wasi) {
```
sbc100, line 689: How about `assert(wasi);` and skip the `if` check?

src/interp/interp-wasi.cc
```diff
@@ -753,7 +756,7 @@ Result WasiRunStart(const Instance::Ptr& instance,
   }
 
   // Unregister memory
-  wasiInstances.erase(instance.get());
+  UnregisterWasiInstance(instance);
```
sbc100, line 759: This needs to be renamed I think.

sbc100: Should this line be removed now that the registration and unregistration now all seems to be taken care in wasm-interp.cc?

src/interp/interp-wasi.cc
tDwtp: It might allready be deleted. The main file gets a delete before this happens, which might not be cleared, if `_start` was not found or used. See `WasiRunStart`. Its just for later, when I might load and clear without any order.

src/interp/interp-wasi.cc
tDwtp: Nope, You might use it elsewhere (in spec-test for example). I did not want to monolith `wasm-interp.cc`.

src/interp/interp-wasi.cc
tDwtp: was for testing and finding a memory-leak culprit. is removed. Thanks for the notice ._.

src/tools/wasm-interp.cc
tDwtp: Thanks for the notice. I did not even know that exists already.

tDwtp:
sorry for the many mails... I am still setting stuff up, I moved and did not have my pc set up untill now.
I still have not set up my dev enironment properly...

src/tools/wasm-interp.cc
tDwtp: @sbc100 minor question, wouldn't it make sense to just ... use all three? Like: Always register by `file_name`, if exists register by `debug_name`, if exists register by `override_name`.
This could be an idea for a revision. And as soon as I findout how to get `debug_name`. (It seems like I need to query it manually from the file? The other approaches used in other files are ... complicated)

PR #2013, opened 2022-10-03
WastLexer: log lexing errors directly

keithw:
Currently, the WastLexer logs errors by calling the Error method on a passed-in WastParser. This makes it inconvenient to log an error in contexts where no WastParser was available (e.g. from WastLexer::NoTrailingReservedChars()). This is newly an issue because of the tokenization change in https://github.com/WebAssembly/spec/pull/1499.

This PR changes/simplifies things so a `WastLexer` receives an `Errors*` parameter on construction, just like a `WastParser` does. That lets the `WastLexer` log errors directly. This also disentangles the lexer and parser a bit (i.e. we no longer need to include `wast-parser.h` in `wast-lexer.cc`).

This also updates the tokens.txt test to include an extra error message that we had been losing.

(Edit: And also a corresponding change to the Emscripten helper and the JavaScript wrappers that call it... I'm not sure this JavaScript code gets tested in the CI, but I did build locally and test the demos, and it appears to work and reflect the changes.)


sbc100 (approved):
Did you mean to include docs/demo/wabt.js in this change?

keithw:
> Did you mean to include docs/demo/wabt.js in this change?

I did... it seemed a good time to do it, since we're changing the signatures of some functions that get called between the post JS and the C++ Emscripten helpers. But I'm definitely not an expert in JavaScript or the JS parts of WABT...

Issue #2014, opened 2022-10-04
cwriter-template not being installed with 1.0.30?


jakirkham:
When running CMake and installing wabt, library loading issues are occurring around `cwriter-template`. This just started occurring with 1.0.30 and did not occur with 1.0.29. For example:

```
wasm2c --help
wasm2c: error while loading shared libraries: libcwriter-template.so: cannot open shared object file: No such file or directory
```

Here is the [attached log]( https://github.com/WebAssembly/wabt/files/9704508/wabt-feedstock_20221003_1_linux_64.txt ) for more details. Also here is [the build script]( https://github.com/conda-forge/wabt-feedstock/blob/3452251fa3ff6d2375280fcc55a5372f66b66b47/recipe/build.sh#L1 ). This example is with Linux, but am seeing the same issue on other platforms.

AFAICT the `cwriter-template` library is not getting installed with everything else. Not sure if an `install` command is missing or if I should be doing something else.

PR #2015, opened 2022-10-04
wasm2c: Don't create a separate library for cwriter-template. NFC

sbc100:
Fixes: #2014

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -495,14 +495,13 @@ if (BUILD_TOOLS)
     ${WABT_SOURCE_DIR}/src/template/wasm2c.bottom.h
     ${WABT_SOURCE_DIR}/src/template/wasm2c.includes.c
     ${WABT_SOURCE_DIR}/src/template/wasm2c.declarations.c
-    )
+  )
 
-  add_library(cwriter-template wasm2c_header_top.cc wasm2c_header_bottom.cc wasm2c_source_includes.cc wasm2c_source_declarations.cc)
+  set(CWRITER_TEMPLATE_SRC wasm2c_header_top.cc wasm2c_header_bottom.cc wasm2c_source_includes.cc wasm2c_source_declarations.cc)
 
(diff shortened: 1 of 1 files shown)


sbc100:
I could just make it `SAT

CMakeLists.txt
```diff
@@ -497,12 +497,11 @@ if (BUILD_TOOLS)
     ${WABT_SOURCE_DIR}/src/template/wasm2c.declarations.c
     )
 
-  add_library(cwriter-template wasm2c_header_top.cc wasm2c_header_bottom.cc wasm2c_source_includes.cc wasm2c_source_declarations.cc)
+    set(CWRITER_TEMPLATE_SRC wasm2c_header_top.cc wasm2c_header_bottom.cc wasm2c_source_includes.cc wasm2c_source_declarations.cc)
```
keithw, line 500: Looks like there is some extra space here at the beginning of the line...

keithw (approved):
lgtm. It would be nice if we had testing to catch this kind of problem before release. :-(

Longer term, I think we would like to include c-writer.cc and the template sources in libwabt itself (rather than only linking them in to the wasm2c executable), so that programs that link against libwabt can use the CWriter class. But that can wait.

CMakeLists.txt
sbc100: > lgtm. It would be nice if we had testing to catch this kind of problem before release. :-(
> 
> Longer term, I think we would like to include c-writer.cc and the template sources in libwabt itself (rather than only linking them in to the wasm2c executable), so that programs that link against libwabt can use the CWriter class. But that can wait.

I actually don't know how this happened because I could repro it locally (locally I don't see any `.so` libraries).  I guess maybe there is some kind of cmake flag that says "build any libraries that are not specifically static, as dynamic".. but i odn't know what it is.

jakirkham:
Thanks Sam and Keith! 🙏

Issue #2016, opened 2022-10-06
wabt 1.0.30 build issue


chenrui333:
👋 trying to build the latest release, but run into some build issue. The error log is as below:

<details>
<summary>build error</summary>

```
  /tmp/wabt-20221003-91990-ggldl9/src/interp/interp-wasi.cc:647:10: fatal error: 'wasi_api.def' file not found
  #include "wasi_api.def"
           ^~~~~~~~~~~~~~
  1 error generated.
```

</details>

full build log, https://github.com/Homebrew/homebrew-core/actions/runs/3178360168/jobs/5179823985
relates to Homebrew/homebrew-core#112291

sbc100:
Oh, its looks like we are not testing with `-DWITH_WASI=ON`.   I'm working on a fix now.   In the mean you could remove that cmake flag?  (Do you actually need/use it?)

PR #2017, opened 2022-10-06
Fix WITH_WASI build

sbc100:
Also, make sure we test this configuration.

Fixes: #2016


.github/workflows/build.yml
```diff
@@ -50,7 +50,7 @@ jobs:
     - name: mkdir
       run: mkdir -p out
     - name: cmake
-      run: cmake .. -G Ninja -DWERROR=ON -Werror=dev
+      run: cmake .. -G Ninja -DWITH_WASI -DWERROR=ON -Werror=dev
```
keithw, line 53: Does this need to be `-DWITH_WASI=ON`?

keithw (approved):
lgtm once the tests pass...

PR #2021, opened 2022-10-19
wasm2c: simd support

wrv:
Hi folks!

I've been working with @hovav, @talg, @shravanrn, and @deian to add SIMD support to wasm2c.

This pull request incorporates [simd-everywhere](https://github.com/simd-everywhere/simde) into the wasm2c output, which maps WASM SIMD C intrinsics to any supported target architecture. 

The `test/simd` directory has some examples to verify everything works okay, but I still need to properly incorporate the SIMD tests in the testsuite.

Any feedback on how to improve this would be appreciated.


keithw:
Thanks, this is awesome!!! SIMD is the one remaining piece to have wasm2c supporting the entire WebAssembly 2.0 specification. Thank you so much for working on this.

I had previously implemented (some of) the test jig changes necessary to run the SIMD spec tests, so I just rebased that on top of your SIMD support (and formatted it), and most of the spec tests are passing! Please see https://github.com/keithw/wabt/tree/ucsd-simd and of course feel free to use anything that you see helpful there.

Once the spec tests are passing, we can do a full code review, but for now here's some high-ish-level comments:
- Are you wedded to having this separate test/simd directory? At first blush it seems like, let's just get the spec tests passing and that's probably good enough as far as what gets included in the repo.
- It might be ideal if the wasm2c output didn't include the SIMDe headers unless the module uses SIMD, but this seems hard to do given wasm2c's mostly single-pass nature. So maybe the next-best thing is to make it take `--enable-simd` (with default off) and only define the preprocessor macro (and include the new headers) if it's on? This isn't ideal since I think `--enable-simd` is the default everywhere else in WABT and of course it's merged into the spec, but seems like the least-bad and easiest option... @sbc100 may have a better idea. (To namespace it a little more nicely, maybe the macro should be something like `WASM_RT_ENABLE_SIMD` instead of `ENABLESIMD`?)
- Maybe we should claim 'v' for type-mangling a v128, since we don't have to worry about void anymore? (A multi-value struct always has at least two fields...)
- It looks like MSVC is barfing on some of these SIMDe function names? Hopefully there's an easy fix there...
- I'm guessing the SIMD load functions will need something like `wasm_asm("" ::"ryx"(result));` to tell the C compiler we care about the value of the load (because we need a deterministic trap if the load is out of range), just like we needed this on the ordinary load functions. Maybe there is a more cross-platform way to do this...?
- This doesn't seem to try to support SIMD load/store on big-endian platforms. Maybe we don't care (especially since our support for big-endian platforms seems to have bitrotted anyway, per https://github.com/WebAssembly/wabt/pull/1969), but it seems like either we should say we don't care (and strip out the big-endian support) or we should try to be consistent between SIMD vs. non-SIMD loads and stores?
- Don't forget to update the wasm2c README to check the SIMD box! :-)

Bottom line, this is great and looks like you did a huge amount of work -- thank you and I'll be excited when we have this merged.

sbc100:
@ngzhian WDYT of this approach?

ngzhian:
> @ngzhian WDYT of this approach?

Using simd-everywhere is a good idea! nemequ@ fyi

sbc100:
Regarding testing, I agree with @keithw that it might be best to use the normal spec test suite rather than writing something custom.  If we do keep those custom tests they should probably live in `test/wasm2c/simd`.  

To experiment with the upstream test suite you can just run `./test/update-spec-tests.py` and this should create whole bunch of tests in `test/wasm2c/spec/`.   You can then run these with `./test/run-tests.py test/wasm2c/spec/`.. or run just one of them at a time. e.g. `./test/run-tests.py test/wasm2c/spec/simd_load.txt`

sbc100:
I also think we could land this even with only a subset of the spec tests passing, on the assumption that we can iterate on it.

src/c-writer.cc
```diff
@@ -3466,39 +4367,121 @@ void CWriter::Write(const SimdLaneOpExpr& expr) {
...
+  Memory* memory = module_->memories[0];
+
+  const char* func = nullptr;
+  switch (expr.opcode) {
+    /* SIMD V128 LoadZero Opcodes */
+    case Opcode::V128Load32Zero: func = "v128_load32_zero"; break;
+    case Opcode::V128Load64Zero: func = "v128_load64_zero"; break;
+
+    default:
+      printf("issue with LoadZeroExpr opcode %s \n", expr.opcode.GetName());
```
sbc100, line 4472: Probably don't want this printf here in the final change.

src/c-writer.cc
```diff
@@ -3466,39 +4367,121 @@ void CWriter::Write(const SimdLaneOpExpr& expr) {
...
+    const char* func = nullptr;
+  // wasm_v128_store16_lane(in_a, res, 2);
+  switch (expr.opcode) {
+  case Opcode::V128Store8Lane: func = "v128_store8_lane"; break;
+  case Opcode::V128Store16Lane: func = "v128_store16_lane"; break;
+  case Opcode::V128Store32Lane: func = "v128_store32_lane"; break;
+  case Opcode::V128Store64Lane: func = "v128_store64_lane"; break;
+  default:
+      printf("issue with SimdStoreLaneExpr opcode %s \n", expr.opcode.GetName());
+      WABT_UNREACHABLE;
```
sbc100, line 4403: Indention looks odd here and in the first line of the function.  Maybe `git clang-format origin/main`?

src/c-writer.cc
```diff
@@ -3848,20 +3850,48 @@ void CWriter::Write(const ConvertExpr& expr) {
...
-    case Opcode::I64Load8S: func = "i64_load8_s"; break;
-    case Opcode::I32Load8U: func = "i32_load8_u"; break;
-    case Opcode::I64Load8U: func = "i64_load8_u"; break;
-    case Opcode::I32Load16S: func = "i32_load16_s"; break;
-    case Opcode::I64Load16S: func = "i64_load16_s"; break;
-    case Opcode::I32Load16U: func = "i32_load16_u"; break;
-    case Opcode::I64Load16U: func = "i64_load16_u"; break;
-    case Opcode::I64Load32S: func = "i64_load32_s"; break;
-    case Opcode::I64Load32U: func = "i64_load32_u"; break;
+    case Opcode::I32Load:
```
sbc100, line 3853: Can you instead wrap this case statement in `// clang-format off` `  `// clang-format on` ?  (So we can keep the brief version)

wrv: Can do!

wrv:
Thanks for the feedback! @keithw, I've incorporated the testing infrastructure you linked to into this most recent commit. It looks there are some SIMD instructions still left to debug, and I also saw you mention NaN support as missing.

> Are you wedded to having this separate test/simd directory?

Nope, we can get rid of those. 

> make wasm2c take --enable-simd at the command line (with default off) and have it define the preprocessor macro itself in the generated header file if that flag is set?

I tried to search for cases in which the SIMD support is checked, but couldn't find any examples. Do you think there's a way to see if any SIMD instructions have been read by the Binary IR reader, and then define the preprocessor macro if so?

> To namespace it a little more nicely, maybe the macro should be something like WASM_RT_ENABLE_SIMD instead of ENABLESIMD?

Changed. 

> Maybe we should claim 'v' for type-mangling a v128, since we don't have to worry about void anymore? 

I took 'o' from the __int128 type found here https://itanium-cxx-abi.github.io/cxx-abi/abi-mangling.html . I'm fine to change it, though.

> It looks like MSVC is very unhappy about something? 

From the link you shared, it looks like it was a missing `/DWASM_RT_ENABLE_SIMD` in the compilation parameters. MSVC also complained it didn't like some of the v128 casting so I removed that.

> Does this need to initialize a v128 local to all zeros?

I don't think it's required because values should be loaded before being operated on, but I included initialization in the most recent commit. 

> It might be nice to take the "amalgamated" SIMDe single-file header and compile into wasm2c as a template at build time, and have wasm2c spit it back out alongside the generated code (maybe at some fixed filename and only when some option is given?), rather than introducing a runtime dependency where the user needs to have a recent-enough SIMDe installed in their include path to compile anything that wasm2c generates.

I'm curious how this compares to wasm2c currently requiring "wasm-rt.h" to compile? 

> Don't forget to update the wasm2c README to check the SIMD box! :-)

Updated ^.^


For adding "something like wasm_asm("" ::"ryx"(result));" and big-endian platform support, part of using simd-everywhere was to abstract out the platform dependence. I added the former to the most recent commit, but I'll look into simd-everywhere to see if it can handle the latter.

Looking at the testing infrastructure, I've ran into two issues:
1. I may be missing something, but it looks like negative values aren't read correctly as constants in run-spec-wasm2c.py. In the `simd_i16x8_q15mulr_sat_s` test I got incorrect expected results. 
2. I'm not sure why, but the `simd_loadX_laneY` tests don't initialize variables?

keithw:
> Thanks for the feedback! @keithw, I've incorporated the testing infrastructure you linked to into this most recent commit.

Awesome! It's looking good so far. I tried to contribute some further changes that you might like; please see https://github.com/keithw/wabt/pull/1 (or https://github.com/keithw/wabt/tree/ucsd-simd) and feel free to take anything you find helpful.

> > make wasm2c take --enable-simd at the command line (with default off) and have it define the preprocessor macro itself in the generated header file if that flag is set?
> 
> I tried to search for cases in which the SIMD support is checked, but couldn't find any examples. Do you think there's a way to see if any SIMD instructions have been read by the Binary IR reader, and then define the preprocessor macro if so?

Probably (by modifying the reader to track this and set a Module-scope boolean in the IR), but it feels a bit gross. I think the easier option may just be to explicitly require an `--enable-simd` on the command line (similar to how we parse for `--enable-multi-memory` and other flags).

> > Maybe we should claim 'v' for type-mangling a v128, since we don't have to worry about void anymore?
> 
> I took 'o' from the __int128 type found here https://itanium-cxx-abi.github.io/cxx-abi/abi-mangling.html . I'm fine to change it, though.

Fair enough! I didn't know there was a standard for this.

> > Does this need to initialize a v128 local to all zeros?
> 
> I don't think it's required because values should be loaded before being operated on, but I included initialization in the most recent commit.

:+1: You're absolutely right for stack vars, but I think the problem happens if we don't initialize locals. E.g. if you have:
```wat
(module (func $uselocal (result i32) (local $x i32) (local.get $x)))
```
wasm2c will turn it into:
```c
static u32 w2c_uselocal(Z_example_instance_t* instance) {
  u32 w2c_x = 0;
  FUNC_PROLOGUE;
  u32 w2c_i0;
  w2c_i0 = w2c_x;
  FUNC_EPILOGUE;
  return w2c_i0;
}
```
and the only thing preventing UB is the `u32 w2c_x = 0;` where the local is initialized to zero.

> > It might be nice to take the "amalgamated" SIMDe single-file header and compile into wasm2c as a template at build time, and have wasm2c spit it back out alongside the generated code (maybe at some fixed filename and only when some option is given?), rather than introducing a runtime dependency where the user needs to have a recent-enough SIMDe installed in their include path to compile anything that wasm2c generates.
> 
> I'm curious how this compares to wasm2c currently requiring "wasm-rt.h" to compile?

Well... I guess it's not that clear, but I think of `wasm-rt.h` (and `wasm-rt-impl.h` and `wasm-rt-impl.c`) as an example runtime that WABT distributes and that the user may want to modify for their use-case, which they're free to do as long as they keep the same interface. (But they can change `wasm-rt.h` to change the type of `externref` if they want, etc.) If you install a Debian package of WABT (or the cmake wabt-development package), it ends up up installing these files in /usr/src/wasm2c/ , and it installs wasm2c in /usr/bin, so you can run wasm2c and then compile its output.

Whereas SIMDe is somebody else's package of a bunch of implementation headers (or one "amalgamated" header if you build it that way) that I don't think **we're** going to install in /usr/include ourselves, but no distributed/packaged version of SIMDe supports Wasm yet, so... how is the user supposed to actually run wasm2c and compile its output if that output depends on something that's only included in the Wabt source repo?

> For adding "something like wasm_asm("" ::"ryx"(result));" and big-endian platform support, part of using simd-everywhere was to abstract out the platform dependence. I added the former to the most recent commit, but I'll look into simd-everywhere to see if it can handle the latter.

We may need to put an `#ifdef __x86_64__` around the asm since I think "x" ix x86-specific... :-(

Re: endianness, that would be nice, although I worry that WABT's technique of storing the memory contents backwards on big-endian systems (to avoid byteswapping) may be too specialized to appear in a library, but, happy to be surprised. Honestly I'm not sure our big-endian code still works anyway. :-(

> Looking at the testing infrastructure, I've ran into two issues:
> 
>     1. I may be missing something, but it looks like negative values aren't read correctly as constants in run-spec-wasm2c.py. In the `simd_i16x8_q15mulr_sat_s` test I got incorrect expected results.

Agreed there's a problem here but I'm not sure where the responsibility lies yet. :-/

> 
>     2. I'm not sure why, but the `simd_loadX_laneY` tests don't initialize variables?

I believe this was a bug in the c-writer implementation; please see https://github.com/keithw/wabt/commit/97b7590280acc93e7d1fa7805fe91e2adc536345 for the fix. (There was a similar issue in the TernaryExpr implementation; see https://github.com/keithw/wabt/commit/90ff14861d8b62b9631f3bcfbdfa1e650c4332b5)

shravanrn:
> Probably (by modifying the reader to track this and set a Module-scope boolean in the IR), but it feels a bit gross.

Thought I'd throw in a minor data point. There are other features I am looking at for library sandboxing (which will be behind a feature flag) that unfortunately does have to do something like the above tracking. I'm mentioning this, in case it affects the current decision either way. For this scenario of SIMD, i am ok with either tracking or `--enable-simd`

keithw:
I definitely don't have a strong view -- if you think scanning can be done reasonably, no objection here. I agree it would also be nice to keep the wasm2c defaults the same as the rest of WABT (which would mean SIMD on by default).

sbc100:
This is looking really good!   Thanks for working on it.

I think maybe we can leave the "automatically detect SIMD usage" as a followup?

Lets couple more minor comments.

wrv:
An update on this:
- I found issues with SIMDe's handling of `f32x4_nearest` and `f64x2_nearest` and have opened a pull request to address those: https://github.com/simd-everywhere/simde/pull/973 
- For compiling the wast tests, I added `-march=native` as well as all the possible `/arch:` MSVC flags to enable architecture specific SIMD because I didn't find a `native` equivalent. On my machine, the MSVC flags that are not for my architecture are ignored, but let me know if there's a better way to do this. 
- The only major issue I'm encountering now is the representation of signed zero, but the [Intel Intrinsics Guide](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=signed-zero%20values) claims that the IEEE 754 standard is not followed for many instructions. 
- I have not been able to figure out how to fix the issue with the 2 failing tests of q15mulr, but given the [message on this previous commit](https://github.com/WebAssembly/simd/commit/770ce643) I'm guessing it has something to do with saturating the results? 

Let me know if there's anything else in particular you'd like me to address so this can be added.

sbc100 (approved):
looks good to me!   Is it ready to land now?

test/run-spec-wasm2c.py
```diff
@@ -320,8 +336,18 @@ def _WriteAssertActionCommand(self, command):
     def _Constant(self, const):
         type_ = const['type']
         value = const['value']
-        if type_ in ('f32', 'f64') and value in ('nan:canonical', 'nan:arithmetic'):
-            assert False
```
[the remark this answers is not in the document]
wrv: From what I've gathered, `nan:arithmetic` and `nan:canonical` are not constants, but some high order bits that are set, so they were only used as test result values, and checked with assert statements (e.g., https://github.com/WebAssembly/wabt/blob/main/test/run-spec-wasm2c.py#L267 and https://github.com/WebAssembly/wabt/blob/main/test/spec-wasm2c-prefix.c#L172). 

I created equivalent assert tests for if those `nan` strings are in any SIMD lanes.

keithw:
Thank you for all this awesome work!

Re: NaNs, the current test seems a little over-tolerant in that if **any** SIMD lane is expecting a NaN, it turns it into a `ASSERT_RETURN_CANONICAL_NAN_F64X2`-type test, which then succeeds as long as **any** SIMD lane actually is a NaN (even if it's not even the same one as was originally supposed to be one).

Here is a possible fix for your consideration: https://github.com/keithw/wabt/pull/2/commits/9649cb0dbe2c549a26d8486777c65674aa8d344d

This basically just treats each SIMD lane as equivalent to a multi-value return and tests them all individually. (I also had to add support to test NaNs in individual multi-value values since we didn't have that already...)

This doesn't alter the result of any tests that are already succeeding.

keithw:
I think we may be at a point where (all of?) the remaining test failures are more likely than not to reflect bugs in some of SIMDe's backend implementations of the Wasm SIMD instructions. (Are we possibly the first people to actually run SIMDe's Wasm intrinsics through the Wasm testsuite?)

E.g. the fact that the floor/ceil/trunc tests pass with `-march=native` but not without it suggests to me that SIMDe's **SSE2** implementations of the Wasm floor/ceil/trunc instructions might be slightly off, but the **SSE4.1** implementations match the Wasm semantics. This is a little frustrating from a testing perspective but probably something everybody has to deal with. (I wonder how V8 and Wasmtime test all these zillions of possible permutations?)

Given this situation, I think we probably have something like two practical options:

1. Omit the ~five failing test .txt files for now, land this PR soon (maybe using an "in-progress" emoji for SIMD in the README so we don't suggest full conformance yet?), and work with SIMDe to fix the test failures in their implementations over time. Ideally SIMDe would run the Wasm testsuite against all their backend implementations of the Wasm instructions (e.g. SSE4.1, SSE2, pure C, ARM, etc.).

2. Modify this PR so that wasm2c implements the SIMD floor/ceil/trunc/nearest/Q15 operations itself in pure C, rather than calling into SIMDe for these. That would get us 100% conformance.

Either one works for me... my tentative suggestion is probably for #1 because we can always do #2 in a follow-on.

ngzhian:
> I think we may be at a point where (all of?) the remaining test failures are more likely than not to reflect bugs in some of SIMDe's backend implementations of the Wasm SIMD instructions. (Are we possibly the first people to actually run SIMDe's Wasm intrinsics through the Wasm testsuite?)
> 
> E.g. the fact that the floor/ceil/trunc tests pass with `-march=native` but not without it suggests to me that SIMDe's **SSE2** implementations of the Wasm floor/ceil/trunc instructions might be slightly off, but the **SSE4.1** implementations match the Wasm semantics. This is a little frustrating from a testing perspective but probably something everybody has to deal with. (I wonder how V8 and Wasmtime test all these zillions of possible permutations?)
> 

https://github.com/simd-everywhere/simde/blob/master/simde/wasm/simd128.h#L8661
There is a special case for SSE4.1 and then a fallback case for SSE2, so you likely nailed the reason.
(I don't think SIMDe runs against the spec test suite.)

> Given this situation, I think we probably have something like two practical options:
> 
> 1. Omit the ~five failing test .txt files for now, land this PR soon (maybe using an "in-progress" emoji for SIMD in the README so we don't suggest full conformance yet?), and work with SIMDe to fix the test failures in their implementations over time. Ideally SIMDe would run the Wasm testsuite against all their backend implementations of the Wasm instructions (e.g. SSE4.1, SSE2, pure C, ARM, etc.).
> 2. Modify this PR so that wasm2c implements the SIMD floor/ceil/trunc/nearest/Q15 operations itself in pure C, rather than calling into SIMDe for these. That would get us 100% conformance.
> 
> Either one works for me... my tentative suggestion is probably for #1 because we can always do #2 in a follow-on.

I encourage you to file a bug on SIMDe with a small repro, they are pretty responsive.
If you are interested in fixing stuff on SIMDE and would like to put up a PR, I can give you pointers on where you look.

keithw:
Here are some more commits that get this "closer" to passing all the GitHub CI checks: https://github.com/keithw/wabt/pull/2

- remove the five failing spec tests for now (simd_f32x4, simd_f32x4_rounding, simd_f64x2, simd_f64x2_rounding.wast, simd_i16x8_q15mulr_sat_s)
- use "x" (instead of "rx") as the asm constraint for SIMD loads (on x86-64), otherwise clang on macOS can optimize out the load which makes the simd_address test fail (since it's supposed to trap on an out-of-bounds load)
- remove the arch-specific compiler flags for now since they seem to produce a lot of logged error messages on MSVC (causing test output mismatches)
- extend the SLOW factor to 10x (instead of 3x) so some of the slooow tests come closer to passing

Even with all this, it's still timing out on all the GitHub Linux tests -- some of these are just making the compiler work really hard and the tests are taking >2 hours. (I'm a little confused why the asan tests take particularly long... afaik these are still just compiling the wasm2c tests with cc, which I think is still gcc, and without any sanitizer flags. But I must be confused somehow on what's happening. The tests do pass on my Ubuntu laptop in ~20 minutes.)

keithw:
Update: I got the tests to go 100% green [as you saw] by switching the asan and ubsan tests to `clang -march=x86-64-2` instead of GCC, and further increasing some timeouts. The tests are still pretty slow; for the main "build (ubuntu-latest)" test, which is still GCC, it's taking an hour (up from 8 minutes without SIMD). The ubsan test (now on clang) took 41 minutes.

keithw:
On the "test duration" issue, I don't want to hold this up indefinitely, but I'm also sad about making the ubuntu-latest test go from ~8 minutes to ~68 minutes and become the critical path for the whole testsuite (which used to take <20 minutes). The problem seems to be the extremely large `run_spec_tests()` function body in the pmin/pmax tests, which GCC just has a hard time with. Splitting the function into like 2 half-length functions looks like it would help a dramatic amount.

I think we could:

1. land this as-is, and try to reduce the test duration later
2. give up on GCC completely and use clang everywhere
3. try to first land a change to `run-spec-wasm2c.py` that declares module instances as globals and then splits up the `run_spec_tests()` into a few sub-functions if it runs over 1,000 lines

My vote would be for #1 or #3, and I don't feel that strongly, especially if @wrv and others are eager to get this in. But >hourlong tests are a drain on everybody's productivity.

sbc100:
> On the "test duration" issue, I don't want to hold this up indefinitely, but I'm also sad about making the ubuntu-latest test go from ~8 minutes to ~68 minutes and become the critical path for the whole testsuite (which used to take <20 minutes). The problem seems to be the extremely large `run_spec_tests()` function body in the pmin/pmax tests, which GCC just has a hard time with. Splitting the function into like 2 half-length functions looks like it would help a dramatic amount.
> 
> I think we could:
> 
> 1. land this as-is, and try to reduce the test duration later
> 2. give up on GCC completely and use clang everywhere (clang doesn't seem to have the same problem with a super-long function body)
> 3. try to first land a change to `run-spec-wasm2c.py` that declares module instances as globals and then splits up the `run_spec_tests()` into a few sub-functions if it runs over 1,000 lines
> 
> My vote would be for #1 or #3, and I don't feel that strongly, especially if @wrv and others are eager to get this in. But >hourlong tests are a drain on everybody's productivity, especially for something that really doesn't need to take an hour. Anybody else have a view?

Does this only effect gcc then?   Are all versions of gcc effected?    Do we run with gcc on CI?   I would rather not have the CI start to take over and hour. 

How about a mode where wasm2c output files with >N are skipped if the compiler is GCC?   The CI systems could use that mode, as could local gcc users who don't want to wait forever for tests.

sbc100:
> I just took a stab at approach #3 (in PR #2077). This seems to do the trick; on my laptop running Ubuntu 22.04 and GCC 11.3, it reduces the time for the simd_f32x4_pmin_pmax.wast test from 16 minutes down to about 70 seconds. The Python is probably... improvable.
> 
> Re: the questions:
> 
> > Does this only effect gcc then?
> 
> Yeah, the long function bodies don't seem to be a problem on clang.
> 
> > Are all versions of gcc effected?
> 
> It seems to affect GCC 11 and 12; I haven't tested on tip-of-tree.
> 
> > Do we run with gcc on CI?
> 
> Yes, right now (on main branch) all the Linux tests use GCC (ubuntu-latest, asan, ubsan, and min-cmake). After this PR as it stands, asan and ubsan are moving to clang, but ubuntu-latest and min-cmake are still on GCC.
> 

Hmm, I wonder if we should always run the wasm2c tests with the same compiler we use to build wasm2c.   i.e. if I run `make clang-debug` and then I run the wasm2c tests I probably want to be using clang there too... not sure how easy it would be to make that work.

keithw:
> Hmm, I wonder if we should always run the wasm2c tests with the same compiler we use to build wasm2c. i.e. if I run `make clang-debug` and then I run the wasm2c tests I probably want to be using clang there too... not sure how easy it would be to make that work.

Probably pretty doable; we would need to set the CC and WASM2C_CFLAGS environment variables in the Makefile's TEST routine. (We probably want to set both of them -- right now if you "make clang-release-asan", it doesn't run the wasm2c tests with clang OR with asan. The GitHub actions do get this right by explicitly setting WASM2C_CFLAGS in the asan and ubsan tests.)

keithw:
Ok great, with the merger of #2078 and #2077, these SIMD changes are down to 37 minutes in the CI.

The critical path is now the asan and ubsan tests, and I think the main reason is basically that each test runs the entire wasm2c testsuite twice (e.g. for test-clang-debug-asan and test-clang-release-asan) with the exact same wasm2c flags each time.

We could split these tests in two (basically asan-debug and asan-release, and ubsan-debug and ubsan-release) to get the whole testsuite down to around 18 minutes. And I think if we're doing that, I guess ideally the "-debug" tests would compile the wasm2c spec tests without "-O2" (or at least with "-O0" appended at the end) to get the additional test coverage?

keithw:
Well, the good news is that with the changes in #2080 and #2081, the CI tests (even with this SIMD PR applied) are down to around 16 minutes end-to-end (https://github.com/keithw/wabt/pull/3 for a test run).

The bad news is that now that we're running the UBSAN tests properly, it's finding some issues in SIMDe. For example, UBSAN believes that SIMDe's implementations of load16_lane, load32_lane, and load64_lane are undefined behavior because they rely on an unaligned access and cast (e.g. https://github.com/simd-everywhere/simde/blob/master/simde/wasm/simd128.h#L7937) instead of using memcpy.

I think for now we can drop these three tests and fix it later, either by upstreaming a fix in SIMDe or using our own implementation. @wrv what do you think is best?

Separately, now that I'm looking more closely at that code, I'm not sure the implementation of load_lane, load_splat, and possible store_lane is quite right in terms of the MEMCHECK (which is only relevant on non-signal-handler platforms). I added a test with explicit memchecks (turning off the signal handler) that some of these seem to be failing (https://github.com/keithw/wabt/actions/runs/3572017180/jobs/6004437082). I'll make a comment on the actual code so you can see what I'm talking about. @wrv could you please take a look?

src/template/wasm2c.declarations.c
```diff
@@ -120,6 +120,187 @@ DEFINE_STORE(i64_store8, u8, u64)
...
+
+#define DEFINE_SIMD_STORE_LANE(name, func, t, lane)                   \
+  static inline void name(wasm_rt_memory_t* mem, u64 addr, t value) { \
+    MEMCHECK(mem, addr, t);                                           \
+    func((v128*)&mem->data[addr], value, lane);                       \
+  }
+
+DEFINE_SIMD_LOAD_FUNC(v128_load, simde_wasm_v128_load, v128);
+
+DEFINE_SIMD_LOAD_FUNC(v128_load8_splat, simde_wasm_v128_load8_splat, v128);
```
keithw, line 164: Should the "t" param here possibly be a u8 instead of a v128? I don't think it needs to MEMCHECK room for an entire v128 if it's only loading a u8 and then splatting it. And similarly for the load16_splat, load32_splat, load32_zero/load64_zero, loadN_lane, storeN_lane, etc.

wrv: Yeah, good catch. I just pushed a fix for that.

sbc100:
Looks like this change needs to be rebased/merged before it can land.

wrv:
> I think for now we can drop these three tests and fix it later, either by upstreaming a fix in SIMDe or using our own implementation.
I've opened up https://github.com/simd-everywhere/simde/pull/974 to address the issue, so I think it's safe to drop the tests for now. 

I've also addressed the MEMCHECK size issue in my recent commits. 

I appreciate the feedback and review this pull request has received so far.

keithw:
@sbc100 would you be able to take another look at #2080 and #2081? I think once those are in (subject to revision), and this one is rebased on top, I'm hopeful the CI times will turn out to be manageable and this PR can be merged soon after. I think it's very close.

@wrv, after we get through that, I imagine we'll need you to actually rebase this PR on top of the current main branch (not just copy commits from it -- actually "git rebase origin/main"). If you prefer you can also click "Allow edits from maintainers" in the PR and one of us can help with some of the final cleanups/rebasing -- your call.

Also: If you submit a trivial PR here for something else and get it merged, then the CI will start running automatically every time you update the PR. (It only needs approval for first-time contributors.) Or you could set up the PLSysSec fork (or your own fork) to run the GitHub actions on push. Might be more convenient so you can rapidly iterate to get them all green, which is a prerequisite for merging.

wrv:
@keithw  Since this PR is from an organization, it seems that I don't have the ability to click "Allow edits from maintainers". I've instead forked the current main branch, reapplied the changes, and got the tests passing in [my own repo](https://github.com/wrv/wabt). Do you recommend I create a pull request from that one so y'all can help with the final clean up? 

I'm excited to get this included and eager to help as best as I can.

wrv:
Just a heads up that I've gotten a response on https://github.com/simd-everywhere/simde/pull/974, so that should help us move forward.

wrv:
Following up to share that https://github.com/simd-everywhere/simde/pull/973 and https://github.com/simd-everywhere/simde/pull/974 have been incorporated into the SIMDe main branch. A new release is in the works: https://github.com/simd-everywhere/simde/issues/861#issuecomment-1387389094. 

@sbc100 @keithw let me know if there's anything I can do to help with https://github.com/WebAssembly/wabt/pull/2080 or https://github.com/WebAssembly/wabt/pull/2081.

wrv:
In total, there are 5 failing tests, with Q15 the only one I'm not sure of the cause. The others are due to floating point signed zero or NAN comparisons. 
1. `simd_f32x4.wast`: max has issues with NANs and signed zero, and min has issues with signed zero.
2. `simd_f32x4_rounding.wast`: trunc has issues with NAN, and nearest with signed zero.
3. `simd_f64x2.wast`: same as simd_f32x4.wast
4. `simd_f64x2_rounding.wast`: ceil has an issue with signed zero, floor with NAN, trunc with NAN, and nearest with signed zero.
5. `simd_i16x8_q15mulr_sat_s.wast`: unclear. I get a call stack exhaustion on Linux testing, and an out-of-bounds memory/table access in Windows.

keithw:
Closing as completed in #2119.

keithw:
I submitted a PR to SIMDe that gets the SIMD tests all passing (on Intel at least): https://github.com/simd-everywhere/simde/pull/1010

PR #2023, opened 2022-10-24
[wasm-objdump] Print segment offsets as unsigned

sbc100:


src/binary-reader-objdump.cc
```diff
@@ -1757,10 +1757,18 @@ void BinaryReaderObjdump::PrintInitExpr(const InitExpr& expr) {
...
+      if (as_unsigned) {
+        PrintDetails(" - init i32=%u\n", expr.insts[0].imm.i32);
+      } else {
+        PrintDetails(" - init i32=%d\n", expr.insts[0].imm.i32);
+      }
       break;
     case InitExprType::I64:
-      PrintDetails(" - init i64=%" PRId64 "\n", expr.insts[0].imm.i64);
+      if (as_unsigned) {
+        PrintDetails(" - init i64=%" PRId64 "\n", expr.insts[0].imm.i64);
```
keithw, line 1768: Is this backwards? (PRId64 for `as_unsigned` and PRIu64 otherwise?)

keithw (approved):
lgtm % comment

It might be nice if we had a dump test showing that all the other init exprs are still printed as signed. (Right now you could make as_unsigned the default and not break anything.)

sbc100:
> lgtm % comment
> 
> It might be nice if we had a dump test showing that all the other init exprs are still printed as signed. (Right now you could make as_unsigned the default and not break anything.)

Good idea. Done.

PR #2024, opened 2022-10-26
Align `memory.copy` arg types with `memory64` proposal

adambratschikaye:
When both the `multi-memory` and `memory64` features are enabled it's possible for `memory.copy` to copy between two memories of different types. In this case, the `memory64` proposal specifies that the `src` and `dst` arguments should correspond to the types of the memory they are accessing, and the `size` argument should correspond to the "minimum" of the two memory types [1]. That is, the `size` should be an `i32` if either of the memories has type `i32` and it should be `i64` if both of the memories have type `i64`.

This existing code just expects all three arguments to match the type of the source memory which works whenever the source and destination memories have the same type. This change adjusts that logic to agree with the `memory64` proposal in the cases where the two memories have differing types.

[1]
https://github.com/WebAssembly/memory64/blob/5bfb70d9888b96a2f3d6412ed3599b91364da610/proposals/memory64/Overview.md?plain=1#L211

### DIFF
--- a/include/wabt/type-checker.h
+++ b/include/wabt/type-checker.h
@@ -99,7 +99,7 @@ class TypeChecker {
   Result OnLocalSet(Type);
   Result OnLocalTee(Type);
   Result OnLoop(const TypeVector& param_types, const TypeVector& result_types);
-  Result OnMemoryCopy(const Limits& limits);
+  Result OnMemoryCopy(const Limits& srclimits, const Limits& dstlimits);
   Result OnDataDrop(Index);
   Result OnMemoryFill(const Limits& limits);
   Result OnMemoryGrow(const Limits& limits);
--- a/src/type-checker.cc
+++ b/src/type-checker.cc
@@ -706,8 +706,16 @@ Result TypeChecker::OnLoop(const TypeVector& param_types,
   return result;
 }
 
-Result TypeChecker::OnMemoryCopy(const Limits& limits) {
-  return CheckOpcode3(Opcode::MemoryCopy, &limits, &limits, &limits);
+Result TypeChecker::OnMemoryCopy(const Limits& src_limits,
+                                 const Limits& dst_limits) {
+  Limits size_limits = src_limits;
+  // The memory64 proposal specifies that the type of the size argument should
+  // be the mimimum of the two memory types.
+  if (src_limits.is_64 && !dst_limits.is_64) {
+    size_limits = dst_limits;
+  }
+  return CheckOpcode3(Opcode::MemoryCopy, &src_limits, &dst_limits,
+                      &size_limits);
 }
 
 Result TypeChecker::OnDataDrop(uint32_t segment) {
--- a/src/shared-validator.cc
+++ b/src/shared-validator.cc
@@ -909,10 +909,11 @@ Result SharedValidator::OnMemoryCopy(const Location& loc,
                                      Var srcmemidx,
                                      Var destmemidx) {
   Result result = CheckInstr(Opcode::MemoryCopy, loc);
-  MemoryType mt;
-  result |= CheckMemoryIndex(srcmemidx, &mt);
-  result |= CheckMemoryIndex(destmemidx, &mt);
-  result |= typechecker_.OnMemoryCopy(mt.limits);
+  MemoryType srcmt;
+  MemoryType dstmt;
+  result |= CheckMemoryIndex(srcmemidx, &srcmt);
+  result |= CheckMemoryIndex(destmemidx, &dstmt);
+  result |= typechecker_.OnMemoryCopy(srcmt.limits, dstmt.limits);
   return result;
 }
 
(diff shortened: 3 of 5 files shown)


adambratschikaye:
This should fix https://github.com/WebAssembly/wabt/issues/2025.

sbc100:
Thanks!

This seems like a gap in the spec tests, although I suppose it really the intersection of two different proposals?   Does one proposal depend on the other?

adambratschikaye:
Yes, I think there is a gap in the testing, but I'm not sure what the policy/expectation/responsibility is with regard to proposals including tests that depend on other proposals. In this case neither proposal actually depends on the other so I don't think it would be possible to add a test in the actual proposal branches of the spec.

It seems like there will always be a gap of this sort when two independent proposals interact before they are standardized?

sbc100:
> Yes, I think there is a gap in the testing, but I'm not sure what the policy/expectation/responsibility is with regard to proposals including tests that depend on other proposals. In this case neither proposal actually depends on the other so I don't think it would be possible to add a test in the actual proposal branches of the spec.
> 
> It seems like there will always be a gap of this sort when two independent proposals interact before they are standardized?

Exactly. In these situations it seems reasonable for wabt to write its own tests, perhaps will a comment in the test to mention that it should be a spec test once one of the proposals lands.

adambratschikaye:
> perhaps write a comment in the test to mention that it should be a spec test once one of the proposals lands.

Yeah, certainly. Just added it.

keithw:
Thanks for this! If you have the cycles, are you up for adding an invalid test as well? It would be good to make sure that we correctly reject cases where the size argument is the wrong type. We can also land this now and defer improving the test until later if you prefer.

include/wabt/type-checker.h
```diff
@@ -172,13 +173,13 @@ class TypeChecker {
                            Type expected3,
                            const char* desc);
   Result CheckOpcode1(Opcode opcode,
-                      const Limits* limits = nullptr,
+                      std::optional<Type> type = std::nullopt,
```
sbc100, line 176: Maybe give this a more meaningful name?  How about `memory_type`?   Previously it was more obvious that the limits argument related to the memory, but `type` on it own in less clear I think.

src/type-checker.cc
```diff
@@ -713,24 +719,34 @@ Result TypeChecker::OnLoop(const TypeVector& param_types,
...
+Result TypeChecker::OnMemoryCopy(const Limits& src_limits,
+                                 const Limits& dst_limits) {
+  std::optional<Type> src_type = GetLimitType(src_limits);
+  std::optional<Type> dst_type = GetLimitType(dst_limits);
+  std::optional<Type> size_type = Type(Type::I32);
+  // The memory64 proposal specifies that the type of the size argument should
+  // be the mimimum of the two memory types.
+  if (src_limits.is_64 && dst_limits.is_64) {
+    size_type = Type(Type::I64);
+  }
```
sbc100, line 731: It seems like we could perhaps avoid the rest of this change if we restructured this method a little bit.

How about something like this:

```
Limits size_limits = src_limits;
// The memory64 proposal specifies that the type of the size argument should
// be the mimimum of the two memory types.
if (src_limits.is_64 && !dst_limits.is_64) {
  size_limits = dst_limits;
}
return CheckOpcode3(Opcode::MemoryCopy, &src_limits, &dst_limits, &size_limits);
```

I think this would avoid change the majority of these function signatures and all the additional use of `std::optional`.

Do you think its makes more sense to be passing around `std::optional<Type>` everywhere instead of `Limits*`?  If so maybe we can seperate out that change and land that first as an NFC? (non-functional change)

adambratschikaye: Yes, I can definitely split them up if you prefer. I was thinking `std::optional<Type>` made more sense then `Limits*` because ultimately the `Type` is the only field in `Limits` that was being used.

include/wabt/type-checker.h
adambratschikaye: Yep, I changed them all to `memory_type`.

src/type-checker.cc
adambratschikaye: I created https://github.com/WebAssembly/wabt/pull/2032 which does just the `Limits* -> std::optional<Type>` change. And I'll rebase this PR on top of it so that it just contains the change in logic.

adambratschikaye: I already rebased https://github.com/WebAssembly/wabt/pull/2032 and this on top of https://github.com/WebAssembly/wabt/pull/2031.

adambratschikaye: Just to be clear: This PR now has two commits, but the first commit is it's own PR as https://github.com/WebAssembly/wabt/pull/2032.

sbc100:
Can we land this change with just 0319eda3477ba3d0429ce58863119e6c87e866f1.. and consider the refactor separately?

adambratschikaye:
> Can we land this change with just [0319eda](https://github.com/WebAssembly/wabt/commit/0319eda3477ba3d0429ce58863119e6c87e866f1).. and consider the refactor separately?

Yep, changed this to now just have the logic change.

adambratschikaye:
Thanks!

PR #2028, opened 2022-10-27
Fix wasm2c examples to print helpful prompts

shravanrn:
Ensure wasm2c examples print something useful when you run them

### DIFF
--- a/wasm2c/examples/rot13/main.c
+++ b/wasm2c/examples/rot13/main.c
@@ -45,6 +45,11 @@ void Z_hostZ_buf_done(struct Z_host_instance_t* instance, u32 ptr, u32 size) {
 }
 
 int main(int argc, char** argv) {
+  /* Make sure there is at least one command-line argument. */
+  if (argc < 2) {
+    printf("Invalid argument. Expected '%s WORD'\n", argv[0]);
+    return 1;
+  }
   /* Initialize the Wasm runtime. */
   wasm_rt_init();
 
--- a/wasm2c/examples/fac/main.c
+++ b/wasm2c/examples/fac/main.c
@@ -5,7 +5,10 @@
 
 int main(int argc, char** argv) {
   /* Make sure there is at least one command-line argument. */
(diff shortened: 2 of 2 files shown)


keithw:
Both lgtm if you think this makes things more approachable. (On rot13, how would you feel about making it print a helpful message only if there are no command-line arguments? But I don't feel strongly if you prefer it this way.)

keithw (approved):

shravanrn:
> print a helpful message only if there are no command-line arguments?

Fixed!

wasm2c/examples/rot13/main.c
```diff
@@ -45,6 +45,11 @@ void Z_hostZ_buf_done(struct Z_host_instance_t* instance, u32 ptr, u32 size) {
 }
 
 int main(int argc, char** argv) {
+  /* No command-line arguments expected. */
+  if (argc > 1) {
```
keithw, line 49: I think the test and error message here is backwards -- the program runs rot13 on its command-line arguments. So I think we probably want something like `if (argc <= 1)` for this test, and then the usage message would tell the user they're expected to supply at least one word that gets rot13'd?

wasm2c/examples/rot13/main.c
```diff
@@ -78,6 +83,8 @@ int main(int argc, char** argv) {
   /* Free the Wasm runtime state. */
   wasm_rt_free();
 
+  printf("Done!\n");
```
keithw, line 86: Not sure we need this if we're always either printing the output or a "usage" message...

shravanrn:
Ah, my bad. Fixed!

wasm2c/examples/rot13/main.c
shravanrn: Ah, my bad. Fixed!

wasm2c/examples/rot13/main.c
shravanrn: Removed!

PR #2031, opened 2022-10-31
Simplify part of type-checker. NFC

sbc100:
Simplify `CheckOpcode1` by avoiding using it for memory.fill.

Simplify and move GetMemoryParam.

### DIFF
--- a/src/type-checker.cc
+++ b/src/type-checker.cc
@@ -317,24 +317,22 @@ Result TypeChecker::PopAndCheck3Types(Type expected1,
   return result;
 }
 
-Result TypeChecker::CheckOpcode1(Opcode opcode,
-                                 const Limits* limits,
-                                 bool has_address_operands) {
-  Result result =
-      PopAndCheck1Type(opcode.GetMemoryParam(
-                           opcode.GetParamType1(), limits,
-                           has_address_operands || opcode.GetMemorySize() != 0),
-                       opcode.GetName());
-  PushType(has_address_operands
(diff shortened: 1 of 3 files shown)


sbc100:
I noticed this extra complexity while reviewing #2024

src/type-checker.cc
```diff
@@ -344,15 +349,11 @@ Result TypeChecker::CheckOpcode3(Opcode opcode,
...
-                                              has_address_operands),
-                        opcode.GetMemoryParam(opcode.GetParamType2(), limits2,
-                                              has_address_operands),
-                        opcode.GetMemoryParam(opcode.GetParamType3(), limits3,
-                                              has_address_operands),
-                        opcode.GetName());
+  Result result = PopAndCheck3Types(
+      GetMemoryParam(opcode.GetParamType1(), limits1, limits1 != nullptr),
+      GetMemoryParam(opcode.GetParamType2(), limits2, limits2 != nullptr),
+      GetMemoryParam(opcode.GetParamType3(), limits3, limits3 != nullptr),
```
keithw, line 355: This feels a bit weird for now (if each time the third argument is really just `secondarg != nullptr`, we could omit the third parameter to GetMemoryParam completely). But I guess it will need to make the decision in a more sophisticated way once this correctly implements memory.copy across different-type memories.

keithw (approved):
lgtm

src/type-checker.cc
sbc100: Hmm.. you are right.. I simplified some more.

PR #2032, opened 2022-11-01
Use a std::optional<Type> for checking memory arguments. NFC

adambratschikaye:
When type checking arguments to memory instructions we currently pass around the `Limits*` of the memory when all that's really needed is the `Type` of that memory. So this refactoring changes those `Limits*` arguments to `std::optional<Type>` instead.

### DIFF
--- a/include/wabt/type-checker.h
+++ b/include/wabt/type-checker.h
@@ -18,6 +18,7 @@
 #define WABT_TYPE_CHECKER_H_
 
 #include <functional>
+#include <optional>
 #include <vector>
 
 #include "wabt/common.h"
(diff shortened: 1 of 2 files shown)


sbc100:
Reading this diff now I'm not actually sure this step in the right direction.   I do see that it is limiting the information we pass around, but its also adding a little complexity to code, and creating more objects without a clear behnifit.

How would you feel about landing a version of #2024 that doesn't include this change.  Especially since #2031 landed I think it could be very small change perhaps?

adambratschikaye:
> How would you feel about landing a version of #2024 that doesn't include this change. Especially since #2031 landed I think it could be very small change perhaps?

Sure, no problem.

PR #2035, opened 2022-11-02
wasm2c: handle duplicate names across module fields

### LINKED ISSUE #2034: wasm2c doesn't handle duplicate debug names
wasm2c seems to have trouble if the module has duplicate debug names. E.g. this module:
```wat
(module
 (memory $x 0)
 (table $x 0 funcref)
)
```
processed with:
```
$ wat2wasm --debug-names -o test.wasm test.wat
$ wasm2c -o test.c test.wasm
```
correctly produces unique-ified names when it declares the elements:
```c
typedef struct Z_test_instance_t {
  wasm_rt_memory_t w2c_x;
  wasm_rt_funcref_table_t w2c_x_0;
} Z_test_instance_t;
```
but doesn't use the right names when accessing them:
```c
static void init_memories(Z_test_instance_t* instance) {
  wasm_rt_allocate_memory(&instance->w2c_x, 0, 65536);
}

static void init_tables(Z_test_instance_t* instance) {
  wasm_rt_allocate_funcref_table(&instance->w2c_x, 0, 4294967295);
}

void Z_test_free(Z_test_instance_t* instance) {
  wasm_rt_free_funcref_table(&instance->w2c_x);
  wasm_rt_free_memory(&instance->w2c_x);
}
```
This can also happen even if the module doesn't have debug names (because NameGenerator uses "e" for both element segments and exceptions, which maybe we should also change, but I think this would be harmless if wasm2c could handle the duplicate names whether they're generated or come from the name section).

keithw:
This change allows wasm2c to handle duplicate names across different module fields (e.g. an elem segment and tag both named "$e0"), by tracking the field type explicitly at each definition and then reference to a GlobalName.

There may be a less intensive way to do this... am not wedded to the approach, but this does seem to fix it.

~~Also fixes an issue in the BinaryReaderIR where tag names were being bound in a different bindings hash, which was leading to a segfault in wasm2c when a named tag was exported (because it couldn't find the tag by name).~~

Adds a regression test.

Fixes #2034.

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -477,7 +477,7 @@ char CWriter::MangleType(Type type) {
 constexpr char CWriter::MangleField(ModuleFieldType type) {
   assert(static_cast<std::underlying_type<ModuleFieldType>::type>(type) <
          std::numeric_limits<char>::max());
-  return static_cast<char>(type);
+  return 'a' + static_cast<char>(type);
 }
 
 // static
@@ -577,13 +577,6 @@ std::string_view StripLeadingDollar(std::string_view name) {
   return name;
 }
 
-std::string_view StripTrailingSigil(std::string_view name) {
-  if (!name.empty()) {
-    name.remove_suffix(1);
-  }
-  return name;
-}
-
 void CWriter::DefineImportName(const Import* import,
                                std::string_view module,
                                std::string_view field_name) {
@@ -625,8 +618,7 @@ void CWriter::DefineImportName(const Import* import,
 template <ModuleFieldType T>
 std::string CWriter::DefineGlobalScopeName(std::string_view name) {
   std::string mangled = std::string(name) + MangleField(T);
-  std::string unique = DefineName(
-      &global_syms_, StripTrailingSigil(StripLeadingDollar(mangled)));
+  std::string unique = DefineName(&global_syms_, StripLeadingDollar(name));
   bool success = global_sym_map_.emplace(mangled, unique).second;
   assert(success);
   return unique;
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -73,30 +73,27 @@ struct LocalName {
   const std::string& name;
 };
 
-template <ModuleFieldType>
 struct GlobalName {
-  explicit GlobalName(const std::string& name) : name(name) {}
+  GlobalName(ModuleFieldType type, const std::string& name)
+      : type(type), name(name) {}
+  ModuleFieldType type;
   const std::string& name;
 };
 
-template <ModuleFieldType T>
-struct ExternalPtr : GlobalName<T> {
-  using GlobalName<T>::GlobalName;
+struct ExternalPtr : GlobalName {
+  using GlobalName::GlobalName;
 };
 
-template <ModuleFieldType T>
-struct ExternalRef : GlobalName<T> {
-  using GlobalName<T>::GlobalName;
+struct ExternalRef : GlobalName {
+  using GlobalName::GlobalName;
 };
 
-template <ModuleFieldType T>
-struct ExternalInstancePtr : GlobalName<T> {
-  using GlobalName<T>::GlobalName;
+struct ExternalInstancePtr : GlobalName {
+  using GlobalName::GlobalName;
 };
 
-template <ModuleFieldType T>
-struct ExternalInstanceRef : GlobalName<T> {
-  using GlobalName<T>::GlobalName;
+struct ExternalInstanceRef : GlobalName {
+  using GlobalName::GlobalName;
 };
 
 struct GotoLabel {
@@ -241,16 +238,11 @@ class CWriter {
   void Write(Index);
   void Write(std::string_view);
   void Write(const LocalName&);
-  template <ModuleFieldType T>
-  void Write(const GlobalName<T>&);
-  template <ModuleFieldType T>
-  void Write(const ExternalPtr<T>&);
-  template <ModuleFieldType T>
-  void Write(const ExternalRef<T>&);
-  template <ModuleFieldType T>
-  void Write(const ExternalInstancePtr<T>&);
-  template <ModuleFieldType T>
-  void Write(const ExternalInstanceRef<T>&);
+  void Write(const GlobalName&);
+  void Write(const ExternalPtr&);
+  void Write(const ExternalRef&);
+  void Write(const ExternalInstancePtr&);
+  void Write(const ExternalInstanceRef&);
   void Write(Type);
   void Write(SignedType);
   void Write(TypeEnum);
@@ -706,48 +698,43 @@ void CWriter::Write(const LocalName& name) {
   Write(local_sym_map_[name.name]);
 }
 
-template <ModuleFieldType T>
-void CWriter::Write(const GlobalName<T>& name) {
-  std::string mangled = name.name + MangleField(T);
+void CWriter::Write(const GlobalName& name) {
+  std::string mangled = name.name + MangleField(name.type);
   assert(global_sym_map_.count(mangled) == 1);
   Write(global_sym_map_.at(mangled));
 }
 
-template <ModuleFieldType T>
-void CWriter::Write(const ExternalPtr<T>& name) {
+void CWriter::Write(const ExternalPtr& name) {
   bool is_import = import_syms_.count(name.name) != 0;
   if (!is_import) {
     Write("&");
   }
-  Write(GlobalName<T>(name.name));
+  Write(GlobalName(name));
 }
 
-template <ModuleFieldType T>
-void CWriter::Write(const ExternalInstancePtr<T>& name) {
+void CWriter::Write(const ExternalInstancePtr& name) {
   bool is_import = import_syms_.count(name.name) != 0;
   if (!is_import) {
     Write("&");
   }
-  Write("instance->", GlobalName<T>(name.name));
+  Write("instance->", GlobalName(name));
 }
 
-template <ModuleFieldType T>
-void CWriter::Write(const ExternalRef<T>& name) {
+void CWriter::Write(const ExternalRef& name) {
   bool is_import = import_syms_.count(name.name) != 0;
   if (is_import) {
-    Write("(*", GlobalName<T>(name.name), ")");
+    Write("(*", GlobalName(name), ")");
   } else {
-    Write(GlobalName<T>(name.name));
+    Write(GlobalName(name));
   }
 }
 
-template <ModuleFieldType T>
-void CWriter::Write(const ExternalInstanceRef<T>& name) {
+void CWriter::Write(const ExternalInstanceRef& name) {
   bool is_import = import_syms_.count(name.name) != 0;
   if (is_import) {
-    Write("(*instance->", GlobalName<T>(name.name), ")");
+    Write("(*instance->", GlobalName(name), ")");
   } else {
-    Write("instance->", GlobalName<T>(name.name));
+    Write("instance->", GlobalName(name));
   }
 }
 
@@ -798,7 +785,7 @@ void CWriter::Write(const LabelDecl& label) {
 
 void CWriter::Write(const GlobalInstanceVar& var) {
   assert(var.var.is_name());
-  Write(ExternalInstanceRef<ModuleFieldType::Global>(var.var.name()));
+  Write(ExternalInstanceRef(ModuleFieldType::Global, var.var.name()));
 }
 
 void CWriter::Write(const StackVar& sv) {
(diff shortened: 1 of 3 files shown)


keithw:
(FWIW, this doesn't seem to change the wasm2c output on modules where it worked already. I tried rebasing #2006 on top of this PR and it didn't change.)

sbc100:
Would it make sense to split out the `BinaryReaderIR` bug fix?  Can the regression test for that bug be written without wasm2c, or was it somehow only effecting wasm2c?  


Still reading the c-writer changes.. in general they seem quite complex for what I would hope would be simple fix, but maybe this is an inherently tricky problem..

src/c-writer.cc
```diff
@@ -678,50 +721,50 @@ void CWriter::Write(const LocalName& name) {
...
+template <ModuleFieldType T>
+void CWriter::Write(const ExternalInstancePtr<T>& name) {
   bool is_import = import_syms_.count(name.name) != 0;
-  if (is_import) {
-    Write("instance->", GetGlobalName(name.name));
-  } else {
-    Write("&instance->", GetGlobalName(name.name));
+  if (!is_import) {
+    Write("&");
   }
```
sbc100, line 747: How about splitting out the removal of `Deref` and `AddressOf` too.. that seems independent maybe?

keithw:
> Would it make sense to split out the `BinaryReaderIR` bug fix? Can the regression test for that bug be written without wasm2c, or was it somehow only effecting wasm2c?

Happy to split out the BinaryReaderIR bug fix, but I haven't been able to write a regression test that breaks the current code without wasm2c. (The other big user of BinaryReaderIR is WatWriter it doesn't try to look up a tag by name.)

> Still reading the c-writer changes.. in general they seem quite complex for what I would hope would be simple fix, but maybe this is an inherently tricky problem..

Agreed. :-( My other idea would be to try to have the NameGenerator (or BinaryReaderIR) guarantee global uniqueness of all names in the module, perhaps by appending things when necessary, but that seems even more intensive than these CWriter changes...

keithw:
Just pushed #2041 which does the easier parts of this one. Hopefully the remaining diff here (currently the second commit in this PR) will become simpler then.

src/c-writer.cc
keithw: Split out to #2041.

test/regress/regress-2034.txt
```diff
@@ -0,0 +1,17 @@
...
+;;; ARGS: --debug-names --enable-exceptions
+(module
+ (func $x)
+ (export "func" (func $x))
+ (global $x i32 (i32.const 14))
+ (table $x (export "tab") 0 funcref)
+ (elem $x)
+ (memory $x (export "mem") 0)
+ (data $x)
+ (tag $x (export "tag"))
```
sbc100, line 11: Do we really not have any spec tests that include elements of different types with the same name?   Its seems odd that we would need to write a wasm2c specific test for this behaviour.  Or is it because we don't run the spec tests with --debug-names enabled normally?

It does seem odd to be that this should even be allowed in the spec.  Wouldn't it be simple to force all names to be globally unique?

sbc100: Sadly it does looks like identifiers are only unique per-index-space: https://webassembly.github.io/spec/core/text/values.html#text-id

keithw: Sure, do you want that in the test or just here in GitHub?

wasm2c was already good about keeping the C identifiers unique (`DefineName` takes care of that), so the output ends up looking like this:

```c
typedef struct Z_regress_2034_0_wasm_instance_t {
  u32 w2c_foo;
  wasm_rt_memory_t w2c_foo_0;
  wasm_rt_funcref_table_t w2c_foo_1;
} Z_regress_2034_0_wasm_instance_t;
```

```
static u32 w2c_foo_4;

static void init_tags(void) {
  w2c_foo_4 = wasm_rt_register_tag(0);
}

static void w2c_foo_5(Z_regress_2034_0_wasm_instance_t*);

static void init_globals(Z_regress_2034_0_wasm_instance_t* instance) {
  instance->w2c_foo = 14u;
}


static void init_memories(Z_regress_2034_0_wasm_instance_t* instance) {
  wasm_rt_allocate_memory(&instance->w2c_foo_0, 0, 65536);
}

static void init_data_instances(Z_regress_2034_0_wasm_instance_t *instance) {
}

static void init_tables(Z_regress_2034_0_wasm_instance_t* instance) {
  wasm_rt_allocate_funcref_table(&instance->w2c_foo_1, 0, 4294967295);
}

static void init_elem_instances(Z_regress_2034_0_wasm_instance_t *instance) {
}

static void w2c_foo_5(Z_regress_2034_0_wasm_instance_t* instance) {
  FUNC_PROLOGUE;
  FUNC_EPILOGUE;
}

/* export: 'func' */
void Z_regress_2034_0_wasmZ_func(Z_regress_2034_0_wasm_instance_t* instance) {
  return w2c_foo_5(instance);
}

/* export: 'tab' */
wasm_rt_funcref_table_t* Z_regress_2034_0_wasmZ_tab(Z_regress_2034_0_wasm_instance_t* instance) {
  return &instance->w2c_foo_1;
}

/* export: 'mem' */
wasm_rt_memory_t* Z_regress_2034_0_wasmZ_mem(Z_regress_2034_0_wasm_instance_t* i
nstance) {
  return &instance->w2c_foo_0;
}

/* export: 'tag' */
const u32 *Z_regress_2034_0_wasmZ_tag = &w2c_foo_4;
```

src/c-writer.cc
```diff
@@ -547,35 +577,65 @@ std::string_view StripLeadingDollar(std::string_view name) {
...
-                                      std::string_view field_name) {
-  std::string mangled = MangleName(module) + MangleName(field_name);
-  import_syms_.insert(name);
-  global_syms_.insert(mangled);
-  global_sym_map_.insert(SymbolMap::value_type(name, mangled));
-  return mangled;
+std::string_view StripTrailingSigil(std::string_view name) {
+  if (!name.empty()) {
+    name.remove_suffix(1);
+  }
```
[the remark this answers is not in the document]
keithw: Done.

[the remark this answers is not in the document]
keithw: (and, method eliminated)

test/regress/regress-2034.txt
keithw: Or if you prefer the diff:

```diff
@@ -585,7 +585,7 @@
 static u32 w2c_foo_4;
 
 static void init_tags(void) {
-  w2c_foo = wasm_rt_register_tag(0);
+  w2c_foo_4 = wasm_rt_register_tag(0);
 }
 
 static void w2c_foo_5(Z_regress_2034_0_wasm_instance_t*);
@@ -596,41 +596,41 @@
 
 
 static void init_memories(Z_regress_2034_0_wasm_instance_t* instance) {
-  wasm_rt_allocate_memory(&instance->w2c_foo, 0, 65536);
+  wasm_rt_allocate_memory(&instance->w2c_foo_0, 0, 65536);
 }
 
 static void init_data_instances(Z_regress_2034_0_wasm_instance_t *instance) {
 }
 
 static void init_tables(Z_regress_2034_0_wasm_instance_t* instance) {
-  wasm_rt_allocate_funcref_table(&instance->w2c_foo, 0, 4294967295);
+  wasm_rt_allocate_funcref_table(&instance->w2c_foo_1, 0, 4294967295);
 }
 
 static void init_elem_instances(Z_regress_2034_0_wasm_instance_t *instance) {
 }
 
-static void w2c_foo(Z_regress_2034_0_wasm_instance_t* instance) {
+static void w2c_foo_5(Z_regress_2034_0_wasm_instance_t* instance) {
   FUNC_PROLOGUE;
   FUNC_EPILOGUE;
 }
 
 /* export: 'func' */
 void Z_regress_2034_0_wasmZ_func(Z_regress_2034_0_wasm_instance_t* instance) {
-  return w2c_foo(instance);
+  return w2c_foo_5(instance);
 }
 
 /* export: 'tab' */
 wasm_rt_funcref_table_t* Z_regress_2034_0_wasmZ_tab(Z_regress_2034_0_wasm_instance_t* instance) {
-  return &instance->w2c_foo;
+  return &instance->w2c_foo_1;
 }
 
 /* export: 'mem' */
 wasm_rt_memory_t* Z_regress_2034_0_wasmZ_mem(Z_regress_2034_0_wasm_instance_t* instance) {
-  return &instance->w2c_foo;
+  return &instance->w2c_foo_0;
 }
 
 /* export: 'tag' */
-const u32 *Z_regress_2034_0_wasmZ_tag = &w2c_foo;
+const u32 *Z_regress_2034_0_wasmZ_tag = &w2c_foo_4;
 
 void Z_regress_2034_0_wasm_init_module(void) {
   assert(wasm_rt_is_initialized());
@@ -650,6 +650,6 @@
 }
 
 void Z_regress_2034_0_wasm_free(Z_regress_2034_0_wasm_instance_t* instance) {
-  wasm_rt_free_funcref_table(&instance->w2c_foo);
-  wasm_rt_free_memory(&instance->w2c_foo);
+  wasm_rt_free_funcref_table(&instance->w2c_foo_1);
+  wasm_rt_free_memory(&instance->w2c_foo_0);
 }
```
```

Comment (keithw) [reply test/regress/regress-2034.txt]: > Do we really not have any spec tests that include elements of different types with the same name? Its seems odd that we would need to write a wasm2c specific test for this behaviour. Or is it because we don't run the spec tests with --debug-names enabled normally?

My current understanding is that:
- the spec tests don't always use symbolic identifiers at all (maybe in half of the tests?)
- when they do, they never reuse the same identifier to refer to different elements within a module (the wasm2c tests pass 100% even with `--debug-names` turned on)
- and even if they did, we wouldn't have caught it because we don't run the wasm2c spec tests with `--debug-names` enabled

If you want to turn on `--debug-names` be default for wasm2c testing, I wouldn't have any objection, although the real way I hit this bug was from names coming out of the NameGenerator (since it generates `$e0` both for tag #0 and for elem segment #0) so I think we'd lose some test coverage either way.

File: src/c-writer.cc [hunk 4]
```diff
@@ -547,35 +577,57 @@ std::string_view StripLeadingDollar(std::string_view name) {
...
   global_syms_.insert(mangled);
-  global_sym_map_.insert(SymbolMap::value_type(name, mangled));
-  return "(*" + mangled + ")";
+  bool success =
+      global_sym_map_.emplace(name + MangleField(type), mangled).second;
+  assert(success);
 }
 
+template <ModuleFieldType T>
 std::string CWriter::DefineGlobalScopeName(std::string_view name) {
```
sbc100, line 619: Does this need to be a template?  Can it not just take ModuleFieldType as a second argument?

src/c-writer.cc
```diff
@@ -67,18 +68,36 @@ struct Label {
...
   const std::string& name;
 };
 
-using LocalName = Name<0>;
-using GlobalName = Name<1>;
-using ExternalPtr = Name<2>;
-using ExternalRef = Name<3>;
-using ExternalInstancePtr = Name<4>;
-using ExternalInstanceRef = Name<5>;
+template <ModuleFieldType>
```
sbc100, line 76: What is the purpose of making this a template?   It doesn't seem to use the template parameter at all.

src/c-writer.cc
keithw: Sure. I guess I was thinking that the value is always known at compile-time, but it can be an argument just as easily. Done.

src/c-writer.cc
keithw: I wanted all the GlobalNames to be different types (templated on the index space / module field) so that `Write(const GlobalName<T>&)` could also be templated (based on the type of GlobalName) so it can look up the correctly suffixed name in the `global_sym_map_`.

If you prefer, we could just have every GlobalName include a member variable of type `ModuleFieldType`, and then `Write(const GlobalName&)` examines that member. Then there's no templates, at the cost of a tiny amount of runtime overhead.

sbc100: Ah I think I understand now.. I personally think I would find that form more readable.  I guess I'm slightly biased against template magic though.   What do you think?   I won't block this PR as it stands if you feel its worth the trade-off.

keithw: I really do not feel strongly. :-) Here it is the way you suggested.

sbc100 (approved):

src/c-writer.cc
```diff
@@ -177,6 +193,7 @@ class CWriter {
   void PopLabel();
 
   static char MangleType(Type);
+  static constexpr char MangleField(ModuleFieldType);
```
sbc100, line 196: Maybe add `constexpr` to `MangleType` for consistency here?

keithw: Sure, done.

PR #2036, opened 2022-11-03
Add .git-blame-ignore-revs

aheejin:
This adds a `.git-blame-ignore-revs` file, which was supported since version 2.23 (Aug 2019), which lets `git blame` ignore specified revisions. Usage examples:
https://github.com/llvm/llvm-project/blob/main/.git-blame-ignore-revs https://github.com/WebAssembly/binaryen/blob/main/.git-blame-ignore-revs

This currently contains a clang-format commit (#1684).


keithw (approved):
lgtm

PR #2037, opened 2022-11-03
Remove some consts for std::string_view

dimitry-:
Also convert a constant from char[] to constexpr std::string_view

Test: builds


dimitry-:
I am not sure what lint wants from me - it fails with "fatal: HEAD does not point to a branch", how do I fix it?

sbc100 (approved):

PR #2038, opened 2022-11-04
wasm2c: Fix 8gb model, signals, memory reserve/commits, stack depth on windows

shravanrn:
Correct handling of Wasm's 8gb model on Windows
- Uses the Windows ExceptionHandler to track access violations (segfaults)
- Fix Windows memory to reserve, but not commit memory up front
- Correct handling of return values in os_mmap functions
- Decouple stack depth handling and heap bounds checking

### DIFF
--- a/src/template/wasm2c.declarations.c
+++ b/src/template/wasm2c.declarations.c
@@ -1,16 +1,16 @@
 
 #define TRAP(x) (wasm_rt_trap(WASM_RT_TRAP_##x), 0)
 
-#if WASM_RT_MEMCHECK_SIGNAL_HANDLER
-#define FUNC_PROLOGUE
-
-#define FUNC_EPILOGUE
-#else
+#if WASM_RT_USE_STACK_DEPTH_COUNT
 #define FUNC_PROLOGUE                                            \
   if (++wasm_rt_call_stack_depth > WASM_RT_MAX_CALL_STACK_DEPTH) \
     TRAP(EXHAUSTION);
 
 #define FUNC_EPILOGUE --wasm_rt_call_stack_depth
+#else
+#define FUNC_PROLOGUE
+
+#define FUNC_EPILOGUE
 #endif
 
 #define UNREACHABLE TRAP(UNREACHABLE)
--- a/wasm2c/wasm-rt-impl.c
+++ b/wasm2c/wasm-rt-impl.c
@@ -25,7 +25,8 @@
 #include <stdlib.h>
 #include <string.h>
 
-#if WASM_RT_MEMCHECK_SIGNAL_HANDLER_POSIX && !WASM_RT_SKIP_SIGNAL_RECOVERY
+#if WASM_RT_MEMCHECK_SIGNAL_HANDLER && !WASM_RT_SKIP_SIGNAL_RECOVERY && \
+    !defined(_WIN32)
 #include <signal.h>
 #include <unistd.h>
 #endif
@@ -48,8 +49,14 @@ typedef struct FuncType {
 
 #if WASM_RT_MEMCHECK_SIGNAL_HANDLER && !WASM_RT_SKIP_SIGNAL_RECOVERY
 static bool g_signal_handler_installed = false;
-static char* g_alt_stack;
+#ifdef _WIN32
+static void* g_sig_handler_handle = 0;
 #else
+static char* g_alt_stack = 0;
+#endif
+#endif
+
+#if WASM_RT_USE_STACK_DEPTH_COUNT
 uint32_t wasm_rt_call_stack_depth;
 uint32_t wasm_rt_saved_call_stack_depth;
 #endif
@@ -67,7 +74,7 @@ static jmp_buf* g_unwind_target;
 
 void wasm_rt_trap(wasm_rt_trap_t code) {
   assert(code != WASM_RT_TRAP_NONE);
-#if !WASM_RT_MEMCHECK_SIGNAL_HANDLER
+#if WASM_RT_USE_STACK_DEPTH_COUNT
   wasm_rt_call_stack_depth = wasm_rt_saved_call_stack_depth;
 #endif
   WASM_RT_LONGJMP(wasm_rt_jmp_buf, code);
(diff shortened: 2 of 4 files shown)


shravanrn:
@sbc100 I tested that the tests pass locally on windows. Any chance you can help me figure out if the CI tests are failing for some unrelated reason?

shravanrn:
@sbc100 Updated with a simple implementation that passes all checks

wasm2c/wasm-rt.h
```diff
@@ -56,35 +52,53 @@ extern "C" {
  * This is usually 10%-25% faster, but requires OS-specific support.
  */
 
-/** Check whether the signal handler is supported at all. */
-#if (defined(__linux__) || defined(__unix__) || defined(__APPLE__)) && \
-    defined(__WORDSIZE) && __WORDSIZE == 64
```
sbc100, line 61: Why not keep this check as is?

src/template/wasm2c.declarations.c
```diff
@@ -1,7 +1,7 @@
 
 #define TRAP(x) (wasm_rt_trap(WASM_RT_TRAP_##x), 0)
 
-#if WASM_RT_MEMCHECK_SIGNAL_HANDLER
+#if !WASM_RT_USE_STACK_DEPTH_COUNT
```
sbc100, line 4: Maybe flip the if/else bodies around do that this can be written as a positive check

wasm2c/wasm-rt.h
```diff
@@ -56,35 +52,53 @@ extern "C" {
...
+#ifndef WASM_RT_SKIP_SIGNAL_RECOVERY
+#define WASM_RT_SKIP_SIGNAL_RECOVERY 0
+#endif
```
sbc100, line 57: Why move these three lines down?

sbc100 (commented):
This looks awesome!

Do I assume we are testing these windows signal handlers as part of our CI when we run the wasm2c tests on the windows bot?

sbc100:
Is the `Uses the Windows try/catch SEH exception to track access violations (segfaults)` part of the description our of date?  Or am I not understanding the try/catch thing.

wasm2c/wasm-rt.h
shravanrn: The `(defined(__linux__) || defined(__unix__) || defined(__APPLE__))` is no longer needed, as we support both POSIX and windows. The `defined(__WORDSIZE) && __WORDSIZE == 64` approach to testing 64-bit architectures does not work on msvc (https://godbolt.org/z/af3E66s6G)

shravanrn:
> Do I assume we are testing these windows signal handlers as part of our CI when we run the wasm2c tests on the windows bot?

This should be the case. The signal handler is on by default on 64-bit windows. From what I can see, the CI on Windows targets amd64.

> Is the Uses the Windows try/catch SEH exception to track access violations (segfaults) part of the description our of date? Or am I not understanding the try/catch thing.

My bad. This is old. Will push a fix now.

src/template/wasm2c.declarations.c
shravanrn: will fix

wasm2c/wasm-rt-impl.c
```diff
@@ -227,67 +258,84 @@ static int os_mprotect(void* addr, size_t size) {
...
 
-    /* Use alt stack to handle SIGSEGV from stack overflow */
-    g_alt_stack = malloc(SIGSTKSZ);
-    if (g_alt_stack == NULL) {
-      perror("malloc failed");
-      abort();
-    }
+static void os_install_signal_handler(void) {
+  /* Use alt stack to handle SIGSEGV from stack overflow */
+  g_alt_stack = malloc(SIGSTKSZ);
```
sbc100, line 273: Unrelated to this change (we can maybe do it as a followup) but is there any reason we can't just declare this as static data `e.g. static char g_alt_stack[SIGSTKSZ];`  ?

wasm2c/wasm-rt-impl.c
```diff
@@ -206,6 +212,31 @@ static void os_print_last_error(const char* msg) {
     printf("%s. No error code.\n", msg);
   }
 }
+
+#if WASM_RT_MEMCHECK_SIGNAL_HANDLER && !WASM_RT_SKIP_SIGNAL_RECOVERY
+
+static LONG os_signal_handler(PEXCEPTION_POINTERS pExceptionInfo) {
+  if (pExceptionInfo->ExceptionRecord->ExceptionCode ==
```
sbc100, line 219: I suppose this is windows-only code, but maybe we should use the wabt naming convension here?   i.e. call this argument just `info` maybe?   (that might allow the conditions below to fit on a single line too).

sbc100 (approved):
Nice!

wasm2c/wasm-rt-impl.c
shravanrn: My worry is that this would unconditionally take up space even for implementations that do not use the signal based recovery. (I have some similar concerns with the current exception handler implementation). The current approach gives us a "pay for what you use" model which is definitely preferable for the Firefox usecase, as we won't use the signal handler in wasm2c

wasm2c/wasm-rt-impl.c
shravanrn: will fix

wasm2c/wasm-rt-impl.c
sbc100: But can't we just boot the declaration behind `WASM_RT_MEMCHECK_SIGNAL_HANDLER && !WASM_RT_SKIP_SIGNAL_RECOVERY` just like these functions are?

shravanrn: oh right. I just remembered there was a breaking change on this in the latest glibc also. 
https://sourceware.org/git/?p=glibc.git;a=blob;f=NEWS;h=85e84fe53699fe9e392edffa993612ce08b2954a;hb=HEAD

SIGSTKSZ is no longer a constant and is redefined as `SIGSTKSZ is redefined to sysconf (_SC_SIGSTKSZ)`. This means you can't have static array definitions of the form `static char g_alt_stack[SIGSTKSZ]` 

Not sure but I think this new glibc is in use in Ubuntu 22.10 (I'm still testing on Ubuntu 22.04LTS so I can't confirm)

sbc100: Ah .. annoying.

PR #2041, opened 2022-11-05
c-writer.cc: Eliminate GetGlobalName (NFC)

keithw:
Require all resolution of GlobalNames to happen in one place (Write(GlobalName)). Also eliminate AddressOf, Deref, GlobalVar.

This is a step towards fixing #2034 (split out from #2035).

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -655,10 +655,7 @@ void CWriter::Write(const LocalName& name) {
 
 void CWriter::Write(const GlobalName& name) {
   // TODO(keithw) deal with duplicates across module fields
-  assert(global_sym_map_.count(name.name) == 1);
-  auto iter = global_sym_map_.find(name.name);
-  assert(iter != global_sym_map_.end());
-  Write(iter->second);
+  Write(global_sym_map_.at(name.name));
 }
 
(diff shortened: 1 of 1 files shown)


src/c-writer.cc
```diff
@@ -673,50 +653,45 @@ void CWriter::Write(const LocalName& name) {
...
-  return iter->second;
-}
-
 void CWriter::Write(const GlobalName& name) {
-  Write(GetGlobalName(name.name));
+  // TODO(keithw) deal with duplicates across module fields
+  assert(global_sym_map_.count(name.name) == 1);
+  auto iter = global_sym_map_.find(name.name);
+  assert(iter != global_sym_map_.end());
+  Write(iter->second);
```
sbc100, line 661: Would it be enough to just do `Write(global_sym_map_.at(name.name))` should be enough here.. no need for `count` or `find` maybe?

I guess this is existing code though, so might not make sense as part of this change.

sbc100 (approved):

src/c-writer.cc
keithw: Sure, that works fine for now. We can worry about how much defensive programming is worthwhile in the next one.

PR #2043, opened 2022-11-07
wasm2c: Uninstall sighandler in os_cleanup_signal_handler

sbc100:


shravanrn:
Nice, I had on my todo list to submit this exact PR sometime today :)

wasm2c/wasm-rt-impl.c
```diff
@@ -296,6 +296,19 @@ static void os_install_signal_handler(void) {
...
 static void os_cleanup_signal_handler(void) {
+  /* Undo what was done in os_install_signal_handler */
+  struct sigaction sa;
+  sa.sa_handler = SIG_DFL;
+  if (sigaction(SIGSEGV, &sa, NULL) != 0 || sigaction(SIGBUS, &sa, NULL)) {
+    perror("sigaction failed");
+    abort();
+  }
+
+  if (sigaltstack(NULL, NULL) != 0) {
```
shravanrn, line 307: I think I have ignored any errors when we remove the signal handler on windows. We should probably get this consistent in the future, but explicitly erroring out is probably the better way to go :+1:

shravanrn (approved):

PR #2044, opened 2022-11-07
wasm2c: Avoid malloc/free for signal handling stack. NFC

sbc100:


sbc100:
This static data should only be part of the binary if `os_install_signal_handler` is used, so folks who opt out of signal handling should not be effected.

sbc100:
Apparently this doesn't work on newer glibc because SIGSTKSZ is not longer constant.

PR #2045, opened 2022-11-07
wasm2c: cleanup function type indexes when calling wasm_rt_free

shravanrn:


wasm2c/wasm-rt-impl.c
```diff
@@ -333,6 +333,15 @@ bool wasm_rt_is_initialized(void) {
...
 
 void wasm_rt_free(void) {
+  for (uint32_t i = 0; i < g_func_type_count; ++i) {
+    free(g_func_types[i].params);
+    free(g_func_types[i].results);
+  }
+
+  g_func_type_count = 0;
+  free(g_func_types);
+  g_func_types = 0;
```
sbc100, line 343: `= NULL` ?

sbc100 (approved):

sbc100 (approved):

sbc100:
Makes me think we should be running these tests with sanitizer enabled maybe?

PR #2047, opened 2022-11-08
wasm2c: Avoid unnecessary heap allocation in wasm_rt_register_func_type. NFC

sbc100:


sbc100:
I noticed this potential optimization when investigating the use of global data in this function.

keithw (approved):
lgtm if the tests pass. I imagine any "production" runtime is going to want a much more sophisticated implementation of `wasm_rt_register_func_type` (e.g. probably you want to be able to GC an individual module, and maybe you need it to be thread-safe, and probably other things I'm not thinking of).

shravanrn:
Windows is a bit finicky with alloca iirc. Looks like its `_alloca` on Windows per [here](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/alloca?view=msvc-170)
Also, the page says its deprecated, so that worries me a little. That said, Windows is unlikely to remove it.

Did a quick test on godbolt. `alloca` seems to return an int, but `_alloca` returns a void* as expected
https://godbolt.org/z/ca49Wdr38

sbc100:
> Windows is a bit finicky with alloca iirc. Looks like its `_alloca` on Windows per [here](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/alloca?view=msvc-170) Also, the page says its deprecated, so that worries me a little. That said, Windows is unlikely to remove it.
> 
> Did a quick test on godbolt. `alloca` seems to return an int, but `_alloca` returns a void* as expected https://godbolt.org/z/ca49Wdr38

Looks like we already depend on windows alloca elsewhere in wabt:
https://github.com/WebAssembly/wabt/blob/ac20a437ada3afc67b2ed3a4fedc335c8201a7f3/src/config.h.in#L58-L60.  So I think I can just do the same here.

keithw (approved):
lgtm % this comment:

FWIW in wasm2c.declarations.c we also define `alloca` to be `_alloca` if `_MSC_VER` is set. Here we're doing it if `_WIN32` is set. I'm not sure which is correct but probably we should be using the same macro in each place?

(Maybe if somebody is using GCC or clang on Win32 they get one of these macros and not the other? In which case, do they get `alloca` or `_alloca`? I have no idea...)

sbc100:
> lgtm % this comment:
> 
> FWIW in wasm2c.declarations.c we also define `alloca` to be `_alloca` if `_MSC_VER` is set. Here we're doing it if `_WIN32` is set. I'm not sure which is correct but probably we should be using the same macro in each place?
> 
> (Maybe if somebody is using GCC or clang on Win32 they get one of these macros and not the other? In which case, do they get `alloca` or `_alloca`? I have no idea...)

I made it consistent across the 3 places we define this.

PR #2048, opened 2022-11-08
Some minor code cleanups in c-writer.cc

dimitry-:
Use `const auto&` in for loops where applicable.
Remove usages of `module_->imports` outside of BeginInstance
Replace `if (vec.size())` with `if (!vec.empty())`

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -1221,7 +1221,7 @@ void CWriter::BeginInstance() {
 
 // Write module-wide imports (funcs & tags), which aren't tied to an instance.
 void CWriter::WriteImports() {
-  if (module_->imports.empty())
+  if (unique_imports_.empty())
     return;
 
   Write(Newline());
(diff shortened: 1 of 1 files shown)


src/c-writer.cc
```diff
@@ -1427,19 +1427,20 @@ void CWriter::WriteDataInitializers() {
...
-      Write(Newline(), "static const u8 data_segment_data_",
-            GlobalName(data_segment->name), "[] = ", OpenBrace());
-      size_t i = 0;
-      for (uint8_t x : data_segment->data) {
-        Writef("0x%02x, ", x);
-        if ((++i % 12) == 0)
-          Write(Newline());
-      }
-      if (i > 0)
+    if (data_segment->data.empty()) {
```
keithw, line 1430: Maybe keep this `if (!data_segment->data.empty())` to preserve the current control flow? The `if (empty) { continue; }` pattern seems more error-prone long-term.

keithw (commented):
Thank you! lgtm % comment

src/c-writer.cc
sbc100: I could go either way here. I tend to prefer early-return/early-continue over indenting an entire block,  but also this part of the change seems different to the rest.

sbc100: In this file at least we seem to be using single line block without curly braces.  We are not consistent about in in wabt right now, but at least in this file it seems very common.   (IIUC the google/chromium style guide allows both).

keithw (approved):

src/c-writer.cc
keithw: Sounds good -- it doesn't seem worth holding this PR up for style issues if we don't have strong (or the same) thinking on the desired style anyway. Will plan to merge now...

Issue #2018, opened 2022-10-11
wast2json fails to parse tail call proposal test file


osa1:
Trying with eca3519d (current main branch), wast2json can parse [return_call.wast](https://github.com/WebAssembly/testsuite/blob/4f77306bb63151631d84f58dedf67958eb9911b9/proposals/tail-call/return_call.wast), but it cannot parse [return_call_indirect.wast](https://github.com/WebAssembly/testsuite/blob/4f77306bb63151631d84f58dedf67958eb9911b9/proposals/tail-call/return_call_indirect.wast). Output:

```
$ ~/wasm/wabt/build/wast2json --enable-all return_call_indirect.wast
return_call_indirect.wast:160:38: error: unexpected token "type", expected an expr.
      (then (return_call_indirect 1 (type $out-i32) (i32.const 0)))
                                     ^^^^
return_call_indirect.wast:48:7: error: undefined function variable "$fac"
      $fac $fac-acc $even $odd
      ^^^^
return_call_indirect.wast:48:12: error: undefined function variable "$fac-acc"
      $fac $fac-acc $even $odd
           ^^^^^^^^
return_call_indirect.wast:48:21: error: undefined function variable "$even"
      $fac $fac-acc $even $odd
                    ^^^^^
return_call_indirect.wast:48:27: error: undefined function variable "$odd"
      $fac $fac-acc $even $odd
                          ^^^^
```

sbc100:
Hmm, it looks like maybe wabt doesn't support tail-call as much as the README claims its does.

In particular it looks like the we don't run the spec tests for tail-call: https://github.com/WebAssembly/wabt/blob/eca3519d738bcd245d46915e9c373c6d58f5b20b/test/update-spec-tests.py#L100

I guess there is some work here to get it all working.

PR #2049, opened 2022-11-08
Add tail-call spec tests and fix related parser issue

sbc100:
Fixes: #2018


keithw (approved):
Wow, I'm glad the fix was so small.

PR #2052, opened 2022-11-11
Honor CC environment variable in test/run-spec-wasm2c.py

sbc100:
I've been having some issues with the recent version of gcc on my system and being able to run `CC=/usr/bin/clang ./test/run-tests.py` has proved useful.

### DIFF
--- a/test/run-spec-wasm2c.py
+++ b/test/run-spec-wasm2c.py
@@ -426,6 +426,7 @@ def main(args):
     default_compiler = 'cc'
     if IS_WINDOWS:
         default_compiler = 'cl.exe'
+    default_compiler = os.getenv('CC', default_compiler)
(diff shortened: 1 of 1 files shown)


keithw (approved):
lgtm.

(Wonder if it might be useful to support CFLAGS too at some point -- I sometimes find myself editing the script to add '-g' when debugging a new feature...)

sbc100:
> lgtm.
> 
> (Wonder if it might be useful to support CFLAGS too at some point -- I sometimes find myself editing the script to add '-g' when debugging a new feature...)

sgtm.  Since its test-only code we can experiment and see what works for us.

keithw:
(Update: I forgot we had WASM2C_CFLAGS which already does this.)

PR #2053, opened 2022-11-11
wasm2c: Use GetReferenceNullValue() helper. NFC

sbc100:
I could do the same for FuncRef above.. but that would turn `isNull` on FuncRef from a one 1-word check to a 3-word check.  Maybe not worth it?


keithw (approved):
lgtm. And agreed, I think fine to keep the Type::FuncRef case as-is. (I guess we could add an explanatory comment if we really wanted...)

Issue #2022, opened 2022-10-24
wat2wasm: incorrectly produces funcref from externref element


mathetake:
I noticed this during catching up with [the latest spec test](https://github.com/WebAssembly/spec/commit/f2b7c60c6726d0857f137d9243107a74f8538668)

Steps to repro:

```
$ cat original.wat             

(module
  (import "exporter" "table" (table $t 2 externref))
  (elem (i32.const 0) externref (ref.null extern)))

$ wasm-tools parse original.wat -o wt.wasm && wasm-tools print wt.wasm -o wt.wat
$ cat wt.wat
(module
  (import "exporter" "table" (table $t 2 externref))
  (elem (;0;) (i32.const 0) externref (ref.null extern)))


$ ~/wabt-1.0.30/bin/wat2wasm original.wat -o wabt.wasm && wasm-tools print wabt.wasm -o wabt.wat
$ cat wabt.wat
(module
  (import "exporter" "table" (table (;0;) 2 externref))
  (elem (;0;) (i32.const 0) funcref (ref.null extern)))
```

So it seems like `wat2wasm` produces the element with the initial byte `0x4` which is fixed to `funcref`.

keithw:
Closing as fixed by #2054.

PR #2054, opened 2022-11-11
Update testsuite

sbc100:
As well as the testsuite update there are two notable changes that come with it here.  These can both be split out an landed first if it makes sense.

1. wasm2c now supports element sections containing externref.  Currently only the null reference is supported.

2. element segments no longer use funcref as the default element type but instead, unless explicitly included in the binary, the element type defaults to the type of the table in which the segment is active.

Fixes: #1612 #2022


src/binary-reader.cc
```diff
@@ -2678,6 +2677,7 @@ Result BinaryReader::ReadElemSection(Offset section_size) {
     }
 
     // For backwards compat we support not declaring the element kind.
+    Type elem_type = Type::Void;
```
keithw, line 2680: Huh, I'm surprised we need to parse these as Type::Void element segments that later get auto-resolved (in the BinaryReaderIr and BinaryReaderInterp and SharedValidator) by looking at the table type. Is there a test case (or other case involving pre-reference-types modules) that requires this...? I don't quite understand the situation where this is needed.


From the current spec it seems like there are exactly 8 options, and every choice either implicitly indicates funcref (0 and 4), explicitly encodes elemkind which must be funcref (1 through 3), or explicitly encodes reftype (5-7). When is it implicit?

src/c-writer.cc
```diff
@@ -1510,8 +1510,18 @@ void CWriter::WriteElemInitializers() {
       continue;
     }
 
-    Write("static const wasm_elem_segment_expr_t elem_segment_exprs_",
-          GlobalName(elem_segment->name), "[] = ", OpenBrace());
+    switch (elem_segment->elem_type) {
+      case Type::FuncRef:
+        Write("static const wasm_elem_segment_expr_t elem_segment_exprs_",
```
keithw, line 1515: Should we rename `wasm_elem_segment_expr_t` now that we're supporting elem segment initialization exprs of both types (and this one is only for funcrefs)? Maybe `wasm_funcref_expr_t` ? Or too much unnecessary churn?

keithw (commented):
Looks awesome. I don't quite understand what modules will need to have the Type::Void elem types with later resolution to the table's type (please see comment).

keithw:
FWIW I think this will close #1612 and #2022.

src/binary-reader.cc
sbc100: I'm not exactly sure what the history is here, but there seems to be circumstances under which elem_type is not read from the binaryen.  Currently we only read and element type `if (flags & (SegPassive | SegExplicitIndex)) {`.

There are certainly tests that fail if we assume FuncRef like the old code did... but I don't have a precise description to hand.

sbc100: Specifically `elem.wast` in the spec tests fails if we default to FuncRef here... investigating more.

keithw: I think the issue here may be in the writer side, specifically wast2json (and ultimately ir.cc ElemSegment::GetFlags). Any externref-typed segment needs to have SegUseElemExprs **and** SegExplicitIndex both set, or else the type is implicitly funcref. (Maybe we should rename `SegUseElemExprs` to `SegUseElemTypeAndExprs` to be clearer...)

src/c-writer.cc
sbc100: Hmm yes!  Thanks for pointing that out..

src/binary-reader.cc
sbc100: You were right.. much nicer fix now..

keithw (approved):
lgtm. The MSVC thing is annoying; I don't have any great ideas other than making it a preprocessor macro, which I guess is not the end of the world.

PR #2055, opened 2022-11-11
Convert type checks to assertions in the interpreter. NFC

sbc100:
All of these checks represent cases where a validation error would prevent the type mismatch.

When debugging #2054 this check actually worked against since it was resulting a false-positive "out-of-bound" error reports when really it was an internal type inconsistency (a bug).

### DIFF
--- a/src/interp/interp.cc
+++ b/src/interp/interp.cc
@@ -208,6 +208,7 @@ Store::Store(const Features& features) : features_(features) {
   roots_.New(ref);
 }
 
+#ifndef NDEBUG
 bool Store::HasValueType(Ref ref, ValueType type) const {
   // TODO opt?
   if (!IsValid(ref)) {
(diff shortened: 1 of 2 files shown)


src/interp/interp.cc
```diff
@@ -208,6 +208,7 @@ Store::Store(const Features& features) : features_(features) {
   roots_.New(ref);
 }
 
+#ifndef NDEBUG
```
keithw, line 211: Not sure I understand the purpose of wrapping this function in NDEBUG if it's only getting called inside `assert` anyway.

src/interp/interp.cc
```diff
@@ -208,6 +208,7 @@ Store::Store(const Features& features) : features_(features) {
...
 bool Store::HasValueType(Ref ref, ValueType type) const {
   // TODO opt?
```
keithw, line 213: Looks like the TODO happened! :-)

keithw (approved):
lgtm % comment assuming tests pass

src/interp/interp.cc
sbc100: Just so it clear to the reader that this is only needed/used in debug builds.

keithw: :+1:

PR #2058, opened 2022-11-12
Install wasm2c source files as data

### LINKED ISSUE #2057: Include wasm-rt* source files in binary release packages
When using `wasm2c`, the `wasm-rt-*` files are necessary to compile/link the resulting program. The WABT binary release packages come with a compiled version of wasm-rt as a static library, but this may not match the target platform for which the `wasm2c` result is used. 

I therefore propose to include the necessary wasm2c runtime files (`wasm-rt-impl.c`, `wasm-rt-impl.h`, and `wasm-rt.h`) in the binary package as source under a `src/wasm2c` subdir. This gives toolchains that use wasm2c the flexibility to compile the wasm2c output for any platform.

Note that the Debian binary package of `wabt` already packages the wasm-rt source files in the WABT package under `/usr/src/wasm2c` (see [the filelist](https://packages.debian.org/sid/amd64/wabt/filelist)).

remko:
Toolchains using wasm2c depend on the wasm-rt files to build the generated programs. Using a packaged binary wasm-rt library is not enough for toolchains that do e.g. cross-compilation, and they need to copy and use the source files instead. Therefore, package the required wasm-rt source files as data for toolchains to use.

Even though some files are already installed in `include`, I deliberately chose to install all the required headers and source files in the data dir, to ensure it's clear these belong together as data for toolchains. I believe it is unusual and confusing that a toolchain should copy some data from the platform's `include` dir and assume it is platform-independent, as this dir should only contain platform-specific headers used for compiling software for the host platform. This is also consistent with e.g. the [Debian package](https://packages.debian.org/sid/amd64/wabt/filelist), which also packages the wasm-rt source files together (in addition to the headers being present in the `include` dir, for tools that use the packaged binary wasm-rt library instead).

Closes #2057.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -393,8 +393,13 @@ IF (NOT WIN32)
       INCLUDES DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
     )
     install(
-      FILES "wasm2c/wasm-rt-impl.h" "wasm2c/wasm-rt.h"
+      FILES "wasm2c/wasm-rt.h"
       TYPE INCLUDE
+      COMPONENT wabt-development
+    )
+    install(
+      FILES "wasm2c/wasm-rt-impl.h" "wasm2c/wasm-rt-impl.c"
+      DESTINATION "${CMAKE_INSTALL_DATADIR}/wabt/wasm2c"
       COMPONENT wabt-development
     )
   endif ()


CMakeLists.txt
```diff
@@ -397,6 +397,11 @@ IF (NOT WIN32)
       TYPE INCLUDE
       COMPONENT wabt-development
     )
+    install(
+      FILES "wasm2c/wasm-rt-impl.h" "wasm2c/wasm-rt.h" "wasm2c/wasm-rt-impl.c"
```
sbc100, line 401: Foes it makes sense to install the headers both in the include path *and* in the wasm2c data path?   Can we just pick one maybe?

remko: @sbc100 I asked myself that question. 

There are 3 approaches a tool can take with wasm2c output:

1. Use the distribution's binary wasm-rt library. This is for tools (or users) that are happy to only compile wasm2c's output for the target platform, and want to only use the same/compatible flags for their program that the distribution used for compiling the wasm-rt files.
2. Use the wasm-rt source files. This is for tools that want to control the compiler and compilation flags for the wasm2c output, want to create a source code package of the output to distribute or use downstream, and/or don't want to depend on the wabt distribution being present for their operation.
3. Use its own wasm-rt implementation.

Given that, where to install `wasm-rt-impl.h` and `wasm-rt.h` (which are currently distributed in `include`)?

- `wasm-rt-impl.h`: This file is only necessary to compile `wasm-rt-impl.c`. By that reasoning, it doesn't belong in the include dir, and should only be in the data dir. However, note that by that reasoning, this file shouldn't exist at all and should be part of `wasm-rt-impl.c`, so I'm not sure why the split was made? Maybe to be able to reuse the 2-3 convenience macros of it in case (3) for your own implementation, although that sounds unlikely to happen (as this makes your generated source depend on a wabt installation just for a couple of macros, and drags in some externals into your implementation too).

- `wasm-rt.h`:
   - If you're in case 1, you just want to call your compiler on the resulting sources and the platform-dependent binary wabt-rt library. So, `wasm-rt.h` belongs in the include dir.
   - If you're in case 2 or 3, `wasm-rt.h` is used as a platform-independent data file, and is copied for your own use (either for further distribution, or for calling your own compiler with your own flags). You could technically copy the file from the `include` dir, but I believe it is unusual and confusing, since you expect this dir to contain (potentially platform-dependent) header files for compiling against installed libraries. 
   
  Since these are different intended use cases (with different requirements) for `wasm-rt.h`, I think the `wasm-rt.h` header should live in both locations, even though it is the same file. Looking at [the Debian package listing](https://packages.debian.org/sid/amd64/wabt/filelist), it seems Debian also follows this reasoning and has the file in 2 places, and since they have very strict packaging policies, I assume they thought there was a good reason to.

sbc100: Ok, I think I understand now. 

Why can't we put `wasm-rt.h` in the include directory (where it should be accessible in both 1, and 2 cases

And for wasm-rt-impl.h I agree it should probably not be in the include directory at all and shipped only in the data directory.

As to why wasm-rt-impl.h exists at all, I'm not sure of the history there, but it also seems reasonable for a project to split stuff into a header, even if its only used in a single source file.  At least, it seems like a decision that package maintainers shouldn't need to be concened with.  Perhaps we can separately consider merging?

sbc100: i.e. can we put the impl header in the data dir and public header in the header dir?

sbc100 (approved):

CMakeLists.txt
remko: @sbc100 The reason I didn't do it is because the purposes are different, and copying data from the include dir (which is only supposed to include headers necessary for compiling against libraries in lib/) is not something that is usually asked from tools.

But I updated the PR to do as you suggest.

PR #2061, opened 2022-11-14
c-writer.cc: replace TODO with assert now that validator catches case

keithw:


sbc100 (approved):

src/c-writer.cc
```diff
@@ -2536,11 +2536,7 @@ void CWriter::Write(const ExprList& exprs) {
         const ElemSegment* src_segment =
             module_->GetElemSegment(inst->segment_index);
 
-        // TODO: This should be an assert because the validator should catch
-        // this, but currently it doesn't.
-        if (dest_table->elem_type != src_segment->elem_type) {
```
shravanrn, line 2541: Fyi, this converts a release assert into a debug assert. Is this by design?

sbc100: Yes, I think so.  The reason is that the validator should already ensure this.  Checking it again in debug builds maybe makes sense just for a sanity check (and as an aid to the reader), but it should be strictly unnecessary.

PR #2062, opened 2022-11-14
wasm2c: allow externref initialization when externref type is not void*

keithw:
Here's one approach to keep externref initializers agnostic to the type of externref while keeping MSVC happy about only using constant expressions in the initializers. (We use `wasm_rt_externref_t` typedef'd to a non-pointer type, so `NULL` doesn't typecheck in that setting.)

An alternative could be to change `wasm_rt_externref_null_value` to a `#define` (instead of a `static const`) but this seems maybe preferable...?

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -272,6 +272,7 @@ class CWriter {
   void WriteGlobalInitializers();
   void WriteDataInitializers();
   void WriteElemInitializers();
+  void WriteElemTableInit(bool, const ElemSegment*, const Table*);
   void WriteExports(WriteExportsKind);
   void WriteInitDecl();
   void WriteFreeDecl();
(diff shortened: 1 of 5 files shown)


src/c-writer.cc
```diff
@@ -1516,13 +1516,20 @@ void CWriter::WriteElemInitializers() {
               GlobalName(elem_segment->name), "[] = ", OpenBrace());
         break;
       case Type::ExternRef:
-        Write("static const wasm_rt_externref_t elem_segment_exprs_",
```
sbc100, line 1519: Why even declare this at all if its not going to be used?

src/template/wasm2c.declarations.c
```diff
@@ -494,20 +494,20 @@ static inline void funcref_table_init(wasm_rt_funcref_table_t* dest,
   }
 }
 
-// Currently we only support initializing externref tables with ref.null.
+// Currently Wasm only supports initializing externref tables with ref.null.
 static inline void externref_table_init(wasm_rt_externref_table_t* dest,
-                                        const wasm_rt_externref_t* src,
+                                        const bool* unused,
```
sbc100, line 499: Why not just drop the second argument for now?

sbc100 (approved):
Seems reasonable for now.

src/c-writer.cc
keithw: Reworked to eliminate this.

src/template/wasm2c.declarations.c

PR #2071, opened 2022-11-16
interp: Remove condition for performing type checks on `Value`

### LINKED ISSUE #2069: interp: Value conditional struct field not present in release build
I'm trying to track down an issue with a modified version of `wasm-interp.cc` I'm using. It's not clear whether it's a bug on my end, or a bug in interp, so I hope someone can point me to what's going wrong. I tried to strip down the interpreter to the minimal reproducing program.

I have the following WebAssembly program:

```wasm
(import "shell" "read" (func $shell_read (param i32 i32)))
(func $run (export "run")  
  (call $shell_read (i32.const 0x8) (i32.const 0x9)))
 ```

I try to run this using the following interpreter:

```c++
namespace wabti = wabt::interp;

wabt::Result run() {
  static wabti::Store store;
  static wabt::Features features;
  static wabt::Errors errors;
  wabti::ModuleDesc desc;
  CHECK_RESULT(wabti::ReadBinaryInterp("input.wasm", input, sizeof(input), 
        wabt::ReadBinaryOptions(features, nullptr, true, true, true), &errors, &desc));
  auto core = wabti::Module::New(store, desc);

  // Bind import
  wabti::RefVec imports;
  auto import = core->desc().imports[0];
  auto ft = *wabt::cast<wabti::FuncType>(import.type.type.get());
  auto func = wabti::HostFunc::New(store, ft,
    [](wabti::Thread &, const wabti::Values &params, wabti::Values &, wabti::Trap::Ptr*) -> wabt::Result {
      std::cout << "cb " << params.size() << " " << params[0].type << std::endl;
      return wabt::Result::Ok;
  });
  imports.push_back(func.ref());

  // Instantiate module
  wabti::Trap::Ptr trap;
  auto corei = wabti::Instance::Instantiate(store, core.ref(), imports, &trap);
  assert(corei);

  // Run
  wabti::Values runParams;
  wabti::Values runResults;
  return store.UnsafeGet<wabti::Func>(corei->funcs()[1])->Call(store, runParams, runResults, &trap, nullptr);
}

int main(int argc, char *argv[]) {
  run();
}
```

When I run this against a Debug build of libwabt, the program works fine, and this is the output:

    cb 2 -1

However, when I run this against a Release build of libwabt, this is what I get:

    cb 1 9

It seems that at least the params parameter is wrong (wrong length, and wrong contents).

I tried running an asan build. In my real program, I get asan errors during module instantiation, but in the minimal program above, I don't get anything.

remko:
Value's `type` field was compiled conditionally on the `NDEBUG` define. This causes problems with programs compiling against libwabt that don't define this macro, as the Value layout no longer matches.

Removing the condition entirely, assuming the performance impact of this is limited (especially if assertions are disabled).

Fixes #2069

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,6 +111,10 @@ check_type_size(size_t SIZEOF_SIZE_T)
 include(TestBigEndian)  # Note: deprecated in CMake 3.20
 test_big_endian(WABT_BIG_ENDIAN)
 
+if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+  set(WABT_DEBUG 1)
+endif ()
+
 configure_file(src/config.h.in include/wabt/config.h @ONLY)
 
 
--- a/include/wabt/interp/interp.h
+++ b/include/wabt/interp/interp.h
@@ -27,6 +27,7 @@
 #include <vector>
 
 #include "wabt/cast.h"
+#include "wabt/config.h"
 #include "wabt/common.h"
 #include "wabt/feature.h"
 #include "wabt/opcode.h"
@@ -580,7 +581,7 @@ struct Value {
   };
 
  public:
-#ifndef NDEBUG
+#ifdef WABT_DEBUG
   Value() : v128_(0, 0, 0, 0), type(ValueType::Any) {}
   void SetType(ValueType t) { type = t; }
   void CheckType(ValueType t) const {
--- a/src/config.h.in
+++ b/src/config.h.in
@@ -22,6 +22,8 @@
 
 #cmakedefine WABT_VERSION_STRING "@WABT_VERSION_STRING@"
 
+#cmakedefine WABT_DEBUG @WABT_DEBUG@
+
 /* TODO(binji): nice way to define these with WABT_ prefix? */
 
 /* Whether <alloca.h> is available */


include/wabt/interp/interp.h
```diff
@@ -589,11 +588,6 @@ struct Value {
     assert(t == type || type == ValueType::Any);
   }
   ValueType type;
-#else
-  Value() : v128_(0, 0, 0, 0) {}
-  void SetType(ValueType) {}
-  void CheckType(ValueType) const {}
-#endif
```
sbc100, line 596: I believe you can leave SetType/CheckType inside the `ifdef`.  The important thing is that the sizeof the struct be the same in both debug and release I think,

remko: @sbc100 Correct.

include/wabt/interp/interp.h
```diff
@@ -588,12 +588,12 @@ struct Value {
     // Alternatively we could modify InterpAlloca to set the type.
     assert(t == type || type == ValueType::Any);
   }
-  ValueType type;
 #else
   Value() : v128_(0, 0, 0, 0) {}
   void SetType(ValueType) {}
   void CheckType(ValueType) const {}
 #endif
+  ValueType type;
```
sbc100, line 596: Can you add comment here.  How about `only used in debug builds but declared here to avoid struct size differing`.

Actually though, you know what, I think you might be correct about putting this in `config.h`.  If we have a release version of libwabt we really want to fix this header to be release too.  

Also, the sizeof values in the interpreter is likely very important to performance of the release build (not sure if anyone actually cares about that).

Would it be tricky to put this in config.h and fix it at library build time?

CMakeLists.txt
```diff
@@ -111,6 +111,10 @@ check_type_size(size_t SIZEOF_SIZE_T)
 include(TestBigEndian)  # Note: deprecated in CMake 3.20
 test_big_endian(WABT_BIG_ENDIAN)
 
+if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+set(WABT_DEBUG 1)
```
sbc100, line 115: Indent two spaces

sbc100 (approved):
Nice!  lgtm

PR #2073, opened 2022-11-18
CI: add CIFuzz action

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.

This is a service offered by OSS-Fuzz where WABT already runs. CIFuzz can help detect catch regressions and fuzzing build issues early, and has a variety of features (see the URL above). In the current PR the fuzzers gets build on a pull request and will run for 300 seconds.

Signed-off-by: David Korczynski <david@adalogics.com>

### DIFF
--- a/.github/workflows/wabt-cifuzz.yml
+++ b/.github/workflows/wabt-cifuzz.yml
@@ -1,0 +1,24 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'wabt'
+        dry-run: false
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'wabt'
+        fuzz-seconds: 300
+        dry-run: false
+    - name: Upload Crash
+      uses: actions/upload-artifact@v3
+      if: failure() && steps.build.outcome == 'success'
+      with:
+        name: artifacts
+        path: ./out/artifacts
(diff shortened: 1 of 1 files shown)


keithw:
I think the failure here is an indication that (some of?) WABT's fuzz-targets at https://github.com/google/oss-fuzz/tree/master/projects/wabt no longer build with the current WABT tree, and as a result, I don't think oss-fuzz is fuzzing WABT anymore. https://oss-fuzz-build-logs.storage.googleapis.com/index.html#wabt suggests the last successful build was Sept. 27. :-(

DavidKorczynski:
> I think the failure here is an indication that (some of?) WABT's fuzz-targets at https://github.com/google/oss-fuzz/tree/master/projects/wabt no longer build with the current WABT tree, and as a result, I don't think oss-fuzz is fuzzing WABT anymore. https://oss-fuzz-build-logs.storage.googleapis.com/index.html#wabt suggests the last successful build was Sept. 27. :-(

FIxed in https://github.com/google/oss-fuzz/pull/9101

keithw:
Filed https://github.com/google/oss-fuzz/pull/9125 to add access to the oss-fuzz dashboard to some of the currently more-active maintainers.

keithw (approved):
Now that we seem to have everything working on the oss-fuzz side (and because this workflow only seems to pick up "new" fuzz failures so it's passing), I'm tempted to merge this but keep the CI status non-required.

The upsides seem like (a) it prevents us from letting the oss-fuzz fuzzers bitrot without us knowing about it, and (b) it may detect "new" fuzz failures that would be introduced by a PR.

Downsides seem like (a) it's yet another line in our testing matrix, and although it only takes 10 minutes so shouldn't on the critical path, I'm not sure if GitHub eventually starts to limit our parallelism, and (b) we now have three essentially independent "fuzz" pathways in WABT that we should probably think about consolidating. (1: the oss-fuzz fuzzers kept in Google's oss-fuzz repo, 2: the "wasm2wat_fuzz" binary in our own repo that's now compiled [but not executed] in our "fuzz" CI tests, and then 3: the ./scripts/fuzz-*.sh scripts and the `fuzz-in` directory that run afl-fuzz.)

keithw:
Merging (as a non-required status check). We should revisit this if it starts to cause problems or delay the overall CI pipeline.

PR #2075, opened 2022-11-24
wasm2c: add a test for __stack_pointer

dimitry-:
The __stack_pointer does not seem to be working
properly, - the test fails.

### DIFF
--- a/test/wasm2c/wast/stack_pointer.wast
+++ b/test/wasm2c/wast/stack_pointer.wast
@@ -1,0 +1,61 @@
+(module
+  (type (;0;) (func (param i32 i32)))
+  (type (;1;) (func (param i32 i32) (result i32)))
+  (func $store_char*__char_ (type 0) (param i32 i32)
+    local.get 0
+    local.get 1
+    i32.store8)
+  (func (export "get_from_stack") (type 1) (param i32 i32) (result i32)
+    (local i32 i32 i32)
+    global.get $__stack_pointer
+    local.tee 2
+    local.set 3
+    local.get 2
+    local.get 1
+    i32.const 15
+    i32.add
+    i32.const -16
+    i32.and
+    i32.sub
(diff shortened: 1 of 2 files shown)


keithw:
This test also fails in the reference interpreter and in WABT's interpreter, so it doesn't *seem* to be a wasm2c problem?

```
spec/interpreter$ ./wasm -i /tmp/stack_pointer.wast 
Result: 0 : i32
Expect: 1 : i32
/tmp/stack_pointer.wast:60.1-60.83: assertion failure: wrong return values
```

```
$ wast2json stack_pointer.wast
$ spectest-interp stack_pointer.json
stack_pointer.wast:60: mismatch in result 0 of assert_return: expected i32:1, got i32:0
stack_pointer.wast:61: mismatch in result 0 of assert_return: expected i32:6, got i32:0
1/3 tests passed.
```

test/wasm2c/wast/stack_pointer.wast
```diff
@@ -0,0 +1,61 @@
+(module
+  (type (;0;) (func (param i32 i32)))
+  (type (;1;) (func (param i32 i32) (result i32)))
+  (func $store_char*__char_ (type 0) (param i32 i32)
+    local.get 0
+    local.get 1
```
sbc100, line 6: The extra/redundant local.get/local.set instructions look like they might be coming from unoptimized llvm output?   Perhaps that test would be more readable if use use `-O1` or `-O2` or run `wasm-opt` on the binary?   

Ideally we could hand-write a smaller/simpler/more readable test?

dimitry-:
Sorry for the delay - I get one day a week to work on this project :)
Background: this code is compiled from c++ code. and if I try O2 or higher llvm cleverly optimises stack operation away.

It might be llvm problem or but in the test - I will investigate further. Closing this until then.

PR #2077, opened 2022-11-28
run-spec-wasm2c.py: split `run_spec_tests` function if it's too long

keithw:
This is intended to improve GCC's long compile times for some of the extremely long SIMD tests that will come soon (#2021).

On my laptop running Ubuntu 22.04, this change reduces the time for the simd_f32x4_pmin_pmax.wast test from 16 minutes to about 70 seconds.

### DIFF
--- a/test/run-spec-wasm2c.py
+++ b/test/run-spec-wasm2c.py
@@ -129,6 +129,7 @@ class CWriter(object):
     def Write(self):
         self._WriteIncludes()
         self.out_file.write(self.prefix)
+        self._WriteModuleInstances()
         self.out_file.write("\nvoid run_spec_tests(void) {\n\n")
         for command in self.commands:
             self._WriteCommand(command)
@@ -147,7 +148,7 @@ class CWriter(object):
         return self.unmangled_names[idx]
 
     def GetModuleInstanceName(self, idx_or_name=None):
-        return self.GetModulePrefix() + '_instance'
+        return self.GetModulePrefix(idx_or_name) + '_instance'
 
     def _CacheModulePrefixes(self):
         idx = 0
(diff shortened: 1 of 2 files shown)


test/run-spec-wasm2c.py
```diff
@@ -147,7 +157,7 @@ def GetModulePrefixUnmangled(self, idx):
         return self.unmangled_names[idx]
 
     def GetModuleInstanceName(self, idx_or_name=None):
-        return self.GetModulePrefix() + '_instance'
+        return self.GetModulePrefix(idx_or_name) + '_instance'
```
sbc100, line 160: Is this related?

test/run-spec-wasm2c.py
```diff
@@ -243,9 +253,15 @@ def _WriteCommand(self, command):
...
         self.out_file.write('%s_init_module();\n' % self.GetModulePrefix())
-        self.out_file.write('%s_instance_t %s;\n' % (self.GetModulePrefix(), self.GetModuleInstanceName()))
         self._WriteModuleInitCall(command, False)
 
+    def _WriteModuleInstances(self):
+        idx = 0
+        for command in self.commands:
+            if IsModuleCommand(command):
+                self.out_file.write('%s_instance_t %s;\n' % (self.GetModulePrefix(idx), self.GetModuleInstanceName(idx)))
+                idx += 1
```
sbc100, line 263: Same here, is this related to the splitting up?

sbc100 (approved):

test/run-spec-wasm2c.py
keithw: Yes, otherwise it's not possible to call GetModuleInstanceName on an idx that isn't the current idx. (Which we need to do now that we're declaring all the module instance structs upfront.) Probably this method should always have been passing the `idx_or_name` argument through.

test/run-spec-wasm2c.py
keithw: Yes, this is most of the first commit. We're declaring all of the instance_t structs upfront (as globals) so that any of the split-up run_spec_tests_**N** functions can refer to any of them.

sbc100 (approved):
lgtm!

keithw:
Sigh, this seems to be an asan false-positive in GCC 9 (which we use in CI) that's gone by GCC 11 and gone in clang.

PR #2078, opened 2022-11-28
wasm2c CI: run asan and ubsan tests on clang 14

keithw:
Move the GitHub CI tests for "asan" and "ubsan" over to using clang as the wasm2c compiler (in addition to using clang to compile wasm2c and the rest of WABT itself). Prerequisite for #2077.

This could probably be done in a more principled way (by modifying the Makefile for "clang-xxx-asan" and "clang-xxx-ubsan") but this matches the current practice in setting WASM2C_CFLAGS explicitly in the GitHub workflow (and is easier).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -81,9 +81,10 @@ jobs:
 
   build-asan:
     name: asan
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     env:
       USE_NINJA: "1"
+      CC: "clang" # used by the wasm2c tests
       WASM2C_CFLAGS: "-fsanitize=address"
     steps:
     - uses: actions/setup-python@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -84,6 +84,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       USE_NINJA: "1"
+      CC: "clang"
```
sbc100, line 87: Maybe add a comment that this is for the benefit of the wasm2c tests?

sbc100 (approved):

.github/workflows/build.yml
keithw: Done.

keithw:
Just bumped the runners for the asan and ubsan tests to `ubuntu-22.04` to get clang 14, which is required to pass the stack exhaustion tests. (`ubuntu-latest` is still sometimes `ubuntu-20.04`, which includes clang 10, which doesn't seem to respect the `'-fno-optimize-sibling-calls'` flag and tail-call-optimizes away the infinite recursion.)

PR #2079, opened 2022-11-29
[WIP] Use length-prefixed init expressions

sbc100:
This change reserves a new byte in (0xc5) in the opcode space for encoding the
size of a const expression.   When this byte is the first instruction in a const
expression the leb that follows is the length in bytes of the entire expression.

See https://github.com/WebAssembly/extended-const/issues/9


keithw:
Given the resolution to https://github.com/WebAssembly/extended-const/issues/9 , is it okay to close this one?

PR #2080, opened 2022-11-29
Make wasm2c output UBSAN-clean (and run w2c CI w/ UBSAN)

keithw:
We had been running the GitHub UBSAN wasm2c tests with -fsanitize=undefined but without -fno-sanitize-recover, meaning some of the spec tests were printing UBSAN error messages but still returning 0, so we weren't seeing the test failures.

The two issues that UBSAN was flagging were:
- fabs and fabsf left-shifting signed literals by 31 or 63 (switched to making them unsigned)
- calling memcpy with a zero length but NULL src (switched to adding an explicit `check_only` parameter to `memory_init` and `LOAD_DATA` that makes them skip the actual memcpy but still enforce the bounds-check, which is required)

Once we reach convergence on #2080/#2081/#2134, I think we're ready to merge #2119 (SIMD).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -105,7 +105,7 @@ jobs:
     env:
       USE_NINJA: "1"
       CC: "clang" # used by the wasm2c tests
-      WASM2C_CFLAGS: "-fsanitize=undefined"
+      WASM2C_CFLAGS: "-fno-sanitize-recover=undefined"
     steps:
     - uses: actions/setup-python@v1
       with:
--- a/src/template/wasm2c.declarations.c
+++ b/src/template/wasm2c.declarations.c
@@ -64,10 +64,12 @@ static inline void load_data(void* dest, const void* src, size_t n) {
     dest_chars[n - i - 1] = cursor;
   }
 }
-#define LOAD_DATA(m, o, i, s)                   \
-  do {                                          \
-    RANGE_CHECK((&m), m.size - o - s, s);       \
-    load_data(&(m.data[m.size - o - s]), i, s); \
+#define LOAD_DATA(m, o, check_only, i, s)         \
+  do {                                            \
+    RANGE_CHECK((&m), m.size - o - s, s);         \
+    if (!check_only) {                            \
+      load_data(&(m.data[m.size - o - s]), i, s); \
+    }                                             \
   } while (0)
 #define DEFINE_LOAD(name, t1, t2, t3)                                  \
   static inline t3 name(wasm_rt_memory_t* mem, u64 addr) {             \
@@ -90,10 +92,12 @@ static inline void load_data(void* dest, const void* src, size_t n) {
 static inline void load_data(void* dest, const void* src, size_t n) {
   memcpy(dest, src, n);
 }
-#define LOAD_DATA(m, o, i, s)      \
-  do {                             \
-    RANGE_CHECK((&m), o, s);       \
-    load_data(&(m.data[o]), i, s); \
+#define LOAD_DATA(m, o, check_only, i, s) \
+  do {                                    \
+    RANGE_CHECK((&m), o, s);              \
+    if (!check_only) {                    \
+      load_data(&(m.data[o]), i, s);      \
+    }                                     \
   } while (0)
 #define DEFINE_LOAD(name, t1, t2, t3)                      \
   static inline t3 name(wasm_rt_memory_t* mem, u64 addr) { \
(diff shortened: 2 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -85,7 +85,7 @@ jobs:
     env:
       USE_NINJA: "1"
       CC: "clang" # used by the wasm2c tests
-      WASM2C_CFLAGS: "-fsanitize=address"
+      WASM2C_CFLAGS: "-fsanitize=address -fno-sanitize-recover"
```
sbc100, line 88: Wow.. under what circumstances do ASAN failures don't cause the program to return non-zero?   I had no idea this extra flag was needed.

sbc100: Looking at the docs it looks like only UBSan has the `no-sanitize-recover` flag and that it designed to be used *instead* of `-fsantitize=`.  So it would be `-fno-sanitize-recover=undefined` ?

keithw: Just took the `-fno-sanitize-recover` off of ASAN. I was really just trying to copy the behavior we already have in CMakeLists.txt, where it appends `-fno-sanitize-recover` for the UBSAN run. (And you're right, only for UBSAN.) Then after #2081 we start automatically propagating the exact flags from CMakeLists.txt into WASM2C_CFLAGS (including `-fno-sanitize-recover`) so we can stop hardcoding it in build.yml.

keithw: Just made it `-fno-sanitize-recover=undefined` per request. It will become automatic with #2081 anyway (will match the exact flags from CMakeLists.txt).

src/template/wasm2c.declarations.c
```diff
@@ -426,7 +430,7 @@ static float wasm_fabsf(float x) {
   if (UNLIKELY(isnan(x))) {
     uint32_t tmp;
     memcpy(&tmp, &x, 4);
-    tmp = tmp & ~(1 << 31);
+    tmp = tmp & ~((uint32_t)1 << 31);
```
sbc100, line 433: Would 1UL work here instead of the cast?

test/wasm2c/add.txt
```diff
@@ -128,10 +128,12 @@ static inline void load_data(void* dest, const void* src, size_t n) {
...
-#define LOAD_DATA(m, o, i, s)                   \
-  do {                                          \
-    RANGE_CHECK((&m), m.size - o - s, s);       \
-    load_data(&(m.data[m.size - o - s]), i, s); \
+#define LOAD_DATA(m, o, check_only, i, s)         \
+  do {                                            \
+    RANGE_CHECK((&m), m.size - o - s, s);         \
+    if (!check_only) {                            \
+      load_data(&(m.data[m.size - o - s]), i, s); \
+    }                                             \
```
sbc100, line 136: I guess the compiler can eliminate this extra condition most of the time?

test/wasm2c/minimal.txt
```diff
@@ -532,14 +536,15 @@ static inline void memory_copy(wasm_rt_memory_t* dest,
...
+                               bool check_only,
                                const u8* src,
                                u32 src_size,
                                u32 dest_addr,
                                u32 src_addr,
                                u32 n) {
   if (UNLIKELY(src_addr + (uint64_t)n > src_size))
     TRAP(OOB);
-  LOAD_DATA((*dest), dest_addr, src + src_addr, n);
+  LOAD_DATA((*dest), dest_addr, check_only, src + src_addr, n);
```
sbc100, line 547: Rather than add an extra argument to the LOAD_DATA macro could we define a separate `memory_check` or do:

```
```
if (check_only)
  RANGE_CHECK
else
  LOAD_DATA
```

Better still, should we just use `RANGE_CHECK` here since IIRC we are not actually trying to load anything.

Actually, maybe that doesn't work since RANGE_CHECK doesn't always do anything... maybe we need something like `FORCE_RANGE_CHECK` here?

src/template/wasm2c.declarations.c
keithw: Sure, either works. Will change it.

test/wasm2c/minimal.txt
keithw: What makes this annoying is that the range that LOAD_DATA needs to check depends (I think) on the endianness -- it's written differently in the different cases. (I'm not 100% confident wasm2c currently works on little-endian platforms, but I think we still nominally care about doing the right thing there.)

How do you feel about the version I just pushed, where we just completely eliminate the check_only parameter from everywhere, and instead we just test "n" in load_data()? This has a tiny runtime cost but I'm guessing it's in the noise in any realistic setting, and it's a lot cleaner/simpler.

(BTW RANGE_CHECK always checks -- it's MEMCHECK that turns into a no-op when using the signal handler.)

sbc100 (approved):
lgtm!

PR #2081, opened 2022-11-29
Split asan/ubsan tests by matrixing + propagating cmake settings to w2c tests

keithw:
~~(This has to be sequenced behind #2080 because it starts running the wasm2c tests with UBSAN.)~~

This PR speeds up the CI build (helpful for SIMD) by matrixing the asan and ubsan CI workflows into parallel release and debug jobs.

It auto-propagates the Make and CMake configuration (e.g. `make test-clang-debug-ubsan`) through to the wasm2c tests (in `run-spec-wasm2c.py`), as discussed at https://github.com/WebAssembly/wabt/pull/2021#issuecomment-1329746204 and later comments.

If we do it this way, it will require a repository configuration change to the required status checks for the main branch. Another option is to just add more named jobs in build.yml without using the matrix feature.

Once we reach convergence on ~~#2080~~/#2081/~~#2134~~, I think we're ready to merge #2119 (SIMD).


CMakeLists.txt
```diff
@@ -111,8 +111,17 @@ check_type_size(size_t SIZEOF_SIZE_T)
...
+endif ()
+  
 if (CMAKE_BUILD_TYPE STREQUAL "Debug")
   set(WABT_DEBUG 1)
+  set(WASM2C_CFLAGS "${WASM2C_CFLAGS} -g -O0")
+endif ()
+
+if (DEFINED CMAKE_OSX_SYSROOT)
+  set(WASM2C_CFLAGS "${WASM2C_CFLAGS} -isysroot ${CMAKE_OSX_SYSROOT}")
 endif ()
```
sbc100, line 125: WASM2C_CFLAGS is only used when running tests right?   Could we put all this stuff inside the `BUILD_TESTS` block to make this clear?

keithw: Done (except for where we set the sanitizer flags -- I kept that one adjacent because that's where we know the full flags).

test/run-spec-wasm2c.py
```diff
@@ -509,7 +509,7 @@ def main(args):
             '--enable-multi-memory': options.enable_multi_memory})
 
         options.cflags += shlex.split(os.environ.get('WASM2C_CFLAGS', ''))
-        cc = utils.Executable(options.cc, *options.cflags, forward_stderr=True,
+        cc = utils.Executable(options.cc, after_args=options.cflags, forward_stderr=True,
```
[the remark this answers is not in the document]
keithw: We want the environment-specified CFLAGS to go to the end of the argument list so that CMake can supply something like WASM2C_CFLAGS="-g -O0" (on a Debug build) and have that override the default of `-O2` in run-spec-wasm2c.py (because it comes later in the compiler invocation).

Another option would be for run-spec-wasm2c.py to check for the existence of the WASM2C_CFLAGS environment variable, and only use '-O2' if that env var is undefined? But I think it's probably still helpful to put the environment variable flags at the end of the argument list because people might want to do other things (e.g. turn off warnings, etc.) so you want the user-specified flags to override the defaults.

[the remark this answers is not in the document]
keithw: Split out into #2134.

CMakeLists.txt
```diff
@@ -618,6 +619,18 @@ if (BUILD_TOOLS)
...
 if (BUILD_TESTS)
+  if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+    set(WASM2C_CFLAGS "${WASM2C_CFLAGS} -g -O0")
+  endif ()
+
+  if (DEFINED CMAKE_OSX_SYSROOT)
+    set(WASM2C_CFLAGS "${WASM2C_CFLAGS} -isysroot ${CMAKE_OSX_SYSROOT}")
+  endif ()
+
+  if(DEFINED ENV{WASM2C_CFLAGS})
```
sbc100, line 630: Space after `if`

sbc100 (approved):

CMakeLists.txt
keithw: Thanks, fixed.

keithw:
Hi @sbc100 -- I think this is ready. Are you able to adjust the repository configuration to require these status checks:
- "sanitize (asan, debug)"
- "sanitize (asan, release)"
- "sanitize (ubsan, debug)"
- "sanitize (ubsan, release)"

and remove these?
- "asan"
- "ubsan"

sbc100:
sure I can do that one this lands.  Should I land this now?

keithw:
Yes please!

PR #2085, opened 2022-12-01
update-spec-tests.py: Include wasm2c version of proposals tests

sbc100:


keithw (approved):
lgtm. We could imagine having a second (longer) `unimplemented` list for wasm2c that reflects the proposals it supports.

PR #2086, opened 2022-12-01
wasm2c: Initial implementation of memory64

sbc100:


wasm2c/wasm-rt-impl.c
```diff
@@ -366,9 +366,9 @@ void wasm_rt_free(void) {
...
 void wasm_rt_allocate_memory(wasm_rt_memory_t* memory,
-                             uint32_t initial_pages,
-                             uint32_t max_pages) {
-  uint32_t byte_length = initial_pages * PAGE_SIZE;
+                             uint64_t initial_pages,
+                             uint64_t max_pages) {
+  uint64_t byte_length = initial_pages * PAGE_SIZE;
 #if WASM_RT_MEMCHECK_SIGNAL_HANDLER
   /* Reserve 8GiB. */
   void* addr = os_mmap(0x200000000ul);
```
keithw, line 374: I think we'll have to forbid using WASM_RT_MEMCHECK_SIGNAL_HANDLER (or find some other strategy?) if a memory address+offset is allowed to be more than $2^{33}$ -- otherwise can't a memory read go beyond the allocated virtual address space?

keithw (commented):
I guess it's understandable, but I wish some of the memory64 tests tested growing a memory to $\ge 2^{33}$ bytes and then reading and writing to those high addresses. I realize it may be difficult to run that test on most architectures.

wasm2c/wasm-rt-impl.c
sbc100: Hmm, yes indeed.   I guess maybe we are going to need a WASM_RT_MEMORY64 macro that is defined in memory64 mode.

sbc100: How do you feel about landing this simple change, as-is, and then continuing to iterate on full wasm64 in followup PRs?  I guess we should probably remove the check mark from the README until later too?

keithw: I don't know how much effort you feel this is worth, but maybe our "dream" solution here would involve:

1. Letting wasm2c emit explicit "sw_checked" versions (i.e. mandatory bounds check) of allocate_memory/grow_memory/free_memory and load and store, depending on decisions made at wasm2c time. @angelamontemayor  has already done this part for use in Fixpoint: https://github.com/WebAssembly/wabt/commit/dfc437bf3c78727a697ff68b41f3bba20084c7b9

2. By default, wasm2c generates "sw_checked" allocate calls and load/store instructions for 64-bit indexed memories, and keeps the allocates and load/store instructions for 32-bit indexed memories the same (either sw-checked or hw-checked depending on whether we have the signal handler).

3. Optionally, we could imagine wasm2c would also avoid sw_checked instructions for memory64 memories that are (a) not imported and (b) have a maximum size specified in the declaration $\le 2^{32}$ bytes.

keithw (approved):
lftm % discussion

wasm2c/wasm-rt-impl.c
sbc100: OK, I decided to revisit this.

For this initial version we don't support `WASM_RT_MEMCHECK_SIGNAL_HANDLER` at all when memory64 is enabled. 

I've used `__builtin_add_overflow` to check for 64-bit overflow when calculating base + offset in ` RANGE_CHECK`.  This means we don't currently support non-clang/non-gcc compiler with wasm2c + memory64.

This keeps this change simple, and we can them improve on things incrementally if we wish.

sbc100:
This PR is actually much simpler now, and passes all the memory64 tests.

wasm2c/wasm-rt-impl.c
```diff
@@ -361,11 +361,15 @@ void wasm_rt_free(void) {
...
-  uint32_t byte_length = initial_pages * PAGE_SIZE;
+                             uint64_t initial_pages,
+                             uint64_t max_pages,
+                             bool is64) {
+  uint64_t byte_length = initial_pages * PAGE_SIZE;
 #if WASM_RT_MEMCHECK_SIGNAL_HANDLER
   /* Reserve 8GiB. */
+  assert(
+      !is64 &&
+      "memotry64 is not yet compatible with WASM_RT_MEMCHECK_SIGNAL_HANDLER");
```
keithw, line 372: "memory64"

keithw:
Still lgtm...

PR #2088, opened 2022-12-01
fix -Wunused-variable warnings when building without asserts

walkingeyerobot:


src/c-writer.cc
```diff
@@ -601,24 +601,20 @@ void CWriter::DefineImportName(const Import* import,
   import_syms_.insert(name);
   import_module_sym_map_.emplace(name, import->module_name);
   global_syms_.insert(mangled);
-  bool success =
-      global_sym_map_.emplace(name + MangleField(type), mangled).second;
-  assert(success);
+  assert(global_sym_map_.emplace(name + MangleField(type), mangled).second);
```
sbc100, line 604: Sadly you can't do this because `global_sym_map_.emplace` has a side effect that we want..

sbc100: You can do this but its a bit ugly:

```
#ifndef NDEBUG
   bool success =
#endif
global_sym_map_.emplace(name + MangleField(type), mangled).second;
```

sbc100: Actually answer from here looks like the nicest way: https://stackoverflow.com/questions/777261/avoiding-unused-variables-warnings-when-using-assert-in-a-release-build

walkingeyerobot: Fair enough! Can I steal `WASM_UNUSED` from binaryen? https://github.com/WebAssembly/binaryen/blob/main/src/compiler-support.h#L37

include/wabt/compiler-support.h
```diff
@@ -0,0 +1,30 @@
...
+// Taken from binaryen's src/compiler-support.h
+// The code might contain TODOs or stubs that read some values but do nothing
+// with them. The compiler might fail with [-Werror,-Wunused-variable].
+// The WASM_UNUSED(varible) is a wrapper that helps to suppress the error.
+#define WASM_UNUSED(expr)                                                      \
+  do {                                                                         \
+    if (sizeof expr) {                                                         \
+      (void)0;                                                                 \
+    }                                                                          \
+  } while (0)
```
sbc100, line 28: I think it would make sense to put this in `src/config.h.in` and call it `WABT_UNUSED`

sbc100: Oh wait.. we already have `WABT_UNUSED`.. i think maybe you can just apply that to definition of `success`?

sbc100: As a followup we should switch from `WABT_UNUSED` to C++17 `[[maybe_unused]]`?

sbc100 (approved):

src/c-writer.cc
```diff
@@ -603,6 +603,7 @@ void CWriter::DefineImportName(const Import* import,
   global_syms_.insert(mangled);
   bool success =
       global_sym_map_.emplace(name + MangleField(type), mangled).second;
+  WABT_UNUSED(success);
```
sbc100, line 606: I think the idea is that it goes on the line with the declaration.. no need for this extra line.

PR #2091, opened 2022-12-02
Use C++17's [[fallthrough]]. NFC

sbc100:


keithw (approved):
lgtm. We might have to add `SYSTEM` to the uvwasi `include_directories` line to get this to compile, if there is implicit fallthrough in uvwasi. (I'm guessing clang won't print the warning for something included with `-isystem`?)

keithw:
Will intend to merge this soon unless there is objection (@sbc100).

PR #2093, opened 2022-12-02
Remove use of alloca. NFC

sbc100:
We can just use a variables sized array in the once place we were using it.


sbc100:
Damn.. I guess MSVC still doesn't do this, even for C++17 :(

PR #2095, opened 2022-12-02
build.yml: add wasm2c check with explicit software memchecks

keithw:
Now that we're using the signal handler on Windows, we were missing test coverage for software-checked memories in wasm2c.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -119,6 +119,24 @@ jobs:
     - run: make test-clang-debug-ubsan
     - run: make test-clang-release-ubsan
 
+  build-wasm2c-memchecked:
+    name: wasm2c-memchecked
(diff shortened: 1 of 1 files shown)


sbc100 (approved):
I'm slightly worried about adding more and more test configs.

If we starts to be a problem we could run just a subset during CI and then do the full set on nightly basis?

keithw:
> If we starts to be a problem we could run just a subset during CI and then do the full set on nightly basis?

Agreed. At least so far GitHub seems happy to give us the parallelism, so I'm hopeful that adding these kinds of tests won't increase our end-to-end CI time much at all. (On that note, are you able to give another look to #2080 and #2081? If we get these in, we'll be in the homestretch to landing the wasm2c SIMD PR (#2021) which would be huge.)

PR #2096, opened 2022-12-02
Use C++11's [[likely]]/[[unlikely]]. NFC

sbc100:
This could improve performance on windows which didn't previously define these macros.


sbc100:
Actually forgot this.. it requires C++20

PR #2104, opened 2022-12-19
src/wast-parser.cc: Accept non abbreviated forms of result for select

yamt:
eg. "select (result i32) (result)"

cf. https://github.com/WebAssembly/spec/pull/1567


sbc100:
What does `select (result i32) (result)` mean semantically?

yamt:
> What does `select (result i32) (result)` mean semantically?

if i read the spec correctly, it's same as `select (result i32)`.

in general `(result A) (result B)` is same as `(result A B)` and the same rule applies to the case of zero.

sbc100 (approved):

PR #2105, opened 2022-12-19
wasm2c-nosandbox: This branch implements stack_pointer and disables some memory checks

dimitry-:
Please only review last 3 commits since first 2 are cherry-pick from not-yet merged PRs

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -39,6 +39,7 @@ extern const char* s_header_top;
 extern const char* s_header_bottom;
 extern const char* s_source_includes;
 extern const char* s_source_declarations;
+extern const char* s_source_declarations_nosandbox;
 
 namespace wabt {
 
@@ -996,7 +997,11 @@ std::string CWriter::GenerateHeaderGuard() const {
 void CWriter::WriteSourceTop() {
   Write(s_source_includes);
   Write(Newline(), "#include \"", header_name_, "\"", Newline());
-  Write(s_source_declarations);
+  if (options_.no_sandbox) {
+    Write(s_source_declarations_nosandbox);
+  } else {
+    Write(s_source_declarations);
+  }
 }
 
 void CWriter::WriteMultivalueTypes() {
@@ -3317,9 +3322,13 @@ void CWriter::Write(const LoadExpr& expr) {
   Memory* memory = module_->memories[module_->GetMemoryIndex(expr.memidx)];
 
   Type result_type = expr.opcode.GetResultType();
-  Write(StackVar(0, result_type), " = ", func, "(",
-        ExternalInstancePtr(ModuleFieldType::Memory, memory->name), ", (u64)(",
-        StackVar(0), ")");
+  if (options_.no_sandbox) {
+    Write(StackVar(0, result_type), " = ", func, "((u64)(", StackVar(0), ")");
+  } else {
+    Write(StackVar(0, result_type), " = ", func, "(",
+          ExternalInstancePtr(ModuleFieldType::Memory, memory->name),
+          ", (u64)(", StackVar(0), ")");
+  }
   if (expr.offset != 0)
     Write(" + ", expr.offset, "u");
   Write(");", Newline());
@@ -3346,8 +3355,13 @@ void CWriter::Write(const StoreExpr& expr) {
 
   Memory* memory = module_->memories[module_->GetMemoryIndex(expr.memidx)];
 
-  Write(func, "(", ExternalInstancePtr(ModuleFieldType::Memory, memory->name),
-        ", (u64)(", StackVar(1), ")");
+  if (options_.no_sandbox) {
+    Write(func, "((u64)(", StackVar(1), ")");
+  } else {
+    Write(func, "(", ExternalInstancePtr(ModuleFieldType::Memory, memory->name),
+          ", (u64)(", StackVar(1), ")");
+  }
+
   if (expr.offset != 0)
     Write(" + ", expr.offset);
   Write(", ", StackVar(0), ");", Newline());
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -178,6 +178,8 @@ class CWriter {
   void WriteCHeader();
   void WriteCSource();
 
+  bool SkipImport(const Import* import) const;
+
   size_t MarkTypeStack() const;
   void ResetTypeStack(size_t mark);
   Type StackType(Index) const;
@@ -273,6 +275,7 @@ class CWriter {
                                         const std::string&);
   void WriteModuleInstance();
   void WriteGlobals();
+  void WriteStackPointerGlobal(const Global& global);
   void WriteGlobal(const Global&, const std::string&);
   void WriteGlobalPtr(const Global&, const std::string&);
   void WriteMemories();
@@ -370,9 +373,19 @@ class CWriter {
 
   std::vector<std::pair<std::string, MemoryStream>> func_sections_;
   SymbolSet func_includes_;
+  const Global* stack_pointer_global_ = nullptr;
 };
 
 constexpr std::string_view kImplicitFuncLabel = "$Bfunc";
+constexpr std::string_view kEnvModuleName = "env";
+constexpr std::string_view kStackPointerName = "$__stack_pointer";
+constexpr std::string_view kStackPointerNameWithoutDollar = "__stack_pointer";
+
+constexpr size_t kStackSize = 64 * 1024;
+constexpr size_t kStackBufferSize = 4 * 1024;
+constexpr std::string_view kStackArrayVariableName = "g_w2c_stack_array";
+constexpr std::string_view kStackOffsetGlobalVariableName =
+    "g_w2c_stack_offset";
 
 size_t CWriter::MarkTypeStack() const {
   return type_stack_.size();
@@ -498,7 +511,7 @@ std::string CWriter::MangleName(std::string_view name) {
   std::string result = "Z_";
 
   if (!name.empty()) {
-    for (unsigned char c : name) {
+    for (char c : name) {
       if ((isalnum(c) && c != kPrefix) || c == '_') {
         result += c;
       } else {
@@ -1121,10 +1134,18 @@ void CWriter::WriteTags() {
   Write(CloseBrace(), Newline());
 }
 
+bool CWriter::SkipImport(const Import* import) const {
+  return options_.no_sandbox && import->module_name == kEnvModuleName &&
+         import->field_name == kStackPointerNameWithoutDollar;
+}
+
 void CWriter::ComputeUniqueImports() {
   using modname_name_pair = std::pair<std::string, std::string>;
   std::map<modname_name_pair, const Import*> import_map;
   for (const Import* import : module_->imports) {
+    if (SkipImport(import)) {
+      continue;
+    }
     // After emplacing, the returned bool says whether the insert happened;
     // i.e., was there already an import with the same modname and name?
     // If there was, make sure it was at least the same kind of import.
@@ -1150,15 +1171,18 @@ void CWriter::ComputeUniqueImports() {
 }
 
 void CWriter::BeginInstance() {
-  if (module_->imports.empty()) {
+  ComputeUniqueImports();
+
+  if (unique_imports_.empty()) {
     Write("typedef struct ", ModuleInstanceTypeName(), " ", OpenBrace());
     return;
   }
 
-  ComputeUniqueImports();
-
   // define names of per-instance imports
   for (const Import* import : module_->imports) {
+    if (SkipImport(import)) {
+      continue;
+    }
     DefineImportName(import, import->module_name, import->field_name);
   }
 
@@ -1219,6 +1243,13 @@ void CWriter::BeginInstance() {
     Write("/* import: '", import->module_name, "' '", import->field_name,
           "' */", Newline());
 
+    if (options_.no_sandbox && import->kind() == ExternalKind::Global &&
+        import->field_name == kStackPointerName) {
+      Write("/* ", kStackPointerName,
+            " implementation is replaced with thread-local array */",
+            Newline());
+      continue;
+    }
     switch (import->kind()) {
       case ExternalKind::Global:
         WriteGlobal(cast<GlobalImport>(import)->global,
@@ -1335,17 +1366,33 @@ void CWriter::WriteModuleInstance() {
 
 void CWriter::WriteGlobals() {
   Index global_index = 0;
-  if (module_->globals.size() != module_->num_global_imports) {
-    for (const Global* global : module_->globals) {
+  for (const Global* global : module_->globals) {
+    if (options_.no_sandbox && global->name == kStackPointerName) {
+      // Remember the global to write it later (it needs to be part of .c file)
+      stack_pointer_global_ = global;
+    } else {
       bool is_import = global_index < module_->num_global_imports;
       if (!is_import) {
         WriteGlobal(*global, DefineGlobalScopeName(ModuleFieldType::Global,
                                                    global->name));
         Write(Newline());
       }
-      ++global_index;
     }
+    ++global_index;
   }
+}
+
+void CWriter::WriteStackPointerGlobal(const Global& global) {
+  assert(options_.no_sandbox);
+  // Ideally we would want to have a guard page here in order to
+  // catch stack-overflows. For now just add 4k as a buffer page.
+  // We also add a range check against kStackSize on GlobalSet,
+  // note that since leaf functions do not set the __stack_pointer
+  // they won't trigger the range check in global.set
+  Write("wasm_thread_local uint8_t ", kStackArrayVariableName, "[", kStackSize,
+        " + ", kStackBufferSize, "];", Newline());
+  Write("wasm_thread_local uintptr_t ", kStackOffsetGlobalVariableName,
+        " = sizeof(", kStackArrayVariableName, ");", Newline());
 }
 
 void CWriter::WriteGlobal(const Global& global, const std::string& name) {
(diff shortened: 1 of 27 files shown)


keithw:
Heya, here are some first impressions. I hope these are helpful to you; it seems like great news for wasm2c if the project can successfully provide benefit to Android in this way.

a. It could be helpful to include a writeup about the intended semantics of this (which of course can evolve over time as necessary). Otherwise, it will be challenging for non-Googlers to intelligently review or contribute to maintaining this. (Most WABT options/modes are implementing WebAssembly proposals in various phases of the proposal process where there's already a public writeup and discussion.)

b. It would also be useful to have a test that demonstrates/tests the intended behavior of this mode. The stack-pointer.txt test in this PR also passes without `--no-sandbox`, so I didn't find it as helpful as I had hoped in understanding "what's different" about this mode.

c. It looks (tentatively) like this code makes two significant changes:
   1. Loads and stores are done as if the memory "data" pointer is NULL and bounds-checks are disabled (effectively, as if WASM_RT_MEMCHECK_SIGNAL_HANDLER is enabled or "size" is infinite).
   2. wasm2c tries to detect if a global represents the stack_pointer in the producer's shadow stack (by looking at the global's debug name) and, if so, omits the global and transforms its gets and sets. For a get, the generated code instead gets the address of a thread-local byte array, plus a thread_local offset. For a set, the generated code does some bounds-checks against the size of the thread-local array and then does the set.

For (1), I think I'm getting it. My personal view is that if possible, it would be nice in terms of maintainability if we can find a way to minimize the number of special-case checks throughout c-writer.cc. At the extreme, I believe this **could** be a runtime-only change (i.e., not changing c-writer.cc or src/templates at all), if the embedder implements `wasm_rt_allocate_memory` to return a memory with data=0 and size=UINT64_MAX and `wasm_rt_grow_memory` and `wasm_rt_free_memory` to do nothing. Maybe you don't love that for performance reasons, in which case, it seems like this PR has already made all the changes necessary in `wasm2c.declarations.nosandbox.c` when it changes DEFINE_LOAD and DEFINE_STORE to ignore the memory and its data pointer. Is it also necessary to change the way LoadExprs and StoreExprs are emitted in c-writer.cc?

For (2), I'm a bit more confused unfortunately.
- I'm not sure identifying the shadow stack pointer by its debug name is viable. LLVM doesn't appear to give a debug name to the stack pointer in its relocatable object files. E.g. if I compile something like:
   ```c
   void func2(int *x);
   void func1(void) { int x; func2(&x); }
   ```
   with `clang -Os -c -o minimal.o minimal.c` (LLVM commit dd109f615), the stack pointer global doesn't seem to have a debug name. (It does get one after linking...) And, custom sections (including the name section) aren't normative or required; it seems a bit fragile to depend on them for correctness.
- If I run `wasm2c --no-sandbox -o minimal_wasm2c.c minimal.o`, it aborts with `wasm2c: /home/keithw/stanford/wabt/src/c-writer.cc:719: void wabt::{anonymous}::CWriter::Write(const wabt::{anonymous}::GlobalName&): Assertion `global_sym_map_.count(mangled) == 1' failed.` (Maybe because the stack pointer is imported under the expected import names, but doesn't have the expected debug name?)
- It seems a bit risky to have `g_w2c_stack_array` and `g_w2c_stack_offset` as thread-local file-scope globals (different from the per-memory/per-instance scope of the formal spec). This seems like it can cause trouble if the host calls an instance's exported function from one thread, the function leaves something in this region of its memory, and then the host calls an exported function on the same instance from a different thread, or on a different instance from the same thread. Maybe we don't expect the Wasm module to treat its shadow stack as persistent across returns to the host, but I don't know of a rule against this. I wonder if the more "WebAssembly-y" design might be to let the module declare or import its memory in the instance as normal (but as an optimization, could fix its size to be the initial value of the stack pointer), and let it hold the shadow stack unchanged? I.e., let that memory be imported/exported as normal, and initialize the stack pointer to be the absolute address of the end of this memory? I think this would match the intended semantics for "well-behaved" modules (every instance would have a different stack pointer and would end up writing into its instance's memory) and might handle some of those edge cases better. Or another "WebAssembly-y" design would be to rewrite the stack_pointer to be an import, and then the host has to supply it at instantiation-time as an absolute pointer to an array that it has allocated for this purpose. (Just like the caller has to allocate the stack when using the clone(2) syscall on Linux.)
- In any event, from a wasm2c perspective, it would be really nice if we could avoid special-casing individual globals and imports. To Wasm, the stack_pointer isn't special at all, and it would be nice to avoid treating it specially in wasm2c if possible. Do you think it would be possible to factor some of this out into a separate Wasm-to-Wasm transformation that happens upfront, and keep the added complexity and special cases out of wasm2c? Is it possible to simply add a relocation for the stack_pointer initialization value, so that the shadow stack would be placed in virtual memory by the same mechanism that places everything else in nonconflicting memory? Or, if not, what about a Wasm-to-Wasm transformation that rewrites stack_pointer as an import, and then the host supplies the import at runtime as a pointer to the end of the instance's stack-sized memory, or to an array that the host declares on its own stack?

src/c-writer.cc
```diff
@@ -3317,9 +3388,13 @@ void CWriter::Write(const LoadExpr& expr) {
   Memory* memory = module_->memories[module_->GetMemoryIndex(expr.memidx)];
 
   Type result_type = expr.opcode.GetResultType();
-  Write(StackVar(0, result_type), " = ", func, "(",
-        ExternalInstancePtr(ModuleFieldType::Memory, memory->name), ", (u64)(",
-        StackVar(0), ")");
+  if (options_.no_sandbox) {
```
keithw, line 3391: Given the changes to wasm2c.declarations.nosandbox.c, could this case be eliminated? (I.e. ignore the memory in the actual load function.)

src/c-writer.cc
```diff
@@ -3346,8 +3421,13 @@ void CWriter::Write(const StoreExpr& expr) {
 
   Memory* memory = module_->memories[module_->GetMemoryIndex(expr.memidx)];
 
-  Write(func, "(", ExternalInstancePtr(ModuleFieldType::Memory, memory->name),
-        ", (u64)(", StackVar(1), ")");
+  if (options_.no_sandbox) {
```
keithw, line 3424: Same comment here (could this be kept the same, given the changes to wasm2c.declarations.nosandbox.c)?

dimitry-:
> Heya, here are some first impressions. I hope these are helpful to you; it seems like great news for wasm2c if the project can successfully provide benefit to Android in this way.
> 
> a. It could be helpful to include a writeup about the intended semantics of this (which of course can evolve over time as necessary). Otherwise, it will be challenging for non-Googlers to intelligently review or contribute to maintaining this. (Most WABT options/modes are implementing WebAssembly proposals in various phases of the proposal process where there's already a public writeup and discussion.)
> 
> b. It would also be useful to have a test that demonstrates/tests the intended behavior of this mode. The stack-pointer.txt test in this PR also passes without `--no-sandbox`, so I didn't find it as helpful as I had hoped in understanding "what's different" about this mode.

I suppose every test that works for no-sandboxed mode will pass in sandboxed one. One use-case I can think of is a module using native libc. But at this point I am not sure how to write that kind of test. But even these might be ok if say wasi provides full (sandboxed) support of libc.

> 
> c. It looks (tentatively) like this code makes two significant changes:
> 
> 1. Loads and stores are done as if the memory "data" pointer is NULL and bounds-checks are disabled (effectively, as if WASM_RT_MEMCHECK_SIGNAL_HANDLER is enabled or "size" is infinite).
> 2. wasm2c tries to detect if a global represents the stack_pointer in the producer's shadow stack (by looking at the global's debug name) and, if so, omits the global and transforms its gets and sets. For a get, the generated code instead gets the address of a thread-local byte array, plus a thread_local offset. For a set, the generated code does some bounds-checks against the size of the thread-local array and then does the set.
> 
> For (1), I think I'm getting it. My personal view is that if possible, it would be nice in terms of maintainability if we can find a way to minimize the number of special-case checks throughout c-writer.cc. At the extreme, I believe this **could** be a runtime-only change (i.e., not changing c-writer.cc or src/templates at all), if the embedder implements `wasm_rt_allocate_memory` to return a memory with data=0 and size=UINT64_MAX and `wasm_rt_grow_memory` and `wasm_rt_free_memory` to do nothing. Maybe you don't love that for performance reasons, in which case, it seems like this PR has already made all the changes necessary in `wasm2c.declarations.nosandbox.c` when it changes DEFINE_LOAD and DEFINE_STORE to ignore the memory and its data pointer. Is it also necessary to change the way LoadExprs and StoreExprs are emitted in c-writer.cc?

That makes sense - thanks for the suggestion. I will see if I can do that by just modifying `wasm_rt` functions.

> 
> For (2), I'm a bit more confused unfortunately.
> 
> * I'm not sure identifying the shadow stack pointer by its debug name is viable. LLVM doesn't appear to give a debug name to the stack pointer in its relocatable object files. E.g. if I compile something like:
>   ```c
>   void func2(int *x);
>   void func1(void) { int x; func2(&x); }
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   with `clang -Os -c -o minimal.o minimal.c` (LLVM commit dd109f615), the stack pointer global doesn't seem to have a debug name. (It does get one after linking...) And, custom sections (including the name section) aren't normative or required; it seems a bit fragile to depend on them for correctness.

This is concerning.. I guess from the wasm code point of view there is nothing special about stack_pointer. But we need a way to recognize it in wasm code. Are there less fragile alternatives for this?

> * If I run `wasm2c --no-sandbox -o minimal_wasm2c.c minimal.o`, it aborts with `wasm2c: /home/keithw/stanford/wabt/src/c-writer.cc:719: void wabt::{anonymous}::CWriter::Write(const wabt::{anonymous}::GlobalName&): Assertion `global_sym_map_.count(mangled) == 1' failed.` (Maybe because the stack pointer is imported under the expected import names, but doesn't have the expected debug name?)
> * It seems a bit risky to have `g_w2c_stack_array` and `g_w2c_stack_offset` as thread-local file-scope globals (different from the per-memory/per-instance scope of the formal spec). This seems like it can cause trouble if the host calls an instance's exported function from one thread, the function leaves something in this region of its memory, and then the host calls an exported function on the same instance from a different thread, or on a different instance from the same thread. Maybe we don't expect the Wasm module to treat its shadow stack as persistent across returns to the host, but I don't know of a rule against this. I wonder if the more "WebAssembly-y" design might be to let the module declare or import its memory in the instance as normal (but as an optimization, could fix its size to be the initial value of the stack pointer), and initialize the stack pointer to be the absolute address of the end of this memory?

I do not belive stack is supposed to be persistent between calls. I was thinking about passing pointers to the stack to a function running in a different thread - but this use-case should work as expected, since we will pass pointers and not offsets.

As the pointer - my initial change was setting a pointer to the end of the stack array - unfortunatelly clang refused to compile it since it cannot do such initialization for `__thread` variables.

> I think this would match the intended semantics for "well-behaved" modules (every instance would have a different stack pointer and would end up writing into its instance's memory) and might handle some of those edge cases better. Or another "WebAssembly-y" design would be to rewrite the stack_pointer to be an import, and then the host has to supply it at instantiation-time as an absolute pointer to an array that it has allocated for this purpose. (Just like the caller has to allocate the stack when using the clone(2) syscall on Linux.)

> * In any event, from a wasm2c perspective, it would be really nice if we could avoid special-casing individual globals and imports. To Wasm, the stack_pointer isn't special at all, and it would be nice to avoid treating it specially in wasm2c if possible. Do you think it would be possible to factor some of this out into a separate Wasm-to-Wasm transformation that happens upfront, and keep the added complexity and special cases out of wasm2c? Is it possible to simply add a relocation for the stack_pointer initialization value, so that the shadow stack would be placed in virtual memory by the same mechanism that places everything else in nonconflicting memory? Or, if not, what about a Wasm-to-Wasm transformation that rewrites stack_pointer as an import, and then the host supplies the import at runtime as a pointer to the end of the instance's stack-sized memory, or to an array that the host declares on its own stack?

This might be tricky since there is going to be no host in no-sandbox mode (that is if I understand the host correctly). Stack-pointer will need to be initialized every time `pthread_create()` for example is called. Considering that the native libc is not aware of the library linking against it running wasm - it won't be able to do anything special about it.

We have also considered replacing `__stack_pointer` with access to native stack. Maybe we want to reevaluate it.

dimitry-:
Some thoughts after another look at the imports/instances and runtime.

We want the compiled module to preserve function signatures from Wasm. Think of `pthread_create` - it expect a function with specific signature `void* (*)(void*)`, but currently during generation wasm2c adds an `instance_t*` parameter to every function. So we probably want to remove this otherwise we will need to track every function pointer/callback. And this means we also want to remove `memory_t` references. This makes changes to the generator a requirement. And it also means that stack_pointer needs to be a global variable.

Offtopic: It is my understanding that Wasm does not have globals in the way C++ understands them, all globals are thread-local, I wonder if it is going to present a problem. Is everything in .data segment treated as thread-local for `wasm` target?

sbc100:
> Some thoughts after another look at the imports/instances and runtime.
> 
> We want the compiled module to preserve function signatures from Wasm. Think of `pthread_create` - it expect a function with specific signature `void* (*)(void*)`, but currently during generation wasm2c adds an `instance_t*` parameter to every function. So we probably want to remove this otherwise we will need to track every function pointer/callback. And this means we also want to remove `memory_t` references. This makes changes to the generator a requirement. And it also means that stack_pointer needs to be a global (or rather thread-local) variable.

Yeah, it sounds like you would an option to remove the "instancing" feature and assume a single global instance instead.

> 
> Offtopic: It is my understanding that Wasm does not have globals in the way C++ understands them, all globals are thread-local, I wonder if it is going to present a problem. Is everything in .data segment treated as thread-local for `wasm` target?

No that won't be a problem.  A wasm global is thread local, but llvm doesn't generate wasm globals at all (aside from a couple of special ones like __stack_pointer).   Global data in the c/c++ sense all lives in linear memory.

keithw:
> Some thoughts after another look at the imports/instances and runtime.
> 
> We want the compiled module to preserve function signatures from Wasm. Think of `pthread_create` - it expect a function with specific signature `void* (*)(void*)`, but currently during generation wasm2c adds an `instance_t*` parameter to every function.

Yeah -- a WebAssembly "module" is sort of like a class in C++. And a module "instance" is like a C++ object of that class. Or another way to look at it would be that the "module" is similar to an executable on disk, and the module "instance" is like a running process.

Wasm functions are like C++ member functions: they act on the state associated with their originating instance. (A C++ method also takes the instance pointer, i.e. "this", as its first parameter in the ABI.) Or somewhat similarly, a userspace function in Linux is executed in the context of a particular kernel process with its VM mappings, namespaces, signal handlers, register file, etc. -- that's basically the "instance" there.

In Wasm, there is one global runtime-wide structure called the "store" that all functions/globals/memories/tables/tags notionally live in. So you can have multiple instances (of the same module, or of different modules) share the same global or memory, as long as one instance declares it and the others import it from there.

> So we probably want to remove this otherwise we will need to track every function pointer/callback. And this means we also want to remove `memory_t` references. This makes changes to the generator a requirement. And it also means that stack_pointer needs to be a global (or rather thread-local) variable.

Honestly I'm not sure I can comment intelligently on the proposed/evolving design without a clearer view of what you're trying to accomplish. Is there any change you could share a writeup? (To the extent this is changing the semantics of WebAssembly, are you open to writing this up as a proposal like the other WABT features?)

I guess I fear there may be a lot of parts of WebAssembly you're not going to like when you get to them (not just modules having multiple instances; there's probably a bunch of things involving tables, imports/exports, exceptions, etc. that don't have an obvious "no sandbox" interpretation).

Candidly, I think the situation I'd love to avoid is ending up with a lot of spaghetti code in c-writer.cc to support what is basically an internal Google experiment that nobody else understands. :-/

> I do not belive stack is supposed to be persistent between calls.

I think the design challenge is... says who? This is not a WebAssembly rule; afaik there's nothing that distinguishes "memory below this index" from "memory above this index" to let us safely say that "memory below this index" is the shadow stack which should be cloned upon `pthread_create` but doesn't need to be persistent, whereas "memory above this index" is the heap which can be shared across threads and does need to be persistent.

I don't doubt that LLVM-produced code acts like this, but it feels a bit icky to have wasm2c starting to depend on the implicit behavior of a particular Wasm producer. :-/

matthias-blume:
Sorry for jumping in here...

There seems to be a bit of an impasse here.  We discussed the situation at our end and decided that we will, for the time being, do our development on a separate branch, while trying to stay reasonably up to date with the main branch and the tip of the tree.  I believe that we could relatively easily come to an agreement anyway, but this approach will let us push forward in the short term without needing to provide lengthy design documents etc., and we can revisit the issue of incorporating our changes into the main branch at a later point.  Hopefully many open issues have been worked out by then and coherent documentation can be written.

All this being said, I wanted to clarify some of the goals and properties of our work relative to "normal" uses of WASM.  Hopefully this will put everybody's mind at ease:

- We have no plans at all to change the existing WASM design, its properties or semantics.
- All our changes will be gated by a flag (--no-sandbox, for the time being further protected by --experimental).  Without these flags there will be no change in behavior in any of the tools within the wabt suite, in particular wasm2c.
- Our goal is to focus on a particular special case of WASM usage where there is only one module (and only a single instance of that module at runtime), one memory, one function table, ..., and where the instructions are the result of compiling C (or C-like) code.
- In this special situation we want to let wasm2c render this module in C in such a way that the output is directly interoperable (to the maximum extent that we can achieve) with plain native code.  There will be no sandbox, no bounds checks, the indirect function table is elided (function pointers are code addresses), and the one memory is "real" memory (memory addresses are relative to 0).  What is crucial here is that all data and functions use the same conventions and representations that native data and functions use.  In particular, it means that in this mode we don't need the module argument in functions, and - indeed - we can't afford to have such an argument since it would render function signatures incompatible with corresponding native signatures.

In any case, as I said above, none of these modifications would kick in unless specifically called upon via flags.

dimitry-:
we moved this to the fork.

PR #2107, opened 2022-12-20
Add relaxed SIMD to wabt.js features

marcusb:


src/emscripten-exported.json
```diff
@@ -1,59 +1,61 @@
...
-"_wabt_write_binary_module",
-"_wabt_write_binary_spec_script",
-"_wabt_write_module_result_get_result",
-"_wabt_write_module_result_release_log_output_buffer",
-"_wabt_write_module_result_release_output_buffer",
-"_wabt_write_text_module",
-"_dummy_workaround_for_emscripten_issue_7073"
-]
+  "_free",
+  "_malloc",
```
sbc100, line 3: why are these lines showing up as different?  did you newlines change maybe?

surma: Looks like a change in indentation to me

sbc100: Oh yes, I'm got to make a change to avoid using json here at all..

sbc100 (approved):

Issue #1906, opened 2022-04-25
Make online wat2wasm respect the enable exception flag


hamishwillee:
Tried https://webassembly.github.io/wabt/demo/wat2wasm/ feeding in some simple Wat with exception definitions:
![image](https://user-images.githubusercontent.com/5368500/165032910-6e6a82be-83b4-4858-b6e7-2810b5e36c99.png)
I do not get the errors if I run the tool manually.

zherczeg:
It does not seems trivial. It seems the exceptions boolean is written into some `p` array, which is probably an (emscripten) emulated memory. If I read it back with `Module["_wabt_exceptions_enabled"](5626592)` (the number seems an address passed to `Module["_wabt_set_exceptions_enabled"]`) I got `1`. Unfortunately there is a lot of generated code here and hard to debug its intention.

hamishwillee:
I thought it more likely an issue in the code that calls "generic" [wat2wasm/](https://webassembly.github.io/wabt/demo/wat2wasm/). Because if you manually call the tool from the command line with exceptions enabled flag you do not get an error.

zherczeg:
There is a lot happening here. The online code is generated from the C source code, and the generator creates some bridge between the js world and the C world. I can confirm that the bridge reads the "exceptions" flag correctly, and store it as value 1 in some internal array called p. I suspect it is a memory like interface, since the location is called `addr`. I am sure the intention is to use this value later, but it does not happen (or an incorrect address is used), and it is hard to debug because the code is generated.

binji:
It looks like the web version just needs to be updated. I rebuilt it locally and your example worked for me!

binji:
https://github.com/WebAssembly/wabt/pull/1907

keithw:
The online wat2wasm demo now seems to unconditionally allow exceptions whether the box is checked or not...

binji:
I noticed that too, I think there may be a few places where the exceptions flag isn't properly checked.

keithw:
We seem to have a (different) bug where a tag element is allowed even without `--enable-exceptions`, but this isn't particular to the online wat2wasm. Closing this issue.

PR #2109, opened 2022-12-23
Enable features in libwabt.js by default

marcusb:
This applies the same defaults for features in the JavaScript-compiled version as the native tools. Without this change, enabling only the "reference_types" feature did not work, as it got turned off due to the dependency on "bulk_memory", which is counter-intuitive (see Features::UpdateDependencies()).


marcusb:
@sbc100

keithw:
Neat... do you happen to know if this also fixes #1906?

marcusb:
> Neat... do you happen to know if this also fixes #1906?

I cannot reproduce any of the problems mentioned there, with or without this patch.

keithw:
Hmm, me too now. Thanks, will close it.

marcusb:
But it did make me think that the demo should be updated, so I changed it to match the native feature defaults as well.

docs/demo/wasm2wat/demo.js
```diff
@@ -65,7 +64,7 @@ editorContainer.ondrop = function(e) {
   readAndCompileFile(file);
 }
 var fileBuffer = null;
-for (var feature of FEATURES) {
+for (var feature of Object.keys(FEATURES)) {
```
sbc100, line 67: Below you use ` for (feature in FEATURES)`.. should we be consistent?

sbc100 (approved):

src/wabt.post.js
```diff
@@ -86,9 +86,8 @@ function allocateCString(s) {
 /// Features
 function Features(obj) {
   this.addr = Module._wabt_new_features();
-  for (var i = 0; i < FEATURES.length; ++i) {
-    var feature = FEATURES[i];
-    this[feature] = obj[feature] | 0;
+  for (feature in FEATURES) {
+    this[feature] = obj[feature] | FEATURES[feature] | 0;
```
marcusb, line 90: oof this is a bug, will rewrite

src/wabt.post.js
```diff
@@ -305,16 +315,16 @@ WasmModule.prototype.toText = function(options) {
   var writeModuleResult_addr =
       Module._wabt_write_text_module(this.module_addr, foldExprs, inlineExport);
 
-  var result = Module._wabt_write_module_result_get_result(
-      writeModuleResult_addr);
+  var result =
+      Module._wabt_write_module_result_get_result(writeModuleResult_addr);
```
sbc100, line 319: Unrelated change?   Maybe revert this is so?

src/wabt.post.js
```diff
@@ -305,16 +315,16 @@ WasmModule.prototype.toText = function(options) {
...
 
   try {
     if (result !== WABT_OK) {
       throw new Error('toText failed.');
     }
 
-    var outputBuffer = new OutputBuffer(
-        Module._wabt_write_module_result_release_output_buffer(
+    var outputBuffer =
+        new OutputBuffer(Module._wabt_write_module_result_release_output_buffer(
```
sbc100, line 327: Ditto

src/wabt.post.js
```diff
@@ -86,28 +100,24 @@ function allocateCString(s) {
...
+Object.keys(FEATURES).forEach(
+    feature => Object.defineProperty(Features.prototype, feature, {
+      enumerable: true,
+      get: function() {
+        return Module['_wabt_' + feature + '_enabled'](this.addr);
+      },
+      set: function(newValue) {
+        Module['_wabt_set_' + feature + '_enabled'](this.addr, newValue | 0);
+      }
+    }));
```
sbc100, line 120: Can you avoid re-indenting this entire block?  It looks like the just the first line is changing?

src/wabt.post.js
```diff
@@ -86,28 +100,24 @@ function allocateCString(s) {
 /// Features
 function Features(obj) {
   this.addr = Module._wabt_new_features();
-  for (var i = 0; i < FEATURES.length; ++i) {
-    var feature = FEATURES[i];
-    this[feature] = obj[feature] | 0;
-  }
+  allFeatures().forEach((v, f) => this[f] = booleanOrDefault(obj[f], v));
```
sbc100, line 103: How about `for (const [f, v] of Object.entries{FEATURES}) { ... }`   ?

docs/demo/wat2wasm/demo.js
```diff
@@ -52,9 +39,9 @@ var base64El = document.getElementById('base64');
 var binaryBuffer = null;
 var binaryBlobUrl = null;
 
-for (var feature of FEATURES) {
-  var featureEl = document.getElementById(feature);
-  features[feature] = featureEl.checked;
+for (const [f, v] of wabt.allFeatures()) {
```
sbc100, line 42: How about `for (const [f, v] of Objet.entries(wasm.FEATURES) } {`?  Then you don't need to new `addFeatures` function at all.

src/wabt.post.js
marcusb: These were from clang-format, the linter complains otherwise.

src/wabt.post.js
marcusb: Emscripten just [doesn't like it](https://github.com/WebAssembly/wabt/actions/runs/3784684407/jobs/6434167954#step:3:1025), I don't know why.

marcusb: think I found a way around it, dropping the `const` :shrug:

docs/demo/wasm2wat/demo.js
```diff
@@ -14,35 +14,19 @@
  * limitations under the License.
  */
 
+WabtModule().then(function(wabt) {
```
sbc100, line 17: Why not keep this down on line 29?

sbc100 (approved):
lgtm assuming you tested locally and it works.

marcusb:
Yes, tested the demos and the wabt.js build locally.

PR #2110, opened 2022-12-23
wast-parser.cc: disallow exception tag unless exceptions enabled

keithw:
Fixes a bug first noticed at #1906.

### DIFF
--- a/test/regress/regress-38.txt
+++ b/test/regress/regress-38.txt
@@ -1,0 +1,11 @@
+;;; TOOL: wat2wasm
+;;; ERROR: 1
+(module
+  (tag $e0 (param i32))
+)
+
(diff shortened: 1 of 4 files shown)


test/regress/regress-38.txt
```diff
@@ -0,0 +1,11 @@
+;;; TOOL: wat2wasm
```
sbc100, line 1: What is 38?  I though these used bug number?

keithw: Most of the regression tests seemed to be just numbered sequentially, so I went with that since we didn't have a bug number for this bug exactly. Can easily renumber it this this PR (done).

sbc100 (approved):

PR #2113, opened 2022-12-29
wasm2c: expose the initial and max size required from imported memory

shravanrn:
Wasm modules may have more static data than the initial memory of the memory provided (this happens particularly when using externally provided memory). Today, wasm2c simply traps if this is the case. With this change, wasm2c's `load_data()` calls `wasm_rt_grow_memory()` to handle this case.

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -226,7 +226,7 @@ class CWriter {
   static const char* GetReferenceTypeName(const Type& type);
   static const char* GetReferenceNullValue(const Type& type);
 
-  enum class WriteExportsKind {
+  enum class CWriterPhase {
     Declarations,
     Definitions,
   };
@@ -287,12 +287,13 @@ class CWriter {
   void WriteDataInitializers();
   void WriteElemInitializers();
   void WriteElemTableInit(bool, const ElemSegment*, const Table*);
-  void WriteExports(WriteExportsKind);
+  void WriteExports(CWriterPhase);
   void WriteInitDecl();
   void WriteFreeDecl();
   void WriteInit();
   void WriteFree();
   void WriteInitInstanceImport();
+  void WriteImportProperties(CWriterPhase);
   void WriteFuncs();
   void Write(const Func&);
   void WriteParamsAndLocals();
@@ -1669,7 +1670,7 @@ void CWriter::WriteElemTableInit(bool active_initialization,
   Write(");", Newline());
 }
 
-void CWriter::WriteExports(WriteExportsKind kind) {
+void CWriter::WriteExports(CWriterPhase kind) {
   if (module_->exports.empty())
     return;
 
@@ -1685,7 +1686,7 @@ void CWriter::WriteExports(WriteExportsKind kind) {
         const Func* func = module_->GetFunc(export_->var);
         mangled_name = ExportName(MangleName(export_->name));
         internal_name = func->name;
-        if (kind == WriteExportsKind::Declarations) {
+        if (kind == CWriterPhase::Declarations) {
           WriteFuncDeclaration(func->decl, mangled_name);
         } else {
           func_ = func;
@@ -1729,7 +1730,7 @@ void CWriter::WriteExports(WriteExportsKind kind) {
         const Tag* tag = module_->GetTag(export_->var);
         mangled_name = ExportName(MangleName(export_->name));
         internal_name = tag->name;
-        if (kind == WriteExportsKind::Declarations) {
+        if (kind == CWriterPhase::Declarations) {
           Write("extern ");
         }
         Write("const u32 *", mangled_name);
@@ -1740,7 +1741,7 @@ void CWriter::WriteExports(WriteExportsKind kind) {
         WABT_UNREACHABLE;
     }
 
-    if (kind == WriteExportsKind::Declarations) {
+    if (kind == CWriterPhase::Declarations) {
       Write(";", Newline());
       continue;
     }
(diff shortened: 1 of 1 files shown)


shravanrn:
@sbc100 @keithw Looks like some of the spec tests expect the wasm memory to not grow when loading the data segment. Unfortunately, the Firefox/RLBox use case may need still need something like this PR. 

**Our use case:** For the RLBox use case we will need to use an external `wasm_rt_memory_t` (as there are expectations on memory alignment). We want to size this memory in the same way as internal memory --- start with a size large enough to hold the data segment. Grow as needed during execution. Unfortunately, this isn't possible as the size of the data segment is not available for query during runtime. 

**Possible fixes:**
1) Something like this patch, but we keep the behavior behind a macro flag
2) We add an extra field to `wasm_rt_memory_t` called `can_grow_during_init` and leave this as `false` by default in wasm2c. For the RLBox use case, we can create the memory with this field set to true
3) We expose functions from wasm2c generated code to query the size of the data segment. 

Any thoughts on how best to proceed (I am leaning towards option 1 or 2)?

sbc100:
> @sbc100 @keithw Looks like some of the spec tests expect the wasm memory to not grow when loading the data segment. Unfortunately, the Firefox/RLBox use case may need still need something like this PR.
> 
> **Our use case:** For the RLBox use case we will need to use an external `wasm_rt_memory_t` (as there are expectations on memory alignment). We want to size this memory in the same way as internal memory --- start with a size large enough to hold the data segment + grow as needed during execution. Unfortunately, this isn't possible as the size of the data segment is not available for query during runtime. Any other option involving estimating the data segment size either wastes memory if we overallocate (a major problem in 32-bit platforms), or leads to unnecessary crashes if we underallocate.
> 
> **Possible fixes:**
> 
> 1. Something like this patch, but we keep the behavior behind a macro flag
> 2. We add an extra field to `wasm_rt_memory_t` called `can_grow_during_init` and leave this as `false` by default in wasm2c. For the RLBox use case, we can create the memory with this field set to true
> 3. We expose functions from wasm2c generated code to query the size of the data segment.
> 
> Any thoughts on how best to proceed (I am leaning towards option 1 or 2)?

Hmm.. to me (3) seems like the most logical way of doing things.  IIUC the wasm module being compiled imports its memory, and should always specify the size of the imported memory such that it is large enough for its initial data segments.  

It makes sense to me that the embedding/loading code should be able see the specification of the memory that is required (min, max, etc).  But perhaps i'm missing why this would be hard to achieve with the way wasm2c works..?

wasm2c/wasm-rt.h
```diff
@@ -51,6 +51,13 @@ extern "C" {
 #define wasm_rt_unreachable abort
 #endif
 
+/**
+ * The default WASM page size
```
sbc100, line 55: "default" seems odd since this is never configurable.

Maybe you could split out this non-functional change that just renames this macro?

shravanrn: Actually, for option 3, this wouldn't be needed, so i think I can revert this change

shravanrn:
@sbc100 I think the issue with option 3 right now that the information is not available to the developer (please correct me if I am wrong here). We could add some functions that expose this information. I have made this change in this PR now. With this change, the following code gets added to output C file

```
u32 Z_my_wasm_moduleZ_envZ_memory_initsize(Z_glue_lib_wasm2c_instance_t* instance){
  return 17;
}
``` 

The embedder/loader can now call this function to get the initial page size (17 in the above example). Does this look ok to you?

sbc100:
> @sbc100 I think the issue with option 3 right now that the initial memory size information is not available to the embedder (please correct me if I am wrong here). We could add some functions that expose this information. I have updated this PR to add these functions. With this change, the following code gets added to C file output by wasm2c
> 
> ```
> u32 Z_my_wasm_moduleZ_envZ_memory_initial(Z_my_wasm_module_instance_t* instance){
>   return 17;
> }
> 
> u32 Z_my_wasm_moduleZ_envZ_memory_max(Z_my_wasm_module_instance_t* instance){
>   return 4294967295;
> }
> ```
> 
> The embedder/loader can now call these function to get the initial page size (17 in the above example) and max page size. Does this look ok to you?

That sounds like the right kind of thing yes.   Could we make these simply u32 symbols rather than function-that-return-u32?

shravanrn:
> That sounds like the right kind of thing yes. Could we make these simply u32 symbols rather than function-that-return-u32?

Updated! let me know if this looks good to land

src/c-writer.cc
```diff
@@ -226,7 +226,7 @@ class CWriter {
   static const char* GetReferenceTypeName(const Type& type);
   static const char* GetReferenceNullValue(const Type& type);
 
-  enum class WriteExportsKind {
+  enum class WriteCStmtKind {
```
sbc100, line 229: I wonder if we can find a better name?   WriterPhase?

sbc100: CWriterPhase?

sbc100 (approved):
I guess we don't have any examples/tests checked in that show how this effects the output?  I don't think we need to block on that since we are working on adding more tests that show the full output.

lgtm % naming

src/c-writer.cc
shravanrn: `CWriterPhase` sounds good

src/c-writer.cc
```diff
@@ -1910,6 +1911,39 @@ void CWriter::WriteInitInstanceImport() {
...
+    }
+
+    Write(";", Newline());
+  };
+
+  for (const Import* import : unique_imports_) {
+    if (import->kind() == ExternalKind::Memory) {
+      const Limits* memory_limits =
+          &(cast<MemoryImport>(import)->memory.page_limits);
+      write_import_prop(import, "initial", memory_limits->initial);
```
keithw, line 1938: Could we maybe call this "min" to match the WebAssembly term? (https://webassembly.github.io/spec/core/syntax/types.html#syntax-limits)

Also, would be nice to publish this for tables as well, but that can be another PR.

shravanrn: Will change to min :+1:

PR #2114, opened 2022-12-30
wasm2c: mark manually formatted code in c-writer

shravanrn:
Prevent clang-format from automatically changing manually formatted regions of code. The file has some existing annotations, but is missing some. Checked on clang-format 10


keithw (approved):
lgtm if tests pass. (Some of these @wrv also did in #2021.)

sbc100 (approved):

PR #2116, opened 2023-01-03
wasm2c: harden set/longjmp to check for uninit jmp_buf

shravanrn:
Use of powerful primitives like longjmp is of some security concern for the Firefox context. While the Firefox use case would leave all setjmp/longjmp use cases as dead code, there is still some concerns. This PR makes the wasm2c generated code more defensive when using `longjmp`. The code now checks that `jmp_buf` is initialized by `setjmp` before use with `longjmp`.

**Costs**: The costs here are minor: (1) a variable initialization during wasm_rt_try, and (2) a conditional check during exceptions or traps (which are usually not the critical path in programs).

### DIFF
--- a/wasm2c/wasm-rt.h
+++ b/wasm2c/wasm-rt.h
@@ -283,9 +283,19 @@ void wasm_rt_load_exception(uint32_t tag, uint32_t size, const void* values);
 WASM_RT_NO_RETURN void wasm_rt_throw(void);
 
 /**
+ * A hardened jmp_buf that allows us to checks if it is initialized before use
+ */
+typedef struct {
+  /* Is the jmp buf intialized? */
+  bool initialized;
+  /* jmp_buf contents */
+  jmp_buf buffer;
+} wasm_rt_jmp_buf;
+
+/**
  * The type of an unwind target if an exception is thrown and caught.
  */
-#define WASM_RT_UNWIND_TARGET jmp_buf
+#define WASM_RT_UNWIND_TARGET wasm_rt_jmp_buf
 
 /**
  * Get the current unwind target if an exception is thrown.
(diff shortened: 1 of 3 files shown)


wasm2c/wasm-rt.h
```diff
@@ -313,11 +322,14 @@ uint32_t wasm_rt_exception_size(void);
...
 #if WASM_RT_MEMCHECK_SIGNAL_HANDLER && !defined(_WIN32)
-#define WASM_RT_SETJMP(buf) sigsetjmp(buf, 1)
+#define WASM_RT_SETJMP_SETBUF(buf) sigsetjmp(buf, 1)
 #else
-#define WASM_RT_SETJMP(buf) setjmp(buf)
+#define WASM_RT_SETJMP_SETBUF(buf) setjmp(buf)
 #endif
 
+#define WASM_RT_SETJMP(buf) \
+  ((buf).initialized = true, WASM_RT_SETJMP_SETBUF((buf).buffer))
```
sbc100, line 331: I've not familiar with this comma seperated syntax in C .. is it part of C89?  Would this be more clear if it was two separate statements withing a `do/while(0)`?

wasm2c/wasm-rt.h
```diff
@@ -282,10 +282,19 @@ void wasm_rt_load_exception(uint32_t tag, uint32_t size, const void* values);
  */
 WASM_RT_NO_RETURN void wasm_rt_throw(void);
 
+/** A hardened setjmp buff that checks if the jmp buffer is initialized before
```
sbc100, line 285: We put the opening `/**` on the line by itself

sbc100 (commented):
lgtm % comments

wasm2c/wasm-rt.h
shravanrn: The comma separate syntax is pretty convenient, as it treats several statements like a single expression and returns the last statements' value. The nice bit is that it allows use to continue to use WASM_RT_SETJMP in if statements `if (WASM_RT_SETJMP(buf) {...}`. In contrast, a `do{} while(0)` impl would not work in the prior example. Fwiw, the comma syntax is already present in wasm2c runtime --- See [here](https://github.com/WebAssembly/wabt/blob/main/wasm2c/wasm-rt-impl.h#L64-L66)

sbc100: Sounds good.

sbc100 (approved):

Issue #2046, opened 2022-11-08
wasm2c use of globals and statics introduce race conditions in concurrent environments


shravanrn:
wasm2c has a number of globals and statics that assume that wasm instances are only embedded in a single threaded host. 

However, when using wasm2c in complex hosts like Firefox, different host threads can spin up multiple sandbox instances concurrently, invoke function in their respective sandboxes and destroy sandboxes. In this use case, a number of these globals and statics can lead to exploitable race conditions --- hence in the fork of wasm2c used in Firefox, we have completely eliminated the use of such global state.

Some of the current uses of global state in order of security severity are

- [wasm_rt_jmp_buf](https://github.com/WebAssembly/wabt/blob/main/wasm2c/wasm-rt-impl.c#L67) which is used in recovering from traps and if exceptions are enabled
- [wasm_rt_call_stack_depth/wasm_rt_saved_call_stack_depth](https://github.com/WebAssembly/wabt/blob/main/wasm2c/wasm-rt-impl.c#L60-L61) which is used if stack depth counting is employed on the target platform
- [wasm exception related state](https://github.com/WebAssembly/wabt/blob/main/wasm2c/wasm-rt-impl.c#L69-L73) which is used if exceptions are enabled
- [s_tag_count](https://github.com/WebAssembly/wabt/blob/main/wasm2c/wasm-rt-impl.c#L136) which is used to allocate tag numbers for exceptions
- [g_func_types/g_func_type_count](https://github.com/WebAssembly/wabt/blob/main/wasm2c/wasm-rt-impl.c#L64-L65) which is used to give unique type indexes to Wasm indirection function types

I got part way into solving this, however I don't have the cycles to completely address at the moment.

shravanrn:
From what I can tell, the above list contains two different types of state
- Module specific state
- Thread specific state

**Module specific state**
Module specific state from the above would be the [g_func_types/g_func_type_count](https://github.com/WebAssembly/wabt/blob/main/wasm2c/wasm-rt-impl.c#L64-L65)

Here is my attempt at solving this
https://github.com/WebAssembly/wabt/compare/main...PLSysSec:wabt:remove-race-conditions-module

[...]

PR #2117, opened 2023-01-03
wasm2c: lock func indexes as they are shared across wasm instances

shravanrn:
For the Firefox use case, we will be creating multiple wasm instances dynamically on concurrent threads. Currently, the  creation of the instances is racy when querying/updating the function type indexes, as this is data shared across wasm instances. This PR fixes this by locking any query/update of the function indexes. 

This PR follows a discussion at https://github.com/WebAssembly/wabt/issues/2046

### DIFF
--- a/wasm2c/wasm-rt-impl.c
+++ b/wasm2c/wasm-rt-impl.c
@@ -42,6 +42,59 @@
 #define alloca _alloca
 #endif
 
+#if !defined(_WIN32)
+
+#include <pthread.h>
+#define WASM_RT_MUTEX_TYPE pthread_mutex_t
+#define WASM_RT_CREATE_MUTEX(lock)        \
+  do {                                    \
+    if (pthread_mutex_init(&(lock), NULL) != 0) \
+      abort();                            \
+  } while (0)
+#define WASM_RT_DESTROY_MUTEX(lock)          \
+  do {                                       \
+    if (pthread_mutex_destroy(&(lock)) != 0) \
+      abort();                               \
+  } while (0)
+#define WASM_RT_MUTEX_LOCK(lock)          \
+  do {                                    \
+    if (pthread_mutex_lock(&(lock)) != 0) \
+      abort();                            \
+  } while (0)
+#define WASM_RT_MUTEX_UNLOCK(lock)          \
+  do {                                      \
+    if (pthread_mutex_unlock(&(lock)) != 0) \
+      abort();                              \
+  } while (0)
+
+#else
+
+#define WASM_RT_MUTEX_TYPE HANDLE
+#define WASM_RT_CREATE_MUTEX(lock)       \
+  lock = CreateMutex(NULL, FALSE, NULL); \
+  do {                                   \
+    if (!lock)                           \
+      abort();                           \
+  } while (0)
+#define WASM_RT_DESTROY_MUTEX(lock) \
+  do {                              \
+    if (CloseHandle(lock) == 0)     \
+      abort();                      \
+    lock = 0;                       \
+  } while (0)
+#define WASM_RT_MUTEX_LOCK(lock)                              \
+  do {                                                        \
+    if (WaitForSingleObject(lock, INFINITE) != WAIT_OBJECT_0) \
+      abort();                                                \
+  } while (0)
+#define WASM_RT_MUTEX_UNLOCK(lock) \
+  do {                             \
+    if (ReleaseMutex(lock) == 0)   \
+      abort();                     \
+  } while (0)
+
+#endif
+
 #define PAGE_SIZE 65536
 #define MAX_EXCEPTION_SIZE PAGE_SIZE
 
@@ -68,6 +121,9 @@ uint32_t wasm_rt_saved_call_stack_depth;
 
 static FuncType* g_func_types;
 static uint32_t g_func_type_count;
+// Mutex for the func type indexes which are shared across multiple Wasm
+// instances, potentially running on different threads
+static WASM_RT_MUTEX_TYPE g_func_type_mutex;
 
 jmp_buf wasm_rt_jmp_buf;
 
@@ -125,23 +181,37 @@ uint32_t wasm_rt_register_func_type(uint32_t param_count,
     func_type.results[i] = va_arg(args, wasm_rt_type_t);
   va_end(args);
 
-  for (i = 0; i < g_func_type_count; ++i)
-    if (func_types_are_equal(&g_func_types[i], &func_type))
-      return i + 1;
+  bool found = false;
+  uint32_t result = 0;
 
-  // This is a new/unseed type.  Copy our stack allocated params/results into
-  // permanent heap allocated space.
-  wasm_rt_type_t* params = malloc(param_size);
-  wasm_rt_type_t* results = malloc(result_size);
-  memcpy(params, func_type.params, param_size);
-  memcpy(results, func_type.results, result_size);
-  func_type.params = params;
-  func_type.results = results;
+  WASM_RT_MUTEX_LOCK(g_func_type_mutex);
 
-  uint32_t idx = g_func_type_count++;
-  g_func_types = realloc(g_func_types, g_func_type_count * sizeof(FuncType));
-  g_func_types[idx] = func_type;
-  return idx + 1;
+  for (i = 0; i < g_func_type_count; ++i) {
+    if (func_types_are_equal(&g_func_types[i], &func_type)) {
+      found = true;
+      result = i + 1;
+      break;
+    }
+  }
+
+  if (!found) {
+    // This is a new/unseed type.  Copy our stack allocated params/results into
+    // permanent heap allocated space.
+    wasm_rt_type_t* params = malloc(param_size);
+    wasm_rt_type_t* results = malloc(result_size);
+    memcpy(params, func_type.params, param_size);
+    memcpy(results, func_type.results, result_size);
+    func_type.params = params;
+    func_type.results = results;
+
+    uint32_t idx = g_func_type_count++;
+    g_func_types = realloc(g_func_types, g_func_type_count * sizeof(FuncType));
+    g_func_types[idx] = func_type;
+    result = idx + 1;
+  }
+
+  WASM_RT_MUTEX_UNLOCK(g_func_type_mutex);
+  return result;
 }
 
 uint32_t wasm_rt_register_tag(uint32_t size) {
@@ -334,6 +404,7 @@ static void os_cleanup_signal_handler(void) {
 #endif
 
 void wasm_rt_init(void) {
+  WASM_RT_CREATE_MUTEX(g_func_type_mutex);
 #if WASM_RT_MEMCHECK_SIGNAL_HANDLER && !WASM_RT_SKIP_SIGNAL_RECOVERY
   if (!g_signal_handler_installed) {
     g_signal_handler_installed = true;
@@ -351,6 +422,8 @@ bool wasm_rt_is_initialized(void) {
 }
 
 void wasm_rt_free(void) {
+  // We don't lock g_func_type_mutex here as we expect there to be no running
+  // wasm instances when this function is called.
   for (uint32_t i = 0; i < g_func_type_count; ++i) {
     free(g_func_types[i].params);
     free(g_func_types[i].results);
@@ -363,6 +436,7 @@ void wasm_rt_free(void) {
 #if WASM_RT_MEMCHECK_SIGNAL_HANDLER && !WASM_RT_SKIP_SIGNAL_RECOVERY
   os_cleanup_signal_handler();
 #endif
+  WASM_RT_DESTROY_MUTEX(g_func_type_mutex);
 }
 
 void wasm_rt_allocate_memory(wasm_rt_memory_t* memory,


wasm2c/wasm-rt-impl.c
```diff
@@ -68,6 +104,9 @@ uint32_t wasm_rt_saved_call_stack_depth;
 
 static FuncType* g_func_types;
 static uint32_t g_func_type_count;
+// Mutex for the func type indexes which are shared across multiple Wasm
+// instances
```
sbc100, line 108: ", potentially running on different threads"

sbc100: Would a valid alternative simply be to make g_func_types and g_func_type_count into TLS?   I guess that would require just as much work since C11 `_Thread_local` keyword cannot be relied on here?

wasm2c/wasm-rt-impl.c
```diff
@@ -351,6 +405,7 @@ bool wasm_rt_is_initialized(void) {
 }
 
 void wasm_rt_free(void) {
+  WASM_RT_MUTEX_LOCK(g_func_type_mutex);
```
sbc100, line 408: Presumably is reasonable to assert that there are no active instances when this is called  (since all best would be off if they were)?  So technically maybe the lock isn't needed here?

wasm2c/wasm-rt-impl.c
```diff
@@ -125,23 +164,37 @@ uint32_t wasm_rt_register_func_type(uint32_t param_count,
     func_type.results[i] = va_arg(args, wasm_rt_type_t);
   va_end(args);
 
-  for (i = 0; i < g_func_type_count; ++i)
-    if (func_types_are_equal(&g_func_types[i], &func_type))
-      return i + 1;
+  bool found = false;
```
sbc100, line 167: I guess you could have used `goto done;` to avoid this extra variable, and the extra indentation of the !found block below.. but this approach doesn't seem much worse.

wasm2c/wasm-rt-impl.c
```diff
@@ -42,6 +42,42 @@
 #define alloca _alloca
 #endif
 
+#if !defined(_WIN32)
+
+#include <pthread.h>
+#define WASM_RT_MUTEX_TYPE pthread_mutex_t
+#define WASM_RT_CREATE_MUTEX(lock) \
+  do {                             \
```
sbc100, line 50: Why not call pthread_mutex_init and pthread_mutex_destroy as part of these macros?   Seems maybe important, especially if you don't use PTHREAD_MUTEX_INITIALIZER.

wasm2c/wasm-rt-impl.c
```diff
@@ -42,6 +42,42 @@
...
+  } while (0)
+#define WASM_RT_DESTROY_MUTEX(lock) \
+  do {                              \
+  } while (0)
+#define WASM_RT_MUTEX_LOCK(lock)        \
+  if (pthread_mutex_lock(&(lock)) != 0) \
+  abort()
```
sbc100, line 57: Does clang-format not indent the abort calls here?

wasm2c/wasm-rt-impl.c
```diff
@@ -42,6 +42,42 @@
...
+#include <pthread.h>
+#define WASM_RT_MUTEX_TYPE pthread_mutex_t
+#define WASM_RT_CREATE_MUTEX(lock) \
+  do {                             \
+  } while (0)
+#define WASM_RT_DESTROY_MUTEX(lock) \
+  do {                              \
+  } while (0)
+#define WASM_RT_MUTEX_LOCK(lock)        \
+  if (pthread_mutex_lock(&(lock)) != 0) \
```
sbc100, line 56: Maybe wrap all these macros using the `do { } while(0)` pattern?

wasm2c/wasm-rt-impl.c
```diff
@@ -42,6 +42,42 @@
...
+
+#else
+
+#define WASM_RT_MUTEX_TYPE HANDLE
+#define WASM_RT_CREATE_MUTEX(lock)       \
+  lock = CreateMutex(NULL, FALSE, NULL); \
+  if (!lock)                             \
+  abort()
+#define WASM_RT_DESTROY_MUTEX(lock) \
+  if (CloseHandle(lock) == 0)       \
```
sbc100, line 70: maybe add `lock = 0`?

sbc100 (commented):
lgtm % comments

wasm2c/wasm-rt-impl.c
shravanrn: > ", potentially running on different threads"
:+1: 

> Would a valid alternative simply be to make g_func_types and g_func_type_count into TLS?

Independent of how we could implement TLS, I think the bigger issue is that TLS wouldn't work here. This is because `g_func_types` and `g_func_type_count` must have the same value when queried from (different instances on) different threads

sbc100: Can you explain why instances need to share `g_func_types`?   Its it an internal detail?

wasm2c/wasm-rt-impl.c
shravanrn: Yup, I considered not having it as well. I don't have a strong preference one way or the other. Let me know if you prefer not having it.

wasm2c/wasm-rt-impl.c
shravanrn: It removes the indent unfortunately

wasm2c/wasm-rt-impl.c
shravanrn: yikes, my bad! Will fix

wasm2c/wasm-rt-impl.c
sbc100: No string preference here, just confirming my assumptions.  Maybe add a comment?

wasm2c/wasm-rt-impl.c
shravanrn: Adding a semi-colon after the abort and wrapping the whole thing in a do-while fixes the clang-formatter. Will make this change

wasm2c/wasm-rt-impl.c
shravanrn: @sbc100 Actually, can we assume C11? If yes, I can probably use [C11 mutex support](https://en.cppreference.com/w/c/thread/mtx_lock). For what its worth, the next PR I plan to submit will be to make use of TLS vars for things like exception handling, stack depth vars etc. which is again much simpler if we have C11.

shravanrn: > Can you explain why instances need to share g_func_types? Its it an internal detail?

@sbc100 I was mostly basing my reasoning on @keithw 's comment from [here](https://github.com/WebAssembly/wabt/issues/2046#issuecomment-1306638191)
Let me know if you believe this is not necessary, in which case, we can probably just move this to some sort of "module state" instead of global state

sbc100: > @sbc100 Actually, can we assume C11? If yes, I can probably use [C11 mutex support](https://en.cppreference.com/w/c/thread/mtx_lock). For what its worth, the next PR I plan to submit will be to make use of TLS vars for things like exception handling, stack depth vars etc. which is again much simpler if we have C11.

I think wasm2c is supposed to generate C89 and I assume that we would want to extend that to the rt impl.. so I think that answer is no, we should avoid anything outside of C89.

sbc100: > > Can you explain why instances need to share g_func_types? Its it an internal detail?
> 
> @sbc100 I was mostly basing my reasoning on @keithw 's comment from [here](https://github.com/WebAssembly/wabt/issues/2046#issuecomment-1306638191) Let me know if you believe this is not necessary, in which case, we can probably just move this to some sort of "module state" instead of global state

Maybe lets wait for Keith to chime in then.  I'm fine with this type of locking, or making types module internal.   Because this only happens during module init I think that locking should have minimal impact.

keithw:
Thanks for looping me in. I guess my view hasn't really changed from the last discussion; briefly:

- I see the runtime implementation included in the WABT repo as basically just an example. It's far from production-grade in many respects. For the limited purposes it's used for, I don't think this is a big problem[^1], and I think it's better to err towards keeping it simple if possible. There's no single answer to what it means to "transform wasm to C," so it seems like it's better for wasm2c to plan for flexibility in the embedder's runtime implementation (and let us debate what the API should be), rather than try to maintain a runtime that meets the needs of a mega-app like Chrome or Firefox. It's going to be hard enough to do that in the actual wasm2c transpiler, which to me is the real shared ground here.
- If Firefox needs module initialization (and trap catching and exception handling) to be thread-safe, it totally makes sense that they would need more sophistication in their runtime. If the proposal is that we should make WABT's own runtime implementation fully re-entrant and bear that burden upstream, **that** seems like the big decision to make, rather than this individual PR. Because it seems to lead down a path where we're adding locks, thread_locals, and the multi-platform equivalents in various places, and then perhaps also trying to test/fuzz the new thread-safety properties we're adding.
- I'm not super into that because I think WABT is a small project that's currently somewhat underwater in terms of complexity, fuzz failures, etc., and I'm wary of biting off more unless we think the change is of global benefit. But if we do decide to do that, as before my suggestion would be to start a second runtime implementation, rather than adding this complexity piecemeal to the "example" runtime and pushing the maintenance burden on everybody. And ideally some Mozilla people will show up to contribute to its maintenance and provide healthy discussion about what the API boundary should be. (Although if I were Firefox... wouldn't I want to use C++ to represent some of these runtime data structures?)

- For the particular use-case in the PR (RLBox in a Firefox-like app), it seems like there may be a nicer approach anyway, for two reasons:
   - The output of `wasm_rt_register_func_type` only has to be unique within the context of "a module and the modules it transitively imports funcrefs and funcref tables from." For RLBox, I assume this is just "the module."
   - If you know that scope upfront, `wasm_rt_register_func_type` can be implemented as a pure function with no side effects, e.g. consult a precomputed/immutable mapping from functypes to unique integers and return the correct value. This might be cleaner than locking the table and mutating it...
  
[^1]: Lots of ISO C and POSIX functions can't be implemented re-entrantly at all afaik, e.g. localtime, getpwnam, seed48, etc. Not saying this is great, but at least the wasm2c runtime functions can be implemented reentrantly if you need to!

wasm2c/wasm-rt-impl.c
keithw: Longer comment below. The (integer) func types have to be unique for different types across the context of "a module and the modules it transitively imports funcrefs and funcref tables from," because this is how `call_indirect` is type-checked at runtime. So basically, two modules need to share g_func_types if one of them imports a funcref or funcref table from the other.

sbc100: I see, that makes sense.  So even for MVP modules that don't include first class funcrefs, we still need to share functype indexes if we want to call_indirect across module boundaries, which, in the general case, we most definitely do.

sbc100:
> Thanks for looping me in. I guess my view hasn't really changed from the last discussion; briefly:
> 
> * I see the runtime implementation included in the WABT repo as basically just an example. It's far from production-grade in many respects. For the limited purposes it's used for, I don't think this is a big problem[1](#user-content-fn-1-18f54a8fe15de09945d87c11a35ed36a), and I think it's better to err towards keeping it simple if possible. There's no single answer to what it means to "transform wasm to C," so it seems like it's better for wasm2c to plan for flexibility in the embedder's runtime implementation (and let us debate what the API should be), rather than try to maintain a runtime that meets the needs of a mega-app like Chrome or Firefox. It's going to be hard enough to do that in the actual wasm2c transpiler, which to me is the real shared ground here.
> * If Firefox needs module initialization (and trap catching and exception handling) to be thread-safe, it totally makes sense that they would need more sophistication in their runtime. If the proposal is that we should make WABT's own runtime implementation fully re-entrant and bear that burden upstream, **that** seems like the big decision to make, rather than this individual PR. Because it seems to lead down a path where we're adding locks, thread_locals, and the multi-platform equivalents in various places, and then perhaps also trying to test/fuzz the new thread-safety properties we're adding.
> * I'm not super into that because I think WABT is a small project that's currently somewhat underwater in terms of complexity, fuzz failures, etc., and I'm wary of biting off more unless we think the change is of global benefit. But if we do decide to do that, as before my suggestion would be to start a second runtime implementation, rather than adding this complexity piecemeal to the "example" runtime and pushing the maintenance burden on everybody. And ideally some Mozilla people will show up to contribute to its maintenance and provide healthy discussion about what the API boundary should be. (Although if I were Firefox... wouldn't I want to use C++ to represent some of these runtime data structures?)

Attempting to answer these points together.   I agree that the wasm-rt implementation that ships with wasm2c can most likely never achieve the goal of being all things to all people but think we can and should strive to make it is as complete and robust as possible, within reason.   I don't think I agree (at least not at this point in time) with idea that it should be considered just "example code", but even that were true, its not a reason to avoid all complexity.   I think we should see it more as a "reference implementation", and strive for production quality where we can.

Obviously, there is a line we will be walking in terms of how much complexity we want to deal with, and how much we want to resist complexity that seems implementation specific.

Perhaps this upstreaming will be futile because there will be some core wasm-rt feature from rlbox that we really don't want upstream and they will ultimately be stuck on a fork anyway.   But the smaller the fork needs to be the better our reference implementation is IMHO.

I think multi-threading is going to widely used/useful and implementing the wasm threads proposal is something we will want to do.  I can imagine perhaps also implementing the new wasi-threads proposal so that we can run the tests coming or of that project.

I suppose I can imagine forking the runtime, as you suggest, into runtime-simple, and runtime-complex, but that has some major downsides in terms of our testing matrix.  I don't think this kind of PR pushes us over the limit where we would want to take on that burden. 

> * For the particular use-case in the PR (RLBox in a Firefox-like app), it seems like there may be a nicer approach anyway, for two reasons:
>   
>   * The output of `wasm_rt_register_func_type` only has to be unique within the context of "a module and the modules it transitively imports funcrefs and funcref tables from." For RLBox, I assume this is just "the module," which means the "func_types" structure only needs to be per-module, and it's only used at module initialization time. So for the RLBox use case, it might be enough to make this structure thread-local.
>   * Or, if you know all the modules upfront, `wasm_rt_register_func_type` could be implemented as a pure function with no side effects, e.g. consult a precomputed/immutable mapping from functypes to unique integers and return the correct value. This might be cleaner than mutating anything at module initialization time.
> 
> ## Footnotes
> 1. Lots of ISO C and POSIX functions can't be implemented re-entrantly at all afaik, e.g. localtime, getpwnam, seed48, etc. Not saying this is great, but at least the wasm2c runtime functions can be implemented reentrantly if you need to! [leftwards_arrow_with_hook](#user-content-fnref-1-18f54a8fe15de09945d87c11a35ed36a)

wasm2c/wasm-rt-impl.c
sbc100: Actually, not that I understand that scope of this index space I do think that "thread local" might be be better approach here.   The reason is that neither tables nor funcrefs are currently shara-able between wasm threads.   So while it makes sense that these type indexes be shared between modules on the same thread, so doesn't make sense to shared them between threads.

So some kind of TLS would probably make more sense here.   In the wasm thread model that we have today, each thread would need to separately instantiate each module.

sbc100:
> If the consensus is to stick with one runtime, and we want that runtime become fully re-entrant, I think it may be possible to solve some of the root problems a little more definitively vs. adding locks incrementally.
> 
> Let me try to summarize my understanding: The purpose of g_func_types is to maintain a mapping from each unique functype to a unique identifier. The identifier is compared at runtime in CALL_INDIRECT to check equality of types. The mapping from functypes to identifiers needs to be consistent across any scope of "a module and its transitive imports," which is why g_func_types is a global right now. But that makes it inconvenient to add new functype->identifier mappings in parallel.
> 
> I think we could move to a world where:
> 
> * we eliminate the `wasm_rt_register_func_type` runtime API endpoint, and the g_func_types data structure, completely
> * instead, wasm2c generate a deterministic identifier for the function type at "transpile-time" instead of module-initialization time
> 
> What if wasm2c serializes the param and result types to an opaque string in the module's .c file, and the func_types identifier becomes a pointer to this string? We can do nice things like always zero-pad to the nearest 2/4/8/16 bytes (which makes a memcmp quite fast, especially given that only one side of a CALL_INDIRECT needs to deref the pointer -- the other side is always a known string). This might let us get out of the whole business entirely.

I like the idea of avoiding the whole registration system.    With this proposal make type checking during call_indirect into a memcpy?  I guess that downside there is that the cost of checking the sig would be O(number of arg)?

Given that each thread has it own instance of each module, do you think that making the mapping thread-local might be simpler for now?   Its not possible to call_indirect from one thread to another so it should be fine for each thread to have its own mapping.



> 
> Godbolt example: https://godbolt.org/z/GvTYTrfE7

keithw:
> I like the idea of avoiding the whole registration system. With this proposal make type checking during call_indirect into a memcpy?

Well, a memcmp, but I think in the common case it would actually be faster than today. Right now the type checking during call_indirect has to fetch two i32s from memory (both of which are opaque identifiers that came from wasm_rt_register_func_type at module-initialization time) and then execute a `cmp` instruction to compare them.

In this proposal, for a function with fewer than ~12 params and results, I think wasm2c would be able to cram everything into 4 bytes. So now the type checking would involve executing a `cmp` instruction between an **immediate** i32 (because now the serialized type is known at compile-time) and a single i32 fetched from memory.

> I guess that downside there is that the cost of checking the sig would be O(number of arg)?

Yes in the asymptote, but in practice it will compile to a single `cmp QWORD` instruction up to ~24 params and results, and I suspect (hope) the overhead of the memcmp is dwarfed by the overhead of the actual indirect function call for any reasonable number of args. There's also the benefit of being able to have one of the sides of the comparison now supplied as an immediate instead of having to compare two runtime-supplied values.

> Given that each thread has it own instance of each module, do you think that making the mapping thread-local might be simpler for now? Its not possible to call_indirect from one thread to another so it should be fine for each thread to have its own mapping.

Hmm, I'm not totally sure I follow here. Having g_func_types be lockless and thread_local seems like it would work fine **for RLBox** because I don't think RLBox cares about modules that import tables/funcrefs from other modules anyway.

But in general, what if the embedder has a module ModA that imports a table from module ModB, and the embedder chooses to run Z_ModA_init_module() on one thread while running Z_ModB_init_module() on a second thread in parallel? If we want our runtime to be re-entrant, I think that means supporting this.

sbc100:
> But in general, what if the embedder has a module ModA that imports a table from module ModB, and the embedder chooses to run Z_ModA_init_module() on one thread while running Z_ModB_init_module() on a second thread in parallel? If we want our runtime to be re-entrant, I think that would mean supporting this.

Ah I see yes, even though instances can't be shared between threads (in the current instance-per-thread model), this type table/data is module-specific rather than instance-specific?  so i can/should be shareable.

shravanrn:
@sbc100 Sorry for the delay in responding. I just read through the thread. Please correct me if I am wrong, but the conclusions here are
- TLS won't work for func indexes
- Ideally, we probably want a static signature to index scheme in place of a counter

These make sense to me. 

As a starting point, can we land this mutex approach for now (this limitation is currently a blocker for wasm2c use in Firefox) ? If the signature scheme makes sense, I can start working on implementing this in parallel to the other wasm2c PRs i am working on.

keithw:
> As a starting point, can we land this mutex approach for now (this limitation is currently a blocker for wasm2c use in Firefox) ?

Hmm, thanks for this. I think it might help us all get on the same page better (for this issue and working on an ongoing basis) if you were able to expand a bit to help us understand Firefox's needs better here. I would love to have Mozilla be using the main branch of wasm2c and contributing on an ongoing basis, but I honestly don't get why these issues would be blockers.

- Does Firefox plan to use the WABT runtime completely unmodified? I can see wanting to upstream things if possible, but even our tiny academic research project has tweaked the runtime implementation to embed wasm2c output in ways that don't always make sense to upstream. I would be surprised if a gigantic mega-app isn't going to need to customize the runtime implementation even a little bit for its own needs, in ways that don't always make sense upstream (maybe because they use Firefox-specific data structures or API calls, etc.).
- If Firefox really wants to do module initialization in parallel across multiple threads... can't they just lock this themselves? Many libraries put the responsibility for locking on the consumer (e.g. https://www.fftw.org/fftw3_doc/Thread-safety.html) instead of doing it internally; it doesn't seem that crazy.
- Isn't Firefox already using wasm2c-generated output and RLBox in release builds shipping today? What changed (are they looking to expand, or...?).
- Is there any public discussion of this on the Mozilla side that might help illuminate the constraints better?

keithw:
>     TLS won't work for func indexes

I think our conclusion was that TLS would work fine for g_func_types **for RLBox**. This data structure is global is to support modules importing funcref tables or funcrefs from other modules, which (AIUI) RLBox modules don't do; they only import functions and only from the host, right?

keithw:
Welp, I went and implemented the "wasm2c-time global functype identifiers" scheme (#2120). Curious to get both of your thoughts, but, given how complicated it ended up being, I can see the argument that the runtime-impl-only-change is easiest here. At least then, runtimes that have more knowledge about the scopes and linking relationships between modules can solve this without locks.

shravanrn:
RLBox doesn't really use interactions across multiple modules, but passes in callbacks from the host which has the same requirements. Callbacks may not be created in the same thread as they are invoked, so TLS wouldn't work in the RLBox setting. 

Here is a simple example

_Wasm rlbox module code_

```C
typedef int(*int_int_func_t)(int); 

int w2c_func(Z_rlbox_instance_t* wasm_instance, u32 /* int_int_func_t */ f, int param) {
  int_int_func_t fn = (int_int_func_t) wasm_instance.w2c_T0.data[f];
  return fn(wasm_instance, param);
}
```

_Firefox code_

```C
Z_rlbox_instance_t* wasm_instance = nullptr;
wasm_rt_funcref_t func_val;

int host_func(Z_rlbox_instance_t* wasm_instance, int a) {
  return a + 1;
}

// Firefox thread 1: create the instance and the callback

Z_rlbox_init_module();
wasm_instance = Z_rlbox_instantiate();

wasm_rt_funcref_t func_val;
func_val.func_type = wasm_rt_register_func_type( /* args */ 2, WASM_RT_I32, WASM_RT_I32, /* return */ 1, WASM_RT_I32);
func_val.func = reinterpret_cast<wasm_rt_function_ptr_t>(host_func);
func_val.module_instance = &wasm_instance;

// Firefox thread 2: set the callback and invoke functions

uint32_t free_index = // unused index in sandbox_callback_table;
wasm_instance.w2c_T0.data[free_index] = func_val;

uint32_t result = w2c_func(wasm_instance, free_index, 5);
assert(result == 6);

```

(responding to the other questions in a new post)

shravanrn:
@keithw Re locks vs signature scheme vs passing in the func_index context --- I think in the long run, no strong preferences and either solution sounds fine to me (with a slight preference for the signature scheme --- i may have a much simpler way to implement this than the approach in the closed PR). 

@keithw @sbc100 However, in the short term, given that the signature prototype proved complicated to build, I think we should probably land the locks in the interrim. I am wary of blocking on designing/implementing a complex solution (either option above), until we have a concrete use case where this complexity is actually necessary. For the Firefox use case, the only extra thing we need right now is security --- we can't ship code with race conditions. Initialization and callbacks are not on the critical path and so locks would be fine.

keithw:
Ok, if you just want to get this in the upstream runtime with the understanding that things will probably change again soon, fine with me I guess.

shravanrn:
sounds good!
@sbc100 @keithw any chance could sign-off on this PR?

wasm2c/wasm-rt-impl.c
shravanrn: (conversation resolved in the main conversation thread)

sbc100 (approved):

shravanrn:
Sorry for the delay, back from travel. I won't land this just yet, given that #2120 has some momentum. 

My plan: If #2120 pending concerns get resolved and we are able to land it, I'll close this PR. If not, I'll land this PR as an interim measure.

PR #2120, opened 2023-01-06
Eliminate wasm_rt_register_func_type in wasm2c runtime API

keithw:
Kicked off by #2117, here is a PR that makes wasm2c serialize each function type, rather than registering function types at module-initialization time in a global registry maintained by the runtime. The serialized function type is the SHA-256 of the mangled param and result types, separated by a space.

At runtime in call_indirect, a known (immediate) function type is compared against the function type stored in a funcref structure. For call_indirects to functions local to the module, or for any call_indirect with a toolchain that merges string constants across compilation units (generally, GCC and clang), this is done by checking the equality of the pointers to each function type. Otherwise, the actual 32-byte values are compared.

The function type IDs can be looked up at runtime with `Z_[modname]_get_func_type`, which matches the API from `wasm_rt_register_func_type` (which goes away with this PR). I added a `callback` example to demo this.

I did a ton of microbenchmarking for a "foreign-module" call_indirect (https://gist.github.com/keithw/aaf17bf55165fe0de522c5c8dbfc6d4d) and the summary is that:
- in the cases I tested, GCC/clang+ld were able to merge the identical string constants even across compilation units, which means that performance is essentially unchanged (we're just testing pointer equality instead of i32 equality). In every case I tested, it was within +/- 1 cycle of the current main branch.[^1]

[^1]: On an AMD 4750U, the performance with clang actually improves over the main branch, from 6 cycles per iteration down to 5 cycles per iteration. GCC with PDO is unchanged, at around 5.25 cycles per iteration. GCC without PDO takes a hit from 5 cycles per iteration up to 6 cycles per iteration. On an Intel i7-9700T, the performance with GCC is unchanged (7 cycles per iteration either way), and clang takes a 1-cycle hit (from 6 cycles to 7 cycles per iteration).)

- on MSVC, this doesn't seem to be possible (because a `static const char * const` can't be used in a static initializer, so the function types have to be static const char arrays). I didn't benchmark it on Windown, but on GCC or clang, there's generally a penalty of around 1 or 1.5 cycles from having to do the 256-bit compare. Again, this is only for call_indirects that cross modules; a "same-module" call_indirect is just the pointer comparison.

wasm2c does the SHA-256 either by using OpenSSL's libcrypto or, if not available or if requested with `cmake -DUSE_INTERNAL_SHA256=ON`, by using a vendored (header-only) PicoSHA2. (For something security-sensitive, it seemed like we should prefer to use a well-known SHA-256 implementation if possible.) There is no runtime dependency on SHA-256 in the wasm2c runtime or generated modules.


keithw:
Closing as I think there may be a much simpler way to make everybody happy here.

keithw:
And, reopening after some the "much simpler way" turned out to be more vexing than I had hoped (see comment).

shravanrn:
@keithw The broad approach looks good to me. However, I think the one change that would be great is if we can pull the type serializer into its own separate file/library so I can pull this in for RLBox also. This would let me avoid re-implementing the type serializer in RLBox to dynamically construct a type index. 

One point I had when I quickly look at the code is that the complexity of the type serializer makes me a bit nervous, however I don't know of a way to make this any simpler. I was hoping to do a closer code review, but I am traveling at the moment, so I may only be able to get to this in a few days.

keithw:
> However, I think the one change that would be great is if we can pull the type serializer into its own separate file/library so I can pull this in for RLBox also.

I think that would be quite doable, but there's a question about how you want to use it. E.g. are you able to link with libwabt to get it? That would be one easy way to make this happen without duplicating anything.

keithw:
Or... maybe the easiest approach here is for wasm2c to stick a bunch of `#define`s in the generated header file, giving a reasonable name for each type in the module, e.g.:
```c
#ifndef Z_w2c_type_param_i32_i64_return_externref
#define Z_w2c_type_param_i32_i64_return_externref (wasm_rt_func_type_t){{.val = 260}, 8})
#endif
```

That way you could just use the type directly in your code (worst case the type doesn't exist in any module and the code fails to compile, which should be an error anyway since there's no way it can successfully be call_indirect'ed), and it doesn't seem like _that_ big of a deal to pollute the global (preprocessor) namespace with these `#define` names.

src/template/wasm2c.declarations.c
```diff
@@ -15,10 +15,19 @@
 
 #define UNREACHABLE TRAP(UNREACHABLE)
 
-#define CALL_INDIRECT(table, t, ft, x, ...)             \
-  (LIKELY((x) < table.size && table.data[x].func &&     \
-          table.data[x].func_type == func_types[ft]) || \
-       TRAP(CALL_INDIRECT),                             \
+static inline bool func_types_eq(wasm_rt_func_type_t a, wasm_rt_func_type_t b) {
+  if (UNLIKELY(a.len != b.len)) {
```
mishermasher, line 19: Drive-by comment, feel free to disregard: if you guaranteed that the `uint64_t` and `const char*` representations never collided (e.g. by always aligning the string allocations to an even byte boundary and setting the LSB of the uint64_t encoding to 1) then const char* equality implies string equality as well, which let's you remove two branches in the common case and just do:

if (a.type.val == b.type.val) {
  return true;
}
if (a.type.val & 1 || a.type.val & 1) {
  return false;
}
if (a.type.len != b.type.len) {
  return false;
}
return !memcmp(a.type.str, b.type.str, a.len);

keithw: Hmm, thanks! I think only **one** of those branches is there today; the "a" argument (including a.len) is always known at compile time. But I think yours saves another branch.

It may be a little tricky to figure out how to arrange this:
- I don't know how to guarantee how the linker aligns a static char* string literal (do you?).
- Right now, every bit in the uint64_t is spoken for.
- What are the rules on the legality of comparing `a.type.val == b.type.val` if "str" is the active member of the union (and if the two members aren't the same length, e.g. on a 32-bit system)? How do we make this safe?

mishermasher: Re alignment: With C11 you have _Alignas(2), but maybe we have to support further back than that. One option might be to simply use "static const uint16_t[]" instead of "static const char[]".

On that note, you could maybe also use \0 termination in the array to save you from having to also store length (saves a bit on cache/memory pressure, maybe).

In terms of validity, this sort of thing is fine when operating over trivial/POD types (you would want to make sure your pointers are 64-bit though and pad them out if not -- maybe use uint64_t regardless to store the value and do a pointer cast when using as a pointer instead). Those are the rules for C++ so I imagine C is also fine with it.

That said, if all bits are spoken for then disregard. I just thought I'd offer this up as I'm starting to look into using wasm2c and indirect call performance will be important.

keithw: Hmm, there may be a much simpler way to do this. What if the functype is a pointer (perhaps to a {length,const char*}, or maybe to four uint64_t's that are the output of a SHA-256), and the condition for equality is that *either* the pointers are identical, or the target of the pointers are identical? And then each module can canonicalize its own function types so that identical types get an identical pointer.

That way, a call_indirect within the module will be a single pointer comparison, end of story. And a call_indirect for a funcref imported from another module will generally involve comparing four uint64_t's (or a memcmp between two length-tagged memory regions).

How would you feel about that... is it enough for "same-module" call_indirects to be extremely fast, and "other-module" call-indirects to be, I dunno, still extremely fast but probably 2 cycles or something?

mishermasher: Sounds good to me! Much cleaner. Thank you!

keithw: Ok done, and thanks for the idea! I think this is a lot simpler, and because of (string) constant merging in the GNU/LLVM toolchains, it basically doesn't affect performance compared with the main branch because pointer equality is almost always the end of the story. Also there's no difference anymore between "large-signature" functions and short-signature functions.

keithw:
I went through a bunch of revisions on this and I think I'm happy where it's landed now. Because of string interning/constant pooling in the GNU/LLVM toolchains, there's essentially no performance impact (no matter the size of the function signature), even when making a call_indirect across modules.

In cases where the actual 256-bit values have to be compared (i.e., the call_indirect crosses modules AND the toolchain didn't merge the string constants across compilation units), it's generally a 1 or 1.5 cycle penalty.

shravanrn:
Sorry catching up after 10 days away

> I think that would be quite doable, but there's a question about how you want to use it. E.g. are you able to link with libwabt to get it? That would be one easy way to make this happen without duplicating anything.

It will be a tough sell to link all of wabt in Firefox, due to the binary bloat. Realistically, this would need to be separated out into its own files. (I am happy to help pull this out if needed). 

That said, I don't think this needs to change the internal library configuration in the wabt project. If we can pull out just the function that generates the func type into its own file which only imports wasm-rt.h and standard C headers, this would be perfectly fine. I am happy to pull just that one file into the runtime we would use for Firefox.

> Other option would be to just duplicate the functionality in the wasm runtime 

Do you mean in the runtime that is checked in? Sure, that would be fine. If you mean duplicate it in the rlbox-wasm runtime, that would be fine if I can easily pull in just these files and not all of wabt (same as prev point). The only thing i'd want to avoid is having a second implementation as it feels like things can get out of sync. 

> Or... maybe the easiest approach here is for wasm2c to stick a bunch of #defines in the generated header file, giving a reasonable name for each type in the module, e.g.:

Unfortunately, this would not be usable by RLBox. RLBox tries to make dealing with wasm'd libraries identical to using non-Wasm'd libraries. Thus when handling callbacks (host exposed functions), RLBox automates things like func_type index lookup by (1) reflecting on the callback function signature and (2) calling the procedure to get the func_type index based on the reflection results. Unfortunately, C++ does not offer a way to pick one of many macros based on reflection. Thus a simple static procedure for func type index generation `u32 get_func_idx(wasm_rt_type_t  ret, wasm_rt_type_t params ... )` would be preferable. 

I'll go through the latest changes to the PR shortly. Some quick thoughts 
- Using strings for comparison and relying on merging for perf is clever.
- I didn't fully get where the hashing is used yet, but I would love to avoid relying on an extra dependency just for this.

I'll have a closer look at the code and get back

shravanrn:
@keithw Just a thought: Could we skip the hashing step altogether and simply output the mangled string? That way, we cam compare with `strcmp`. Since the string merging will most likely reduce this to a pointer compare, it seems like we would maintain the perf benefits without pulling in a dependency for hashing.

keithw:
> @keithw Basic question: Could we skip the hashing step altogether and simply output the mangled string? That way, we can compare with `strcmp`. Since the string merging will most likely reduce this to a pointer compare, it seems like we would maintain the perf benefits without pulling in a dependency for hashing.

I guess I'd love to avoid thinking about the uncertain cost of comparing unbounded-length strings at runtime. (The original version of this PR of course had that issue, but worked hard to fit "most" function types in a single uint64_t -- at the cost of a somewhat complex serializer that is gone now.)

It seems comforting to know that the worst case is a single 256-bit compare (which is a single cycle when compiling for contemporary x86-64 with AVX-2, and about 1.5 cycles on x86-64 otherwise). Of course all this only applies when the linker wasn't smart enough to merge the string constants in the first place (but I'm not sure how reliable that happens for arbitrary-length string constants -- again it seems more comfortable to know they're always 32 bytes exactly).

> EDIT: just saw the concern about cross module perf. If this is a concern and we want to stick with hashing, I think PicoSHA2 maybe a nicer option, just so that wabt runtime doesn't pull in more external dependencies than necessary (in essence, avoiding all parts of openssl would be my preferred option :) )

Well, not sure if this was clear, but the wasm2c **runtime** has zero added dependencies from this. No hashing happens at runtime -- only at wasm2c-time.

The wasm2c tool does end up needing to do SHA-256 somehow. In the current version of this PR, CMake tries to find OpenSSL, but if it's configured not to via the command line (or if it's not available in the places that CMake looks), wasm2c ends up using the vendored (internal) PicoSHA2 and the resulting wasm2c doesn't end up with any external library dependency. So if you want a wasm2c free of OpenSSL dependencies, that's pretty easy to arrange (and either way the runtime and compiled code have no hashing at all, whether OpenSSL or internal). How do you feel about that / does it work for you?

deian:
Cool stuff. Personally I think the other option is safer and simpler.

Bringing OpenSSL in the TCB seems super scary and for a bunch of use cases would conflict with system that ship with crypto libs (NSS, libsodium, etc.). It's also not obvious what the performance implications are beyond these micro benchmarks (in contrast, the simpler lock approach already met demands of workloads where latency matters).

keithw:
Hmm, I don't understand this objection. It's not going to conflict with any other crypto libs -- as I said, the runtime is completely unaffected. The runtime doesn't do any hashing.

If you don't want to link your wasm2c (tool) with OpenSSL to get SHA-256, you don't have to -- it will use the internal PicoSHA2. If you want this to always use the internal PicoSHA2, honestly that's fine with me -- I'm not wedded to how we get SHA-256 in the tool. It seemed preferable to use an established implementation when possible, but I don't feel strongly about it.

shravanrn:
> If you'd prefer to always use the internal PicoSHA2 ... that would be fine with me 

Sure, I think we are on the same page. Let's go with the PicoHash as the default standard in wasm2c. 

To clarify our concerns re OpenSSL:
- wasm2c choice of mechanism to generate func indexes has to be replicated in the rlbox runtime (we would like to literally use the same files and code to avoid any discrepancies)
- so any wasm2c compile dependency shows up as a rlbox runtime dependency
- we would like to avoid OpenSSL dependency in the rlbox runtime

src/c-writer.cc
```diff
@@ -212,6 +213,8 @@ class CWriter {
   std::string DefineLocalScopeName(std::string_view);
   std::string DefineStackVarName(Index, Type, std::string_view);
 
+  static void SerializeFuncType(const FuncType&, std::string&);
```
shravanrn, line 216: If we can move this function into its own .cc file that depends only on wasm-rt.h and sha256.h, that would be really helpful for the rlbox use case. We could probably move this functionality into the current `sha256.c` file and rename that to something like `func-type-serialize.c`.

CMakeLists.txt
```diff
@@ -100,6 +100,15 @@ check_include_file("unistd.h" HAVE_UNISTD_H)
 check_symbol_exists(snprintf "stdio.h" HAVE_SNPRINTF)
 check_symbol_exists(strcasecmp "strings.h" HAVE_STRCASECMP)
 
+find_package(OpenSSL)
```
shravanrn, line 103: My preference is that we can default to picohash unless users want to opt into openssl. At the least, we probably want a way to avoid using OpenSSL explicitly from the command line.

src/c-writer.cc
keithw: Ok cool, let's iterate on this and I think we can find something that works for everybody. The most challenging part is going to be about mangling the individual types (e.g. how to turn an i32 into "i") in a way that's convenient. Right now, that means including wabt/type.h, which brings in wabt/base-types.h, wabt/string-format.h, and (this last one is probably optional) wabt/config.h. So, a small number of headers, but not as small as you asked for. (Unfortunately I don't think wasm2c/libwabt can depend on wasm-rt.h -- that's crossing a boundary between transpiler and runtime.)

Q: I pushed a version of this (with documentation in the wasm2c/README.md) -- is it acceptable to you? If so, awesome. :smile: 

Q: If not, would you prefer a more explicit serialization that doesn't depend on anything in libwabt, and that you could feasibly implement outside? We could make the input to the hash function be something like "(param i32 f32 i32) (result i64)" (or whatever), and then you'd hash that string with whatever SHA-256 implementation you prefer. Would that be preferable?

[Q: A third option (but more work for you and I suspect not one that makes you happy) would be to stop passing funcrefs between the host and a wasm2c-generated module. Instead you'd have to make a shim module that imports the host function (directly), and imports the library's table, and fills in the funcref entry in the table. Then you'd run that shim module through wasm2c, and then just use it at runtime. That way all the functypes would be generated by wasm2c. But I don't want you to have to change your whole application here; I think we can find a way for you to generate functypes at runtime that isn't a major burden.]

CMakeLists.txt
keithw: Ok great, I added a `-DUSE_INTERNAL_SHA256=ON` option that makes this explicit.

keithw:
OK @shravanrn, I thought about this a bunch and have a proposed approach that you might actually like. If RLBox is comfortable calling the current `wasm_rt_register_func_type` at runtime, how about if we make wasm2c generate basically the exact same API as part of the transpiled module, so you can call `Z[modname]_get_func_type( 2, 1, WASM_RT_I32, WASM_RT_F32, WASM_RT_I64 )` at runtime and get back the corresponding wasm_rt_func_type_t (or NULL if it doesn't exist in the module). You wouldn't have to change the way RLBox does reflection you wouldn't ever have to do any SHA-256ing.

keithw:
OK, done -- I hope this is what you were looking for! wasm2c now generates a `Z_[modname]_get_func_type` function whose API mirrors the `wasm_rt_register_func_type` call that RLBox is used to. For RLBox, my understanding is that passing a callback dynamically to the module will now be largely unchanged. I added a `callback` demo that gives an example. If the [module takes a callback at runtime](https://github.com/WebAssembly/wabt/blob/w2c-elim-global-func-types/wasm2c/examples/callback/callback.wat), then [here's what the host-side code](https://github.com/WebAssembly/wabt/blob/w2c-elim-global-func-types/wasm2c/examples/callback/main.c) to pass in the type-checked callback can look like.

shravanrn (approved):

shravanrn:
@keithw This option looks great! Thanks for all your work accommodating the RLBox use case as well. I think we can ship this!

keithw:
Wonderful! Pleasure doing business with you -- I'm glad we were able to converge.

@sbc100 Do you want to weigh in on/review this before merger?

wasm2c/examples/fac/fac.c
```diff
@@ -586,13 +591,13 @@ DEFINE_TABLE_SET(externref)
...
 DEFINE_TABLE_FILL(externref)
 
-static bool s_module_initialized = false;
-
-static u32 func_types[1];
+#if defined(__GNUC__) || defined(__clang__)
+#define FUNC_TYPE_T(x) static const char* const x
+#else
+#define FUNC_TYPE_T(x) static const char x[]
+#endif
```
sbc100, line 598: Can we not just use `static const char*` in both cases, avoiding the macro completely?

wasm2c/examples/fac/fac.h
```diff
@@ -29,9 +29,9 @@ typedef struct Z_fac_instance_t {
   char dummy_member;
 } Z_fac_instance_t;
 
-void Z_fac_init_module(void);
 void Z_fac_instantiate(Z_fac_instance_t*);
 void Z_fac_free(Z_fac_instance_t*);
+wasm_rt_func_type_t Z_fac_get_func_type(uint32_t param_count, uint32_t result_count, ...);
```
sbc100, line 34: Does this need to be a public function?

sbc100 (commented):
I don't love the extra dependency, but otherwise I guess the overall approach nicer in the longer term.

.gitmodules
```diff
@@ -13,3 +13,6 @@
 [submodule "third_party/uvwasi"]
 	path = third_party/uvwasi
 	url = https://github.com/nodejs/uvwasi
+[submodule "third_party/picosha2"]
+	path = third_party/picosha2
+	url = https://github.com/okdshin/PicoSHA2
```
sbc100, line 18: Doesn't adding this dependency make much harder to compile wasm2c output?   Seems like a fairly be downside.

wasm2c/examples/fac/fac.c
keithw: Unfortunately I couldn't figure out how -- on MSVC, a `static const char* const` doesn't seem to be a constant expression and can't be used in an initializer, which is how we represent elem segments. (On the other hand, I think C requires a `static const char[]` to have a distinct address so it can't be merged with another identical string constant.)

wasm2c/examples/fac/fac.h
keithw: I think so -- the purpose is to help the host generate a funcref that can be passed to Wasm code (as in the new "callback" example). The function isn't called by any of the wasm2c-generated code.

.gitmodules
keithw: Hmm, I don't think so -- it's only used by the wasm2c tool, not its output. (To be clear, in this PR at present, libwabt can be compiled either to link against libcrypto or with this vendored implementation, and if doing the latter, it just gets compiled in; it doesn't create a dependency.)

There isn't an added dependency for the wasm2c output, the runtime, or the host.

sbc100: Thanks for clarifying, that sounds much more reasonable.

keithw:
Ok, merging soon barring further concerns/objections. If there is unexpected trouble I hope people will let us know and we can of course revisit. I'm going to email the Debian/Ubuntu and Fedora package maintainers to give a heads-up about the (optional) added libcrypto dependency.

sbc100:
I wonder if SHA1 would be good enough (and slightly faster and less memory intensive at runtime)?  It seems like if it good enough to avoid collisions it git it should be more than good enough for this purpose.

keithw:
My understanding was that the security community has been sharply discouraging any "new" use of SHA-1 for a while now. (https://www.zdnet.com/article/its-the-end-of-sha-1-and-i-feel-fine/ is from six years ago.) Even Git has been working on a transition to SHA-256. I think it would raise some eyebrows if we picked SHA-1 in 2023. :-/

In terms of runtime cost, there's some added burden from holding 32 bytes in the binary instead of 20, but it seems nominal since I basically wasn't expecting those 32 bytes to ever get accessed. (If the wasm2c output is compiled via clang or GCC, call_indirect is generally just comparing the pointers without dereferencing them.)

sbc100:
> My understanding was that the security community has been sharply discouraging any "new" use of SHA-1 for a while now. (https://www.zdnet.com/article/its-the-end-of-sha-1-and-i-feel-fine/ is from six years ago.) Even Git has been working on a transition to SHA-256. I think it would raise some eyebrows if we picked SHA-1 in 2023. :-/

If this were some kind of crypto application I would agree.  I don't really see this as having the same bar for unbreakability though.  What is the threat model you are worried about?   I guess building a collisions detection system is probably more work that just using SHA-256?   .. so maybe you are right.  

Part of me does wonder if this solution is a little too clever/complex for its own good.. on the other hand avoiding the central/global type registry is a really nice property.

This change lgtm though.. I think these concerns a minor.

keithw:
All fair points. Re: threat model, I guess I see it as (aspirationally) the same as with a browser-based Wasm engine, in that the Wasm module might be trying really hard to escape the sandbox with a very clever (but invalid) call_indirect. I don't think we're there yet, but I see that as what we're aiming for. So would tend to favor the boring/safe choices.

Merging now, and let's hope this complexity doesn't become a recurring pain point. If so we can certainly revisit.

sbc100:
> All fair points. Re: threat model, I guess I see it as (aspirationally) the same as with a browser-based Wasm engine, in that the Wasm module might be trying really hard to escape the sandbox with a very clever (but invalid) call_indirect.

But presumably no system would ever just trust the compiled/binaries produced by wasm2c.   Any system using wasm2c would also be *running* wasm2c an not consuming the compiled code.. so it could put in place some kind of global collision detection system... but I guess that is what we had in the runtime before.

sbc100:
I also think maybe if we enforced a sensible limit of the size of the types it could make collision forging impossible, but I'm not a hash algorithm expert so its probably best not to guess at this stuff.

deian:
I don't think a sha1 pre-image attack is realistic here (I'd be more concerned about compiler bugs if an attacker that can pull of a pre-image attack is in your attacker model), but that's just my 2c

PR #2121, opened 2023-01-07
c-writer.cc: compute type values internally when possible

keithw:
This PR minimizes calls from wasm2c-generated modules to the runtime's `wasm_rt_register_func_type` function.

- Modules that don't import or export funcrefs end up never calling the runtime; they just assign deduplicated type values internally at wasm2c-time. (These are modules that have no imports or exports that are funcref-typed globals, funcref-typed tables, or functions with a funcref-typed param or result.)
- Other modules only call the runtime for non-duplicate internal types.

I think this gets us a significant distance in many practical situations towards allowing the runtime to eliminate the global functype registry; most modules just don't need it (and if you know your modules don't need it, your runtime doesn't need to implement `wasm_rt_register_func_type`). And this is vastly simpler than #2120 so perhaps easier to stomach.

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -211,6 +211,7 @@ class CWriter {
   std::string DefineGlobalScopeName(ModuleFieldType, std::string_view);
   std::string DefineLocalScopeName(std::string_view);
   std::string DefineStackVarName(Index, Type, std::string_view);
+  static bool ModuleSharesTypes(const Module& module);
 
   void Indent(int size = INDENT_SIZE);
   void Dedent(int size = INDENT_SIZE);
@@ -372,6 +373,8 @@ class CWriter {
 
   std::vector<std::pair<std::string, MemoryStream>> func_sections_;
   SymbolSet func_includes_;
+
+  bool module_shares_types_;
 };
 
 constexpr std::string_view kImplicitFuncLabel = "$Bfunc";
@@ -1054,6 +1057,39 @@ void CWriter::WriteFuncTypes() {
     return;
   }
 
+  // 1) find duplicate types in this module (TODO improve O(N^2) search)
+  std::vector<uint32_t> type_values;
+  type_values.reserve(module_->types.size());
+
+  for (size_t i = 0; i < module_->types.size(); i++) {
+    FuncType* this_type = cast<FuncType>(module_->types[i]);
+    bool duplicate_found = false;
+    for (size_t j = 0; j < i; j++) {
+      const FuncType* other_type = cast<FuncType>(module_->types[j]);
+      if (this_type->sig == other_type->sig) {
+        type_values.push_back(j);
+        duplicate_found = true;
+        break;
+      }
+    }
+    if (!duplicate_found) {
+      type_values.push_back(i);
+    }
+  }
+
+  // 2) if module doesn't share types externally, we can assign type values now
+  if (!module_shares_types_) {
+    Write(Newline());
+    Writef("static const u32 func_types[%" PRIzd "] = {",
+           module_->types.size());
+    for (size_t i = 0; i < type_values.size(); ++i) {
+      Write(type_values[i], ", ");
+    }
+    Write("};", Newline());
+    return;
+  }
+
+  // 3) otherwise, assign type values at module-initialization time
   Write(Newline());
   Writef("static u32 func_types[%" PRIzd "];", module_->types.size());
   Write(Newline());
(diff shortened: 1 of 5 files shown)


shravanrn:
I think this would break the RLBox usecase. See example [here](https://github.com/WebAssembly/wabt/pull/2117#issuecomment-1373966803).

The challenge for RLBox is twofold:
1. The embedder/host has to have the same registry as the wasm instance
2. This registry must be queryable at runtime so the host can dynamically lookup the index type of the callback in the example

This works fine with point 1, but not point 2.

This approach could work, if you also add a static array with the mappings as part of the output,  So the wasm2c generated file would need to also have something like

```C

struct FuncTypeMapping {
 uint32_t func_index;
 FuncType functype;
};

const FuncTypeMapping module_mappings [] = {
  {0, {4, 1, WASM_RT_I32, WASM_RT_I32, WASM_RT_I32, WASM_RT_I32, WASM_RT_I32}},
  {1, {1, 0, WASM_RT_I32}}
  {2, {0, 0}}
}; 
```

shravanrn:
Alternately, this PR + moving `g_func_types` and `g_func_type_count` as module specific globals that are passed in to `wasm_rt_register_func_type` is also fine.  `g_func_types` and `g_func_type_count` would have to be non static so they can be queried externally.

keithw:
Hmm, interesting. Can you give a little more info re: the broader goals of that example, and maybe we can figure out if RLBox can benefit from this?

On the specific example: it's not super-kosher for the host to reach inside the instance structure and frob the private members (e.g. the `wasm_instance.w2c_T0.data[free_index] = func_val` line). The internal structure of the instance is private to the module; only the exports are supposed to be accessible, via the accessor functions that act on the instance. If table T0 is being imported or exported, then this PR won't change anything (but wouldn't help either).

Bigger picture: I think if we want RLBox to benefit from this optimization, are you able to help me understand why the callback type has to be dynamic here? That's what's really putting you in the regime of needing to speak the same language of types as your modules. Why doesn't the module simply import a function (with known type) instead of basically importing a funcref with dynamic type?

And if it really has to be dynamic... do you know if the callback type necessarily matches any of the types in the module?

keithw:
(In any event -- if this one isn't sparking joy, how do you feel about #2120?)

shravanrn:
> Why doesn't the module simply import a function (with known type) instead of basically importing a funcref with dynamic type?

This happens because our use case is library sandboxing. One of the underlying requirements of our usecase is that there is no change to the library code. And by default libraries don't mark any imports as imports aren't a thing outside of Wasm

Instead, libraries typically require external functions to be passed in as callbacks of a given signature. 

A common example is an error callback that is invoked when an error occurs (libjpeg does this)

So the original libjpeg code may look like this

```C
typedef void(*void_void_func_t)(int); 

void_void_func_t err_func = NULL;

void jpeg_init(void_void_func_t p) {
  err_func = p;
}

jpeg_header jpeg_parse_header(char* source){
  // parse from source
  if ( /* error */) {
    err_func();
  }
}

jpeg_image jpeg_parse_image(char* source){
  // parse from source
  if ( /* error */) {
    err_func();
  }
}
```

Application code would look like this

```
void on_error(int error_code) {
  abort();
} 

  jpeg_init(&on_error);
  jpeg_parse_header(argv[1]);
  jpeg_parse_image(argv[1]);
}
```

Once we compile the library to wasm, the `on_error` function in the app is rewritten to be compatible with wasm2c (first arg takes the wasm instance etc.). Then the application puts on_error into the call indirect table and passes in indexes.

sbc100:
> On the specific example: it's not really kosher for the host to reach inside the instance structure and frob the private members (e.g. the `wasm_instance.w2c_T0.data[free_index] = func_val` line).

In this example the instance should be exporting its table in order for the host to inject a function pointer.    

If you don't export the table, but you modify anyway then I think you can break the wasm semantics.  For example, when the table the wasm-opt can make make certain assumptions.  For example, if a table is not exports and module contains no `table.set` instructions, its perfectly valid for the optimizer to convert indirect calls to direct calls where the table index is known.   In other words, you can't modify a table like this unless its exported.

keithw:
Closing in favor of #2120 which is much more comprehensive and effectively includes this PR.

shravanrn:
>  In other words, you can't modify a table like this unless its exported.

Yup, just to close the loop here, yup, we plan to mark this table as exported :)

PR #2122, opened 2023-01-07
wasm2c: eliminate wasm_rt_register_tag()

keithw:
This lets exception tags be pre-assigned at compile-time (link-time?), instead of making the runtime keep an incrementing counter. 

Since every declared tag is unique (unlike with function types which are only unique if the types differ), this is pretty feasible to do in advance -- the tag value just becomes a pointer to a static dummy variable that belongs to the module. We can count on the OS not to load two modules in the same address of virtual memory!

Another step towards reducing the need for the runtime to keep global state.

If we merge #2120 and this one, we'll end up eliminating everything that [modname]_init_module does -- that function could go away as the modules wouldn't need any runtime initialization.

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -1085,40 +1085,12 @@ void CWriter::WriteTags() {
   for (const Tag* tag : module_->tags) {
     bool is_import = tag_index < module_->num_tag_imports;
     if (!is_import) {
-      Write("static u32 ",
+      Write("static const char ",
             DefineGlobalScopeName(ModuleFieldType::Tag, tag->name), ";",
             Newline());
     }
     tag_index++;
   }
-
-  if (module_->tags.empty()) {
-    return;
-  }
-
-  Write(Newline(), "static void init_tags(void) ", OpenBrace());
-
-  tag_index = 0;
-  for (const Tag* tag : module_->tags) {
-    bool is_import = tag_index < module_->num_tag_imports;
-    if (!is_import) {
-      const FuncDeclaration& tag_type = tag->decl;
-      Index num_params = tag_type.GetNumParams();
-      Write(GlobalName(ModuleFieldType::Tag, tag->name),
-            " = wasm_rt_register_tag(");
-      if (num_params == 0) {
-        Write("0");
-      } else if (num_params == 1) {
-        Write("sizeof(", tag_type.GetParamType(0), ")");
-      } else {
-        Write("sizeof(struct ", MangleTagTypes(tag_type.sig.param_types), ")");
-      }
-
-      Write(");", Newline());
-    }
-    tag_index++;
-  }
-  Write(CloseBrace(), Newline());
 }
 
 void CWriter::ComputeUniqueImports() {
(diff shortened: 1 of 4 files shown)


src/c-writer.cc
```diff
@@ -1085,40 +1085,12 @@ void CWriter::WriteTags() {
   for (const Tag* tag : module_->tags) {
     bool is_import = tag_index < module_->num_tag_imports;
     if (!is_import) {
-      Write("static u32 ",
+      Write("static const char ",
```
sbc100, line 1088: Perhaps add a comment here?  Something like "tags are identified and compared by their address alone, so the type used here and the data stored at the address are irrelevant. "?

sbc100: I wonder if we should also at `typedef Tag char` and use that so avoid readers being confused by the use of   char here and the use of `char*` below (when I see that I wonder what string data is stored, and if its null terminated, etc).

Then the comment above could be attached to the typedef.

sbc100 (approved):

src/c-writer.cc
keithw: Done -- I made the actual variable a wasm_tag_placeholder_t (which is private to the .c file) and the public type is a wasm_rt_tag_t (typedef'd to void* in wasm-rt.h). Hope that's compatible with what you had in mind, otherwise happy to amend.

src/c-writer.cc
```diff
@@ -1085,40 +1085,18 @@ void CWriter::WriteTags() {
   for (const Tag* tag : module_->tags) {
     bool is_import = tag_index < module_->num_tag_imports;
     if (!is_import) {
-      Write("static u32 ",
+      // Tags are identified and compared solely by their (unique) address.
+      // The data stored in this variable is never read.
+      if (tag_index == module_->num_tag_imports) {
+        Write(Newline());
+        Write("typedef bool wasm_tag_placeholder_t;", Newline());
```
sbc100, line 1092: Maybe `char` here?   IIRC bool could be bigger than a char?  Also, can we use `bool` in this C89 output?

keithw: Sure, char it is! (FWIW I think wasm2c's output has a pretty enmeshed dependency on C99 at present. CALL_INDIRECT is a variadic macro, we use the "inline" keyword, we have declarations after statements within the function bodies for load and store, we use "ll" integer constants, we use for loops with a variable declaration, etc. I don't see a reason we couldn't make it C89-clean but it would be some work.)

sbc100: Perhaps I meant C99.. I can't remember what goal we had exactly.   I guess one the problem is that MSVC never really implemented C99 IIRC.   Is `bool` and `stdbool.h` in C99?

keithw: _Bool/bool/stdbool.h were added in C99, yeah. And I think you're right about MSVC. :-/

shravanrn: Fwiw, I have always found the MSVC implemented enough of C99 to be useful (experience from other transpilers). I think MSVC recently fixed some of the c99 pre-processor handling as well, so I suspect C99 in general won't be a problem on recent MSVC compilers at least.

PR #2126, opened 2023-01-12
wasm2c rt: mark temp storage for traps/exceptions as thread_local

keithw:
This PR marks the temporary storage used to handle traps, exceptions, and the call-stack limit as thread-local (when the runtime is compiled as C11 or later so `thread_local` is available).

Other than this, I _think_ the only remaining sources of "MT-unsafety" in the runtime are:

- the function type registry (removed in #2120)
- the use of sigaltstack to handle segfaults. I think we will need to refactor the runtime API for this because sigaltstack needs to be called on a per-thread basis (and the alt-stack may need to be malloc'ed and freed on a per-thread basis if we can't make it a static thread-local array).

(Also, it would be nice to have a multithreaded test of the runtime to make sure it really does work re-entrantly.)

### DIFF
--- a/wasm2c/wasm-rt.h
+++ b/wasm2c/wasm-rt.h
@@ -51,6 +51,12 @@ extern "C" {
 #define wasm_rt_unreachable abort
 #endif
 
+#if defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L)
+#define WASM_RT_THREAD_LOCAL _Thread_local
+#else
+#define WASM_RT_THREAD_LOCAL
+#endif
+
 /**
  * Enable memory checking via a signal handler via the following definition:
  *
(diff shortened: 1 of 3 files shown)


wasm2c/wasm-rt.h
```diff
@@ -116,7 +122,7 @@ extern "C" {
 #endif
 
 /** Current call stack depth. */
-extern uint32_t wasm_rt_call_stack_depth;
+extern WASM_RT_THREAD_LOCAL uint32_t wasm_rt_call_stack_depth;
```
sbc100, line 125: Does this variable need to be external/public?  Can this be moved to line 5 of `wasm2c.declarations.c` and declared as a `static` an internal name?

Since that is the only use of `WASM_RT_THREAD_LOCAL`  in the external/public header perhaps that means we can avoid defining `WASM_RT_THREAD_LOCAL` publicly?

sbc100 (approved):

shravanrn:
Looks great! The defining of WASM_RT_THREAD_LOCAL to empty string when not on C11 strikes me as a bit counter-intuitive, but we can fix that as we go forward

PR #2119, opened 2023-01-04
wasm2c: simd support (v2)

wrv:
This is a properly rebased version of https://github.com/WebAssembly/wabt/pull/2021 that the maintainers can edit.

PR #2134, opened 2023-01-26
run-spec-wasm2c.py: move WASM2C_CFLAGS to end of compiler command line

keithw:
This allows CMake to override flags like -O2 by appending -O0 from the env var.

This is helpful for allowing a "debug" test of wasm2c output, since this way we
can disable optimization by adding `-O0` to the environment variable (as used in #2081).

Split out from #2081 per request.

Once we reach convergence on #2080/#2081/#2134, I think we're ready to merge #2119 (SIMD).

### DIFF
--- a/test/utils.py
+++ b/test/utils.py
@@ -40,7 +40,7 @@ class Executable(object):
     def __init__(self, exe, *before_args, **kwargs):
         self.exe = exe
         self.before_args = list(before_args)
-        self.after_args = []
+        self.after_args = kwargs.get('after_args', [])
         self.basename = kwargs.get('basename',
                                    os.path.basename(exe)).replace('.exe', '')
         self.error_cmdline = kwargs.get('error_cmdline', True)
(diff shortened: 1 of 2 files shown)


test/utils.py
```diff
@@ -40,7 +40,7 @@ class Executable(object):
     def __init__(self, exe, *before_args, **kwargs):
         self.exe = exe
         self.before_args = list(before_args)
-        self.after_args = []
+        self.after_args = kwargs.get('after_args', [])
```
sbc100, line 43: Do we still need `*before_args`?  If not we could instead just rename `before_args` to `arfter_args`?

sbc100 (commented):
lgtm % comment

test/utils.py
keithw: Hmm, I guess you're right. Done.

sbc100 (approved):

keithw:
Hmm, that should have been squashed -- I didn't realize this was possible. Apologies for the inelegant 2 commits. :-(

PR #2137, opened 2023-02-01
wasm2c: handle duplicate names between params/locals and labels (NFC)

### LINKED ISSUE #2136: wasm2c local names for params can conflict with names for labels
This module (assembled with `--debug-names`) can cause an assertion failure in wasm2c:

```wat
(module
 (func (param $B0 i32)
       (block
	(br 0))))
```

The problem is that the `$B0` param conflicts with the generated name of the block label. This is similar to #2035. We may need to split out labels from `DefineLocalScopeName` into maybe a `DefineLabelName` or something. :-/

keithw:
Fixes #2136

(My plan had been to follow this up with a substantive PR that "prettifies" the names, which may take some time to get consensus on. This is on the road to a wasm2c "multi-c-output" mode, which will probably require the symbols in the .c file to be globally unique and no longer static so they can be compiled in parallel and then linked.)


test/regress/regress-2136.txt
```diff
@@ -0,0 +1,11 @@
+;;; TOOL: run-spec-wasm2c
+;;; ARGS: --debug-names
+(module
+ (func $B0 (export "$B0") (param $B0 i32)
```
sbc100, line 4: I'm not huge fan of these tests call "regress-XXX" .. can you name the test something like `test/wasm2c/unique-names.txt` or something like that?

Or do you refer this style?

src/c-writer.cc
```diff
@@ -215,7 +224,9 @@ class CWriter {
                         std::string_view module_name,
                         std::string_view field_name);
   std::string DefineGlobalScopeName(ModuleFieldType, std::string_view);
-  std::string DefineLocalScopeName(std::string_view);
+  std::string DefineLocalScopeName(std::string_view, bool);
```
sbc100, line 227: Can we give the new argument here useful name?

src/c-writer.cc
```diff
@@ -712,9 +743,16 @@ void CWriter::Write(std::string_view s) {
...
-  assert(local_sym_map_.count(name.name) == 1);
-  Write(local_sym_map_[name.name]);
+void CWriter::Write(const ParamName& name) {
+  std::string mangled = name.name + kParamSuffix;
+  assert(local_sym_map_.count(mangled) == 1);
+  Write(local_sym_map_[mangled]);
+}
+
+void CWriter::Write(const LabelName& name) {
+  std::string mangled = name.name + kLabelSuffix;
```
sbc100, line 753: Would it make the output more readable if this was a prefix such `l_` for labels and `p_` for params?    We can do that as part of your larger/followup change too of course.

test/regress/regress-2136.txt
```diff
@@ -0,0 +1,11 @@
...
+       (block
+	(br 0))))
+
+(assert_return (invoke "$B0" (i32.const 5)))
+(;; STDOUT ;;;
+1/1 tests passed.
```
sbc100, line 10: Once again I kind of which the full output was shown here, so that I could see that the change looks like in terms of its effect on the output.  IIRC we have an outstanding change to make that happen so it doesn't need to happen here.

test/regress/regress-2136.txt
keithw: Sure, happy to. I don't really have a preference; was just trying to match past practice.

src/c-writer.cc
keithw: Sure, it's `is_label` in the function definition. I can add a name to the declaration too if you want, although the other declarations don't seem to have them...

src/c-writer.cc
keithw: This doesn't affect the output -- it's only an internal tag to keep these names distinct in the symbol maps. The output is still (as close as possible to) the original name.

src/c-writer.cc
sbc100: I don't know why that is, we would probably add them everywhere.. but this one in particular is especially non-intuitive without a name.

test/regress/regress-2136.txt
keithw: We have the three output tests (add.txt, minimal.txt, hello.txt) but that's it right now -- I don't think we have anything pending right now. This PR doesn't change the output (except in cases of name collision) so it won't affect those.

Happy to make more output tests if you want. (I mean we could lock in the full output of every spec test if you want, but, it would probably mean a lot of noise whenever we change the output.)

sbc100 (approved):
lgtm % comments

test/regress/regress-2136.txt
keithw: Renamed to test/wasm2c/duplicate-names.txt

src/c-writer.cc

keithw:
(I think this is ready to land now -- I can't merge it because of the required status checks.)

shravanrn:
Huh, this is weird. Has the status check frozen or something?

Issue #2019, opened 2022-10-12
wasm2c roadmap ideas


keithw:
Now that wasm2c has almost caught up to the current Wasm spec, maybe it's a good time to brainstorm about the roadmap from here and see what everything else thinks is useful/worth prioritizing. Here are some possible items and thoughts to get the discussion going:

- [ ] A WASI implementation for Unix-ish hosts (already basically on approach in #2002) -- will be awesome to have this in
- [ ] Some sort of continuous performance regression testing. Basically an operationalized/ongoing version of the benchmarks at https://kripken.github.io/blog/wasm/2020/07/27/wasmboxc.html
- [ ] Increased safety when modules link with other modules. Right now it's pretty easy to create a segfault (or worse) by having one module import something by name, and another module exports something with the same name but an incompatible type, because right now wasm2c makes "optimistic" extern declarations for everything that's imported. We could improve this by forcing the wasm2c-generated code to actually include the header from the module it wants to import from (so the C compiler will enforce type correctness), instead of making an optimistic declaration itself (#1908).
- [ ] SIMD support. This is the last remaining piece to get full conformance with WebAssembly 2.0, and it would be kind of cool to say we conform to the whole spec before it's finalized. I think I've done most of the scaffolding work, but still need to implement all those instructions, ideally in a way that spits out generated code that's both (a) inlined to SSE2 intrinsics when available, but also (b) with backup implementations in pure C for everywhere else, and then (c) ideally we'd test both backends in the CI.
- [ ] Get Mozilla using the upstream wasm2c as part of the Firefox build process. Would probably be great to have a production "customer"; I know they had wanted bulk memory support (which we now have) but I think they probably also depend on a bunch of features in the UCSD/rlbox fork. We could work with Mozilla/UCSD to get them transitioned over to the main branch.
[...]

PR #2142, opened 2023-02-04
wasm2c: prettify/change name mangling

keithw:
The ultimate bike-shedding PR. :-) Currently sequenced behind #2137.

This refactors the wasm2c name-mangling in two big ways:

1. Removing the 'Z_' prefix and trying to make the names somewhat ergonomic/pretty. Previously the `factorial` export from a `fac` module looked like this:
    
    ```
    u32 Z_facZ_factorial(Z_fac_instance_t*, u32);
    ```
    
    After this PR, it looks like this:
    
    ```
    u32 w2c_fac_factorial(w2c_fac*, u32);
    ```
    
   The name of the instance is now `w2c_fac` and the exported function is simply `w2c_` + module_name + `_` + export_name.

   It gets less pretty if there are forbidden chars in the module or export name, because those still have to be escaped/mangled (and an underscore in a module name now has to be escaped), but I think in the common case these are an ergonomic improvement.

   Symbols defined by wasm2c itself (including `instantiate`/`free`/`get_func_type` and the imported memory properties) are now prefixed with `wasm2c_` ~~`wasm_` (joining the `wasm_rt_...` family)~~ to avoid conflicting with anything defined by the module (prefixed with `w2c_`).

2. Using globally unique (module-prefixed) names for functions, types, segments, and tags, even though these are currently static (internal-linkage) symbols in the .c output. This is preparation for a future "multiple .c output" option where these symbols will need to have external linkage (#2019).

Internally, this tries to tidy up the name-handling logic into more-understandable routines (e.g. ClaimName, FindUniqueName) that encapsulate the raw manipulation of the symbol set and map.


shravanrn:
Will review this as soon as I get a chance. But broadly, the new naming scheme looks way nicer! :)

test/wasm2c/add.txt
```diff
@@ -797,25 +797,25 @@ static u32 w2c_add(Z_test_instance_t* instance, u32 w2c_p0, u32 w2c_p1) {
...
+void wasm_instantiate_test(w2c_test* instance) {
   assert(wasm_rt_is_initialized());
 }
 
-void Z_test_free(Z_test_instance_t* instance) {
+void wasm_free_test(w2c_test* instance) {
 }
 
-wasm_rt_func_type_t Z_test_get_func_type(uint32_t param_count, uint32_t result_count, ...) {
+wasm_rt_func_type_t wasm_get_func_type_test(uint32_t param_count, uint32_t result_count, ...) {
```
sbc100, line 811: In general this change looks great, but I'm not totally sure about this new `wasm_` prefix.

Firstly is there some reason you don't put the `test` name right after the `prefix` like we do for other `w2c_` generated functions?

e.g.

```
wasm_instantiate_test
wasm_free_test
wasm_get_func_type_test
```

vs:

```
wasm_test_instantiate
wasm_test_free
wasm_test_get_func_type
```

The latter seems more natural to me and is more in keeping with the previous generated code.

Also, using the raw `wasm_` seems little odd, I'm not sure why exactly.  IIUC we do need to do different prefixes so that these admin function can't overlap with user functions?    How about using the full `wasm2c_` for the admin functions and the abbreviated `w2c_` for user function?

keithw: > In general this change looks great, but I'm not totally sure about this new `wasm_` prefix.
> 
> Firstly is there some reason you don't put the `test` name right after the `prefix` like we do for other `w2c_` generated functions?

I think I was worried about, e.g., what if the module is named "rt" and exports a function named "free" or something, and then wasm2c will generate a function named "wasm_rt_free" which will conflict with the runtime.

> Also, using the raw `wasm_` seems little odd, I'm not sure why exactly. IIUC we do need to do different prefixes so that these admin function can't overlap with user functions? How about using the full `wasm2c_` for the admin functions and the abbreviated `w2c_` for user function?

Sure, I think if we use `wasm2c_` for the admin functions, we can put the module name right after as you suggest because there's no risk of colliding with "rt". Then we'd end up with three prefixes/families of functions:

1. A module's own exports/functions/types/tags would look like `w2c_[modname]_[name]`
2. "Admin" wasm2c-generated functions would look like `wasm2c_[modname]_instantiate`.
3. Runtime functions would look like `wasm_rt_allocate_memory`, etc.

sbc100: Sounds good to me!

keithw: Done!

shravanrn: Nitpick: Having a distinction between `wasm2c` and `w2c` is a bit wierd to me, but also I am not able to come up with anything better, so we could go with this :+1: 

I think clean slate, I would have suggested something like

w2c_/wasm_w2c_ --- wasm2c admin functions
wasm_rt_ --- wasm runtime functions 
wasm_func_ --- wasm module's functions

But, this would have the downside of renaming pretty much everything.

sbc100: One issue with using `wasm_func_` as a prefix is that every element type would then need its own prefix too, e.g. `wasm_global_`, `wasm_table_`, `wasm_memory_`...

I agree that having both `w2c_` and `wasm2c_` is a bit weird, but it seems like the best worst option in this case.

sbc100:
Can you update the PR description with the message you would like to include when landing this?  (or if you prefer you would post that as a comment)

test/wasm2c/hello.txt
```diff
@@ -856,47 +857,46 @@ static void w2c__start(Z_test_instance_t* instance) {
...
-  instance->Z_wasi_snapshot_preview1_instance = Z_wasi_snapshot_preview1_instance;
+static void init_instance_import(w2c_test* instance, struct w2c_wasi0x5Fsnapshot0x5Fpreview1* w2c_wasi0x5Fsnapshot0x5Fpreview1_instance){
+  instance->w2c_wasi0x5Fsnapshot0x5Fpreview1_instance = w2c_wasi0x5Fsnapshot0x5Fpreview1_instance;
 }
-
-void Z_test_instantiate(Z_test_instance_t* instance, struct Z_wasi_snapshot_preview1_instance_t* Z_wasi_snapshot_preview1_instance) {
+void wasm2c_test_instantiate(w2c_test* instance, struct w2c_wasi0x5Fsnapshot0x5Fpreview1* w2c_wasi0x5Fsnapshot0x5Fpreview1_instance) {
   assert(wasm_rt_is_initialized());
-  init_instance_import(instance, Z_wasi_snapshot_preview1_instance);
+  init_instance_import(instance, w2c_wasi0x5Fsnapshot0x5Fpreview1_instance);
```
sbc100, line 880: Given how common underscores are likely to be in module names this seems like a somewhat of a regression.

I'm not sure what we can do about?  Some other kind of escaping mechanism for modules with underscores in their name?  Use double underscores to mark the start and end of the module name?

keithw:
> Can you update the PR description with the message you would like to include when landing this? (or if you prefer you would post that as a comment)

Sure, I think I had imagined it would be the current commit message for the first (main) commit:

>     wasm2c: prettify/change name-mangling
> 
> This refactors the wasm2c name-mangling in two big ways:
> 
> 1) Removing the 'Z_' prefix and trying to make the names somewhat
>    ergonomic/pretty. Previously the `factorial` export from a `fac`
>    module looked like this:
> 
>    ```
>    u32 Z_facZ_factorial(Z_fac_instance_t*, u32);
>    ```
> 
>    After this commit, it looks like this:
> 
>    ```
>    u32 w2c_fac_factorial(w2c_fac*, u32);
>    ```
> 
> 2) Using globally unique (module-prefixed) names for functions,
>    types, segments, and tags, even though they are currently static
>    (internal-linkage) symbols in the .c output. This is preparation
>    for a future "multiple .c output" option where these symbols
>    will need to have external linkage.

wdyt?

sbc100:
> > Can you update the PR description with the message you would like to include when landing this? (or if you prefer you would post that as a comment)
> 
> Sure, I think I had imagined it would be the current commit message for the first (main) commit:
> 
> > ```
> > wasm2c: prettify/change name-mangling
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This refactors the wasm2c name-mangling in two big ways:
> > 
> > 1. Removing the 'Z_' prefix and trying to make the names somewhat
> >    ergonomic/pretty. Previously the `factorial` export from a `fac`
> >    module looked like this:
> >    ```
> >    u32 Z_facZ_factorial(Z_fac_instance_t*, u32);
> >    ```
> >    
> >    
> >        
> >          
> >        
> >    
> >          
> >        
> >    
> >        
> >      
> >    After this commit, it looks like this:
> >    ```
> >    u32 w2c_fac_factorial(w2c_fac*, u32);
> >    ```
> > 2. Using globally unique (module-prefixed) names for functions,
> >    types, segments, and tags, even though they are currently static
> >    (internal-linkage) symbols in the .c output. This is preparation
> >    for a future "multiple .c output" option where these symbols
> >    will need to have external linkage.
> 
> wdyt?

Maybe add something about the new `wasm2c_` prefix as distinct from the `w2c_` prefix?

test/wasm2c/hello.txt
keithw: We can definitely prettify this at the expense of making the mangling rules more complicated.

One possibility would be: In all names, we escape leading underscores, trailing underscores, and underscores that follow underscores. Additionally, in a module name, we transform a single underscore into a double underscore. The boundary between module name and export name is the first single underscore after `w2c_`.

So `wasi_snapshot_preview1` would become `wasi__snapshot__preview1`. Unfortunately an export named `_start` would become `0x5Fstart` which is a little uglier but maybe still acceptable?

keithw:
> Maybe add something about the new `wasm2c_` prefix as distinct from the `w2c_` prefix?

Done in 07d9a2d7bf751e6a15c7cba426259eb3c3717ded

test/wasm2c/hello.txt
keithw: Just pushed a new commit that does this -- hope it looks cleaner.

wasm2c/README.md
```diff
@@ -143,34 +143,52 @@ You can take a look at the all of these files in
...
+typedef uint32_t u32;
+typedef int32_t s32;
+typedef uint64_t u64;
+typedef int64_t s64;
+typedef float f32;
+typedef double f64;
+
+#if defined(WASM_RT_ENABLE_SIMD)
+typedef simde_v128_t v128;
+#endif
```
sbc100, line 174: Why expand the `...` here?

test/run-spec-wasm2c.py
```diff
@@ -101,11 +101,21 @@ def MangleTypes(types):
...
+
+    # NOTE(keithw): forced escapes for '0x[hexdigit]' not implemented here
+    pattern = b'([^_a-zA-Z0-9])'
+    return re.sub(pattern, Mangle, s.encode('utf-8')).decode('utf-8')
+
+
+def MangleModuleName(s):
+    # double underscores
+    return MangleName(re.sub(b'(_)', b'__',
+                             s.encode('utf-8')).decode('utf-8'))
```
sbc100, line 118: This roundtrip through encode and decode looks odd.

Can we not just stay in decoded the whole time by doing `re.sub('(_)', '__', s)`?

src/c-writer.cc
```diff
@@ -4631,7 +4843,7 @@ void CWriter::WriteCHeader() {
   Write("#define ", guard, Newline());
   Write(Newline());
   Write(s_header_top);
-  Write(Newline());
+  ReserveExportNames();
```
sbc100, line 4846: Should this happen before we even start writing the header?

Why remove the newline `Write(Newline());`

wasm2c/README.md
keithw: Nice catch... I had just pasted in the current code and didn't match the `...` from before. Fixed in bdcd93dffdbd1138b79714978cf783c88078956c

src/c-writer.cc
```diff
@@ -507,6 +526,37 @@ constexpr char CWriter::MangleField(ModuleFieldType type) {
   return 'a' + static_cast<char>(type);
 }
 
+// remove risky characters for pasting into a C-style comment
+static std::string SanitizeForComment(std::string_view str) {
```
sbc100, line 530: This seems like a separate change perhaps?   Can we land this separately?

wasm2c/examples/rot13/main.c
```diff
@@ -20,62 +20,46 @@
 #include "rot13.h"
 
 /* Define structure to hold the imports */
-struct Z_host_instance_t {
+typedef struct w2c_host {
   wasm_rt_memory_t memory;
   char* input;
-};
+} w2c_host;
```
sbc100, line 26: Is adding this extra typedef part of the this change too?   Or was this file just out-of-date?

test/run-spec-wasm2c.py
keithw: Yes, done in 3e5e00f9438390b4e0a368f397323573f9c24eb6

src/c-writer.cc
keithw: In general we're getting double newlines in the generated output in a bunch of places, and this was one of them. It's tricky to handle systematically because some of the functions only write output if some nonobvious condition is true, so you don't want to unconditionally write a newline. I suspect what we may want is a new token and stream state that queues up a single extra Newline, but this can wait until later.

wasm2c/examples/rot13/main.c
keithw: I think it probably should have been a typedef all along (so main.c can refer to it as `w2c_host` instead of `struct w2c_host`, and to match the wasm2c generated code which always typedef's the name of the struct), but it's not critical...

sbc100: Any chance we can split that out as well? .. sorry for being so pedantic, but it helps my rather slow brain understand PR much better when its focused on just one thing.   Feel free to push back on this if you think its too much.

keithw: Sure, no problem. Stand by...

keithw: Done in a5a81648e39ae961a7173f00f34be52d0f5f97e4. I'm guessing there's going to be more in c-writer.cc that you'll want split out, which I may try to get to first to save you the trouble...

keithw: (done for now)

src/c-writer.cc
keithw: (I restored all the old `Write(Newline())`s and will split that out to a later PR.)

wasm2c/README.md
```diff
@@ -455,7 +458,7 @@ module doesn't use any globals, memory or tables.
 The most interesting part is the definition of the function `fac`:
 
 ```c
-static u32 w2c_fac(Z_fac_instance_t* instance, u32 w2c_p0) {
+static u32 w2c_fac_fac_0(w2c_fac* instance, u32 w2c_p0) {
```
sbc100, line 461: What is the `_0` suffix on the function name here?

sbc100: Oh I see we have an internal name and then an exported name that simply forwards its arg to the internal one.

As a followup should we potentially just export the function itself rather than a forwarding function?

src/c-writer.cc
```diff
@@ -527,87 +546,205 @@ std::string CWriter::MangleTagTypes(const TypeVector& types) {
...
+  auto my_isalpha = [](uint8_t ch) {
+    return (ch >= 'A' && ch <= 'Z') || (ch >= 'a' && ch <= 'z');
+  };
+  auto my_isdigit = [](uint8_t ch) { return (ch >= '0' && ch <= '9'); };
+  auto my_isalnum = [&](uint8_t ch) {
+    return my_isalpha(ch) || my_isdigit(ch);
+  };
+  auto my_ishexdigit = [&](uint8_t ch) {
+    return my_isdigit(ch) || (ch >= 'A' && ch <= 'F');
+  };
```
sbc100, line 611: How about simply declaring these as normal static C functions outside this method body?

I guess its mostly aesthetic, but unless we actually need to capture stuff from the current function I'm not a huge fan of inner lambda functions like this.

sbc100 (approved):

wasm2c/README.md
keithw: Yeah, exactly. In this case the names had to be deconflicted because the ID of the function itself is $fac, and also the export is named "fac", so the internal function gets the `_0` suffix.

I think trying to "directly" export a function like this may be doable but a bit complicated. :-( The internal functions are named and called based on their generated name or debug name (e.g. $f0 or $fac), whereas the exports are named based on their export name (in this case, also "fac"). The same function can be exported multiple times under different names, so the module could export this function as "fac" and "fac2" and "fac3" and they would all have to be in the .h file.

If you think this is valuable, I think what we'd try to do is detect a "direct" export (an exported function where the export name is identical to the name of the function itself) and handle it specially by declaring it directly in the `.h` file, and then for any other export of the same function, we just forward back to the original implementation. I can see a pathway here but it might be best done in another PR if you're amenable. :-/

src/c-writer.cc
keithw: :+1:

keithw: Done in 6974baef56330fb2d550aa2ede8a7f4fbfe71d0d

keithw: Thank you!

wasm2c/README.md
sbc100: I yes, that makes total sense.  I think continue to separate internal names from export names makes sense give that context.   I don't think we need to special case for when they happen to match.

I do wonder if we could avoid declaring extra function though.. maybe function exports could be some kind of symbol alias instead?   I guess that runs into visibility issues, and is kind of compiler-specific.   Its a shame C doesn't really have anything like aliases built-in.

keithw: Yeah, I think if we were willing to limit ourselves to GCC and clang, we could use `__attribute__((alias))` no problem. But I don't know how to do this on MSVC. :-/

PR #2146, opened 2023-02-23
wasm2c: multiple .c outputs

yhdengh:
This PR is sequenced after #2142.

This enables wasm2c to have multiple .c outputs, which allows parallel
compilation of the wasm2c outputs. This is useful when the input WASM module is
big.

wasm2c takes the number of .c outputs as an argument to `-j` (default to
1). If the number is equal to 1, the .c output does not change except for two
new macro declarations and the ordering of declarations and definitions. If the
number is greater than 1, wasm2c outputs change in following ways:

1) wasm2c outputs a `[module-name]-impl.h` that includes any module-wide
declarations, including:
  * content of `WriteSourceTop()`
  * function type declarations
  * tag types
  * tag declarations
  * function declarations
  * data segments and elem segments declarations

  Any `static` declaration become `extern` in this header.

2) wasm2c outputs `[module-name].i.c` for i = [0, number of .c outputs). Any
module-wide material is written to `[module-name].0.c`, including:
  * function types, tags, data segments, elem segments
  * imports and exports
  * module initialization, instantiation and free

3) For each function, wasm2c picks which `[module-name].i.c` to go in based on the
SHA-256 hash of the function's name. Since SHA-256 is consistent across runs of
wasm2c, only one .c output changes and need to be recompiled if only one
function changes, or one one function is added or moved.


src/c-writer.cc
```diff
@@ -4819,23 +4948,45 @@ void CWriter::WriteCHeader() {
...
+  WriteDataInitializerDecls();
+  WriteElemInitializerDecls();
+
+  /* Write #include impl header to every output stream if there are multiple .c
+   * outputs*/
+  if (c_streams_.size() > 1) {
+    assert(header_impl_name_.size() > 0);
+    for (auto& stream : c_streams_) {
+      stream_ = stream;
+      Write("#include \"", header_impl_name_, "\"", Newline());
```
keithw, line 4970: This may be kind of annoying, but I ran some quick tests with `run-spec-wasm2c.py -j64`, and the overhead of running GCC with optimization on a ton of C outputs where the only contents are `#include "[test].0-impl.h"` is fairly significant. :-( (E.g. the try-catch test takes about 11.5 seconds on my laptop with `-j64`, compared with 1.7 seconds if I manually delete the `#include` line from the empty files. Those are both running GCC serially 64x3=192 times, most of which are on near-empty source files. Versus... about 0.5 seconds in single-output mode.)

Is there any way we can gracefully defer printing the `#include [modname].0-impl.h` line until the first function is being printed to that output stream? (Maybe we can check the stream offset in `WriteFuncs`?) Might need some care if we're also trying to write the `/* Automatically generated */` header. :-/

src/c-writer.cc
```diff
@@ -4819,23 +4948,45 @@ void CWriter::WriteCHeader() {
 }
 
 void CWriter::WriteCSource() {
-  stream_ = c_stream_;
+  /* Write the "top" to h_impl stream */
+  stream_ = h_impl_stream_;
   Write("/* Automatically generated by wasm2c */", Newline());
```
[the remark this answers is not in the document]
yhdengh: If we write `/*Automatically generated*/` header to every .c output, then I think only macros related to function type declarations need to go in -impl.h. Would it be cleaner to separate that part from `wasm2c.declarations.h`, and only write that to -impl.h?

src/c-writer.cc
yhdengh: Then, this would be like if some function is printed to a certain .c output, we need to write `/*Automatically generated*/` header and `#include -impl.h` to the .c output.

keithw:
Kind of weird that the base branch of this PR didn't get automatically updated with `w2c-pretty-names` was auto-merged into the `main` branch. I guess GitHub doesn't handle this case when it happens across forks. :-(

src/c-writer.cc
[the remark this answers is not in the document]
keithw: I was really just thinking we'd re-print the actual `/* Automatically generated by wasm2c */` line by itself (given that the individual C files are also automatically generated...).

src/c-writer.cc
keithw: Yeah, that makes sense to me.

keithw: It occurs to me that I guess formally it's invalid to have a completely empty compilation unit in C (https://stackoverflow.com/questions/26541150/warning-iso-c-forbids-an-empty-translation-unit). This is triggering a warning/error in GCC when I try to compile with `-Wpedantic`.

What do you think is the best way to handle this? I guess we could add a pass at the end over all the streams and, I guess, if they're completely empty, maybe print something like `/* Empty wasm2c generated file */` and then a dummy typedef?

src/c-writer.cc
```diff
@@ -1527,7 +1603,7 @@ void CWriter::WriteFuncDeclarations() {
   for (const Func* func : module_->funcs) {
     bool is_import = func_index < module_->num_func_imports;
     if (!is_import) {
-      Write("static ");
+      Write(InternalSymbolDeclScope());
```
keithw, line 1606: Because this is the only place `InternalSymbolDeclScope` gets called, and because I think `extern` is the default for function declarations, I *think* this could just be a call to `InternalSymbolScope` and we could eliminate the `InternalSymbolDeclScope` method entirely.

src/c-writer.cc
```diff
@@ -172,15 +172,22 @@ int GetShiftMask(Type type) {
...
-        c_stream_(c_stream),
+        c_streams_(std::move(c_streams)),
         h_stream_(h_stream),
-        header_name_(header_name) {
+        h_impl_stream_(h_impl_stream),
+        header_name_(header_name),
+        header_impl_name_(header_impl_name) {
     module_prefix_ = MangleModuleName(options_.module_name);
+    if (c_streams_.empty()) {
+      c_streams_.push_back(h_stream);
```
sbc100, line 189: Why would `c_streams` be empty here?  Don't we pass in just one in the case of `-j1`?

src/c-writer.cc
```diff
@@ -2300,15 +2425,42 @@ void CWriter::Write(const Func& func) {
...
+  /*
+   * Choose the stream that this function goes in, based on its name.
+   * We need a hash function that's consistent across runs of
+   * wasm2c (which std::hash doesn't guarantee) because if only one function
+   * changes across runs, or one function is added or removed, we want only one
+   * stream to change and need to be recompiled. Here we use "SHA-256/64"
+   * because it's sufficient and already available in wasm2c.
+   */
+  std::string hashed_function_name;
+  sha256(func.name, hashed_function_name);
```
sbc100, line 2437: Why not simply sort the function by name and divide the results in on N equal sized buckets?   It would result in the nice property that each of the N files would have have exactly M/N functions in it.  It seems like a more obvious and straight forward approach.

It would still be the case the changing just one function would result in only one file changing.   We would loose the property when adding a removing functions.. but I don't think that sounds like a particularly important case (I could be wrong?).

I'd also rather not increase our dependency the external sha256 codebase simply because we've already pulled it in.

src/tools/wasm2c.cc
```diff
@@ -127,6 +131,11 @@ int ProgramMain(int argc, char** argv) {
   InitStdio();
   ParseOptions(argc, argv);
 
+  if (s_num_outputs < 1 || s_num_outputs > 1024) {
+    fprintf(stderr, "Number of output files must be between 1 and 1024.\n");
+    exit(1);
```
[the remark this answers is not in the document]
yhdengh: Updated in the latest commit.

src/tools/wasm2c.cc
```diff
@@ -127,6 +131,11 @@ int ProgramMain(int argc, char** argv) {
   InitStdio();
   ParseOptions(argc, argv);
 
+  if (s_num_outputs < 1 || s_num_outputs > 1024) {
+    fprintf(stderr, "Number of output files must be between 1 and 1024.\n");
```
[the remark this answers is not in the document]
yhdengh: Updated in the latest commit.

src/c-writer.cc
yhdengh: Updated in the latest commit.

src/c-writer.cc
yhdengh: It is indeed the case that for the majority of users of wasm2c, sorting functions by name and dividing the results into equal buckets is the more straight forward approach. However, in our use case of wasm2c, we expect users to develop their programs in an incremental way, and therefore remaining the function distribution stable when a single function is added or removed is important to us.

Would it be better to add a field to `WriteCOptions` where callers can specify their own functions for deciding how functions are distributed across buckets, and the default behavior is dividing functions based on function name sorting?

sbc100: Have you measured this to have a real effect on the programs you are working on?   Are you saying you have multiple large wasm programs that you rebuild frequently, with only difference being that addition of removal of a single function?   What kind of large programs are we talking about here?  How long are you seeing the compile times?  Are you using some kind of build system that doesn't use timestamps for change detection?

yhdengh: In our use case, we take a single WebAssembly module as input from the user, using wasm2c+clang to transform that to x86 for execution.

We wrote a program that contains an entry function, some helper functions. The entry function invokes the backend of Clang/LLVM for compiling c source code into x86. We compiled this program to a WebAssembly module. wasm2c + clang (c to x86) on this module would take ~2 hours without any optimization flag and ~7 hours with "-O2".

In this case, we would hope that adding or removing functions from the part we wrote ourselves can be done by only recompiling a small number of .c outputs.

src/c-writer.cc
```diff
@@ -1551,7 +1647,11 @@ void CWriter::WriteFuncDeclarations() {
   for (const Func* func : module_->funcs) {
     bool is_import = func_index < module_->num_func_imports;
     if (!is_import) {
-      Write("static ");
+      if (InternalSymbolScope()) {
```
keithw, line 1650: I *think* this whole stanza could just be `Write(InternalSymbolScope())` (because `extern` is already the default for a function declaration, so no need for the `if` and explicit `Write("extern ")`).

src/c-writer.cc
```diff
@@ -413,6 +467,11 @@ class CWriter {
   SymbolSet func_includes_;
 
   std::vector<std::string> unique_func_type_names_;
+
+  std::function<std::vector<size_t>(const std::vector<Func*>&,
+                                    const size_t,
+                                    const size_t)>
+      name_to_index;
```
keithw, line 474: Underscore after `name_to_index`?

keithw (commented):
lgtm % comments

src/c-writer.cc
```diff
@@ -170,17 +170,59 @@ int GetShiftMask(Type type) {
...
-        c_stream_(c_stream),
+        c_streams_(std::move(c_streams)),
         h_stream_(h_stream),
-        header_name_(header_name) {
+        h_impl_stream_(h_impl_stream),
+        header_name_(header_name),
+        header_impl_name_(header_impl_name) {
     module_prefix_ = MangleModuleName(options_.module_name);
+    if (c_streams_.size() == 1)
+      assert(!options.name_to_index);
```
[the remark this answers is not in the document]
yhdengh: I think for the current implementation of `WriteFuncs`, if there is only a single .c output, `name_to_index_` has to return an all-zero vector. So when there is only a single .c output, either we add a if statement to `WriteFuncs` to not use `name_to_index_` , or we always bind `name_to_index_` with default behavior (which is what I do in f4eadb8).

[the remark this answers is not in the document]
keithw: :+1:

src/c-writer.cc
```diff
@@ -170,17 +170,63 @@ int GetShiftMask(Type type) {
...
         h_stream_(h_stream),
-        header_name_(header_name) {
+        h_impl_stream_(h_impl_stream),
+        header_name_(header_name),
+        header_impl_name_(header_impl_name) {
     module_prefix_ = MangleModuleName(options_.module_name);
+    if (c_streams_.size() != 1 && options.name_to_index) {
+      name_to_index_ = options.name_to_index;
+    } else {
+      name_to_index_ = std::bind(&default_name_to_index, std::placeholders::_1,
```
keithw, line 227: Now that `default_name_to_index` is an ordinary function, I think we can get rid of the call to `std::bind` and just say `name_to_index_ = default_name_to_index;`

include/wabt/c-writer.h
```diff
@@ -26,11 +28,20 @@ class Stream;
 
 struct WriteCOptions {
   std::string_view module_name;
+  /* name_to_index takes list of all functions in the module, number of imported
+   * functions, and number of .c outputs as argument, returns a vector where
+   * vector[i] the index of the .c output that funcs[i] goes into. */
+  std::function<std::vector<size_t>(const std::vector<Func*>&,
```
keithw, line 34: I'm realizing that this signature allows the passed-in `name_to_index` function to mutate the functions if it wants. :-/ Is there any clean way to prevent this? We might have to pass a pair of const_iterators and a count of imports, which doesn't seem ideal but I haven't thought of a better way.

keithw (approved):
I think this is good by me at this point, but obviously I'm working closely with @yhdengh and don't want to overstep.

@sbc100, what are your thoughts / did you want to review again?

keithw:
(Gentle call for objections/further review before merger.)

include/wabt/c-writer.h
```diff
@@ -26,11 +28,23 @@ class Stream;
...
 struct WriteCOptions {
   std::string_view module_name;
+  /* name_to_index takes const iterators to begin and end of a list of all
+   * functions in the module, number of imported functions, and number of .c
+   * outputs as argument, returns a vector where vector[i] the index of the .c
+   * output that funcs_begin + i goes into. */
+  std::function<std::vector<size_t>(
+      std::vector<Func*>::const_iterator funcs_begin,
+      std::vector<Func*>::const_iterator funcs_end,
+      const size_t,
```
sbc100, line 38: No need for `const` on function arguments. 

Can you give these last two arguments meaningful names (I know you describe what they mean in comments but having names is still useful I think).

include/wabt/c-writer.h
```diff
@@ -26,11 +28,23 @@ class Stream;
...
+      const size_t)>
+      name_to_index;
```
sbc100, line 40: How about `name_to_output_file_index`.   Or `output_file_mapping`?   (Since `name` and `index` are both heavily overloaded terms in wabt)

src/tools/wasm2c.cc
```diff
@@ -79,6 +80,9 @@ static void ParseOptions(int argc, char** argv) {
         s_outfile = argument;
         ConvertBackslashToSlash(&s_outfile);
       });
+  parser.AddOption(
+      'j', "multiple-outputs", "NUM", "Number of output files to write",
```
sbc100, line 84: I'm not sure about the use of `j` here.. since since that wasm2c process is not actually running multiple threads, right?    How about `-m` /`--multi-file=N`?

include/wabt/c-writer.h
```diff
@@ -26,11 +28,23 @@ class Stream;
 
 struct WriteCOptions {
   std::string_view module_name;
+  /* name_to_index takes const iterators to begin and end of a list of all
+   * functions in the module, number of imported functions, and number of .c
+   * outputs as argument, returns a vector where vector[i] the index of the .c
+   * output that funcs_begin + i goes into. */
```
sbc100, line 34: Perhaps add "Only called when --multi-file is used"

test/run-spec-wasm2c.py
```diff
@@ -537,6 +537,9 @@ def main(args):
     parser.add_argument('--disable-bulk-memory', action='store_true')
     parser.add_argument('--disable-reference-types', action='store_true')
     parser.add_argument('--debug-names', action='store_true')
+    parser.add_argument('-j', '--num-outputs', metavar='COUNT',
```
sbc100, line 540: I think whatever we call this option it should be the same for both `run-spec-wasm2c.py` and `wasm2c` itself.   I actually like `--num-outputs` if you want to go with that.   Perhaps we can just skip the short arg version?

test/run-spec-wasm2c.py
```diff
@@ -595,11 +598,18 @@ def main(args):
...
         for i, wasm_filename in enumerate(cwriter.GetModuleFilenames()):
             wasm_filename = os.path.join(out_dir, wasm_filename)
-            c_filename = utils.ChangeExt(wasm_filename, '.c')
-            args = ['-n', cwriter.GetModulePrefixUnmangled(i)]
-            wasm2c.RunWithArgs(wasm_filename, '-o', c_filename, *args)
+            c_filename_input = utils.ChangeExt(wasm_filename, '.c')
+            c_filenames = []
+            if options.num_outputs > 1:
+                for j in range(options.num_outputs):
+                    c_filenames.append(utils.ChangeExt(wasm_filename, '_' + str(j) + '.c'))
```
sbc100, line 605: How about `base = os.path.splitext(c_filename_input)[0])` and then within the loop you can do  `f'{base}_{j}.c'`.

(setting more than just the extension seems like an abuse of utils.ChangeExt).

test/wasm2c/spec-multi-output/call.txt
```diff
@@ -0,0 +1,6 @@
+;;; TOOL: run-spec-wasm2c
+;;; STDIN_FILE: third_party/testsuite/call.wast
+;;; ARGS*: -j4
```
sbc100, line 3: I think `--num-outputs=4` would make these tests more readable.

src/c-writer.cc
```diff
@@ -170,17 +172,63 @@ int GetShiftMask(Type type) {
   // clang-format on
 }
 
+/* This function is the default behavior for name_to_index_. For single .c
```
sbc100, line 175: Nit: '/*' on a lint by itself to start multi-line comment

sbc100 (approved):
lgtm with some final comments

src/tools/wasm2c.cc
yhdengh: Edited in 660bf3daf8b8f33d02e37b74bf947289e920f797. `--num-outputs` is used by both `wasm2c` and `run-spec-wasm2c.py`.

src/c-writer.cc
```diff
@@ -264,6 +315,8 @@ class CWriter {
   static const char* GetReferenceNullValue(const Type& type);
   static const char* GetCTypeName(const Type& type);
 
+  const char* InternalSymbolScope();
```
keithw, line 318: This one too -- can InternalSymbolScope be a const method?

src/c-writer.cc
```diff
@@ -247,6 +296,8 @@ class CWriter {
 
   static void SerializeFuncType(const FuncType&, std::string&);
 
+  std::string GetGlobalName(ModuleFieldType, const std::string&);
```
keithw, line 299: Can this be a const method?

keithw (approved):
Two small comments from me and then let's ship it!

PR #2147, opened 2023-02-25
c-writer.cc: sanitize module/field names for use in C comments (NFC)

keithw:
This is to prevent the module from including malicious (or accidentally malicious) data in the name of an export or import to escape a comment.

Otherwise I think a module could name an export something like `abc */ int w2c_mod_func() { malicious_code(); }; /*` (and use a later export to close the comment) and inject arbitrary C code into the wasm2c output.


sbc100 (approved):
Maybe add a test?

keithw:
Thanks, added a test with some edge-case import and export names -- hope this is along the lines of what you had in mind.

Issue #2153, opened 2023-02-28
Build with BUILD_FUZZ_TOOLS=ON fails: fatal error: 'wabt/binary-reader-logging.h' file not found


yurivict:
```
===>  Building for wabt-1.0.32
[  0% 7/137] /usr/local/libexec/ccache/c++ -D__STDC_FORMAT_MACROS=1 -D__STDC_LIMIT_MACROS=1 -Dwabt_fuzz_EXPORTS  -O2 -pipe -fstack-protector-strong -isystem /usr/local/include -fno-strict-aliasing  -isystem /usr/local/include -Wold-style-cast -fno-exceptions -O2 -pipe -fstack-protector-strong -isystem /usr/local/include -fno-strict-aliasing  -isystem /usr/local/include  -DNDEBUG -fPIC   -Wall -Wextra -Wno-unused-parameter -Wpointer-arith -Wuninitialized -fsanitize=fuzzer,address -std=c++17 -MD -MT CMakeFiles/wabt-fuzz.dir/src/binary-reader-logging.cc.o -MF CMakeFiles/wabt-fuzz.dir/src/binary-reader-logging.cc.o.d -o CMakeFiles/wabt-fuzz.dir/src/binary-reader-logging.cc.o -c /usr/ports/www/wabt/work/wabt-1.0.32/src/binary-reader-logging.cc
FAILED: CMakeFiles/wabt-fuzz.dir/src/binary-reader-logging.cc.o 
[...]

PR #2154, opened 2023-02-28
Fix for BUILD_FUZZ_TOOLS=ON

sbc100:
Fixes: #2153

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -88,7 +88,7 @@ jobs:
       WASM2C_CFLAGS: "-march=x86-64-v2" # currently required for SIMDe to pass some tests on x86-64
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


keithw (approved):
lgtm. I think longer term we probably will want to merge the oss-fuzz fuzzers into the wabt repo (https://github.com/google/oss-fuzz/tree/master/projects/wabt) and perhaps replace the current "fuzz" target with them -- oss-fuzz has a richer collection of wabt fuzzers than what we currently have in-tree (which is just `wasm2wat-fuzz` linked against libwabt compiled with `-fsanitize=fuzzer,address`).

It would also be nice to get the fuzz-on-PR target working (https://github.com/WebAssembly/wabt/pull/2073) which I think we may be able to do now that oss-fuzz is running regularly again so in theory it will only find "new" regressions.

PR #2161, opened 2023-03-07
Deprecate wasm-c-api on big-endian and fix rot13 test

### LINKED ISSUE #1972: Rot13 test fails on s390x
When running the unittests on an s390x Docker image, the Rot13 test fails.

Here is a link to a GitHub Actions workflow that shows this failure: https://github.com/alexreinking/wabt/runs/8052934337?check_suite_focus=true

SoniEx2:
This includes #1971 and fixes #1972

Rationale in https://github.com/WebAssembly/wabt/issues/1972#issuecomment-1782130770

### DIFF
--- a/scripts/Dockerfile
+++ b/scripts/Dockerfile
@@ -5,11 +5,12 @@ WORKDIR /ws
 ARG DEBIAN_FRONTEND=noninteractive
 ENV TZ=UTC
 
-COPY . .
-
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && \
     echo $TZ > /etc/timezone && \
     apt-get update && \
-    apt-get install -y git g++ cmake ninja-build python3 file && \
-    cmake -G Ninja -S . -B build -DCMAKE_BUILD_TYPE=Release && \
+    apt-get install -y git g++ cmake ninja-build python3 file
+
+COPY . .
+
+RUN cmake -G Ninja -S . -B build -DCMAKE_BUILD_TYPE=Release && \
     cmake --build build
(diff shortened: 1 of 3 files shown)


scripts/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM debian:latest
+
+WORKDIR /ws
+
+ARG DEBIAN_FRONTEND=noninteractive
+ENV TZ=UTC
+
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && \
+    echo $TZ > /etc/timezone && \
```
sbc100, line 9: Do you know why this TZ hacking is needed.. seems a little odd.

sbc100 (commented):
If thats all it takes to support these platforms its seems reasonable to me.

scripts/Dockerfile
SoniEx2: no idea, ask #1971 ? this PR is based on that one.

SoniEx2:
this should not have broken the main build... where's the typo? >.<

SoniEx2:
found the typo XD

SoniEx2:
yeah this is sadly blocked on WebAssembly/wasm-c-api#180...

SoniEx2:
it's possible these were always broken, but, uh, anyone wanna bisect these remaining failures?

SoniEx2:
perhaps cc @rathann about i386/i686? also @alexreinking do you know anything about using a cross-compiler with these? running the whole compiler under qemu seems unnecessarily slow...

rathann:
> perhaps cc @rathann about i386/i686? also @alexreinking do you know anything about using a cross-compiler with these? running the whole compiler under qemu seems unnecessarily slow...

x86_64 builds of gcc can compile for i686 without qemu or cross-compilation.

PR #2162, opened 2023-03-08
wasm2c: implement the extended const expressions proposal

keithw:


src/c-writer.cc
```diff
@@ -1057,22 +1059,27 @@ void CWriter::Write(const StackVar& sv) {
   }
 }
 
-void CWriter::Write(Type type) {
+// static
+const char* CWriter::GetCTypeName(const Type& type) {
```
sbc100, line 1063: Unrelated to this PR but what do you think about making these helpers into normal static function rather than method?

It would avoid the awkard `// static` comments and improve locality of reference (there really is no need to declare this function up on ling 265 right?

Maybe as a followup we could convert them all.

src/c-writer.cc
```diff
@@ -1199,12 +1206,77 @@ void CWriter::WriteGetFuncTypeDecl() {
...
+          ")" + mini_stack.at(mini_stack.size() - 2) + ")" + op + "((" +
+          std::string(GetCTypeName(binexpr->opcode.GetParamType2())) + ")" +
+          mini_stack.at(mini_stack.size() - 1) + ")";
+      mini_stack.resize(mini_stack.size() - 2);
+      mini_stack.push_back(move(combination));
+    } else {
+      // Leaf node (nullary const expression)
+      Stream* existing_stream = stream_;
+      MemoryStream terminal_stream;
+      stream_ = &terminal_stream;
```
sbc100, line 1262: Neat trick.. I would not have though of temporarily overriding the output `stream_` like that.

sbc100 (approved):
Great!

Update the README mabye?

PR #2163, opened 2023-03-08
decompiler.cc: passive data segments do not have offsets

### LINKED ISSUE #2164: wasm-decompile fails on wasm files containing passive data segments
Error message:
`wasm-decompile: /home/kangm/code/wabt/src/decompiler.cc:688: std::__cxx11::string wabt::Decompiler::InitExp(const ExprList&): Assertion `!el.empty()' failed.`

PR:
https://github.com/WebAssembly/wabt/pull/2163

mannkdev:
Passive data segments do not have offsets, thus the offset ExprList is empty:
https://github.com/WebAssembly/wabt/blob/main/src/binary-reader.cc#L2812

This causes wasm-decompile to fail when trying to decompile wasm files containing passive data segments, as it tries to call InitExp on the empty offset, causing an assertion in InitExpr to fail:
https://github.com/WebAssembly/wabt/blob/main/src/decompiler.cc#L773
https://github.com/WebAssembly/wabt/blob/main/src/decompiler.cc#L688

The fix is to skip printing of the non-existent offset for passive data segments.

Fixes #2164

### DIFF
--- a/src/decompiler.cc
+++ b/src/decompiler.cc
@@ -770,7 +770,11 @@ struct Decompiler {
 
     // Data.
     for (auto dat : mc.module.data_segments) {
-      s += cat("data ", dat->name, "(offset: ", InitExp(dat->offset), ") =");
+      s += cat("data ", dat->name,
+               dat->kind != SegmentKind::Passive
+                   ? cat("(offset: ", InitExp(dat->offset))
+                   : "(passive",
+               ") =");
       auto ds = BinaryToString(dat->data);
       if (ds.size() > target_exp_width / 2) {
         s += "\n";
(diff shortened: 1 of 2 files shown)


keithw:
Looks reasonable...
- Can you please add a test (probably in test/decompile)?
- (For active segments, should this also print out the memidx if nonzero?)

keithw:
@mannkdev Do you have cycles to respond to the above? Would be great to get this merged. Thank you for working on this!

mannkdev:
Hi, sorry for the late response. Yes I will try to look within the next week, thanks.

keithw:
Added a test. Thank you for catching and fixing this!

keithw (approved):
lgtm

PR #2168, opened 2023-03-13
Add new tests

keithw:
Fixes a wasm2c bug where instantiation was initializing memories before tables. The spec requires tables to be initialized before memories (newly tested in the multi-memory [linking0.wast](https://github.com/WebAssembly/multi-memory/blob/main/test/core/multi-memory/linking0.wast) test).

Remaining missing tests are:

- multi-memory/memory_copy1.txt (currently blank)
- relaxed-simd/relaxed_dot_product.txt (we don't support the i16x8.relaxed_dot_i8x16_i7x16_s instruction)

And for wasm2c only:

- five SIMD tests (https://github.com/WebAssembly/wabt/pull/2021#issuecomment-1396386135)
- all relaxed-simd and tail-call tests


sbc100 (approved):
Were these tests recently added to the testsuite repo?   i..e is the the result of running 
update-spec-tests.py after the recent update?

keithw:
> Were these tests recently added to the testsuite repo? i..e is the the result of running update-spec-tests.py after the recent update?

Yeah, exactly.

(BTW are you able to take another look at #2146 while I have you?)

PR #2169, opened 2023-03-14
BinaryReaderIR: limit nesting depth to avoid stack overflow

### LINKED ISSUE #2151: Very deep nested labels can cause BinaryReaderIR to crash (stack overflow in Module destructor)
oss-fuzz has discovered that if a module includes an extremely deep series of nested blocks (e.g. a huge number of nested `if` instructions, each in the parent's "true" block), that `wasm2wat` will crash when trying to run the `Module` destructor, just because each `Block` recursively runs the destructor of its child `IfExpr` which runs the destructor of its child `Block`, etc., and eventually all those nested stack frames cause a stack overflow.

(This is [oss-fuzz #53958](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=53958), currently still under embargo.)

I guess I sort of have two ideas for how we can fix this:
1. make `BinaryReaderIR` establish a maximum nesting depth (a `static constexpr BinaryReaderIR::kMaxNestingDepth` member variable?) that would be enforced in `BinaryReaderIR::PushLabel`. This seems to be allowed by https://webassembly.github.io/spec/core/appendix/implementation.html#structure ("the nesting depth of structured control instructions"), although it's a little weird that BinaryReaderIR doesn't seem to enforce *any* implementation-defined limits right now, so this would seemingly be a first for us?
2. somehow make `wabt::intrusive_list`, `Expr`, and `Block` a lot more clever and figure out how to destroy (and traverse?) these deeply nested data structures without literal recursion

Idea 1 seems a lot easier, but less satisfying. I'm guessing some limit like 16,384 is probably safe (both in terms of stack space as well as accommodating all reasonable modules). But it would be interesting to know what other major Wasm implementations do here or if there's a cleaner way to solve this...

keithw:
Fixes #2151.

Summary: oss-fuzz has discovered that if a module includes an extremely deep series of nested blocks (e.g. a huge number of nested if instructions, each in the parent's "true" block), that wasm2wat will crash when trying to run the Module destructor, just because each Block recursively runs the destructor of its child IfExpr which runs the destructor of its child Block, etc., and eventually all those nested stack frames cause a stack overflow. (This is [oss-fuzz #53958](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=53958).)

This PR fixes the problem the dumb/easy way, by making BinaryReaderIR enforce a cap on nesting depth that's likely to be safe given a reasonable stack size. I couldn't easily find what the major Wasm engines do here, but 1,024 seems like a reasonable value.

### DIFF
--- a/src/binary-reader-ir.cc
+++ b/src/binary-reader-ir.cc
@@ -91,6 +91,8 @@ class CodeMetadataExprQueue {
 };
 
 class BinaryReaderIR : public BinaryReaderNop {
+  static constexpr size_t kMaxNestingDepth = 1024;  // max depth of label stack
+
  public:
   BinaryReaderIR(Module* out_module, const char* filename, Errors* errors);
 
@@ -350,9 +352,9 @@ class BinaryReaderIR : public BinaryReaderNop {
  private:
   Location GetLocation() const;
   void PrintError(const char* format, ...);
-  void PushLabel(LabelType label_type,
-                 ExprList* first,
-                 Expr* context = nullptr);
+  Result PushLabel(LabelType label_type,
+                   ExprList* first,
+                   Expr* context = nullptr);
   Result BeginInitExpr(ExprList* init_expr);
   Result EndInitExpr();
   Result PopLabel();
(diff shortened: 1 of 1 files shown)


keithw:
It would be nice to have a regression test for this, but I think we'd probably have to produce it programmatically at runtime because otherwise it would be like hundreds of kilobytes (like the oss-fuzz reproducer is).

sbc100 (approved):

PR #2171, opened 2023-03-15
c-writer.cc: Add local symbol prefix.

yhdengh:
This adds `kLocalSymbolPrefix` which is used for names of params, locals and stack vars. This allows c-writer to not assign `global_syms_` to `local_syms_` for writing each individual function, since local names can't duplicate global names.

This speeds up the execution of `wasm2c` on large inputs. `clang.wasm` is a 75M WASM module. Before this PR, `wasm2c` on `clang.wasm` would take ~63 minutes.
```
time ./build/wasm2c clang.wasm -o clang.c --enable-multi-memory
real    63m42.780s
user    63m15.818s
sys     0m24.107s
```
After this PR, it takes ~2 minutes.
```
time ./build/wasm2c clang.wasm -o clang.c --enable-multi-memory
real    2m12.889s
user    2m8.853s
sys     0m3.983s
```

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -828,7 +828,7 @@ std::string CWriter::DefineGlobalScopeName(ModuleFieldType type,
                          ExportName(StripLeadingDollar(name)));
 }
 
-/* Names for params, locals, and stack vars are formatted as "w2c_" + name. */
+/* Names for params, locals, and stack vars are formatted as "var_" + name. */
 std::string CWriter::DefineLocalScopeName(std::string_view name,
                                           bool is_label) {
   return ClaimUniqueName(
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -428,6 +428,7 @@ static constexpr char kParamSuffix =
 static constexpr char kLabelSuffix = kParamSuffix + 1;
 
 static constexpr char kSymbolPrefix[] = "w2c_";
+static constexpr char kLocalSymbolPrefix[] = "var_";
 static constexpr char kAdminSymbolPrefix[] = "wasm2c_";
 
 size_t CWriter::MarkTypeStack() const {
@@ -830,9 +831,9 @@ std::string CWriter::DefineGlobalScopeName(ModuleFieldType type,
 /* Names for params, locals, and stack vars are formatted as "w2c_" + name. */
 std::string CWriter::DefineLocalScopeName(std::string_view name,
                                           bool is_label) {
-  return ClaimUniqueName(local_syms_, local_sym_map_,
-                         is_label ? kLabelSuffix : kParamSuffix, name,
-                         kSymbolPrefix + MangleName(StripLeadingDollar(name)));
+  return ClaimUniqueName(
+      local_syms_, local_sym_map_, is_label ? kLabelSuffix : kParamSuffix, name,
+      kLocalSymbolPrefix + MangleName(StripLeadingDollar(name)));
 }
 
 std::string CWriter::DefineParamName(std::string_view name) {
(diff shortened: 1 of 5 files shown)


src/c-writer.cc
```diff
@@ -2360,7 +2361,7 @@ void CWriter::PushFuncSection(std::string_view include_condition) {
 void CWriter::Write(const Func& func) {
   func_ = &func;
   // Copy symbols from global symbol table so we don't shadow them.
```
sbc100, line 2363: I guess this comment is not longer accurate?

I guess its not possible for a global symbol could start with `var_`?

sbc100 (commented):
Wow!  Impressive speedup.

Do you know where the huge speedup comes from?  Was is just making N copies of the global symbol map?

src/c-writer.cc
```diff
@@ -428,6 +428,7 @@ static constexpr char kParamSuffix =
 static constexpr char kLabelSuffix = kParamSuffix + 1;
 
 static constexpr char kSymbolPrefix[] = "w2c_";
```
sbc100, line 430: Should we rename this `kGlobalSymbolPrefix`?

src/c-writer.cc
```diff
@@ -428,6 +428,7 @@ static constexpr char kParamSuffix =
...
+static constexpr char kLocalSymbolPrefix[] = "var_";
```
sbc100, line 431: I guess unlike the other prefixes, this one is 100% internal so changing it is not a user-visible change?

src/c-writer.cc
yhdengh: Global symbols should either start with `w2c_` or `wasm2c_`, and therefore local symbols would not shadow global symbols. Deleted this comment in [ce882d6](https://github.com/WebAssembly/wabt/commit/ce882d6a23803390ab098d908c1df2bac501651a).

src/c-writer.cc
yhdengh: Done in ce882d6a23803390ab098d908c1df2bac501651a.

src/c-writer.cc
yhdengh: Yes. So should I change `kLocalSymbolPrefix` to be a private field of `CWriter`?

yhdengh:
Based on our perf report, when executing `wasm2c` on `clang.wasm`, 90% of the time was spent on assigning `global_syms_` to `local_syms_`. So, yes, this huge speedup comes from not making N copies of `global_syms_`. But how much this PR speeds up in `wasm2c` would, I guess, differ from module to module, since each module has a different size of `global_syms_`.

src/c-writer.cc
sbc100: So sorry, I just meant that if we decided to change `var_` to `local_` in the figure it would not be user visible?

This change lgtm.

sbc100 (approved):
lgtm but maybe @keithw wants to take a look too. 

Those performance numbers are hard to argue with though!  Nice work.

src/c-writer.cc
```diff
@@ -830,9 +831,9 @@ std::string CWriter::DefineGlobalScopeName(ModuleFieldType type,
 /* Names for params, locals, and stack vars are formatted as "w2c_" + name. */
```
keithw, line 831: Let's update this comment too...

keithw (approved):
looks great to me % comment

src/c-writer.cc
yhdengh: Done in f4427d296c80dd2a44f5242ec656a4fdaa0732db

keithw: Great, setting to auto-merge.

PR #2173, opened 2023-03-15
binary-reader.cc: expand max local count and fix error message

### LINKED ISSUE #1593: Incorrect check for max allowed local count
Max allowed number of locals is `UINT32_MAX`, however WABT rejects it:

```sh
$ cat max_locals.wat
  (module binary
    "\00asm" "\01\00\00\00"
    "\01\04\01\60\00\00"       ;; Type section
    "\03\02\01\00"             ;; Function section
    "\0a\0a\01"                ;; Code section

    ;; function 0
    "\08\01"
    "\ff\ff\ff\ff\0f\7f"       ;; 0xFFFFFFFF i32
    "\0b"                      ;; end
  )
$ wat2wasm max_locals.wat
max_locals.wat:1:4: error: error in binary module: @0x0000001c: local count must be < 0x10000000
  (module binary
   ^^^^^^
```

My guess is that this check should use `<=`:
https://github.com/WebAssembly/wabt/blob/92e4cb9df0fab16083e32bf1b0fda0c4b6a55695/src/binary-reader.cc#L2429

Also I believe error message is one 0 off (should be `< 0x100000000`)

keithw:
On the tour of cleaning up what look like relatively easy old issues. (The error message was missing one '0'.)

Fixes #1593

### DIFF
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -2770,8 +2770,8 @@ Result BinaryReader::ReadCodeSection(Offset section_size) {
       Index num_local_types;
       CHECK_RESULT(ReadIndex(&num_local_types, "local type count"));
       total_locals += num_local_types;
-      ERROR_UNLESS(total_locals < UINT32_MAX,
-                   "local count must be < 0x10000000");
+      ERROR_UNLESS(total_locals <= UINT32_MAX,
+                   "local count must be < 0x100000000");
       Type local_type;
       CHECK_RESULT(ReadType(&local_type, "local type"));
       ERROR_UNLESS(IsConcreteType(local_type), "expected valid local type");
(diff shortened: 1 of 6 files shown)


src/binary-reader.cc
```diff
@@ -2770,8 +2770,8 @@ Result BinaryReader::ReadCodeSection(Offset section_size) {
       Index num_local_types;
       CHECK_RESULT(ReadIndex(&num_local_types, "local type count"));
       total_locals += num_local_types;
-      ERROR_UNLESS(total_locals < UINT32_MAX,
-                   "local count must be < 0x10000000");
+      ERROR_UNLESS(total_locals <= UINT32_MAX,
+                   "local count must be < 0x100000000");
```
sbc100, line 2774: How about `"%x", UINT32_MAX` ?

sbc100 (approved):

Issue #198, opened 2016-10-31
wasm-interp arguments


pannous:
It would be nice if we could call wasm-interp with a specific function and arguments

binji:
Agreed. Currently it's only really used for running the spec interpreter's testsuite, but this would be useful for quick tests.

PR #1167, opened 2019-09-25
Run wasm-interp on specific (exported) functions

amirbawab:
Closes: #198 

Minor changes allowing 'wasm-interp' to be executed for specific exported
wasm functions. Example syntax:

```
./wasm-interp                              \
-r "my_func" -a "i32:1" -a "i32:2", etc... \
file.wasm

./wasm-interp          \
-r "func_A" -a "f32:1" \
-r "func_A" -a "f32:2" \
-r "func_B" -a "i64:1" \
file.wasm
```

Currently supported types: i32, i64, f32, f64

PR #2176, opened 2023-03-18
implementation of --run-export and --argument

raphamorim:
Reference: https://github.com/WebAssembly/wabt/pull/1167 cc @keithw 
Closes: https://github.com/WebAssembly/wabt/issues/198

Allows `wasm-interp` to be execute specific exported wasm functions by name.

Quick example:

```lisp
(module (func (export "func_sum") (param i32) (param i32) (result i32)
  local.get 0
  local.get 1
  i32.add
  return))
```

```bash
$ ./wasm-interp sum.wasm -r func_sum -a "i32:8" -a "i32:5" 
func_sum(i32:8, i32:5) => i32:13
```

with `--trace`

```bash
$ ./wasm-interp sum.wasm -r func_sum -a "i32:8" -a "i32:5" --trace
>>> running export "func_sum":
#0.    0: V:2  | local.get $2
#0.    8: V:3  | local.get $2
#0.   16: V:4  | i32.add 8, 5
#0.   20: V:3  | drop_keep $2 $1
#0.   32: V:1  | return
func_sum(i32:8, i32:5) => i32:13
```

Supported types: i32, i64, f32, f64

## Overview

```bash
 ./wasm-interp --help
usage: wasm-interp [options] filename [arg]...

  read a file in the wasm binary format, and run in it a stack-based
  interpreter.

examples:
  # parse binary file test.wasm, and type-check it
  $ wasm-interp test.wasm

  # parse test.wasm and run all its exported functions
  $ wasm-interp test.wasm --run-all-exports

  # parse test.wasm, run the exported functions and trace the output
  $ wasm-interp test.wasm --run-all-exports --trace

  # parse test.wasm and run all its exported functions, setting the
  # value stack size to 100 elements
  $ wasm-interp test.wasm -V 100 --run-all-exports

  # parse test.wasm, run specific exported function by name with argument
  $ wasm-interp test.wasm -r "func_sum" -a "i32:8" -a "i32:5"

options:
      --help                                   Print this help message
      --version                                Print version information
  -v, --verbose                                Use multiple times for more info
      --enable-exceptions                      Enable Experimental exception handling
      --disable-mutable-globals                Disable Import/export mutable globals
      --disable-saturating-float-to-int        Disable Saturating float-to-int operators
      --disable-sign-extension                 Disable Sign-extension operators
      --disable-simd                           Disable SIMD support
      --enable-threads                         Enable Threading support
      --enable-function-references             Enable Typed function references
      --disable-multi-value                    Disable Multi-value
      --enable-tail-call                       Enable Tail-call support
      --disable-bulk-memory                    Disable Bulk-memory operations
      --disable-reference-types                Disable Reference types (externref)
      --enable-annotations                     Enable Custom annotation syntax
      --enable-code-metadata                   Enable Code metadata
      --enable-gc                              Enable Garbage collection
      --enable-memory64                        Enable 64-bit memory
      --enable-multi-memory                    Enable Multi-memory
      --enable-extended-const                  Enable Extended constant expressions
      --enable-relaxed-simd                    Enable Relaxed SIMD
      --enable-all                             Enable all features
  -V, --value-stack-size=SIZE                  Size in elements of the value stack
  -C, --call-stack-size=SIZE                   Size in elements of the call stack
  -t, --trace                                  Trace execution
  -r, --run-export=FUNCTION                    Run exported function by name
  -a, --argument=ARGUMENT                      Add argument to a exported function execution
      --wasi                                   Assume input module is WASI compliant (Export  WASI API the the module and invoke _start function)
  -e, --env=ENV                                Pass the given environment string in the WASI runtime
  -d, --dir=DIR                                Pass the given directory the the WASI runtime
      --run-all-exports                        Run all the exported functions, in order. Useful for testing
      --host-print                             Include an importable function named "host.print" for printing to stdout
      --dummy-import-func                      Provide a dummy implementation of all imported functions. The function will log the call and return an appropriate zero value.
```

```bash

```

### DIFF
--- a/src/tools/wasm-interp.cc
+++ b/src/tools/wasm-interp.cc
@@ -40,7 +40,7 @@
 using namespace wabt;
 using namespace wabt::interp;
 
-struct FunctionCallee {
+struct FunctionCall {
   std::string name;
   Values args;
 };
@@ -62,7 +62,7 @@ static bool s_host_print;
 static bool s_dummy_import_func;
 static Features s_features;
 static bool s_wasi;
-static std::vector<FunctionCallee> s_run_exports;
+static std::vector<FunctionCall> s_run_exports;
 static std::vector<std::string> s_wasi_env;
 static std::vector<std::string> s_wasi_argv;
 static std::vector<std::string> s_wasi_dirs;
@@ -95,7 +95,7 @@ examples:
   $ wasm-interp test.wasm -r "func_sum" -a "i32:8" -a "i32:5"
 )";
 
-Result ParseArgument(std::string argument, Value& val) {
+Result ParseWasmValue(std::string argument, Value& val) {
   Result result = Result::Ok;
 
   size_t cindex;
@@ -158,7 +158,7 @@ static void ParseOptions(int argc, char** argv) {
   parser.AddOption('r', "run-export", "FUNCTION",
                    "Run exported function by name",
                    [](const std::string& argument) {
-                     FunctionCallee func;
+                     FunctionCall func;
                      func.name = argument;
                      s_run_exports.push_back(func);
                    });
@@ -170,7 +170,7 @@ static void ParseOptions(int argc, char** argv) {
                          !s_run_exports.empty(),
                          "Cannot find a function execution for argument '%s'\n",
                          argument.c_str());
-                     ERROR_EXIT_UNLESS(Succeeded(ParseArgument(argument, val)),
+                     ERROR_EXIT_UNLESS(Succeeded(ParseWasmValue(argument, val)),
                                        "Failed to parse argument '%s'\n",
                                        argument.c_str());
                      s_run_exports.back().args.push_back(val);
@@ -210,7 +210,7 @@ static void ParseOptions(int argc, char** argv) {
 
 Result RunSpecificExports(const Instance::Ptr& instance,
                           Errors* errors,
-                          std::vector<FunctionCallee>& callees) {
+                          std::vector<FunctionCall>& calls) {
   Result result = Result::Ok;
 
   auto module = s_store.UnsafeGet<Module>(instance->module());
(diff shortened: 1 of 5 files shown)


src/tools/wasm-interp.cc
```diff
@@ -75,8 +90,49 @@ static const char s_description[] =
   # parse test.wasm and run all its exported functions, setting the
   # value stack size to 100 elements
   $ wasm-interp test.wasm -V 100 --run-all-exports
+
+  # parse test.wasm, run specific exported function by name with argument
+  $ wasm-interp test.wasm -r "func_sum" -a "i32:8" -a "i32:5"
 )";
 
+Result ParseArgument(std::string argument, Value& val) {
```
sbc100, line 98: How about `ParseWasmValue` or `ParseFunctionCallArgument`?

src/tools/wasm-interp.cc
```diff
@@ -39,6 +40,19 @@
 using namespace wabt;
 using namespace wabt::interp;
 
+struct FunctionCallee {
```
sbc100, line 43: The naming if this struct seems a little odd.  How about just just `FunctionCall` since this seems like more than just the callee and also includes a set of argument bindings

sbc100 (commented):
Nice!  Can we a test for these new arguments?

src/tools/wasm-interp.cc
raphamorim: updated [0fcea09](https://github.com/WebAssembly/wabt/pull/2176/commits/0fcea09a9c5edc4daf130bbf81b08b837ebe734c)

src/tools/wasm-interp.cc
raphamorim: Hmm, make sense. Wanted to rename the old struct from the reference PR but also agree that Callee is off. Updated [0fcea09](https://github.com/WebAssembly/wabt/pull/2176/commits/0fcea09a9c5edc4daf130bbf81b08b837ebe734c)

sbc100 (approved):

src/tools/wasm-interp.cc
```diff
@@ -132,6 +208,39 @@ static void ParseOptions(int argc, char** argv) {
...
+                          Errors* errors,
+                          std::vector<FunctionCall>& calls) {
+  Result result = Result::Ok;
+
+  auto module = s_store.UnsafeGet<Module>(instance->module());
+  auto&& module_desc = module->desc();
+
+  for (auto&& export_ : module_desc.exports) {
+    if (export_.type.type->kind != ExternalKind::Func) {
+      continue;
```
sbc100, line 221: Should this be a fatal error?

raphamorim: Not sure, that would make sense as well but I've decided to follow the already existent pattern https://github.com/WebAssembly/wabt/blob/494d145ec5425e71e2520d5b0427aacd6d2744ad/src/tools/wasm-interp.cc#L251

raphamorim: hmm, but this case we are explicitly declaring we want to run an external function by name. Isn't like is iterating... Think fatal would make more sense, should I update @sbc100 ?

sbc100: I think the difference is that "--run-all-exports" really means "--run-all-exported-functions" so in that case it makes sense to skip things that a non-functions.  Bug if you explicitly say `--run-export=foo` I would expect that to fail if `foo` is not a function.

raphamorim: I agree, updated the code and added an example. Please let me know anything

sbc100 (approved):

PR #2177, opened 2023-03-21
Silence warnings on GCC 12

### LINKED ISSUE #2020: warning: ‘index’ may be used uninitialized
Compiling WABT 1.0.30 with gcc 12.2.0, I get the following warning:

```
/usr/include/c++/12/bits/stl_pair.h:665:43: warning: ‘index’ may be used uninitialized [-Wmaybe-uninitialized]
  665 |              || (!(__y.first < __x.first) && __x.second < __y.second); }
      |                 ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
src/wat-writer.cc: In function ‘wabt::Result src/wat-writer.cc:1717:9: note: ‘index’ was declared here
 1717 |   Index index;
```

keithw:
These changes allow the build to succeed with `-DWERROR=ON` on GCC 12.

Fixes #2020

### DIFF
--- a/src/type-checker.cc
+++ b/src/type-checker.cc
@@ -513,12 +513,11 @@ Result TypeChecker::OnCallIndirect(const TypeVector& param_types,
 
 Result TypeChecker::OnIndexedFuncRef(Index* out_index) {
   Type type;
-  Result result = PeekType(0, &type);
+  CHECK_RESULT(PeekType(0, &type));
+  Result result = Result::Ok;
   if (!(type == Type::Any || type.IsReferenceWithIndex())) {
     TypeVector actual;
-    if (Succeeded(result)) {
-      actual.push_back(type);
-    }
+    actual.push_back(type);
     std::string message =
         "type mismatch in call_ref, expected reference but got " +
         TypesToString(actual);
--- a/src/wat-writer.cc
+++ b/src/wat-writer.cc
@@ -1722,7 +1722,7 @@ void WatWriter::WriteInlineExports(ExternalKind kind, Index index) {
 }
 
 bool WatWriter::IsInlineExport(const Export& export_) {
-  Index index;
+  Index index{};
   switch (export_.kind) {
     case ExternalKind::Func:
       index = module.GetFuncIndex(export_.var);
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -773,7 +773,7 @@ void CWriter::DefineImportName(const Import* import,
                                std::string_view module,
                                std::string_view field_name) {
   std::string name;
-  ModuleFieldType type;
+  ModuleFieldType type{};
 
   switch (import->kind()) {
     case ExternalKind::Func:
--- a/src/tools/spectest-interp.cc
+++ b/src/tools/spectest-interp.cc
@@ -189,9 +189,8 @@ int LaneCountFromType(Type type) {
 }
 
 ExpectedValue GetLane(const ExpectedValue& ev, int lane) {
-  int lane_count = LaneCountFromType(ev.lane_type);
   assert(ev.value.type == Type::V128);
-  assert(lane < lane_count);
+  assert(lane < LaneCountFromType(ev.lane_type));
 
   ExpectedValue result;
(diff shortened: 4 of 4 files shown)


src/type-checker.cc
```diff
@@ -172,10 +172,10 @@ Result TypeChecker::GetThisFunctionLabel(Label** label) {
 
 Result TypeChecker::PeekType(Index depth, Type* out_type) {
   Label* label;
+  *out_type = Type::Any;
```
sbc100, line 175: Why is this needed?

src/wat-writer.cc
```diff
@@ -1743,6 +1743,9 @@ bool WatWriter::IsInlineExport(const Export& export_) {
     case ExternalKind::Tag:
       index = module.GetTagIndex(export_.var);
       break;
+
+    default:
+      WABT_UNREACHABLE;
```
sbc100, line 1748: Doesn't adding this default case mean we don't get nice errors when we don't have full case coverage?  (e..g when we add a new extern kind?)

src/type-checker.cc
keithw: One of the callsites of PeekType (TypeChecker::OnIndexedFuncRef) assumes that it will initialize the `Type*` argument even in case of an `Error` result, but if the `Error` is because of a failure in the `CHECK_RESULT(TopLabel(&label));` line, that won't happen. Moving this line higher makes sure that the `Type*` gets initialized even in that kind of error.

sbc100: Isn't the callsite that is wrong then?  I don't think we should need to write to the output variable at all if we are returning an error

src/wat-writer.cc
sbc100: Well, that is shame.  I'd rather not loose those warnings.   Kind of shame that gcc can't figure that out but clang can.

What happens if you add explicit `Fist` and `Last` cases here instead (with a comment explaining why)?

sbc100: Is this a bug in gcc?

keithw: > What happens if you add explicit `Fist` and `Last` cases here instead (with a comment explaining why)?

GCC (and probably clang) complains about duplicate labels in the switch statement...

src/type-checker.cc
keithw: OK, let me revisit this and see how that ends up looking.

src/wat-writer.cc
sbc100: If GCC can't give the error/warning fidelity we want here perhaps we should simply disable that particular warning under GCC?

src/type-checker.cc
keithw: Here's an attempt to fix that callsite instead (which now adds some additional logged error messages to two tests).

src/wat-writer.cc
keithw: In the latest version, I kept the switches as-is and default-initialized the variable instead (in two places). It seems like it would be nice not to have to suppress "may be used uninitialized" warnings completely...?

test/typecheck/bad-callref-int32.txt
```diff
@@ -12,4 +12,7 @@
 out/test/typecheck/bad-callref-int32.txt:6:6: error: type mismatch in call_ref, expected reference but got [i32]
     (call_ref (i32.const 10)
      ^^^^^^^^
+out/test/typecheck/bad-callref-int32.txt:6:6: error: type mismatch at end of function, expected [] but got [i32]
+    (call_ref (i32.const 10)
+     ^^^^^^^^
```
sbc100, line 17: I'm no quite sure I understand why we now see more errors.  Is this a good thing?

sbc100 (approved):
lgtm % comment

test/typecheck/bad-callref-int32.txt
keithw: I don't think I understand the function-references implementation well enough to give a really considered answer, but I ended up finding a more minimal diff that also matches the current error messages 100%.

PR #2178, opened 2023-03-23
Switch preferred SHA-256 implementation from libcrypto -> libsodium

keithw:
The OpenSSL SHA256() implementation has been giving us heartburn when used from a multithreaded environment -- in some cases it can leak memory and make the code fail asan. (Under the hood it's calling pthread_once and mallocing a context structure the first time it's called, etc.)

This switches wabt's "preferred" SHA-256 implementation to search for libsodium, also a well-depended-on crypto implementation.

Unchanged by this PR:
- If libsodium isn't found, the vendored PicoSHA2 remains as the automatic fallback.
- Vendored PicoSHA2 can be forced with -DUSE_INTERNAL_SHA256.

Of our tools, only wasm2c ends up actually linking with libsodium.so because it's the only tool that actually uses SHA-256.

cc: @yhdengh

### DIFF
--- a/cmake/FindSodium.cmake
+++ b/cmake/FindSodium.cmake
@@ -1,0 +1,297 @@
+# Written in 2016 by Henrik Steffen Gaßmann <henrik@gassmann.onl>
+#
+# To the extent possible under law, the author(s) have dedicated all
+# copyright and related and neighboring rights to this software to the
+# public domain worldwide. This software is distributed without any warranty.
+#
+# You should have received a copy of the CC0 Public Domain Dedication
+# along with this software. If not, see
+#
+#     http://creativecommons.org/publicdomain/zero/1.0/
+#
+########################################################################
+# Tries to find the local libsodium installation.
+#
+# On Windows the sodium_DIR environment variable is used as a default
+# hint which can be overridden by setting the corresponding cmake variable.
+#
+# Once done the following variables will be defined:
+#
+#   sodium_FOUND
+#   sodium_INCLUDE_DIR
+#   sodium_LIBRARY_DEBUG
+#   sodium_LIBRARY_RELEASE
+#
+#
+# Furthermore an imported "sodium" target is created.
+#
+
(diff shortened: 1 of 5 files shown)


cmake/FindSodium.cmake
```diff
@@ -0,0 +1,297 @@
+# Written in 2016 by Henrik Steffen Gaßmann <henrik@gassmann.onl>
```
sbc100, line 1: Is this not something that should be shipped with libsodium?

sbc100:
Is there any reason you can't or don't want to switch to `-DUSE_INTERNAL_SHA256`?

keithw:
> Is there any reason you can't or don't want to switch to `-DUSE_INTERNAL_SHA256`?

Sigh. Maybe we should just do that. In general there are two reasons I haven't been thrilled to do that:

- My understanding is that the policy (for good reason) in most FOSS OSes (Debian, Ubuntu, Fedora, etc.) is that programs should link dynamically with crypto/security libraries rather than vendoring them

keithw:
Yeah, we switched to -DUSE_INTERNAL_SHA256 in our own use, but I figure this problem may affect others who use libwabt (or at least the WriteC API) in a multithreaded context.

Given how much of a pain these external dependencies are, maybe we should just strip out the dependency completely and make PicoSHA2 our only implementation. In general I've been reluctant to do that because:

- For software that's included with a GNU/Linux distribution, it's generally preferable (and I think distros often try to mandate) for wabt to link dynamically with a crypto library that the OS maintains and updates, rather than linking statically, because this is how the distro distributes security updates. They don't want to have to track down every package that vendors part of a library and manually update that dependency and then rebuild all those packages every time there's a bug found in the library.

- It feels safer to use a well-known and well-depended-on crypto implementation (e.g. from OpenSSL or libsodium) than "a random GitHub repo" (no offense PicoSHA2).

All that said, we are making pretty minimal/incidental use of SHA-256 so maybe we should just get over it and go with the easy route.

PR #2181, opened 2023-03-28
Avoid O(N^2) complexity when generating many local names

### LINKED ISSUE #2180: wasm2c hangs (an infinite loop) on certain inputs and cannot finish execution.
### Describe the bug
Certain `hang.wasm` causes `wasm2c` an infinite loop. `wasm2c` tries to access a memory that is not permitted instead of providing `type mismatch` error for a while.

`wasm2c --version`: 1.0.32 (git~1.0.32-46-g47a589a1)

--------------------------

### Content of the file that causes the issue:

`vim hang.wasm`:
```
^@asm^A^@^@^@^A^G^A`^B{^?^A~^C^B^A^@^@^L^A^H^@^@^@#^@^@^@^@^@^?
        ^A^G^Aàÿÿ^O^?^K

```

`cat hang.wasm`: 
```
asm`{~
     #
	???
```
--------------------------

### Steps to reproduce:

Here is the file for the bug: 

[hang.wasm.txt](https://github.com/WebAssembly/wabt/files/11082815/hang.wasm.txt) (Remove .txt extension and save as .wasm)

* Install and build WABT
* Run wasm2c on the test file: `wasm2c hang.wasm`
* Observe the crash: It will be observed that `wasm2c` cannot finish the execution.
	`wasm-validate hang.wasm` output: `Segmentation fault`
	`gdb wasm-validate` and subsequent `run hang.wasm` output: 
```
	Program received signal SIGSEGV, Segmentation fault.
	0x000000000047d8c6 in std::vector<wabt::TypeChecker::Label, std::allocator<wabt::TypeChecker::Label> >::_M_realloc_insert<wabt::LabelType&, std::vector<wabt::Type, std::allocator<wabt::Type> > const&, std::vector<wabt::Type, std::allocator<wabt::Type> > const&, unsigned long> (this=this@entry=0x7fffffffdd10, __position=__position@entry=non-dereferenceable iterator for std::vector, __args=@0x7fffffffd8d8: 0, __args=@0x7fffffffd8d8: 0, __args=@0x7fffffffd8d8: 0, __args=@0x7fffffffd8d8: 0) at /usr/lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/stl_vector.h:1762
1762		return (__len < size() || __len > max_size()) ? max_size() : __len;
```
--------------------------

### Expected Behavior:
The expected output  should be an error like other related tools would provide as below. To cite an example, run `wasm2wat` on the test file:
`wasm2wat hang.wasm`:

 The output will be:
```
Expected_Behavior/hang.wasm:0000027: error: type mismatch in implicit return, expected [i64] but got []
```
--------------------------

### Additional information
A combination of afl-fuzz++4.03a and Wasmlike, an Xsmith-based random program generator produced the snippet of code that caused the crash. https://www.flux.utah.edu/project/xsmith

keithw:
Because generated local names each have a unique index, and the index
is part of the generated name, they can't conflict with other generated local names.
As a result, NameGenerator::GenerateAndBindLocalNames() can avoid an O(N^2)
loop where each generated local name is checked against all previous local names.

(We could also do this for generated global/function/type/table/etc. names, but I think
every other type of named entity is different in that it has to exist syntactically in
the Wasm module, so any module that requires a lot of time to process would also
have to be large itself. Whereas a module can conjure up a bajillion params or locals
just by having a large count.)

Fixes #2180

### DIFF
--- a/src/generate-names.cc
+++ b/src/generate-names.cc
@@ -192,20 +192,27 @@ void NameGenerator::MaybeUseAndBindName(BindingHash* bindings,
 }
 
 void NameGenerator::GenerateAndBindLocalNames(Func* func) {
-  std::vector<std::string> index_to_name;
-  MakeTypeBindingReverseMapping(func->GetNumParamsAndLocals(), func->bindings,
-                                &index_to_name);
-  for (size_t i = 0; i < index_to_name.size(); ++i) {
-    const std::string& old_name = index_to_name[i];
-    if (!old_name.empty()) {
+  std::vector<bool> has_name(func->GetNumParamsAndLocals());
+  for (const auto& [name, binding] : func->bindings) {
+    has_name.at(binding.index) = true;
+  }
+  BindingHash generated_local_names;
+  for (size_t i = 0; i < has_name.size(); ++i) {
+    if (has_name[i]) {
       continue;
     }
-
     const char* prefix = i < func->GetNumParams() ? "p" : "l";
     std::string new_name;
-    GenerateAndBindName(&func->bindings, prefix, i, &new_name);
-    index_to_name[i] = new_name;
+    unsigned disambiguator = 0;
+    do {
+      GenerateName(prefix, i, disambiguator++, &new_name);
+    } while (func->bindings.find(new_name) != func->bindings.end());
+    // Because the generated local names each have a unique index, and the index
+    // is part of the generated name, we know they can't conflict with other
+    // generated local names.
+    generated_local_names.emplace(move(new_name), i);
   }
+  func->bindings.merge(generated_local_names);
 }
 
 Result NameGenerator::BeginBlockExpr(BlockExpr* expr) {


sbc100 (approved):

src/generate-names.cc
```diff
@@ -192,20 +192,27 @@ void NameGenerator::MaybeUseAndBindName(BindingHash* bindings,
...
+    unsigned disambiguator = 0;
+    do {
+      GenerateName(prefix, i, disambiguator++, &new_name);
+    } while (func->bindings.find(new_name) != func->bindings.end());
+    // Because the generated local names each have a unique index, and the index
+    // is part of the generated name, we know they can't conflict with other
+    // generated local names.
+    generated_local_names.emplace(move(new_name), i);
   }
+  func->bindings.merge(generated_local_names);
```
sbc100, line 215: I'm struggling the see where the N^2 is the old code.

Doesn't the old code just call `GenerateAndBindName` N times, and isn't `GenerateAndBindName` just doing a hash lookup in loop just like this?   I don't see how delaying the merging of generated_local_names into the bindings is going to speed this up.

keithw: The advantage here was supposed to be that the hash lookup is only checking for uniqueness against the fixed set of names in the names section (func->bindings) but not against the growing collection of newly generated local names.

Unfortunately, now that I'm trying to benchmark this rigorously, I can't replicate the speedup and I suspect I f'ed up the initial measurement that had made me hopeful about this approach. :-( I think we should probably give up on this PR and let #2182 be the "real" fix for this issue.

sbc100: Yeah, I don't imagine the number items in the hash would effect its lookup time... at least not in a material way, right?

PR #2184, opened 2023-03-28
Build arm64 for test pipeline

raphamorim:


.github/workflows/build.yml
```diff
@@ -78,7 +78,28 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: [arm64]
+    steps:
+    - uses: actions/checkout@v1
+      with:
+        submodules: true
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v2
```
sbc100, line 92: I guess github doesn't have support for arm64 runners yet?  Thats too bad.   Maybe once this is fixed we can move to real arm64 hardware: https://github.com/actions/runner-images/issues/5631

raphamorim: No it doesn't support, in another issue doesn't look something they're keen to implement soon https://github.com/actions/runner-images/issues/2552 . That would be the best, with Docker is takes forever, will close this PR.

raphamorim:
Sad because wanted to compare these timeouts the build gets with few tests with the ones I get locally but the build is too long 🙈

PR #2188, opened 2023-04-06
Support current Ubuntu 22.04 CMake version

### LINKED ISSUE #2187: Ubuntu 22.04 doesn't have CMake 3.16
I don't know to what degree that matters, but #1968 disabled building on latest LTS Ubuntu (22.04), which only has CMake 3.14 at the moment. Ubuntu tends to lag behind Debian.

penzn:
Closes #2187

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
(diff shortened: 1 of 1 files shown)


penzn:
I can't add reviewers here.

/CC @sbc100 @keithw

PR #2190, opened 2023-04-07
wasm2c: expose the initial and max size required from imported tables

shravanrn:
Follow up from PR #2113 to expose the min and max size of imported tables in the same way as imported memories


src/c-writer.cc
```diff
@@ -2471,15 +2471,19 @@ void CWriter::WriteImportProperties(CWriterPhase kind) {
...
+    } else if (import->kind() == ExternalKind::Table) {
+      limits = &(cast<TableImport>(import)->table.elem_limits);
+    } else {
+      continue;
     }
+
+    write_import_prop(import, "min", limits->initial);
+    write_import_prop(
+        import, "max",
+        limits->has_max ? limits->max : std::numeric_limits<uint32_t>::max());
```
keithw, line 2486: We may need to do something more complicated here because if `max` is unspecified, the implicit `max` depends on whether it's a table/memory32/memory64. I think it's:
- $2^{32} - 1$ elements (for a table)
- $2^{16}$ pages (for a memory.page_limits without `is_64` set)
- $2^{48}$ pages (for a memory.page_limits with `is_64` set)

(And, the types of the `min` and `max` variables probably should be u64 if limits.is_64 is set, and u32 otherwise.)

keithw (commented):
lgtm % comment

Would you be willing to add a simple output test (similar to the current test/wasm2c/add.txt) so we can track how this appears in the generated output?

sbc100:
> lgtm % comment
> 
> Would you be willing to add a simple output test (similar to the current test/wasm2c/add.txt) so we can track how this appears in the generated output?

I wonder if we should create a single kitchensink example/test for this purpose?  (Doesn't have to be part of this PR, just something to consider going forward).

src/c-writer.cc
keithw: (I wonder if it's easier just to not declare a `max` variable at all if the limits doesn't have a max? I.e. can your reflection system handle that situation?)

shravanrn: hmm... the reflection can handle this, but it would add a lot of unnecessary hoops to jump through to make things work. From what I can see, the path of least resistance here would be 

- Make the min/max of memory a u64 and default to a max of 2^16 or 2^48 based on `is_64` . (Multiple types here will again make this harder to handle so I think sticking with u64 will be simpler)
- Keep the min/max of table a u32 and default to a max of 2^32 - 1

Does this seem reasonable?

shravanrn:
> Would you be willing to add a simple output test (similar to the current test/wasm2c/add.txt) so we can track how this appears in the generated output?

Sure, will look into this

src/c-writer.cc
keithw: Works for me. At some point we may want to export an additional `is_64` variable because these are different types of memories (afaik an i32-indexed memory can't satisfy an import asking for an i64-indexed memory and vice versa, even if the max sizes are capped).

src/c-writer.cc
```diff
@@ -2472,13 +2472,23 @@ void CWriter::WriteImportProperties(CWriterPhase kind) {
...
-      write_import_prop(import, "min", memory_limits->initial);
-      write_import_prop(import, "max",
-                        memory_limits->has_max
-                            ? memory_limits->max
-                            : std::numeric_limits<uint32_t>::max());
+      const Limits* limits = &(cast<MemoryImport>(import)->memory.page_limits);
+      // We use u64 so we can handle both 32-bit and 64-bit memories
+      const uint64_t default_max = limits->is_64
+                                       ? (static_cast<uint64_t>(2) << 48)
+                                       : (static_cast<uint64_t>(2) << 16);
```
keithw, line 2479: Maybe `static_cast<uint64_t>(1)` (or just `1UL`) instead of 2? E.g.  1 << 16 = 65,536.

shravanrn: yikes, good catch. Unless there is a strong preference, I'll leave it as uint64_t though, as I think this keeps the intent clearer than UL.

shravanrn:
@keithw @sbc100 Could one of you sign off on this when you have a chance :)

test/wasm2c/call-indirect-import.txt
```diff
@@ -0,0 +1,932 @@
...
+    local.get 0
+    call_indirect (type 1))
+  (func $callee (type 1))
+  (func $__original_main (type 2) (result i32)
+    i32.const 1
+    call $invoke
+    i32.const 0)
+  (func $main (type 3) (param i32 i32) (result i32)
+    call $__original_main)
+  (elem (;0;) (i32.const 1) func $callee))
```
sbc100, line 19: This looks like it was compiler-generated, but could be simplified quite a bit if it was more hand-written.

For example, I don't think the `__linear_memory` is used at all?   And you don't need the `__original_main` / `main` split you?  (can just put the body of `__original_main` into `main` and delete `__original_main`?)

sbc100 (commented):
lgtm % comment

test/wasm2c/export-names.txt
```diff
@@ -844,8 +845,9 @@ static void init_instance_import(w2c_test* instance, struct w2c_0x5Cmodule* w2c_
   instance->w2c_0x5Cmodule_import0x200x2A0x2F = w2c_0x5Cmodule_import0x200x2A0x2F(w2c_0x5Cmodule_instance);
 }
 
-const u32 wasm2c_test_min_0x5Cmodule_import0x200x2A0x2F = 0;
-const u32 wasm2c_test_max_0x5Cmodule_import0x200x2A0x2F = 4294967295;
+const u64 wasm2c_test_min_0x5Cmodule_import0x200x2A0x2F = 0;
+const u64 wasm2c_test_max_0x5Cmodule_import0x200x2A0x2F = 65536;
+const u8 wasm2c_test_is64_0x5Cmodule_import0x200x2A0x2F = 0;
```
sbc100, line 850: Would it make more sence for `_is64` , `_min` and `_max` to be suffixes?

Or perhaps we should define a wasm_rt_limits struct that can contain all three of these?   I think this could be a followup since this change a clearly incremental on what we have already.

test/wasm2c/call-indirect-import.txt
shravanrn: I had left the linear memory in, so we could see the codegen for that import as well. I can rename the text to `check-import` to make it clear we are importing both tables and memory. I'll remove the duplicate mains.

sbc100: Sounds good!

test/wasm2c/export-names.txt
shravanrn: Makes sense. We can track this in a separate bug I think

shravanrn:
@sbc100 let me know if this looks good now

sbc100 (approved):

PR #2191, opened 2023-04-08
wasm2c: Fix warning of missing std namespace in call to move

shravanrn:


keithw (approved):
lgtm. I wonder whether the permissiveness of relying on ADL here depends on the compiler -- funny we weren't getting the warning in CI or on gcc 12 or clang 15.

PR #2193, opened 2023-04-08
wasm2c: Generate and checkin prebuilt source code in source folder

shravanrn:
In anticipation of Firefox moving to use upstream wasm2c, I am trying to simplify the build process for wasm2c.  

Wasm2c uses CMake generates source code during the build to transform source templates into strings in order to make development easy. However, generating source code on the fly with CMake requires us to jump though hoops in the Firefox build (also Firefox does not use CMake, so baking in a reliance on CMake to build causes added inconvenience).

This PR moves us to a model which hopefully give us the best of both worlds. The template files are still built by CMake, but are done in the src/prebuilt folder and checked in as part of source. This should allow us to continue using the template model during development, but also keep this easy to use in Firefox which can use the checked in pre-generated files.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -59,6 +59,8 @@ jobs:
       if: matrix.os == 'windows-latest'
     - name: build
       run: cmake --build out
+    - name: check if generated files are up-to-date
+      run: python ./scripts/check_clean.py
     - name: unittests
       run: cmake --build out --target run-unittests
     - name: c-api-tests
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -254,12 +254,13 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${WABT_SOURCE_DIR}/cmake)
 set(TEMPLATE_CMAKE ${WABT_SOURCE_DIR}/scripts/gen-wasm2c-templates.cmake)
 
 add_custom_command(
-  OUTPUT wasm2c_header_top.cc wasm2c_header_bottom.cc wasm2c_source_includes.cc wasm2c_source_declarations.cc
+  OUTPUT gen-wasm2c-prebuilt
 
-  COMMAND ${CMAKE_COMMAND} -D out="wasm2c_header_top.cc" -D in="${WABT_SOURCE_DIR}/src/template/wasm2c.top.h" -D symbol="s_header_top" -P ${TEMPLATE_CMAKE}
-  COMMAND ${CMAKE_COMMAND} -D out="wasm2c_header_bottom.cc" -D in="${WABT_SOURCE_DIR}/src/template/wasm2c.bottom.h" -D symbol="s_header_bottom" -P ${TEMPLATE_CMAKE}
-  COMMAND ${CMAKE_COMMAND} -D out="wasm2c_source_includes.cc" -D in="${WABT_SOURCE_DIR}/src/template/wasm2c.includes.c" -D symbol="s_source_includes" -P ${TEMPLATE_CMAKE}
-  COMMAND ${CMAKE_COMMAND} -D out="wasm2c_source_declarations.cc" -D in="${WABT_SOURCE_DIR}/src/template/wasm2c.declarations.c" -D symbol="s_source_declarations" -P ${TEMPLATE_CMAKE}
+  COMMAND ${CMAKE_COMMAND} -D out="${WABT_SOURCE_DIR}/src/prebuilt/wasm2c_header_top.cc" -D in="${WABT_SOURCE_DIR}/src/template/wasm2c.top.h" -D symbol="s_header_top" -P ${TEMPLATE_CMAKE}
+  COMMAND ${CMAKE_COMMAND} -D out="${WABT_SOURCE_DIR}/src/prebuilt/wasm2c_header_bottom.cc" -D in="${WABT_SOURCE_DIR}/src/template/wasm2c.bottom.h" -D symbol="s_header_bottom" -P ${TEMPLATE_CMAKE}
+  COMMAND ${CMAKE_COMMAND} -D out="${WABT_SOURCE_DIR}/src/prebuilt/wasm2c_source_includes.cc" -D in="${WABT_SOURCE_DIR}/src/template/wasm2c.includes.c" -D symbol="s_source_includes" -P ${TEMPLATE_CMAKE}
+  COMMAND ${CMAKE_COMMAND} -D out="${WABT_SOURCE_DIR}/src/prebuilt/wasm2c_source_declarations.cc" -D in="${WABT_SOURCE_DIR}/src/template/wasm2c.declarations.c" -D symbol="s_source_declarations" -P ${TEMPLATE_CMAKE}
+  COMMAND ${CMAKE_COMMAND} -E touch gen-wasm2c-prebuilt
 
   DEPENDS ${WABT_SOURCE_DIR}/src/template/wasm2c.top.h
   ${WABT_SOURCE_DIR}/src/template/wasm2c.bottom.h
@@ -267,7 +268,12 @@ add_custom_command(
   ${WABT_SOURCE_DIR}/src/template/wasm2c.declarations.c
 )
 
-set(CWRITER_TEMPLATE_SRC wasm2c_header_top.cc wasm2c_header_bottom.cc wasm2c_source_includes.cc wasm2c_source_declarations.cc)
+add_custom_target(gen-wasm2c-prebuilt-target DEPENDS gen-wasm2c-prebuilt)
+
+set(CWRITER_TEMPLATE_SRC ${WABT_SOURCE_DIR}/src/prebuilt/wasm2c_header_top.cc
+                         ${WABT_SOURCE_DIR}/src/prebuilt/wasm2c_header_bottom.cc
+                         ${WABT_SOURCE_DIR}/src/prebuilt/wasm2c_source_includes.cc
+                         ${WABT_SOURCE_DIR}/src/prebuilt/wasm2c_source_declarations.cc)
 
 add_custom_target(everything)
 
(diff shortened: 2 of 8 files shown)


keithw:
I don't really feel strongly here, but it would sure be nice to keep prebuilt files out of our repo if possible (since as I recall we were always forgetting to update these on checkin in the past they used to be in the repo). OTOH, it would be great to have Firefox using the mainline wasm2c. Are any of these alternate approaches acceptable to Mozilla?

1. We provide a Python script in the repo (or an executable in C that gets built?) that produces the prebuilt files, or
2. We build the prebuilt files as a workflow artifact and publish them via GitHub on every push.

If not, I agree this is probably the next-best thing...

shravanrn:
@keithw **TLDR:** Perhaps option 2, but there are some caveats (details below). The in-repo source is probably the preferred option, but maybe we can add to CI checks to prevent out-of-date files? Thoughts?

**Details**

>     1. We provide a Python script in the repo (or an executable in C that gets built?) that produces the prebuilt files, or

In theory, something like this is well supported in the Mozilla build, and I wrote this sort of python script in my attempt to integrate this with the build system. Unfortunately, this is an odd case for Mozilla's build as
- this python script generates source for wasm2c
- which then must be built and invoked to generated C source for Firefox
- which must then be compiled into Mozilla source

This added layer of nesting (generating source for a source generator) seems to hit an odd edge case in the build system where a python script doesn't quite work out. Longer term, this build system edge case should/will get fixed, but I was hoping we could employ a simpler solution for now like the one proposed.

> We build the prebuilt files as a workflow artifact and publish them via GitHub on every push.

While this is not the common/preferred path in the Mozilla source tree, this could work. The caveat here is that the Workflow artifacts of old pushes can't be deleted as Firefox is likely not going to follow the latest commit of wasm2c. I would expect it to be periodically updated, but lag up to 6 to 9 months behind on commits (unless there are security fixes)

> As I recall we were always forgetting to update these on checkin in the past when they used to be in the repo

Fair point! Fwiw, this updates the generated source anytime you build after modifying the template files. So hopefully this is less likely. We could add a CI step to generate source and make sure there are no changes? (Not super sure i know how to do this of the top of my head, but something like this should be possible)

Anyway, let me know your thoughts of preferred path given the above.

sbc100:
I think having a pre-commit check in CI to make sure these files are up-to-date sounds like a reasonable solution.

src/prebuilt/wasm2c_header_top.cc
```diff
@@ -0,0 +1,57 @@
...
+#endif
+)w2c_template"
+R"w2c_template(
+#ifdef __cplusplus
+)w2c_template"
+R"w2c_template(extern "C" {
+)w2c_template"
+R"w2c_template(#endif
+)w2c_template"
+;
```
sbc100, line 57: Can we modify the generator such that these files end with a newline.

sbc100:
For the pre-commit check you could use something like this script in emscripten: https://github.com/emscripten-core/emscripten/blob/main/tools/check_clean.py

If we run a script like that after the build it will fail if there are any changes to the prebuilt files.

shravanrn:
> For the pre-commit check you could use something like this script in emscripten: https://github.com/emscripten-core/emscripten/blob/main/tools/check_clean.py
> 
> If we run a script like that after the build it will fail if there are any changes to the prebuilt files.

Oh that example is perfect. I have added it as a lint check for now. Not sure if you wanted it as a git pre-commit hook or something else as well. Let me know.

@sbc100 Also just an fyi that to ensure timestamps aren't messed up I needed to use cmake copy "ONLY_IF_DIFFERENT" flag, which is available only in cmake 3.21. Let me know if there are any concerns with bumping the cmake version as I have done in the updated PR. If so, I can write some code to manually do the same thing with the current cmake version.

.github/workflows/build.yml
```diff
@@ -23,6 +23,7 @@ jobs:
         sudo apt-get install clang-format
     - run: flake8
     - run: ./scripts/clang-format-diff.sh
+    - run: python ./scripts/check_clean.py
```
sbc100, line 26: I think this needs to run after the build, right?  (i.e. it probably needs to move to "build")

sbc100:
Why do the timestamps matter in this case?

shravanrn:
> Why do the timestamps matter in this case?

@sbc100 Oh sorry, i should have worded that better. 
Typically, when running `make` twice, we'd expect the second run of `make` to act like a noop. However, without the `copy_if_different`, the build always rebuilds the template files and updates the timestamps. This happens even if the template files have not been modified. Thus every call `make` rebuilds and relink the source files generated from the templates, even when not necessary. The `copy_if_different` fixes this.

.github/workflows/build.yml
shravanrn: Good point. Will fix.

sbc100:
> > Why do the timestamps matter in this case?
> 
> @sbc100 Oh sorry, i should have worded that better. Typically, when running `make` twice, we'd expect the second run of `make` to act like a noop. However, without the `copy_if_different`, the build always rebuilds the template files and updates the timestamps. This happens even if the template files have not been modified. Thus every call `make` rebuilds and relink the source files generated from the templates, even when not necessary. The `copy_if_different` fixes this.

So why was this not needed before?  Why can't the dependency system in cmake handle only rebuilding the outputs if the inputs change?

shravanrn:
> So why was this not needed before? Why can't the dependency system in cmake handle only rebuilding the outputs if the inputs change?

@sbc100 So previously, the template was converted to generated files as a CMake custom command that produced specific files. CMake is good about only building if dependencies changed in this setting. This entire setup works if you are generating source files that aren't checked in as we were doing before. This approach, however, has one annoying characteristic when generating source that needs to be checked in --- it deletes the generated source when we run `make clean`. So if a developer working on any part of wabt does a `make clean` the deleted generated source is now a part of the changes in their working tree, which could be confusing. Moving from a CMake custom command to a CMake custom target fixes this, but unfortunately, the custom target does not have the "Build only when inputs changed" behavior, so we have to compensate for that. Happy to try an alternate approach if you have one in mind.

sbc100:
> > So why was this not needed before? Why can't the dependency system in cmake handle only rebuilding the outputs if the inputs change?
> 
> @sbc100 So previously, the template was converted to generated files as a CMake custom command that produced specific files. CMake is good about only building if dependencies changed using this approach. Overall, this approach works if you are generating source files that aren't checked in as we were doing before.
> 
> This same approach, however, has one annoying characteristic when generating source that needs to be checked in --- it deletes the generated source when we run `make clean`. So if a developer working on any part of wabt does a `make clean` the deleted generated source is now a part of the changes in their working tree, which could be confusing.
> 
> Moving from a CMake custom command to a CMake custom target ensures the generated files aren't delated on a `make clean`, but unfortunately, the custom target does not have the "Build only when inputs changed" behavior, so we have to compensate for that. I'm happy to try an alternate approach to get this behavior if you have one in mind.

With the new system when is the generator run?    Is it run whenever you run `ninja`/`make` or only if the input files are newer than the generated code?

shravanrn:
> With the new system when is the generator run? Is it run whenever you run ninja/make or only if the input files are newer than the generated code?

@sbc100 With the new system, the generator is run to a tmp file every time make is run, but the tmp file is copied over only if the contents are newer. Let me know if you have concerns with this approach

sbc100:
> > With the new system when is the generator run? Is it run whenever you run ninja/make or only if the input files are newer than the generated code?
> 
> @sbc100 With the new system, the generator is run to a tmp file every time make is run, but the tmp file is copied over only if the contents are newer. Let me know if you have concerns with this approach

It seems unfortunate that running a no-op `ninja` or `make` would not be no-op.   Ideally it would be good if running `ninja` a second times doesn't do anything at all.

shravanrn:
> It seems unfortunate that running a no-op `ninja` or `make` would not be no-op. Ideally it would be good if running `ninja` a second times doesn't do anything at all.

This is now fixed in the new PR. This approach keeps the second build a noop. It also updates the in-tree sources only if the template file has changed.

This is possible as we create a dummy output file that depends on the input templates (leveraging cmake's dependency tracking). The generated files are then created as ``byproducts'' while building the dummy file. These byproducts are not removed during a make clean though. 

(There is a minor detail that it updates the timestamp of the generated files on a clean build, but this doesn't interfere with any usual git or build workflow)

@sbc100 Does this approach seem reasonable

sbc100:
> @sbc100 Does this approach seem reasonable

Sounds good!  Thanks for your patience on this.

CMakeLists.txt
```diff
@@ -254,20 +254,26 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${WABT_SOURCE_DIR}/cmake)
...
+  COMMAND ${CMAKE_COMMAND} -E touch gen-wasm2c-prebuilt
 
   DEPENDS ${WABT_SOURCE_DIR}/src/template/wasm2c.top.h
   ${WABT_SOURCE_DIR}/src/template/wasm2c.bottom.h
   ${WABT_SOURCE_DIR}/src/template/wasm2c.includes.c
   ${WABT_SOURCE_DIR}/src/template/wasm2c.declarations.c
 )
 
-set(CWRITER_TEMPLATE_SRC wasm2c_header_top.cc wasm2c_header_bottom.cc wasm2c_source_includes.cc wasm2c_source_declarations.cc)
+add_custom_target(gen-wasm2c-prebuilt-target DEPENDS gen-wasm2c-prebuilt)
```
sbc100, line 271: Do you still need this extra target or can you just do `add_dependencies(wabt gen-wasm2c-prebuilt)` below?

sbc100 (approved):

CMakeLists.txt
shravanrn: I had tried that earlier and it doesn't quite seem to work. I think `add_dependencies` works only on targets and not files?

sbc100: Interesting, lgtm in that case.  (I do wonder if we can maybe avoid have both a custom command and a custom target here, but I'm not sure it worth spending too much time on).

sbc100 (approved):

PR #2194, opened 2023-04-10
wasm2c: add signature for the wasm2c trap handler if defined

shravanrn:


sbc100 (approved):
I guess we don't any tests for this in the repo?   Might be good to add one?  (does't have to be part of this PR).

shravanrn:
> I guess we don't any tests for this in the repo? Might be good to add one? (does't have to be part of this PR).

I'll create a bug for now. I can definitely write a custom test for this, but would you or @keithw have thoughts on the best way to add this to the existing harness. The test probably has to look something like this.

```
**Wasm module**
memory size = 1
func foo that accesses memory index 2

**Compilation**
Compile above module with
- flag `-DWASM_RT_TRAP_HANDLER=trap_handler`
- A main func that invokes the foo
- A `trap_handler` func

Test succeeds if trap_handler is called and fails otherwise.
```

PR #2196, opened 2023-04-11
wasm2c: Separate the macros for allocation and bounds checks strategies

shravanrn:
(Edited for the latest push)

The wasm2c runtime implementation today blends a couple of concepts together that make it hard to use in 32-bit environment. These are the
1. The memory allocation/growth strategy - **mmap** (memories supporting in-place growth) vs **malloc** (memories that may be moved to grow)
2. The OOB prevention strategy - **Guardpages** (which requires **mmap** memory, 64-bit platforms and WasmMemory32) vs **Boundschecks**. I am planing to add other strategies in the future. 

When using wasm2c in Firefox with RLBox, we want the following configurations
1. **mmap** always as RLBox does not support moveable memories
2. **Guardpages** on 64-bit and **BoundsChecks** on 32-bit (as guard pages are not supported in 32-bit)

Since wasm2c blends this concepts together these attributes are currently controlled by one macro `WASM_RT_MEMCHECK_SIGNAL_HANDLER` which says either use (**mmap** and **Guardpages**, supported only on a 64-bit platform) or (**malloc** and **BoundsChecks**)

This means we cannot get the 32-bit config we are using in Firefox with our wasm2c fork: **mmap** and **BoundsChecks** 

I want to propose decoupling these concepts each controllable with their own macros. We can do sanity checks to ensure disallowed combinations don't go through  (e.g., **malloc** with **Guardpages**). We can then eliminate the `WASM_RT_MEMCHECK_SIGNAL_HANDLER` altogether.  

I have also added some optional code to translate WASM_RT_MEMCHECK_SIGNAL_HANDLER to use the new macros we have added. We can add this if we want to preserve compat with this previous macro, but I am happy to remove this.

@keithw Heads up on this one, as this changes the `wasm-rt.h` header (But the optional compat code can preserve the old behavior if we want this)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -109,7 +109,7 @@ jobs:
     env:
       USE_NINJA: "1"
       CC: "clang" # used by the wasm2c tests
-      WASM2C_CFLAGS: "-march=x86-64-v2 -fsanitize=address -DWASM_RT_MEMCHECK_SIGNAL_HANDLER=0"
+      WASM2C_CFLAGS: "-march=x86-64-v2 -fsanitize=address -DWASM_RT_USE_MMAP=0"
     steps:
     - uses: actions/setup-python@v1
       with:
--- a/src/template/wasm2c.declarations.c
+++ b/src/template/wasm2c.declarations.c
@@ -41,7 +41,7 @@ static inline bool func_types_eq(const wasm_rt_func_type_t a,
     TRAP(OOB);
 #endif
 
-#if WASM_RT_MEMCHECK_SIGNAL_HANDLER
+#if WASM_RT_MEMCHECK_GUARD_PAGES
 #define MEMCHECK(mem, a, t)
 #else
 #define MEMCHECK(mem, a, t) RANGE_CHECK(mem, a, sizeof(t))
--- a/wasm2c/wasm-rt-impl.c
+++ b/wasm2c/wasm-rt-impl.c
@@ -25,8 +25,7 @@
 #include <stdlib.h>
 #include <string.h>
 
-#if WASM_RT_MEMCHECK_SIGNAL_HANDLER && !WASM_RT_SKIP_SIGNAL_RECOVERY && \
-    !defined(_WIN32)
+#if WASM_RT_INSTALL_SIGNAL_HANDLER && !defined(_WIN32)
 #include <signal.h>
 #include <unistd.h>
 #endif
@@ -45,7 +44,7 @@
 #define PAGE_SIZE 65536
 #define MAX_EXCEPTION_SIZE PAGE_SIZE
 
-#if WASM_RT_MEMCHECK_SIGNAL_HANDLER && !WASM_RT_SKIP_SIGNAL_RECOVERY
+#if WASM_RT_INSTALL_SIGNAL_HANDLER
 static bool g_signal_handler_installed = false;
 #ifdef _WIN32
 static void* g_sig_handler_handle = 0;
@@ -167,7 +166,7 @@ static void os_print_last_error(const char* msg) {
   }
 }
 
-#if WASM_RT_MEMCHECK_SIGNAL_HANDLER && !WASM_RT_SKIP_SIGNAL_RECOVERY
+#if WASM_RT_INSTALL_SIGNAL_HANDLER
 
 static LONG os_signal_handler(PEXCEPTION_POINTERS info) {
   if (info->ExceptionRecord->ExceptionCode == EXCEPTION_ACCESS_VIOLATION) {
@@ -211,7 +210,7 @@ static void os_print_last_error(const char* msg) {
   perror(msg);
 }
 
-#if WASM_RT_MEMCHECK_SIGNAL_HANDLER && !WASM_RT_SKIP_SIGNAL_RECOVERY
+#if WASM_RT_INSTALL_SIGNAL_HANDLER
 static void os_signal_handler(int sig, siginfo_t* si, void* unused) {
   if (si->si_code == SEGV_ACCERR) {
     wasm_rt_trap(WASM_RT_TRAP_OOB);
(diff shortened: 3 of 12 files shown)


wasm2c/wasm-rt.h
```diff
@@ -60,52 +60,165 @@ extern "C" {
...
+ * consumption.
  */
+#ifndef WASM_RT_MEMORY_ALLOCATION_MMAP
+#define WASM_RT_MEMORY_ALLOCATION_MMAP 0
+#endif
 
-#ifndef WASM_RT_SKIP_SIGNAL_RECOVERY
-#define WASM_RT_SKIP_SIGNAL_RECOVERY 0
+#ifndef WASM_RT_MEMORY_ALLOCATION_MALLOC
+#define WASM_RT_MEMORY_ALLOCATION_MALLOC 0
```
sbc100, line 99: As far as I can tell this is never actually used.   How about we just have a single macro (`WASM_RT_MEMORY_ALLOCATION_MMAP`) with malloc being used when its not set.   We could even use a shorter name like `WASM_RT_USE_MMAP`?

wasm2c/wasm-rt.h
```diff
@@ -60,52 +60,165 @@ extern "C" {
...
+ * This defaults to GUARD as this is the fasest option, iff the requirements of
+ * GUARD --- 64-bit platforms, MMAP allocation strategy, no 64-bit memories ---
+ * are met. This falls back to BOUNDS otherwise.
+ */
+#ifndef WASM_RT_MEMORY_CHECK_GUARD
+#define WASM_RT_MEMORY_CHECK_GUARD 0
+#endif
+
+#ifndef WASM_RT_MEMORY_CHECK_BOUNDS
+#define WASM_RT_MEMORY_CHECK_BOUNDS 0
```
sbc100, line 135: Same here.. since this is basically a binary setting can we just have a single macro?  

Doesn't the existing `WASM_RT_MEMCHECK_SIGNAL_HANDLER` means that same thing as the new `WASM_RT_MEMORY_CHECK_GUARD`?

wasm2c/wasm-rt.h
```diff
@@ -60,52 +60,165 @@ extern "C" {
...
+ * that use their own signal handler, but nevertheless want to use MMAP
+ * allocation and GUARD checks can do so by allow thing runtime to
+ * TRAP_OR_ABORT.
+ */
+#ifndef WASM_RT_TRAP_MUSTTRAP
+#define WASM_RT_TRAP_MUSTTRAP 0
+#endif
+
+#ifndef WASM_RT_TRAP_CANABORT
+#define WASM_RT_TRAP_CANABORT 0
```
sbc100, line 186: Again this seems like it could/should be just a single macro.  Something like "WASM_RT_HANDLE_TRAPS"?   FF would than set that to 0 to signal that it is going to handle traps in its own existing signal handler?

wasm2c/wasm-rt.h
shravanrn: Sure, I can make this change!

wasm2c/wasm-rt.h
shravanrn: Binary setting - will fix!

> WASM_RT_MEMCHECK_SIGNAL_HANDLER vs WASM_RT_MEMORY_CHECK_GUARD

Perhaps, but part of the problem is that the terminology here is confusing, and hence I am not even sure what the WASM_RT_MEMCHECK_SIGNAL_HANDLER  should really be. WASM_RT_MEMORY_CHECK_GUARD on the other hand is clear --- Semantically, it indicates that OOB's should be **prevented** with guard pages (vs bounds checks). Whether we use a signal handler is more of an internal detail when implementing the behavior needed for `WASM_RT_MEMORY_CHECK_GUARD` and `MustTrap`. By decoupling the check from the handling, this can be much clearer.

wasm2c/wasm-rt.h
shravanrn: Binary setting --- will fix!

wasm2c/wasm-rt.h
shravanrn: Re the binary setting on this --- I do intend to add other enforcement mechanisms here that are neither guard pages or bounds checks. To avoid making another breaking change in the future, it may be useful to leave this is a separate macros. 

Alternately, if you think we are better off leaving this as binary for now, and making the change later, that is fine too. Let me know the preferred path.

shravanrn:
@sbc100 I simplified this PR and updated the description in the first comment. Does this look better?

.github/workflows/build.yml
```diff
@@ -107,7 +107,7 @@ jobs:
     env:
       USE_NINJA: "1"
       CC: "clang" # used by the wasm2c tests
-      WASM2C_CFLAGS: "-march=x86-64-v2 -fsanitize=address -DWASM_RT_MEMCHECK_SIGNAL_HANDLER=0"
+      WASM2C_CFLAGS: "-march=x86-64-v2 -fsanitize=address -WASM_RT_MEMORY_ALLOCATION_MMAP=0"
```
sbc100, line 110: Missing `-D` here I think

src/template/wasm2c.declarations.c
```diff
@@ -41,7 +41,7 @@ static inline bool func_types_eq(const wasm_rt_func_type_t a,
     TRAP(OOB);
 #endif
 
-#if WASM_RT_MEMCHECK_SIGNAL_HANDLER
+#if WASM_RT_MEMORY_CHECK_GUARD
```
sbc100, line 44: Somehow `WASM_RT_MEMORY_CHECK_GUARD` doesn't scan in the way that I would like it to.    Not a huge deal, of course.   How about `WASM_RT_USE_GUARD_PAGES`?

wasm2c/wasm-rt-impl.c
```diff
@@ -281,31 +280,49 @@ void wasm_rt_init(void) {
...
 }
 
 void wasm_rt_free(void) {
-#if WASM_RT_MEMCHECK_SIGNAL_HANDLER && !WASM_RT_SKIP_SIGNAL_RECOVERY
+#if WASM_RT_INSTALL_SIGNAL_HANDLER
   os_cleanup_signal_handler();
 #endif
 }
 
+#if WASM_RT_MEMORY_ALLOCATION_MMAP
```
sbc100, line 296: Likelywise what do you think of `WASM_RT_USE_MMAP`?   This read like the kind of cmake/autoconf configuration macro that I grew up on so maybe I'm biased?

sbc100 (commented):
lgtm % more bikeshedding

sbc100 (approved):

PR #2197, opened 2023-04-12
wasm2c: Add optional WASM_RT_GROW_FAILED_HANDLER to be notified of memory growth failures

shravanrn:
Memory growth failures (when `wasm_rt_grow_memory` returns -1) cause the libc used by wasm toolchains to return a 0 during malloc. This is usually ignored by most applications, and they continue execution resulting in corrupted internal state.

For the use in Firefox, we need a hook to be notified of memory growth failures, so we can annotate subsequent crash reports with this information.

### DIFF
--- a/wasm2c/README.md
+++ b/wasm2c/README.md
@@ -322,7 +322,13 @@ an i32 or i64 address.
 of pages. If there isn't enough memory to do so, or the new page count would be
 greater than the maximum page count, the function must fail by returning
 `0xffffffff`. If the function succeeds, it must return the previous size of the
-memory instance, in pages.
+memory instance, in pages. The host can optionally be notified of failures by
+defining a function the signature `void wasm2c_grow_failed_handler()` and
+compiling the runtime with the with macro definition `#define
+WASM_RT_GROW_FAILED_HANDLER wasm2c_grow_failed_handler`. It is recommended that
+you add this macro definition via a compiler flag
+(`-WASM_RT_GROW_FAILED_HANDLER=wasm2c_grow_failed_handler` on clang/gcc).
+
 
 `wasm_rt_free_memory` frees the memory instance.
 
--- a/wasm2c/wasm-rt-impl.c
+++ b/wasm2c/wasm-rt-impl.c
@@ -70,6 +70,10 @@ static WASM_RT_THREAD_LOCAL wasm_rt_jmp_buf* g_unwind_target;
 extern void WASM_RT_TRAP_HANDLER(wasm_rt_trap_t code);
 #endif
 
+#ifdef WASM_RT_GROW_FAILED_HANDLER
+extern void WASM_RT_GROW_FAILED_HANDLER();
+#endif
+
 void wasm_rt_trap(wasm_rt_trap_t code) {
   assert(code != WASM_RT_TRAP_NONE);
 #if WASM_RT_USE_STACK_DEPTH_COUNT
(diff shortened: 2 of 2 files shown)


wasm2c/wasm-rt-impl.c
```diff
@@ -345,7 +349,8 @@ void wasm_rt_allocate_memory(wasm_rt_memory_t* memory,
 #endif
 }
 
-uint64_t wasm_rt_grow_memory(wasm_rt_memory_t* memory, uint64_t delta) {
+static uint64_t wasm_rt_grow_memory_helper(wasm_rt_memory_t* memory,
```
sbc100, line 352: How about just calling this `grow_memory`.. or `grow_memory_impl`, no need to namespace/prefix it since its static.

sbc100 (commented):
Would your use case be satisfied by something that was just `WASM_RT_ABORT_ON_MEMORY_GROWTH_FAILURE` ?  Or do you need to do more than just abort?

On alternative to the macro would be to use a weak symbol:

```
if ((ret == -1 && wasm2c_grow_failed_handler) {
  wasm2c_grow_failed_handler();
}
```

although I guess one downside of that is that `__attribute__(weak)` is not portable..

shravanrn:
> Would your use case be satisfied by something that was just `WASM_RT_ABORT_ON_MEMORY_GROWTH_FAILURE` ? Or do you need to do more than just abort?

We actually don't abort at the moment. We simply make a note on future crashes if there was a memory growth failure prior to the crash. This is in the off chance that the sandboxed code can actually handle a malloc failure. We may move to abort in the future though. Either way, I think giving the embedder the flexibility of what to do is probably required.

> On alternative to the macro would be to use a weak symbol:
> 
> ```
> if ((ret == -1 && wasm2c_grow_failed_handler) {
>   wasm2c_grow_failed_handler();
> }
> ```
> 
> although I guess one downside of that is that `__attribute__(weak)` is not portable..

wasm2c/wasm-rt-impl.c
shravanrn: Will fix

wasm2c/README.md
```diff
@@ -322,7 +322,13 @@ an i32 or i64 address.
 of pages. If there isn't enough memory to do so, or the new page count would be
 greater than the maximum page count, the function must fail by returning
 `0xffffffff`. If the function succeeds, it must return the previous size of the
-memory instance, in pages.
+memory instance, in pages. The host can optionally be notified of failures by
+defining a function the signature `void wasm2c_grow_failed_handler()` and
+compiling the runtime with the with macro definition `#define
+WASM_RT_GROW_FAILED_HANDLER wasm2c_grow_failed_handler`. It is recommended that
+you add this macro definition via a compiler flag
+(`-WASM_RT_GROW_FAILED_HANDLER=wasm2c_grow_failed_handler` on clang/gcc).
```
sbc100, line 330: How about:

```
The host can optionally be notified of failures by
compiling the runtime with `WASM_RT_GROW_FAILED_HANDLER`
defined to the name of a handler function.  The handler function
should be a void function taking no arguments.  
e.g. `-DWASM_RT_GROW_FAILED_HANDLER=my_handler`
```

sbc100 (approved):

wasm2c/README.md
shravanrn: Oops, looks like I had the "automerge on approve" set. I'll fix this in a new PR

PR #2204, opened 2023-04-20
Fix #2203

### LINKED ISSUE #2203: wasm2c exception handling is broken
```
;;; TOOL: run-spec-wasm2c
;;; ARGS*: --enable-exceptions
(module
  (tag $e0)
  (func $throw (throw $e0))
  (func (export "break-throw") (result i32)
    (try $outer (result i32)
      (do
        (try (result i32)
          (do
            (block $a
              (try (do (br $a)) (delegate $outer))
            )
            (call $throw)
            (i32.const 0)
          )
          (catch $e0 (i32.const 1))
        )
      )
      (catch $e0 (i32.const 2))
    )
  )
)

(assert_return (invoke "break-throw") (i32.const 1))
(;; STDOUT ;;;
1/1 tests passed.
;;; STDOUT ;;)
```

SoniEx2:
(didn't test locally but should work)

Fixes #2203

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -1086,9 +1086,9 @@ void CWriter::Write(const GotoLabel& goto_label) {
               StackVar(amount - i - 1), "; ");
       }
     }
-
-    assert(try_catch_stack_.size() >= label->try_catch_stack_size);
   }
+
+  assert(try_catch_stack_.size() >= label->try_catch_stack_size);
 
   if (try_catch_stack_.size() != label->try_catch_stack_size) {
     const std::string& name =
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -1088,13 +1088,13 @@ void CWriter::Write(const GotoLabel& goto_label) {
     }
 
     assert(try_catch_stack_.size() >= label->try_catch_stack_size);
+  }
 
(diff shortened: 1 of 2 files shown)


src/c-writer.cc
```diff
@@ -1088,13 +1088,13 @@ void CWriter::Write(const GotoLabel& goto_label) {
     }
 
     assert(try_catch_stack_.size() >= label->try_catch_stack_size);
```
keithw, line 1090: Should this `assert` also move outside the `if (label->HasValue()) {` ?

keithw (commented):
lgtm. Thank you for working on this! Could you please submit the test to the exception-handling repo so it can get picked up by the testsuite repo and included in the WABT testsuite (https://github.com/WebAssembly/wabt/issues/2203#issuecomment-1516767126)?

Alternately, I think it would be fine to amend this PR to include that as a one-off test if you just want this to get merged ASAP, and we can merge the upstream testsuite fix down the road. Either way it would be nice to make sure this scenario gets tested so different implementations all agree on the behavior of mixing `br` with `try`/`catch`.

keithw (approved):
lgtm. Will be great when the upstream repo includes some tests of br interacting with try/catch. Thank you for testing and fixing this.

PR #2207, opened 2023-04-21
c-writer.cc: cleanup name lookups to happen consistently in one place

keithw:
This is a prelude to fixing a wasm2c exception-handling issue reported by @SoniEx2 on the Wasm Discord (and related to #2203).

PR #2208, opened 2023-04-21
c-writer.cc: Correctly handle label names when branching out of try block

### LINKED ISSUE #2205: wasm2c exception handling is more broken than we thought
```wast
  (func (export "break-try")
    (try (do (br 0)) (delegate 0))
  )
```

keithw:
~Sequenced behind #2207.~ Fixes #2205. wasm2c was making a LabelDecl with the raw Wasm name instead of the safe/unique C name. Because the label's C name is generated in `BeginTry` but the LabelDecl doesn't come until later, it's probably easiest just to re-retrieve the C name at that point. Or we could have `BeginTry` actually return the generated C name (along with the mark)...

The exception-handling proposal could benefit from better test coverage here (https://github.com/WebAssembly/exception-handling/issues/210) but this is definitely our bug!

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -2837,7 +2837,9 @@ void CWriter::WriteTryCatch(const TryExpr& tryexpr) {
   Write(CloseBrace(), Newline()); /* end of try-catch */
 
   ResetTypeStack(mark);
-  Write(LabelDecl(label_stack_.back().name));
+  assert(!label_stack_.empty());
+  assert(label_stack_.back().name == tryexpr.block.label);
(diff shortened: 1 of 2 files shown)


shravanrn:
Hmm... I'm afraid I don't understand this well enough to help review this, so I'll leave it up to @sbc100 
Ping me, if you want me to dig in and figure it out

keithw:
@sbc100 are you able to give this a review?

sbc100 (approved):
Rubberstamp lgtm.  I'm afraid I don't fulling understand this change either, but if this test passed after and fails before this sgtm.

keithw:
Thanks!

PR #2212, opened 2023-04-22
Replace alloca.h with stdlib.h on FreeBSD

### LINKED ISSUE #2210: [wasm2c] alloca.h
When I try to build on FreeBSD wasm2c tests fail due to lack of `alloca.h` headers (using wasm2c should be problematic because of that). I'm not sure if this happens on other platforms. There are a few potential ways to fix it, for example include `stdlib.h` instead or make the include conditional, either via a preprocessor check in place or by generating the template via CMake.

penzn:
Closes #2210


keithw (approved):
lgtm. (had to make to update the checked-in version of the prebuilt template)

PR #2221, opened 2023-04-27
wasm2c: provide option to disable exception handling in the runtime

shravanrn:
In some of the audits prior to landing wasm2c in upstream Firefox, we saw that exception and tag handling for Wasm is implemented with globals, longjmp etc. While we don't use/plan to use `--enable-exceptions` at the moment, and I can't see anything immediately wrong with the runtime exception code, I think we need the ability to disable this code altogether in the runtime to safely say we aren't exposed to any bugs in this and to ensure other bugs in the application don't use this code as gadgets. This commit adds a macro to disable exception support in the wasm2c runtime. Naturally the support is still enabled by default so existing/new users are not confused, but for the Firefox case, I think we would need something like this macro to land this upstream.

### DIFF
--- a/wasm2c/wasm-rt-impl.c
+++ b/wasm2c/wasm-rt-impl.c
@@ -60,11 +60,13 @@ WASM_RT_THREAD_LOCAL uint32_t wasm_rt_saved_call_stack_depth;
 
 WASM_RT_THREAD_LOCAL wasm_rt_jmp_buf g_wasm_rt_jmp_buf;
 
+#if !WASM_RT_SKIP_EXCEPTION_HANDLING
 static WASM_RT_THREAD_LOCAL wasm_rt_tag_t g_active_exception_tag;
 static WASM_RT_THREAD_LOCAL uint8_t g_active_exception[MAX_EXCEPTION_SIZE];
 static WASM_RT_THREAD_LOCAL uint32_t g_active_exception_size;
 
 static WASM_RT_THREAD_LOCAL wasm_rt_jmp_buf* g_unwind_target;
+#endif
 
 #ifdef WASM_RT_TRAP_HANDLER
 extern void WASM_RT_TRAP_HANDLER(wasm_rt_trap_t code);
@@ -87,6 +89,8 @@ void wasm_rt_trap(wasm_rt_trap_t code) {
   WASM_RT_LONGJMP(g_wasm_rt_jmp_buf, code);
 #endif
 }
+
+#if !WASM_RT_SKIP_EXCEPTION_HANDLING
 
 void wasm_rt_load_exception(const wasm_rt_tag_t tag,
                             uint32_t size,
@@ -126,6 +130,8 @@ uint32_t wasm_rt_exception_size(void) {
 void* wasm_rt_exception(void) {
   return g_active_exception;
 }
+
+#endif
 
 #ifdef _WIN32
 static void* os_mmap(size_t size) {
--- a/wasm2c/wasm-rt.h
+++ b/wasm2c/wasm-rt.h
@@ -196,6 +196,10 @@ extern WASM_RT_THREAD_LOCAL uint32_t wasm_rt_call_stack_depth;
 
 #endif
 
+#ifndef WASM_RT_SKIP_EXCEPTION_HANDLING
+#define WASM_RT_SKIP_EXCEPTION_HANDLING 0
+#endif
+
 #if defined(_MSC_VER)
 #define WASM_RT_NO_RETURN __declspec(noreturn)
 #else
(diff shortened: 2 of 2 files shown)


keithw:
Hmm, unfortunately I think you know my view here. :-( I can't feel strongly about any individual runtime-only hook/macro, but how do you and Mozilla see this evolving over the long term?

It would be awesome to have Firefox using the latest wasm2c (I'm using Firefox right now!), but, I feel a bit like we're frogs in a pot with each new hook. It totally makes sense that Firefox, as a Wasm embedder, has particular needs and is nervous about call sites to longjmp. But at some point, the runtime API is designed so that embedders can choose the runtime behavior they want. It's hard to reason about the interaction of these hooks in the upstream runtime if our only tool for configuration is a bunch of binary preprocessor macros to accommodate the universe of possible embedder desires, especially if we extend to desires of the form "I'm nervous about this part of the code existing," especially if we can't know when it's safe to remove a hook from the API, and especially if we're not going to test most of the configurations.

Is there a discussion over on the Mozilla side that would help understand more of the context about how they're thinking about this? I really don't want to be obstructionist, but, like, the Firefox project is 80,000 times bigger than WABT. How many hooks do they think it's reasonable to ask upstream to maintain for Firefox before they start bearing more of the cost downstream?

Some good paths forward seem like:
- We figure out a better way for embedders to configure the upstream runtime that isn't a hard-to-test ensemble of binary preprocessor macros.
- Mozilla maintains a local wasm-rt-impl.c that meets Firefox's particular needs, while still using wasm2c off the shelf
- Mozilla keeps putting hooks into the upstream wasm-rt-impl.c so Firefox can use it without maintaining its own runtime implementation, but also Mozilla devotes some SWE resources on WABT maintenance and keeping up with the Wasm proposals in a way that benefits the overall project and userbase (this would be a wonderful outcome!)

wdyt?

kripken:
I would also like to understand the concern here more. But reading the PR, I had an idea: the change ifdefs out functions to avoid a risk of them being called. Could the functions instead be defined as weakly linked? Then a user could provide versions that override them (and immediately abort). I think that adds less complexity and seems like it would be as safe, and could be a general customization mechanism.

sbc100:
I assume that the thing being disabled here is the wasm exception handling proposal, and not the general trap handling that would be needed by wasm MVP?

If so, perhaps we should move this code into the auto-generated part of the module, since it really should depend on the whether `--enable-exceptions` is use used when wasm2c is run.   We already have WASM_RT_ENABLE_SIMD which is used in the template code.. perhaps we could added WASM_RT_ENABLE_EXCEPTION_HANDLING in the same manor and have this code be conditional inside the template instead?

I noticed that `WASM_RT_ENABLE_SIMD` is defined on the command line today but it should really be baked into the generated output I think.

shravanrn:
> I assume that the thing being disabled here is the wasm exception handling proposal, and not the general trap handling that would be needed by wasm MVP?

Yup!

> If so, perhaps we should move this code into the auto-generated part of the module, since it really should depend on the whether --enable-exceptions is use used when wasm2c is run. 

This makes sense. I was in part considering this route as well and I'm happy to move to this if this is reasonable for all.

> Could the functions instead be defined as weakly linked? 

I think this is, In theory, a reasonable option. In practice these compiler/linker sensitive tricks tends to cause issues across the range of OS and compiler toolchains that have to be supported (windows os, mingw toolchains differences have proved challenging in the past). I would really like to avoid this route if possible.

> Hmm, unfortunately I think you know my view here. :-( I can't feel strongly about any individual runtime-only hook/macro, but how do you and Mozilla see this evolving over the long term?

Yup. I've laid out my position as well, so I'll skip that part. But regarding the long term evolution, I have completely the opposite take here. Issues such as this are precisely some of the benefits that come from having wasm2c used in a large production application like Firefox beyond the obvious feature additions contributed such as Simd or support for different OSes, ongoing work on perf etc. 

To be clear, my point here is that issued such as these are not "individual macros/hooks" but "necessary knobs" to make wasm2c deployable in production. This is unlikely to be unique for the Firefox use case either... Other production workloads that come up in the future are likely going to need similar things.

As for the question of adding knobs in upstream projects, or for that matter adding knobs surgically  in upstream*wasm runtimes to disable code/features*, this is not somehow unique to wasm2c. When we were using lucet(fastly's production computer and runtime) in Firefox, we chose to disable the signal handler code there due to the complexity and code surface, even though fastly was using this in their production code. This turned out to be a good decision for Firefox, as a few months in, there was a cve in the signal handler code which would have otherwise required an emergency patch.

> if our only tool for configuration is a bunch of preprocessor macros ...  if we're not going to test most of the configurations.

These sound like limitations in the current wasm2c development models than anything to do with upstream uses. No one is saying this has to be through preprocessor macros. That's just how the code is today and we can move away from that if you want. 

>  ask upstream to maintain for Firefox before they start bearing more of the cost themselves?

My goal here is not be a burden upstream. The fact that the wasm2c code if deployed in Firefox, is covered by Mozilla's bug bounty program, the fact that I'm on the hook for any cves that emerge in what we deploy, as well as their subsequent patching etc. which sissy means we have to pay close attention to the code, audits and limit potentially buggy code etc; and the fact that all of this is fed back into wasm2c is awesome and my personal view is that this is a good enough reason to care. Wasm2c use in Firefox is also the SOLE reason why I contribute to wasm2c, willy added Simd support and a bunch of folks contribute to / intend to contribute to / use and test wasm2c. My personal view again is that if this is somehow not viewed as a benefit in the upstream wasm2c, I am happy to maintain a fork of wasm2c or another tool that can be used in Firefox, so we can avoid any issues of burden upstream.

shravanrn:
@sbc100 I have implemented your suggestion in a new PR https://github.com/WebAssembly/wabt/pull/2222 since it looks completely different than what is here.  Could you have a look when you get the chance?

We can continue to use this thread for any other continued administrative discussions.

keithw:
Hi @shravanrn -- I like the way this is evolving with your later PR! We seem to be taking the direction of "figure out a better way for embedders to configure the upstream runtime."

Can you link to the audits you mentioned? It would be helpful to see the conversations on the Mozilla side and understand better how they're think about things.

shravanrn:
>  I like the way this is evolving with your later PR

Sounds good. This would work for the Firefox use case too.

> Can you link to the audits you mentioned?

I should clarify that this is an informal first-party audit conducted by us rather than external formal third-party audits. So there isn't a formal document I can link you to. However, here is a rough set of notes I have on the subject. Look for the "TODOS" as an indication of hardening directions for wasm2c

https://docs.google.com/document/d/10djK7HkHy5O-b_nV1Ef36Fby7V-myHof0nnhig46utI/edit?usp=drivesdk

PR #2223, opened 2023-04-30
wasm2c: reduce the max exceptions size

shravanrn:
Per suggestion of @sbc100 in https://github.com/WebAssembly/wabt/pull/2222#discussion_r1181155780, reducing the size of maximum exceptions

### DIFF
--- a/wasm2c/wasm-rt-impl.c
+++ b/wasm2c/wasm-rt-impl.c
@@ -42,7 +42,7 @@
 #endif
 
 #define PAGE_SIZE 65536
-#define MAX_EXCEPTION_SIZE PAGE_SIZE
+#define MAX_EXCEPTION_SIZE 256
 
 #if WASM_RT_INSTALL_SIGNAL_HANDLER
 static bool g_signal_handler_installed = false;


wasm2c/wasm-rt-impl.c
```diff
@@ -42,7 +42,7 @@
 #endif
 
 #define PAGE_SIZE 65536
-#define MAX_EXCEPTION_SIZE PAGE_SIZE
+#define MAX_EXCEPTION_SIZE 4096
```
sbc100, line 45: This is the used to store the value types that make up the exception.   In all the cases I know of so far this is just a single `i32`.   Larger exception types can be imagined but 4096 seems like rather a excessive.   How about just 256 for now (i.e. 64 i32s or 32 i64s)?   We have a trap in place if this ever get exceeded.

sbc100 (commented):
lgtm % comment

wasm2c/wasm-rt-impl.c
keithw: 256 is fine with me.

shravanrn: fixed

shravanrn:
Could one of you approve the PR when you get a chance?

keithw (approved):

Issue #2224, opened 2023-05-03
wasm2c SIMD final approach checklist


keithw:
simd-everywhere has just merged fixes to pass the Wasm testsuite (https://github.com/simd-everywhere/simde/pull/1010), so I think we're close to being able to close this out, cut a release, and check the box on webassembly.org. This will also be the first time in a while that wasm2c is passing 100% of the current Wasm core testsuite (which we should enjoy until tail-calls get merged).

Here's my thinking/proposal on next steps to get there:

- [x] Update our simde submodule pointer
- [x] Add the five missing tests
- [x] Check the SIMD box in the wasm2c README
- [x] Augment the IR to track whether SIMD is actually used in the module (later we can do the same for exceptions)
- [x] Include SIMD support in the generated code only if the module uses it + SIMD is enabled
- ~Maybe controversial: When SIMD is being used, include the simde header in the generated code (like we do with the scalar operations) so the user doesn't need to have a compatible dev repo of SIMDe in their include path to compile the wasm2c output.~
   - ~~Create the "amalgamated" simd128.h header along with the other source templates~~
   - ~~Check in the "amalgamated" header as a prebuilt file to keep Firefox and others happy (similar to #2193)?~~
- [x] Eliminate WASM_RT_ENABLE_SIMD preprocessor macro and conditionally `#include <simde/wasm/simd128>` instead.
   - [x] CWriter tracks whether SIMD is used in the module's external interface to decide whether the `#include` should be in the header file or only the C file (avoid pulling all of SIMDe simd128.h into the embedder's namespace unless necessary)
- [x] Cut a 1.0.33 release
- [x] Submit a PR to WebAssembly/website that wasm2c 1.0.33 supports SIMD (and extended-const)

@wrv @sbc100 @shravanrn wdyt?

mr-c:
> Maybe controversial: When SIMD is being used, include the simde header in the generated code (like we do with the scalar operations) so the user doesn't need to have a compatible dev repo of SIMDe in their include path to compile the wasm2c output. 

This is fine by me, and from the Debian side it would be nice to have a switch to easily use an installed copy of SIMDe instead of the bundled amalgated version.

keithw:
> This is fine by me, and from the Debian side it would be nice to have a switch to easily use an installed copy of SIMDe instead of the bundled amalgated version.

Sounds reasonable... maybe `--external-simde` or something?

mr-c:
[...]

PR #2226, opened 2023-05-04
Track use of SIMD and exceptions in IR, and CWriter includes support only if used

keithw:
This is tasks 4 and 5 from #2224.

### DIFF
--- a/include/wabt/ir.h
+++ b/include/wabt/ir.h
@@ -1245,6 +1245,12 @@ struct Module {
   BindingHash memory_bindings;
   BindingHash data_segment_bindings;
   BindingHash elem_segment_bindings;
+
+  // Feature tracking for a subset of supported features
+  struct {
+    bool simd = false;
+    bool exceptions = false;
+  } features;
 };
 
 enum class ScriptModuleType {
--- a/include/wabt/ir.h
+++ b/include/wabt/ir.h
@@ -1246,11 +1246,13 @@ struct Module {
   BindingHash data_segment_bindings;
   BindingHash elem_segment_bindings;
 
-  // Feature tracking for a subset of supported features
+  // For a subset of features, the BinaryReaderIR tracks whether they are
+  // actually used by the module. wasm2c (CWriter) uses this information to
+  // limit its output in some cases.
   struct {
     bool simd = false;
     bool exceptions = false;
-  } features;
+  } features_used;
 };
 
 enum class ScriptModuleType {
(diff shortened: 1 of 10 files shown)


keithw:
Another way to do this (which would be more systematic but a bigger change) would be to go through the BinaryReader and replace queries to `options.features_.[something]_enabled()` with a call to a new function that checks if the feature is enabled and, if so, does a callback to the BinaryReaderDelegate (e.g. the BinaryReaderIR) tellling it that that feature is in use. Happy to do it this way if anybody thinks better, but I think the way here is probably less intrusive.

src/c-writer.cc
```diff
@@ -1791,10 +1791,10 @@ void CWriter::WriteCallIndirectFuncDeclaration(const FuncDeclaration& decl,
 }
 
 void CWriter::WriteFeatureMacros() {
-  if (options_.features->exceptions_enabled()) {
+  if (module_->features.exceptions) {
```
shravanrn, line 1794: I thought the plan was to check both the command line switch and if the feature was in use?

src/binary-reader-ir.cc
```diff
@@ -526,6 +526,15 @@ Result BinaryReaderIR::OnFuncType(Index index,
   auto func_type = std::make_unique<FuncType>();
   func_type->sig.param_types.assign(param_types, param_types + param_count);
   func_type->sig.result_types.assign(result_types, result_types + result_count);
+
+  module_->features.simd |=
```
shravanrn, line 530: I'm don't understand the code/standard well enough to know if the places this condition is tracked is sufficient. The upside I think is that even if we miss one, that is not security bug and easy enough to fix. So this works. Does FuncTypes also cover externrefs etc.?

shravanrn (changes requested):

src/c-writer.cc
keithw: The command-line switches are enforced at read-time, so this can't get set unless it's enabled...

shravanrn: ah gotcha

shravanrn (approved):

include/wabt/ir.h
```diff
@@ -1245,6 +1245,12 @@ struct Module {
   BindingHash memory_bindings;
   BindingHash data_segment_bindings;
   BindingHash elem_segment_bindings;
+
+  // Feature tracking for a subset of supported features
```
sbc100, line 1249: Perhaps be a little more explicit about why this separate from supported features.   How about "For a subset of features we track whether or not they are actually used in the module.  This is used by wasm2c to limit the output in certain cases."

And maybe call it `features_used` or something like that?

sbc100 (approved):

include/wabt/ir.h
keithw: Sure, done.

PR #2227, opened 2023-05-05
wasm2c: use standard C type names in headers

keithw:
Fulfills an old TODO. Also a step towards cleaning up/minimizing the wasm2c-generated header to let us include SIMDe in a clean way in one place. (I think it will be possible to eliminate the wasm2c.top.h and wasm2c.bottom.h templates completely, and merge wasm2c.includes.c and wasm2c.declarations.c.)

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -315,6 +315,7 @@ class CWriter {
   static const char* GetReferenceTypeName(const Type& type);
   static const char* GetReferenceNullValue(const Type& type);
   static const char* GetCTypeName(const Type& type);
+  static const char* GetImplTypeName(const Type& type);
 
   const char* InternalSymbolScope() const;
 
@@ -1142,6 +1143,23 @@ void CWriter::Write(const StackVar& sv) {
 const char* CWriter::GetCTypeName(const Type& type) {
   // clang-format off
   switch (type) {
+  case Type::I32: return "uint32_t";
+  case Type::I64: return "uint64_t";
+  case Type::F32: return "float";
+  case Type::F64: return "double";
+  case Type::V128: return "simde_v128_t";
+  case Type::FuncRef: return "wasm_rt_funcref_t";
+  case Type::ExternRef: return "wasm_rt_externref_t";
+  default:
+    WABT_UNREACHABLE;
+  }
+  // clang-format on
+}
+
+// static
+const char* CWriter::GetImplTypeName(const Type& type) {
+  // clang-format off
+  switch (type) {
   case Type::I32: return "u32";
   case Type::I64: return "u64";
   case Type::F32: return "f32";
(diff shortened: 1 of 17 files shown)


src/c-writer.cc
```diff
@@ -1142,11 +1142,11 @@ void CWriter::Write(const StackVar& sv) {
...
-  case Type::I32: return "u32";
-  case Type::I64: return "u64";
-  case Type::F32: return "f32";
-  case Type::F64: return "f64";
-  case Type::V128: return "v128";
+  case Type::I32: return "uint32_t";
+  case Type::I64: return "uint64_t";
+  case Type::F32: return "float";
+  case Type::F64: return "double";
+  case Type::V128: return "simde_v128_t";
```
shravanrn, line 1149: Baking in a simde type seems mildly heavy-weight. My first preference would be to have a define just for this,  but probably not a big deal either way.

test/wasm2c/add.txt
```diff
@@ -98,6 +76,17 @@ u32 w2c_test_add(w2c_test*, u32, u32);
 
 #define UNREACHABLE TRAP(UNREACHABLE)
 
+typedef uint8_t u8;
```
shravanrn, line 79: Having these definitions in the source is a bit confusing now. On the one hand, we have functions using the C types for params and returns. While these short types seem to be defined exclusively for params in DEFINE_LOAD, DEFINE_STORE etc.  Does it make sense to eliminate the short types altogether by also changing the params of DEFINE_LOAD and DEFINE_STORE.

shravanrn (changes requested):

test/wasm2c/add.txt
keithw: We certainly can if that's what people want! It would be a more intrusive change, and it seemed mostly harmless (to me) to keep these typedefs in the generated .c file (vs. leaking them into the embedder's namespace). But I don't have a strong view one way or another if people prefer to just bite the bullet now.

src/c-writer.cc
shravanrn: Fair enough, we can leave this as simde_v128_t

test/wasm2c/add.txt
shravanrn: I think we can probably just remove it tbh. But I guess if the main goal was to avoid leaking the types into to the embedder, I'm fine either way

sbc100 (commented):
I kind of like the explicit and short wasm type names myself, but i can see how perhaps we don't want them in the public interface.

I do think there is perhaps some advantage to knowing that an argument corresponds to a particular wasm type though.    I guess we could using `wasm_u32` in the public interface and the shorter `u32` in the internal one?    I don't feel strongly though.

sbc100:
lgtm, if the two of you think this is the way to go.

keithw:
Okay, in an effort to make everybody mostly happy, how about this version, which
- uses standard C type names (e.g. "uint32_t") in the public interface, and
- uses the explicit/short Wasmy type names (e.g. "u32") consistently in the implementation files?

I was mostly looking for a way to minimize the public header template and honestly don't feel that strongly about this TODO either; if it's not something we really like I'm happy to drop it.

sbc100 (commented):
I think maybe we should just remove the TODO.. and stick with our own custom types, so its clear what they mena.  Switching to `uint32_t` seems like reduction in the clarity of the interface (at least to me it).

shravanrn:
> I think maybe we should just remove the TODO.. and stick with our own custom types, so its clear what they mena. Switching to `uint32_t` seems like reduction in the clarity of the interface (at least to me it).

Sure, I don't have any strong opinions either way here.

keithw:
Ok, let's just not do this then.

PR #2228, opened 2023-05-05
Don't include windows.h for color printing if !HAVE_WIN32_VT100

shravanrn:
We don't need windows.h if color printing is disabled i.e., HAVE_WIN32_VT100 is not defined

@keithw @sbc100 Unfortunately a blocking bug for use in Firefox as we temporarily need to build wasm2c without relying on windows.h

### DIFF
--- a/src/color.cc
+++ b/src/color.cc
@@ -21,8 +21,10 @@
 #include "wabt/common.h"
 
 #if _WIN32
+#if HAVE_WIN32_VT100
 #include <io.h>
 #include <windows.h>
+#endif
 #elif HAVE_UNISTD_H
 #include <unistd.h>
 #endif


keithw (approved):
lgtm if tests pass.

src/color.cc
```diff
@@ -21,8 +21,10 @@
 #include "wabt/common.h"
 
 #if _WIN32
+#if HAVE_WIN32_VT100
```
sbc100, line 24: I don't think you need both here.. wouldn't `HAVE_WIN32_VT100` be enough on its own?

shravanrn: Yeah true. I can make this change in a new PR

shravanrn: In https://github.com/WebAssembly/wabt/pull/2231

PR #2229, opened 2023-05-06
wasm2c: Only add simd and exception code if features are used

shravanrn:
Eliminate feature macros and directly emit simd and exception code when necessary. 

This PR is only so that we get an idea of the approach here. I'll have to cleanup up the rebase once dependencies and merge conflict PRs land. For now, have a look at only the latest commit in the commits tab (as the files changed tab has changes from the dependencies as well)

@sbc100 @keithw I was thinking we could go something like this to address point 1 from https://github.com/WebAssembly/wabt/pull/2222#issuecomment-1531954856 
Personally, I am ok either way. We could choose to do something like this to simplify the output for wasm2c when simd or exceptions are not used. Alternately, we could stick with the feature macros that we are using currently. Let me know your thoughts

Dependencies: https://github.com/WebAssembly/wabt/pull/2226

Merge collisions: https://github.com/WebAssembly/wabt/pull/2227

On the SIMD checklist, this relates to task 5 from https://github.com/WebAssembly/wabt/issues/2224


keithw:
I rebased this into #2236 -- wdyt?

PR #2231, opened 2023-05-08
Cleanup use of HAVE_WIN32_VT100 in color.cc

shravanrn:
@sbc100 Addressing the comment https://github.com/WebAssembly/wabt/pull/2228/files/9ede8fe37999d10ecda449afe43312d04133718d#r1187855033 from https://github.com/WebAssembly/wabt/pull/2228

**Edit:** Unfortunately, i think we can't make this change as HAVE_WIN32_VT100 is defined only in win32 platforms
https://github.com/WebAssembly/wabt/blob/main/CMakeLists.txt#L121

If you prefer I can change the CMAKE to define it always, or we could leave the nested _WIN32 and HAVE_WIN32_VT100 in place. Happy to go either way

### DIFF
--- a/src/color.cc
+++ b/src/color.cc
@@ -20,11 +20,9 @@
 
 #include "wabt/common.h"
 
-#if _WIN32
 #if HAVE_WIN32_VT100
 #include <io.h>
 #include <windows.h>
-#endif
 #elif HAVE_UNISTD_H
 #include <unistd.h>
 #endif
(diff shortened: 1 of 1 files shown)


shravanrn:
@sbc100 Oh actually i think we can't make this change as HAVE_WIN32_VT100 is defined only in win32 platforms
https://github.com/WebAssembly/wabt/blob/main/CMakeLists.txt#L121

If you prefer I can change the CMAKE to define it always, or we could leave the nested _WIN32 and HAVE_WIN32_VT100 in place. Happy to go either way

keithw (approved):

src/color.cc
```diff
@@ -20,11 +20,9 @@
 
 #include "wabt/common.h"
 
-#if _WIN32
```
sbc100, line 23: CI seems to be passing..  

If this doesn't work we can just drop this PR and leave things as is

shravanrn: Oh, huh. I think I just learnt that `#if SOME_UNDEFINED_MACRO` correctly evaluates to false in C. Cool, will land this then.

Issue #1766, opened 2021-11-21
wasm2c supporting threads and other features


rhobro:
Hi there,

I was trying to convert a WASM to C when wasm2c exited saying that the WASM file used threads.

Would it be possible to enable this feature in wasm2c?

Thanks

kripken:
It is possible in principle, but would take some work:

1. Implement support for lowering wasm atomics into C atomics in wasm2c.
2. Implement runtime support for the pthreads API.

rhobro:
As long as it is possible haha. @kripken

Out of interest, what are atomics? Are they the same as the concurrency-safe variables I've come across in languages such as Java and Go?

Secondly, is pthreads part of Emscripten? When looking for existing issues on the matter, I think I came across an issue mentioning pthreads.

rhobro:
https://github.com/WebAssembly/wabt/issues/1645

kripken:
> Out of interest, what are atomics? Are they the same as the concurrency-safe variables I've come across in languages such as Java and Go?

They are lower-level than most languages. Basically they are used to implement the language-level features.

https://github.com/WebAssembly/threads/blob/main/proposals/threads/Overview.md#atomic-memory-accesses

> Secondly, is pthreads part of Emscripten?

Yes, Emscripten has stable support for pthreads:

https://emscripten.org/docs/porting/pthreads.html

That uses a bunch of JS for things not supported in wasm yet, like creating web workers and handling the inability to block on the main thread. It works on the web and in node, but atm not anywhere else AFAIK.

lars-t-hansen:
> It is possible in principle, ...

Wasm has semantics for data races and mixed atomic/non-atomic accesses and partly-aliased accesses where C has only UB, so there's a fairly big caveat here that for an efficient and idiomatic translation, the wasm has to be well-behaved according to C shared-memory semantics to avoid UB.

rhobro:
As long as it is possible.

So the onus is on the WASM to prevent UB? That makes sense.

But being able to compile the WASM code to C and then compile that to machine code should have a decent performance boost.

lars-t-hansen:
> So the onus is on the WASM to prevent UB? That makes sense.

The onus is on the Wasm producer (source language compiler + binaryen + linker + ...) to produce wasm that is of a form that will allow wasm2c to produce UB-free C code.

rhobro:
Ok cool 

How long would it take to implement this feature?

rossberg:
[...]

PR #2233, opened 2023-05-10
wasm2c: partial support for atomic memory ops

shravanrn:
This PR starts the work for support for the threads proposal. In this initial PR, I am only starting with support for the  Atomic memory operations part of the spec https://github.com/WebAssembly/threads/blob/cc01bf0d17ba3fb1dc59fb7c5c725838aff18b50/proposals/threads/Overview.md?plain=1#L228

Since there is a large amount of support needed, it seemed sensible to land this in chunks.

Broadly, there are three types of operations
1. atomics (load, store, add, sub, cmp_xchg)
2. fence
3. conditional variables (wait, notify)

This PR focuses on implementing operations 1 and 2 and relies on atomic builtins provided by gcc, clang and icc https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html
and the MSVC equivalents 
https://learn.microsoft.com/en-us/cpp/intrinsics/intrinsics-available-on-all-architectures?view=msvc-170

On godbolt, it appears that this has support from 

on x86_64
- starting gcc 4.7
- starting clang 3.1
- at least icc 13.0.1

on arm/arm64
- at least gcc 5.4
- at least clang 9

Unfortunately, we can't use C11 operations as casting to/from atomics from/to a regular pointer in the Wasm heap is undefined behavior in C https://stackoverflow.com/questions/55299525/casting-pointers-to-atomic-pointers-and-atomic-sizes --- this means we can't use these ops on pointers to Wasm's linear memory

**Next steps**

- Operation 3 needs further thinking on the best way to implement this, so we can continue to support different compilers and OSes. Currently the best way forward is to implement this, seems to be to implement this on top of atomic ops, similar to [this](There's https://developers.redhat.com/articles/2022/12/06/implementing-c20-atomic-waiting-libstdc#how_can_we_implement_atomic_waiting_). Thoughts welcome.

- This code is tested with a subset of the thread atomic tests. These tests are included in-tree, with unsupported tests (relying on wait/notify) commented out. We can start testing with the full testsuite, once wait/notify is also supported.

- It would be nice to find more efficient ways to perform atomic alignment checks instead of an explicit conditional

### DIFF
--- a/include/wabt/ir.h
+++ b/include/wabt/ir.h
@@ -1252,6 +1252,7 @@ struct Module {
   struct {
     bool simd = false;
     bool exceptions = false;
+    bool threads = false;
   } features_used;
 };
 
--- a/src/binary-reader-ir.cc
+++ b/src/binary-reader-ir.cc
@@ -788,6 +788,7 @@ Result BinaryReaderIR::OnOpcode(Opcode opcode) {
     return AppendExpr(std::move(metadata));
   }
   module_->features_used.simd |= (opcode.GetResultType() == Type::V128);
+  module_->features_used.threads |= (opcode.GetPrefix() == 0xfe);
   return Result::Ok;
 }
 
--- a/src/template/wasm2c_atomicops.declarations.c
+++ b/src/template/wasm2c_atomicops.declarations.c
@@ -1,0 +1,238 @@
+#if defined(_MSC_VER)
+
+#include <intrin.h>
+
+// Use MSVC intrinsics
+
+// For loads and stores, its not clear if we can rely on register width loads
+// and stores to be atomic as reported here
+// https://learn.microsoft.com/en-us/windows/win32/sync/interlocked-variable-access?redirectedfrom=MSDN
+// or if we have to reuse other instrinsics
+// https://stackoverflow.com/questions/42660091/atomic-load-in-c-with-msvc
+// We reuse other intrinsics to be cautious
+#define atomic_load_u8(a, v) _InterlockedOr8(a, 0)
+#define atomic_load_u16(a, v) _InterlockedOr16(a, 0)
+#define atomic_load_u32(a, v) _InterlockedOr(a, 0)
+#define atomic_load_u64(a, v) _InterlockedOr64(a, 0)
+
+#define atomic_store_u8(a, v) _InterlockedExchange8(a, v)
+#define atomic_store_u16(a, v) _InterlockedExchange16(a, v)
+#define atomic_store_u32(a, v) _InterlockedExchange(a, v)
+#define atomic_store_u64(a, v) _InterlockedExchange64(a, v)
+
+#define atomic_add_u8(a, v) _InterlockedExchangeAdd8(a, v)
+#define atomic_add_u16(a, v) _InterlockedExchangeAdd16(a, v)
+#define atomic_add_u32(a, v) _InterlockedExchangeAdd(a, v)
+#define atomic_add_u64(a, v) _InterlockedExchangeAdd64(a, v)
+
+#define atomic_sub_u8(a, v) _InterlockedExchangeAdd8(a, -(v))
+#define atomic_sub_u16(a, v) _InterlockedExchangeAdd16(a, -(v))
+#define atomic_sub_u32(a, v) _InterlockedExchangeAdd(a, -(v))
+#define atomic_sub_u64(a, v) _InterlockedExchangeAdd64(a, -(v))
+
+#define atomic_and_u8(a, v) _InterlockedAnd8(a, v)
+#define atomic_and_u16(a, v) _InterlockedAnd16(a, v)
+#define atomic_and_u32(a, v) _InterlockedAnd(a, v)
+#define atomic_and_u64(a, v) _InterlockedAnd64(a, v)
+
+#define atomic_or_u8(a, v) _InterlockedOr8(a, v)
+#define atomic_or_u16(a, v) _InterlockedOr16(a, v)
+#define atomic_or_u32(a, v) _InterlockedOr(a, v)
+#define atomic_or_u64(a, v) _InterlockedOr64(a, v)
+
+#define atomic_xor_u8(a, v) _InterlockedXor8(a, v)
+#define atomic_xor_u16(a, v) _InterlockedXor16(a, v)
+#define atomic_xor_u32(a, v) _InterlockedXor(a, v)
+#define atomic_xor_u64(a, v) _InterlockedXor64(a, v)
+
+#define atomic_exchange_u8(a, v) _InterlockedExchange8(a, v)
+#define atomic_exchange_u16(a, v) _InterlockedExchange16(a, v)
+#define atomic_exchange_u32(a, v) _InterlockedExchange(a, v)
+#define atomic_exchange_u64(a, v) _InterlockedExchange64(a, v)
+
+// clang-format off
+#define atomic_compare_exchange_u8(a, expected_ptr, desired) _InterlockedCompareExchange8(a, desired, *(expected_ptr))
+#define atomic_compare_exchange_u16(a, expected_ptr, desired) _InterlockedCompareExchange16(a, desired, *(expected_ptr))
+#define atomic_compare_exchange_u32(a, expected_ptr, desired) _InterlockedCompareExchange(a, desired, *(expected_ptr))
+#define atomic_compare_exchange_u64(a, expected_ptr, desired) _InterlockedCompareExchange64(a, desired, *(expected_ptr))
+// clang-format on
+
+#define atomic_fence() _ReadWriteBarrier()
+
+#else
+
+// Use gcc/clang/icc intrinsics
+#define atomic_load_u8(a) __atomic_load_n((u8*)(a), __ATOMIC_SEQ_CST)
+#define atomic_load_u16(a) __atomic_load_n((u16*)(a), __ATOMIC_SEQ_CST)
+#define atomic_load_u32(a) __atomic_load_n((u32*)(a), __ATOMIC_SEQ_CST)
+#define atomic_load_u64(a) __atomic_load_n((u64*)(a), __ATOMIC_SEQ_CST)
+
+#define atomic_store_u8(a, v) __atomic_store_n((u8*)(a), v, __ATOMIC_SEQ_CST)
+#define atomic_store_u16(a, v) __atomic_store_n((u16*)(a), v, __ATOMIC_SEQ_CST)
+#define atomic_store_u32(a, v) __atomic_store_n((u32*)(a), v, __ATOMIC_SEQ_CST)
+#define atomic_store_u64(a, v) __atomic_store_n((u64*)(a), v, __ATOMIC_SEQ_CST)
+
+#define atomic_add_u8(a, v) __atomic_fetch_add((u8*)(a), v, __ATOMIC_SEQ_CST)
+#define atomic_add_u16(a, v) __atomic_fetch_add((u16*)(a), v, __ATOMIC_SEQ_CST)
+#define atomic_add_u32(a, v) __atomic_fetch_add((u32*)(a), v, __ATOMIC_SEQ_CST)
+#define atomic_add_u64(a, v) __atomic_fetch_add((u64*)(a), v, __ATOMIC_SEQ_CST)
+
+#define atomic_sub_u8(a, v) __atomic_fetch_sub((u8*)(a), v, __ATOMIC_SEQ_CST)
+#define atomic_sub_u16(a, v) __atomic_fetch_sub((u16*)(a), v, __ATOMIC_SEQ_CST)
+#define atomic_sub_u32(a, v) __atomic_fetch_sub((u32*)(a), v, __ATOMIC_SEQ_CST)
+#define atomic_sub_u64(a, v) __atomic_fetch_sub((u64*)(a), v, __ATOMIC_SEQ_CST)
+
+#define atomic_and_u8(a, v) __atomic_fetch_and((u8*)(a), v, __ATOMIC_SEQ_CST)
+#define atomic_and_u16(a, v) __atomic_fetch_and((u16*)(a), v, __ATOMIC_SEQ_CST)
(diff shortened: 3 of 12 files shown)


src/template/wasm2c_atomicops.declarations.c
```diff
@@ -0,0 +1,114 @@
+
+#define DEFINE_ATOMIC_LOAD(name, t1, t2, t3)                            \
+  static inline t3 name(wasm_rt_memory_t* mem, u64 addr) {              \
+    MEMCHECK(mem, addr, t1);                                            \
+    t1 result;                                                          \
+    wasm_rt_memcpy(&result, &mem->data[addr], sizeof(t1));              \
+    result = __atomic_load_n((t1*) &mem->data[addr], __ATOMIC_SEQ_CST); \
```
sbc100, line 7: I suppose the alternative would be to use C11 atomics (https://en.cppreference.com/w/c/atomic).  Those might be portable to MSVC?  But then would be increasing the minimum C version from C99 for C11.. maybe this is OK if its only for modules that need threading support?

src/template/wasm2c_atomicops.declarations.c
```diff
@@ -0,0 +1,114 @@
+
```
sbc100, line 1: Remove newline.

sbc100 (commented):
Nice work!

I assume there re some tests in the threading proposal repo that we can start running perhaps?

src/template/wasm2c_atomicops.declarations.c
shravanrn: I think I would be ok with that approach as well. My thinking is for now, we can stick with the current approach as it supports older compilers reasonably also. Support for msvc using msvc intrinsics doesn't look too bad either. If for whatever reason, msvc intrinsics become hard to incorporate, we can just use c11 atomics.

shravanrn: Also looks like msvc got support for c11 atomics fairly recently. So we probably want to go through the intrinsic route if we want to support some of the older msvc compilers.

https://devblogs.microsoft.com/cppblog/c11-atomics-in-visual-studio-2022-version-17-5-preview-2/

shravanrn:
@sbc100 Unfortunately, the tests in the proposal repo have all tests in a single wast including for wait/notify which is not implemented.

https://github.com/WebAssembly/testsuite/blob/7ef86ddeed81458f9031a49a40b3a3f99c1c6a8a/proposals/threads/atomic.wast#L427

So, to run a subset of tests, I have made a local copy of the wast with the unsupported tests removed. Let me know if this works

I have also added the missing alignment checks from the last PR

@sbc100 @keithw I am not thrilled with this approach of performing alignment checks, as I think it will incur non-trivial overhead. Open to any ideas.

src/template/wasm2c_atomicops.declarations.c
shravanrn: I did play with C atomics as well. Unfortunately, we can't use the C atomic ops as

- c atomic ops only works on atomic types `_Atomic`
- casting to/from atomics from/to a regular pointer in the Wasm heap is undefined behavior in C https://stackoverflow.com/questions/55299525/casting-pointers-to-atomic-pointers-and-atomic-sizes

shravanrn: Latest PR implements MSVC instrincs as well

shravanrn:
PR now updates to support msvc also

src/binary-reader-ir.cc
```diff
@@ -795,6 +795,7 @@ Result BinaryReaderIR::OnAtomicLoadExpr(Opcode opcode,
                                         Index memidx,
                                         Address alignment_log2,
                                         Address offset) {
+  module_->features_used.atomic_ops = true;
```
keithw, line 798: I think it _might_ be possible to detect this in one place in BinaryReaderIR::OnOpcode (by checking if opcode.GetPrefix() = 0xfe, or checking if the opcode is in the range from MemoryAtomicNotify to I64AtomicRmw32CmpxchgU) rather than in the individual expr callbacks.

keithw: Or if you wanted to skip the feature _tracking_ and just restore the Features member of the wasm2c options and make it included based on whether `--enable-threads` is given (under the assumption that this won't be the default for a long time), also fine by me.

src/template/wasm2c_atomicops.declarations.c
```diff
@@ -0,0 +1,238 @@
...
+#define atomic_compare_exchange_u16(a, expected_ptr, desired) __atomic_compare_exchange_helper((u16*)(a), expected_ptr, desired)
+#define atomic_compare_exchange_u32(a, expected_ptr, desired) __atomic_compare_exchange_helper((u32*)(a), expected_ptr, desired)
+#define atomic_compare_exchange_u64(a, expected_ptr, desired) __atomic_compare_exchange_helper((u64*)(a), expected_ptr, desired)
+// clang-format on
+
+#define atomic_fence() __atomic_thread_fence(__ATOMIC_SEQ_CST)
+
+#endif
+
+#define ATOMIC_ALIGNMENT_CHECK(addr, t1) \
```
keithw, line 123: Hmm, I don't *think* this ought to be required at runtime (do the tests fail without it?).

Alignment is an immediate, so I think theoretically this was all checked at validation-time (in our case, in SharedValidator::OnAtomicLoad which calls SharedValidator::CheckAtomicAlign).

test/wasm2c/spec/threads/atomic.txt
```diff
@@ -0,0 +1,6 @@
+;;; TOOL: run-spec-wasm2c
+;;; STDIN_FILE: test/wasm2c/spec/threads/atomic_supported.wast
```
keithw, line 2: In practice this can be done directly in the .txt file (instead of needing to reference a separate file); e.g. https://github.com/WebAssembly/wabt/blob/1.0.30/test/wasm2c/old-spec/select.txt

wasm2c/wasm-rt-impl.c
```diff
@@ -417,6 +417,8 @@ const char* wasm_rt_strerror(wasm_rt_trap_t trap) {
       return "Invalid call_indirect";
     case WASM_RT_TRAP_UNCAUGHT_EXCEPTION:
       return "Uncaught exception";
+    case WASM_RT_TRAP_UNALIGNED:
```
keithw, line 420: I don't *think* we need a new runtime trap because I believe this is checked at validation-time? But I might be confused -- do the tests ever raise this or test for this trap?

src/template/wasm2c_atomicops.declarations.c
shravanrn: The tests fail without it. Per the spec, unaligned indexes in atomic ops have to trap. The linear memory indexes are not known at compile time, so I think the check has to be at runtime

wasm2c/wasm-rt-impl.c
shravanrn: it seems like this is a runtime trap, and so I think we may require this

src/template/wasm2c_atomicops.declarations.c
keithw: Ah, too bad.

src/template/wasm2c_atomicops.declarations.c
```diff
@@ -0,0 +1,238 @@
...
+  if ((addr & (sizeof(t1) - 1)) != 0) {  \
```
keithw, line 124: This might be simpler as `if (addr % sizeof(t1))` ?

keithw: Or, maybe better, `if (UNLIKELY(addr % sizeof(t1)))` ?

shravanrn: Fixed!

include/wabt/ir.h
```diff
@@ -1252,6 +1252,7 @@ struct Module {
   struct {
     bool simd = false;
     bool exceptions = false;
+    bool atomic_ops = false;
```
keithw, line 1255: Any chance we can keep these names in sync with the names in feature.def (which here would be `threads`)?

keithw (approved):
lgtm % comments.

Bigger picture question: do you think it's going to be feasible to support the proposal in full? There was an earlier conversation at https://github.com/WebAssembly/wabt/issues/1766 that seems pretty pessimistic. :-( I haven't thought carefully about it though or looked at the C spec.

shravanrn:
> lgtm % comments.
> 
> Bigger picture question: do you think it's going to be feasible to support the proposal in full? There was an earlier conversation at #1766 that seems pretty pessimistic. :-( I haven't thought carefully about it though or looked at the C spec.

yup, I think the quick summary is that #1766 is a more elaborate version of the statement "lowering wasm atomics into C atomics will cause undefined behavior" that I noted earlier.

I believe we are safe from that since we are using compiler intrinsics, since we aren't violating C's aliasing rules of pointers between different types. Overall, wasm2c has reads/writes using memcpy (i.e., it does not assume anything about the type of the source/dest apart from it being a byte buffer, and assumes nothing about overlaps), and has atomic reads/writes using compiler intrinsics which once again assumes nothing about the address, except possibly platform specific alignment restrictions (which are typically weaker or the same as Wasm's requirements anyway). So overall, I believe we are fine here. However, we can continue to monitor UBSan, to make sure things remain good. I don't think msvc currently supports UBsan, but once that is enabled, it would be good to get that on CI as well.

src/template/wasm2c_atomicops.declarations.c
```diff
@@ -0,0 +1,238 @@
...
+  if (UNLIKELY(addr % sizeof(t1))) {     \
+    TRAP(UNALIGNED);                     \
+  }
+
+#define DEFINE_ATOMIC_LOAD(name, t1, t2, t3)               \
+  static inline t3 name(wasm_rt_memory_t* mem, u64 addr) { \
+    MEMCHECK(mem, addr, t1);                               \
+    ATOMIC_ALIGNMENT_CHECK(addr, t1);                      \
+    t1 result;                                             \
+    wasm_rt_memcpy(&result, &mem->data[addr], sizeof(t1)); \
```
turbolent, line 133: What is the reason for having this additional memcpy before the atomic load?

src/template/wasm2c_atomicops.declarations.c
```diff
@@ -0,0 +1,238 @@
...
+    result = atomic_load_##t1(&mem->data[addr]);           \
+    wasm_asm("" ::"r"(result));                            \
```
turbolent, line 135: What is the reason for having this forced read? Is it always necessary? What happens if it is not performed?

turbolent:
@shravanrn Great work! I'm currently implementing similar support in w2c2 and was wondering about some of the code

src/template/wasm2c_atomicops.declarations.c
keithw: wasm2c forces liveness of every load result (not just for atomic ops) because WebAssembly requires a deterministic trap on OOB load/store. We don't want the compiler to optimize the read out (e.g. if the value is never observed) and prevent the SIGSEGV/SIGBUS from getting triggered. There's probably more efficient ways we could be doing this, but this is what we've got to avoid a software bounds-check on every load.

See https://github.com/WebAssembly/wabt/issues/1925 and https://github.com/WebAssembly/wabt/pull/1939

src/template/wasm2c_atomicops.declarations.c
shravanrn: In general, moving data from a source and destination of different types in C (apart from the basic supported things like int to long conversions) should be done by memcpy. This is because casting the `data` byte buffer to multiple target types can trigger UB. Put simply, if you have both expression `u32* a = (u32*) &mem->data[addr]` and `u64* b = (u64*) &mem->data[addr]` in the C file output by wasm, then the C compiler compiling that is allowed to compile & optimize assuming a & b don't alias. However, Wasm makes no such guarantees.

shravanrn:
> @shravanrn Great work! I'm currently implementing similar support in w2c2 and was wondering about some of the code

@turbolent Also, have a look at the currently open PRs. I am changing this to move away from compiler intrinsics to C11 features.

squk:
@shravanrn I'm curious if you have the following expression types working in wasm2c(AtomicWait, AtomicNotify)?https://github.com/WebAssembly/wabt/blob/c24a216237e2731cc71885230bf7de5a9dfac5c7/src/c-writer.cc#L3973-L3976

They are currently unimplemented in the `main` branch and I was hoping a fork of yours has these implemented so they can be merged upstream? I'm attempting to use threads `wasm2c`(https://github.com/google/leveldb/blob/068d5ee1a3ac40dabd00d211d5013af44be55bea/util/env_posix.cc#L817-L832).

shravanrn:
@squk You can try the patch from this PR https://github.com/WebAssembly/wabt/pull/2268 
Note that the wasm2c might have changed a bit since then, so you may have to rebase this to get it to work. Fwiw, note that this PR is not likely to be exactly what we land in wasm2c production (per the comments in that PR), but I did test that this PR works correctly with threads at the time.

squk:
@shravanrn Thanks. I also found these branches 
https://github.com/WebAssembly/wabt/compare/main...PLSysSec:wabt:atomicops_part2_refactor
https://github.com/WebAssembly/wabt/compare/main...PLSysSec:wabt:atomic_sharedmem3

I'm curious, are the atomic memory operations something that's being actively worked on?

shravanrn:
> @shravanrn Thanks. I also found these branches 
> https://github.com/WebAssembly/wabt/compare/main...PLSysSec:wabt:atomicops_part2_refactor
> https://github.com/WebAssembly/wabt/compare/main...PLSysSec:wabt:atomic_sharedmem3
> 
> I'm curious, are the atomic memory operations something that's being actively worked on? 

Given my current availability, I think I'm unlikely to get to this in the next couple of months, especially because I don't know of, or currently have, use cases being blocked by wasm threads support. This may change in the future, in which case, I'll try to prioritize implementation. You're welcome to give the implementation a shot if you have cycles.

PR #2236, opened 2023-05-12
wasm2c: include simd128.h and wasm-rt-exceptions.h where necessary

keithw:
- Adapts @shravanrn's #2229.
- Directly include SIMDe's simd128.h and wasm-rt-exceptions.h where needed.
- Only includes simd128.h in the header file (and embedder's namespace) if the module requires it -- otherwise it only goes in the .c (or -impl.h) file.
- Would check boxes 6 and 7 in #2224.


src/c-writer.cc
```diff
@@ -1790,15 +1802,50 @@ void CWriter::WriteCallIndirectFuncDeclaration(const FuncDeclaration& decl,
...
-    Write("#define WASM_RT_ENABLE_SIMD", Newline(), Newline());
+
+  if (simd_used_in_header_) {
+    WriteV128Decl();
   }
 }
 
+void CWriter::WriteV128Decl() {
+  Write("#include <simde/wasm/simd128.h>", Newline(), Newline());
+  Write("#ifndef WASM_RT_SIMD_TYPE_DEFINED", Newline(),
```
shravanrn, line 1843: Do you need the guard given that you also have simd_used_in_header_?

shravanrn (approved):
lgtm % comment

src/c-writer.cc
```diff
@@ -1790,15 +1802,50 @@ void CWriter::WriteCallIndirectFuncDeclaration(const FuncDeclaration& decl,
...
+       std::any_of(module_->imports.begin(), module_->imports.end(),
+                   [](const auto& x) {
+                     return x->kind() == ExternalKind::Func &&
+                            func_uses_simd(cast<FuncImport>(x)->func.decl.sig);
+                   }) ||
+       std::any_of(module_->exports.begin(), module_->exports.end(),
+                   [&](const auto& x) {
+                     return x->kind == ExternalKind::Func &&
+                            func_uses_simd(module_->GetFunc(x->var)->decl.sig);
+                   }));
```
sbc100, line 1826: How come we have have all this nice modern style of C++ but we still have to type of out `.begin()` and `.end()` everywhere..  :(

Also, why does the third lambda have the `&` capture specifier?   (sorry I'm still not used to reading C++ lambdas).

sbc100 (commented):
lgtm.. I'll let @shravanrn approve since its based on his PR.

src/c-writer.cc
keithw: :-) I think if we upgraded to C++20 we could probably use `std::ranges::any_of`, which finally gets rid of the .begin() and .end().

The third lambda is the only one that uses the `module_` member, so it needs to capture it (and capturing implicitly with `&` seemed like the simplest way to do that).

src/c-writer.cc
keithw: Unfortunately I think we still need the guard, given that this is sometimes going in the .h file, and C forbids duplicate typedefs even if they're identical. If the embedder includes two module .h files, we don't want a collision. :-/

PR #2239, opened 2023-05-21
Upgrade to SIMDe 0.7.6 + a few fixes

mr-c:
Highlights:
- [wasm/simd128: fix altivec_p7 version of wasm_f64x2_pmin](https://github.com/simd-everywhere/simde/commit/96d6e5392eb3baa72b8934594925d7c6754a729b)
- [wasm/simd128: add missing unsigned functions](https://github.com/simd-everywhere/simde/commit/ea5e283ba99735355fc9687044cdc9f8102d0386)
- [wasm simd128: skip SIMDE_CONVERT_VECTOR_ impementations on PowerPC](https://github.com/simd-everywhere/simde/commit/4de999a0ad2c15b64efc6f51b15f21435f87ba2e)
- [wasm simd128: more powerpc fixes](https://github.com/simd-everywhere/simde/commit/7cb569198771366da65c9999c02883e4c2bb4b44)
- [fix for detecting F16C CPU support for MSVC 2022+ with AVX2](https://github.com/simd-everywhere/simde/commit/f9cf4675d101eef2dd70dabbb4d6806b405eac29)
- [gh-actions: put wasm64 under CI](https://github.com/simd-everywhere/simde/commit/6702820b6be1e3f8b2fb31c14d6d94205a516bfa) (all SIMDe tests pass in `memory64` mode!)


keithw:
lgtm.

Longer term: Given that we're not vendoring the simd128.h header into wasm2c, I think nominally the simde submodule is only there so wabt can run the tests. Maybe (?) the "ideal" situation would be where:

- wasm2c/README.md documents the SIMDe dependency and describes the minimum simde version supported (this seems valuable no matter what we do)
- the person compiling wasm2c's output is encouraged to install SIMDe from wherever is convenient to them (Debian, Homebrew, etc.)
- maybe in the longer term, wabt removes the submodule and just has the CI fetch SIMDe from Ubuntu/Homebrew/choco, just like how we currently get ninja ?

keithw (approved):

sbc100:
> lgtm.
> 
> Longer term: Given that we're not vendoring the simd128.h header into wasm2c, I think nominally the simde submodule is only there so wabt can run the tests. Maybe (?) the "ideal" situation would be where:
> 
> * wasm2c/README.md documents the SIMDe dependency and describes the minimum simde version supported (this seems valuable no matter what we do)
> * the person compiling wasm2c's output is encouraged to install SIMDe from wherever is convenient to them (Debian, Homebrew, etc.)
> * maybe in the longer term, wabt removes the submodule and just has the CI fetch SIMDe from Ubuntu/Homebrew/choco, just like how we currently get ninja ?

I'd prefer to keep the submodule, since it seems like the simplest and most direct way to express the dependency.

Could we add a `#warning` if the simde header file doesn't match the version that we are submoduling it?  e.g. `#warning "the included simde header is out of sync with the expectations of this wasm2c-generated code.. YMMV"`

PR #2251, opened 2023-06-08
binary-reader.cc: detect more malformed modules

keithw:
This PR lets the BinaryReader detect malformed modules in the binary format that:

- are missing END opcodes at the end of functions
- have too many END opcodes at the end of functions
- have DELEGATE opcodes outside a TRY
- have ELSE opcodes outside an IF

Previously we were only catching some of these situations later in validation. This is a prelude for a future PR that will make the spectest interpreter require that `assert_malformed` fail to parse the module (currently `assert_malformed` is treated the same as `assert_invalid`, i.e. the command passes as long as validation fails).

### DIFF
--- a/test/regress/regress-28.txt
+++ b/test/regress/regress-28.txt
@@ -15,6 +15,5 @@ section(CODE) {
   }
 }
 (;; STDERR ;;;
-out/test/regress/regress-28/regress-28.wasm:000001c: error: type mismatch at end of function, expected [] but got [any]
-000001c: error: EndFunctionBody callback failed
+0000019: error: function body shorter than given size
 ;;; STDERR ;;)
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -22,6 +22,7 @@
 #include <cstdint>
 #include <cstdio>
 #include <cstring>
+#include <stack>
 #include <vector>
 
 #include "wabt/config.h"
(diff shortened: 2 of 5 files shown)


test/regress/regress-28.txt
```diff
@@ -15,6 +15,5 @@ section(CODE) {
   }
 }
 (;; STDERR ;;;
-out/test/regress/regress-28/regress-28.wasm:000001c: error: type mismatch at end of function, expected [] but got [any]
-000001c: error: EndFunctionBody callback failed
+0000019: error: function body shorter than given size
```
sbc100, line 18: Why did we loose the filename here?

sbc100 (approved):

test/regress/regress-28.txt
keithw: Hmm, nice catch. :-/ It turns out that right now _every_ error that's caught inside the `BinaryReader` itself is getting saved without a filename, because the `BinaryReader` doesn't know anything about filenames and when it creates an `Error`, it sets the error's `Location` where only the `offset` is set (the `filename` field is empty).

We can change this if you want -- e.g.
- we could make `BinaryReader` take an optional filename in its constructor that's just used for error logging, or
- we could make BinaryReaderIR/BinaryReaderInterp::OnError method modify the error's Location before saving it, i.e. from their own `GetLocation()` function (which produces a richer Location with the filename filled in)

I just tried this and it ended up changing a ton of the test diagnostic output, but it's definitely doable either way.

Hope you're okay if we merge this now and can revisit the filenames in error logging in a future PR.

sbc100: yeah I'm not too bothered about that.

PR #2252, opened 2023-06-08
spectest-interp: assert_malformed must error in reader alone

keithw:
In the spectest interpreter, currently `assert_malformed` is treated the same as `assert_invalid`. This PR requires that `assert_malformed` modules must fail in the reader (not later in validation).

This also fixes an issue where `assert_invalid` wasn't validating text modules, so spectest-interp wasn't working properly for something like `(assert_invalid (module quote "...") "")`. There aren't a lot of tests like this because usually there's no need for `(module quote)` unless it's going to be in an `(assert_malformed)`, but the memory64 tests [have one](https://github.com/WebAssembly/memory64/blob/main/test/core/address.wast#L213-L219) because overlarge offsets are moving from a malformed failure to a validation failure. So this is a prerequisite for #2253.

~~This is logically dependent on #2251 (the tests won't pass until #2251 is merged).~~

### DIFF
--- a/src/tools/spectest-interp.cc
+++ b/src/tools/spectest-interp.cc
@@ -25,6 +25,7 @@
 #include <vector>
 
 #include "wabt/binary-reader-ir.h"
+#include "wabt/binary-reader-nop.h"
 #include "wabt/binary-reader.h"
 #include "wabt/cast.h"
 #include "wabt/common.h"
@@ -275,7 +276,7 @@ TypedValue GetLane(const TypedValue& tv, Type lane_type, int lane) {
   return result;
 }
 
-bool ValidIR(const std::string& filename) {
+bool CheckIR(const std::string& filename, bool validate) {
   std::vector<uint8_t> file_data;
 
   if (Failed(ReadFile(filename, &file_data))) {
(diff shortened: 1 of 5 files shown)


src/tools/spectest-interp.cc
```diff
@@ -288,17 +289,29 @@ bool ValidIR(const std::string& filename) {
...
+  if (!validate) {
+    return true;
+  }
+
   return Succeeded(
       ValidateModule(&module, &errors, ValidateOptions{s_features}));
 }
 
+bool WellformedIR(const std::string& filename) {
+  return CheckIR(filename, false);
```
sbc100, line 308: Is "well formed" something that exists in the spec, or something we a creating there?

keithw: It seems to be the spec's term (e.g. "A byte sequence is a well-formed encoding of a module if and only if it is generated by the grammar." from 5.1). If you think it's better to match the script language, we could call these functions "InvalidIR" and "MalformedIR" (instead of "ValidIR" and "WellformedIR") and just negate the callsites?

sbc100 (approved):

src/tools/spectest-interp.cc
```diff
@@ -288,17 +289,29 @@ bool ValidIR(const std::string& filename) {
   ReadBinaryOptions options(s_features, s_log_stream.get(), kReadDebugNames,
                             kStopOnFirstError, kFailOnCustomSectionError);
 
-  Errors errors;
   wabt::Module module;
+  Errors errors;
```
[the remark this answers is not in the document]
keithw: Thanks, done.

Issue #2192, opened 2023-04-08
Copy/paste bug in ReadU64LEB128


kssreeram:
In file `wabt/src/leb128.cc`, function `ReadU64Leb128`:

This snippet is incorrect. It appears to have been copied from `ReadU32Leb128`:
```
else if (p + 9 < end && (p[9] & 0x80) == 0) {
  // The top bits set represent values > 32 bits.
  if (p[9] & 0xf0) {
    return 0;
  }
  *out_value = LEB128_10(uint64_t);
  return 10;
}
```

The right code is:
```
else if (p + 9 < end && (p[9] & 0x80) == 0) {
  // The top bits set represent values > 64 bits.
  if (p[9] & 0xfe) {
    return 0;
  }
  *out_value = LEB128_10(uint64_t);
  return 10;
}
```

The number of used bits in byte 10 equals 64 - 9*7 = 1 bit. Only one bit in the final byte is a part of the 64-bit value. Therefore, all the remaining 7 bits of the last byte must be zero. The check for that is `if (p[9] & 0xfe) {`.

kssreeram:
I'll give it a shot this weekend.

PR #2253, opened 2023-06-08
memory64: when enabled, check offset range at validation-time

keithw:
Before memory64, the "offset" in a load/store expression was a u32, and we enforced this in the WastParser and BinaryReader. After memory64, the "offset" becomes a u64 syntactically, and the validator checks that it's <= UINT32_MAX for i32 memories.

We hadn't been correctly allowing these very large offsets in the text format (even when memory64 was enabled and the memory was i64).

(This change also eliminates the "memories" member in the BinaryReader. The BinaryReader no longer needs to keep track of the memories and their types to check well-formedness.)

~~This PR is logically dependent on #2252 (I don't think the tests will pass until #2252 is merged).~~

All of this is a long prelude to eventually fixing #2192 with a good regression test!

### DIFF
--- a/include/wabt/shared-validator.h
+++ b/include/wabt/shared-validator.h
@@ -104,12 +104,36 @@ class SharedValidator {
   Result OnLocalDecl(const Location&, Index count, Type type);
 
   Result OnAtomicFence(const Location&, uint32_t consistency_model);
-  Result OnAtomicLoad(const Location&, Opcode, Var memidx, Address align);
-  Result OnAtomicNotify(const Location&, Opcode, Var memidx, Address align);
-  Result OnAtomicRmwCmpxchg(const Location&, Opcode, Var memidx, Address align);
-  Result OnAtomicRmw(const Location&, Opcode, Var memidx, Address align);
-  Result OnAtomicStore(const Location&, Opcode, Var memidx, Address align);
-  Result OnAtomicWait(const Location&, Opcode, Var memidx, Address align);
+  Result OnAtomicLoad(const Location&,
+                      Opcode,
+                      Var memidx,
+                      Address align,
+                      Address offset);
+  Result OnAtomicNotify(const Location&,
+                        Opcode,
+                        Var memidx,
+                        Address align,
+                        Address offset);
+  Result OnAtomicRmwCmpxchg(const Location&,
+                            Opcode,
+                            Var memidx,
+                            Address align,
+                            Address offset);
+  Result OnAtomicRmw(const Location&,
+                     Opcode,
+                     Var memidx,
+                     Address align,
+                     Address offset);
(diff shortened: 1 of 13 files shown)


include/wabt/shared-validator.h
```diff
@@ -104,12 +104,36 @@ class SharedValidator {
...
-  Result OnAtomicNotify(const Location&, Opcode, Var memidx, Address align);
-  Result OnAtomicRmwCmpxchg(const Location&, Opcode, Var memidx, Address align);
-  Result OnAtomicRmw(const Location&, Opcode, Var memidx, Address align);
-  Result OnAtomicStore(const Location&, Opcode, Var memidx, Address align);
-  Result OnAtomicWait(const Location&, Opcode, Var memidx, Address align);
+  Result OnAtomicLoad(const Location&,
+                      Opcode,
+                      Var memidx,
+                      Address align,
+                      Address offset);
```
sbc100, line 111: I wonder if would make sense to bundle `memidx`, `align`, `offset` into some kind of `MemLocation` struct?  Are there other places in the code that want to reason about these three as a unit?

src/shared-validator.cc
```diff
@@ -518,6 +518,18 @@ Result SharedValidator::CheckAlign(const Location& loc,
...
 }
 
+Result SharedValidator::CheckOffset(const Location& loc,
+                                    Address offset,
+                                    const Limits& limits) {
+  if (limits.is_64 || (offset <= UINT32_MAX)) {
+    return Result::Ok;
+  }
+
+  PrintError(loc, "offset must be less than or equal to 0xffffffff");
```
sbc100, line 528: I has a slight preference for flipping the logic to we error out early on wasm32 and return `Ok` otherwise.

sbc100 (commented):
lgtm % comments

include/wabt/shared-validator.h
keithw: I think it would make a lot of sense, but it would be a pretty intensive refactor.

There's a lot of callsites in the BinaryReader (and its delegates, e.g. BinaryReaderIR/Interp/etc.) that pass these three in the form where `memidx` is an `Index` and `align` is in the binary (log2) form.

And there's a lot of callsites in the SharedValidator and in the IR expr constructors that pass these three in the form where memidx is a `Var` and `align` is in the expanded form.

So I think the most elegant/comprehensive treatment might have some sort of (name TBD) MemLocationRaw, used in all the binary readers, and a (name also TBD) MemLocationCooked that's used in the SharedValidator and IR?

Happy to do this now if you want but it might make sense for a separate PR.

src/shared-validator.cc
keithw: Done.

sbc100 (approved):

PR #2254, opened 2023-06-09
Add build and publishing of wabt tools with wasi

### LINKED ISSUE #2101: build and publish a wasi binary on release
If wabt was compiled for a wasi target, any wasi runtime can run it which provides flexibility both for ad-hoc CLI use (broader than the binaries on the release page) as well embedded use (without dynamic libraries)

For example, wazero could use this to keep dev dependencies light for wat based tests and also spec tests. Right now, we need an offline installation step and shell forking to use `wat2wasm`

anuraaga:
Fixes #2101 

This adds building and publishing of wabt tools as wasi. This allows the tools to run anywhere a wasm runtime is available, allowing wabt itself to realize the potential of wasm for cross-platform distribution.

Checked the release flow on fork here

https://github.com/anuraaga/wabt/releases/tag/v0.0.5-test

The binary can be tried out, for example if Go is installed,

```
wget https://github.com/anuraaga/wabt/releases/download/v0.0.5-test/wabt-v0.0.5-test-wasi.tar.gz
tar -xf wabt-v0.0.5-test-wasi.tar.gz
go run github.com/tetratelabs/wazero/cmd/wazero@latest run -mount wabt-v0.0.5-test:/work wabt-v0.0.5-test/bin/wasm2wat /work/bin/wasm2wat | less
```


.github/workflows/build.yml
```diff
@@ -80,6 +80,31 @@ jobs:
...
+      - uses: actions/checkout@v3
+        with:
+          submodules: true
+      - name: build-tools
+        run: |
+          docker run -di --name wasi-sdk -v $(pwd):/src --workdir /src ghcr.io/webassembly/wasi-sdk:wasi-sdk-20 bash
+          docker exec wasi-sdk cmake -S . -B out -G Ninja -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTS=0
+          # We only build tools, not runtime, for WASI build
+          docker exec wasi-sdk cmake --build out --config Release \
+          --target wast2json-copy-to-bin \
```
anuraaga, line 96: This can be made simpler with edits to the cmake config, for an initial iteration I tried to be as least invasive as possible. But cmake could be tweaked to either

- Collect tools build, without library build, into one target
- Provide an option to exclude wasm-rt, which I think is the only target that doesn't build with wasi due to use of setjmp

I don't know if there's much use case for publishing the library as wasi so perhaps first option is simpler. Or can just stick with this approach that doesn't need any build config tweaks

sbc100: Yeah, lets figure out to make the `all` target build like it currently does for `emscripten`.    Perhaps that means adding an option to disable wasm-rt then lets do that.  (or we could do it based on detecting that `longjmp` exists?

.github/workflows/build_release.yml
```diff
@@ -87,3 +87,59 @@ jobs:
...
+        asset_name: ${{ steps.archive.outputs.tarball }}
+        tag: ${{ github.ref }}
+
+    - name: upload shasum
+      uses: svenstaro/upload-release-action@v2
+      with:
+        repo_token: ${{ secrets.GITHUB_TOKEN }}
+        file: ./${{ steps.archive.outputs.shasum }}
+        asset_name: ${{ steps.archive.outputs.shasum }}
+        tag: ${{ github.ref }}
```
sbc100, line 145: Can we avoid duplicated these last 3 jobs?  Perhaps we can factor it out into a named job that can be called from both places (I assume github actions supports that).

sbc100 (commented):
lgtm % comments

.github/workflows/build.yml
anuraaga: @sbc100 Thanks for the header check idea, that worked out quite well I think. One other tweak was link flags to libwasm.so, please verify the changes in cmakelists

.github/workflows/build_release.yml
anuraaga: Sure, extracted a reusable action

CMakeLists.txt
```diff
@@ -469,6 +472,9 @@ if (BUILD_LIBWASM)
     target_compile_options(wasm PRIVATE $<$<COMPILE_LANGUAGE:CXX>:-Wno-old-style-cast>)
     target_compile_definitions(wasm PRIVATE "WASM_API_EXTERN=__attribute__((visibility(\"default\")))")
   endif ()
+  if (CMAKE_SYSTEM_NAME STREQUAL "WASI")
+    set_target_properties(wasm PROPERTIES LINK_FLAGS "-mexec-model=reactor")
```
anuraaga, line 476: BTW, I tried filing https://github.com/llvm/llvm-project/issues/63207

sbc100: What we really want here I think is to make libwasm be a static library in this case.

Perhaps the simplest solution here would be to set `-sBUILD_LIBWASM=0`  .. otherwise we could need to somehow auto-detect SHARED vs STATIC or add yet another option.. which doesn't seem worth it.

anuraaga: Thanks, I disabled BUILD_LIBWASM and reverted this change - actually I noticed it seems the default archives do not contain libwasm (presumably since it'd generally be hard to get a dynamic library built on CI to work outside CI) so this shouldn't have any negative effect

.github/actions/upload-release-archive/action.yml
```diff
@@ -0,0 +1,37 @@
+name: Upload release archive
```
sbc100, line 1: Does this need to in subdirectory?  Or can it just be `.github/actions/upload-release-archive.yml`

.github/actions/upload-release-archive/action.yml
```diff
@@ -0,0 +1,37 @@
...
+    description: The OS name being built, used in the archive name
+    required: true
+runs:
+  using: composite
+  steps:
+    - name: archive
+      id: archive
+      shell: bash
+      run: |
+        OSNAME=$(echo ${{ inputs.os }} | awk '{print tolower($0)}')
```
sbc100, line 14: It looks like `| sed 's/-latest//')` changed to `| awk '{print tolower($0)}` here?

.github/actions/upload-release-archive/action.yml
anuraaga: Unfortunately looks like can't

https://github.com/anuraaga/wabt/actions/runs/5238734587/jobs/9457913174

image

.github/workflows/build_release.yml
```diff
@@ -58,32 +58,23 @@ jobs:
...
       with:
-        repo_token: ${{ secrets.GITHUB_TOKEN }}
-        file: ./${{ steps.archive.outputs.tarball }}
-        asset_name: ${{ steps.archive.outputs.tarball }}
-        tag: ${{ github.ref }}
+        submodules: true
+
+    - name: built-tools
+      run: |
+        docker run -di --name wasi-sdk -v $(pwd):/src --workdir /src ghcr.io/webassembly/wasi-sdk:wasi-sdk-20 bash
```
sbc100, line 74: This first line looks like ti run an interactive bash shell.. and the deletes (`-d`) the container afterwards..  what is the purpose of that?

.github/actions/upload-release-archive/action.yml
anuraaga: Sorry thought `runner.os` produced the same, but it was `linux` instead of `ubuntu`, reverted and kicked off build to verify

https://github.com/anuraaga/wabt/actions/runs/5238769469

.github/workflows/build_release.yml
anuraaga: `-d` detaches rather than deletes

https://docs.docker.com/engine/reference/commandline/run/

This allows it to stay running for the other `exec` commands. TBH, I just copied the same pattern as the emscripten build but it seems reasonable - I did try just setting the GHA job itself to use wasi-sdk as its container but the lack of a `git` binary caused source checkout to fail

.github/workflows/build_release.yml
sbc100: Great!

.github/workflows/build_release.yml
sbc100: sgtm

sbc100 (approved):

.github/workflows/build_release.yml
```diff
@@ -58,32 +58,23 @@ jobs:
...
+        docker exec wasi-sdk cmake -S . -B out -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=install -DBUILD_TESTS=0 -DBUILD_LIBWASM=0
+        docker exec wasi-sdk cmake --build out --config Release --target install
 
-    - name: upload shasum
-      uses: svenstaro/upload-release-action@v2
+    - uses: ./.github/actions/upload-release-archive
```
sbc100, line 78: I seem to remember there was a simple way to have a build action upload an artifact during CI (different to adding release artifacts).     Might be cool to enable that so we can download and try out the binaries that get generated by this PR?

Otherwise LGTM!

anuraaga: Yeah it's pretty simple using the upload-artifact step

https://github.com/actions/upload-artifact

In that case I would add the step here, guard the release upload steps to only happen on a release, and use this action from `build.yml`.

I can try the change but I would probably only have it added to wasi then for now - the other steps in `build.yml` don't actually generate a release archive. I think in general it is a good idea to do so as it reduces the chance of CI passing + release failing, so I took the approach with wasi, but I don't want to get too invasive on non-wasi parts of the workflow in this PR if it's OK. So should we just add artifact uploading in `build.yml` for wasi?

sbc100: sgtm

.github/actions/release-archive/action.yml
```diff
@@ -0,0 +1,47 @@
...
+    description: Whether to upload the archives to the release
+runs:
+  using: composite
+  steps:
+    - name: archive
+      id: archive
+      shell: bash
+      run: |
+        OSNAME=$(echo ${{ inputs.os }} | sed 's/-latest//')
+        VERSION=${{ github.ref_name }}
```
anuraaga, line 17: Note this significant change, `ref_name` is still the tag for release build but the branch name for normal one

anuraaga (commented):
Thanks @sbc100, added uploading of wasi artifacts to CI

https://github.com/anuraaga/wabt/actions/runs/5239525600

image
https://github.com/anuraaga/wabt/releases/tag/v0.0.9-test

anuraaga:
Apologies, I had only verified push but not PR on my fork, for a PR the ref name has an unusable slash it seems. Will send a fix

`mv: cannot move 'install' to 'wabt-2254/merge': No such file or directory`

PR #2256, opened 2023-06-09
src/leb128.cc: fix detection of too-big u64 LEB128s

### LINKED ISSUE #2192: Copy/paste bug in ReadU64LEB128

keithw:
Fixes #2192

### DIFF
--- a/test/regress/bad-u64-leb128.txt
+++ b/test/regress/bad-u64-leb128.txt
@@ -1,0 +1,25 @@
+;;; TOOL: run-interp-spec
+;;; ARGS: --enable-memory64
+(assert_malformed
+  (module binary
+    "\00asm" "\01\00\00\00"
+    "\01\04\01\60\00\00"             ;; Type section
+    "\03\02\01\00"                   ;; Function section
+    "\05\03\01\04\00"                ;; Memory section (flags: i64)
+    "\0a\13\01"                      ;; Code section
+    ;; function 0
+    "\11\00"                         ;; local type count
+    "\42\00"                         ;; i64.const 0
+    "\28"                            ;; i32.load
+    "\02"                            ;; alignment 2
+    "\ff\ff\ff\ff\ff\ff\ff\ff\ff\02" ;; 2^64 (invalid)
+    "\1a"                            ;; drop
+    "\0b"                            ;; end
+  )
+  "malformed u64 LEB128"
+)
+(;; STDOUT ;;;
+out/test/regress/bad-u64-leb128.txt:4: assert_malformed passed:
+  0000020: error: unable to read u64 leb128: load offset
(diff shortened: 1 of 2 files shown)


test/regress/bad-u64-leb128.txt
```diff
@@ -0,0 +1,25 @@
+;;; TOOL: run-interp-spec
+;;; ARGS: --enable-memory64
```
sbc100, line 2: Can we write this test without depending on `--enable-memory64`.

How about just a simple `i64.const` instruction?

Also, I guess this points to gap in the spec tests?

sbc100 (commented):
lgtm % comment

test/regress/bad-u64-leb128.txt
keithw: > Can we write this test without depending on --enable-memory64

Unfortunately not, because the bug was in `ReadU64Leb128`. :-( Those are only used by memory64 (for memory min, memory max, and load/store offset -- but the first two are in units of pages, so limited to 2^48 anyway). The const literals are using signed Leb128s, e.g. `ReadS64Leb128`.

keithw: Let me try to submit this to the memory64 proposal repo...

sbc100: OK.. maybe add a comment to that effect.

sbc100 (approved):

test/regress/bad-u64-leb128.txt
keithw: Submitted PR here: https://github.com/WebAssembly/memory64/pull/38

keithw: > OK.. maybe add a comment to that effect.

Done.

PR #2257, opened 2023-06-09
wasm2c: fix bug in windows atomic load macro

shravanrn:


sbc100 (commented):
I guess this would be a compile error?  Are we not compiling this output as part of CI?

sbc100 (approved):

PR #2260, opened 2023-06-12
Use tag or sha for tarball name

anuraaga:
I realized branch name can have arbitrary characters problematic for filenames, especially `/`.

Fixes https://github.com/WebAssembly/wabt/pull/2254#issuecomment-1586600199


sbc100 (approved):
If that works then lgtm

PR #2261, opened 2023-06-18
Do not prescribe static build of libraries

nagy:
This allows it to set from the outside whether a static or a shared library type should be built.
This can be done by specifying this variable on the command line https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html . This is helpful for package maintainers who want to choose the type of library themselves.

To my understanding, this means, that the library will be built as shared, as that is the default, but if you want to stick with static being the default, we can add a line like so:
```cmake
option(BUILD_SHARED_LIBS "Build static libraries (.lib/.a)" OFF)
```


keithw:
No objection from me... I don't think this changes the default behavior (it still builds libwabt, libgtest, libgtest_main, and libwasm-rt-impl as static, probably because we don't have a BUILD_SHARED_LIBS variable).

@alexreinking do you have thoughts?

Ideally we would also find a solution to #1602 (for the other direction -- currently we're forcing libwasm to be shared).

CMakeLists.txt
```diff
@@ -425,7 +425,7 @@ endif ()
 if (HAVE_SETJMP_H)
   set(WASM_RT_FILES "wasm2c/wasm-rt-impl.h" "wasm2c/wasm-rt-impl.c" "wasm2c/wasm-rt-exceptions-impl.c")
 
-  add_library(wasm-rt-impl STATIC ${WASM_RT_FILES})
+  add_library(wasm-rt-impl ${WASM_RT_FILES})
```
sbc100, line 428: I don't think this particular library ever wants to be shared.   I could perhaps be wrong.

CMakeLists.txt
```diff
@@ -672,11 +672,11 @@ if (BUILD_TESTS)
     )
 
     # gtest
-    add_library(gtest STATIC
+    add_library(gtest
       third_party/gtest/googletest/src/gtest-all.cc
     )
 
-    add_library(gtest_main STATIC
+    add_library(gtest_main
```
sbc100, line 679: Same with these test-only libraries, lets just leave them as static for now.  (and the fuzzer one).

sbc100 (commented):
Should we also remove the explicit `SHARED` from libwasm on line 463?

nagy:
> Should we also remove the explicit `SHARED` from libwasm on line 463?

Sounds reasonable to me. I can remove it also if you want.

sbc100:
> > Should we also remove the explicit `SHARED` from libwasm on line 463?
> 
> Sounds reasonable to me. I can remove it also if you want.

Yes please remove the explicit (and I assume redundant `SHARED` from line 463)

sbc100 (approved):

nagy:
The windows build fails now. Maybe that was the reason for the `SHARED`. I don't have a windows machine to debug this.

sbc100:
> The windows build fails now. Maybe that was the reason for the `SHARED`. I don't have a windows machine to debug this.

But I thought it was being claimed that the default was `SHARED`?  In which case, removing it should be a no-op, no?

alexreinking:
The default, when `BUILD_SHARED_LIBS` is not set, is for `add_library(foo src.cpp ...)` to be a _static_ library, not a _shared_ library.

CMakeLists.txt
```diff
@@ -463,7 +463,7 @@ endif ()
 
 # libwasm, which implenents the wasm C API
 if (BUILD_LIBWASM)
-  add_library(wasm SHARED ${WABT_LIBRARY_SRC} src/interp/interp-wasm-c-api.cc)
+  add_library(wasm ${WABT_LIBRARY_SRC} src/interp/interp-wasm-c-api.cc)
   target_link_libraries(wasm PUBLIC wabt)
   target_include_directories(wasm PUBLIC third_party/wasm-c-api/include)
   if (COMPILER_IS_MSVC)
```
alexreinking, line 469: I'm guessing the problem with removing `SHARED` is with these definitions... you don't want `__declspec(dllexport)` on static library code; those macros should just be empty.

alexreinking:
I don't think this PR actually engages with static/shared library distinctions on Windows. I agree in principle that wabt should be buildable both ways, but it's not as simple as just dropping the `STATIC` and `SHARED` keywords.

CMakeLists.txt
sbc100: `WASM_API_EXTERN` should only be set when the library is being built at shared, yes.  So we should had a check here if we want to make it flexible.

Still.. I don't understand why this library isn't defaulting to `SHARED` if that is the cmake default.

nagy: I may have misunderstood the cmake default in my initial statement. It may as well be, that the default is to build statically, in which case this would would indeed cause a change.

alexreinking:
> I don't have a windows machine to test this out but these changes work fine on linux for me.

Sure, but there's no reason to believe that "works on Linux" implies "works on Windows". It would be nice if that were the case, of course...

The major issue with Windows is symbol visibility: in shared libraries, everything is hidden by default and C++ classes that expose standard classes can't be exported because most of the standard library (e.g. `std::string`, `std::vector<>`) is not itself exportable. With static libraries, there's no visibility issue.

keithw:
Closing as dormant. I think we'd still happily take a change like this (making shared the default is fine by me) if we can get the CI tests to pass across the board.

PR #2262, opened 2023-06-20
C++20 compatibility fixes

shravanrn:
Minor changes to ensure wabt compiles under the C++ 20 standard

### DIFF
--- a/include/wabt/common.h
+++ b/include/wabt/common.h
@@ -166,7 +166,7 @@ Dst WABT_VECTORCALL Bitcast(Src&& value) {
 
 template <typename T>
 void ZeroMemory(T& v) {
-  WABT_STATIC_ASSERT(std::is_pod<T>::value);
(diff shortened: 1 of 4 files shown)


sbc100 (approved):

keithw:
lgtm -- should we actually upgrade to using -std=c++20?

shravanrn:
> lgtm -- should we actually upgrade to using -std=c++20?

I'm happy to move it over to c++20 if we think it's a good idea.


(Context for others) Right now the only feature that will rely on c++ 20 is wasm2c runtime support for thread atomic wait/notify. My current plan was to make this an optional feature and switch to c++20 only when this flag is enabled. 

However moving over to c++20 outright will certainly make testing easier.

sbc100:
I think changing min requirement to C++20 would be need more/wider discussion and investigation.  Specifically we would need to know how to would effect the portability of wabt.  For now, just fixing the warnings seems like a good step forward.

PR #2263, opened 2023-06-21
wasm2c: add cpp guard in wasm2c exceptions header

shravanrn:


wasm2c/wasm-rt-exceptions.h
```diff
@@ -17,6 +17,10 @@
 #ifndef WASM_RT_EXCEPTIONS_H_
 #define WASM_RT_EXCEPTIONS_H_
 
+#ifdef __cplusplus
```
keithw, line 20: Maybe this should go after the `#include "wasm-rt.h"` ?

keithw (approved):

wasm2c/wasm-rt-exceptions.h
shravanrn: fixed

PR #2265, opened 2023-06-27
wasm-objdump: Fix local numbering in disassembly

### LINKED ISSUE #2264: wasm-objdump has incorrect local numbering
The module below has two function types declared, `0: () -> ()` and `1: (i32) -> ()`. It also has a single function defined with the second function type i.e. `(i32) -> ()`. That function has a single local defined. Since it has a parameter, my understanding is that the local should have the index 1 in the "Code Disassembly" section of wasm-objdump, but instead has the index 0:

```
module.wasm:	file format wasm 0x1

Section Details:

Type[2]:
 - type[0] () -> nil
 - type[1] (i32) -> nil
Function[1]:
 - func[0] sig=1 <f1>
Code[1]:
 - func[0] size=4 <f1>
Custom:
 - name: "name"
 - func[0] <f1>
 - func[0] local[0] <paramA>
 - func[0] local[1] <localA>
Custom:
 - name: "linking"
  - symbol table [count=1]
   - 0: F <f1> func=0 [ binding=global vis=default ]

Code Disassembly:

00002a func[0] <f1>:
 00002b: 01 7e                      | local[0] type=i64
 00002d: 0b                         | end
```

Removing the first type (`() -> ()`) results in the local having the index 1, as I'd expect.

Looking through the code, I believe the issue is that in `binary-reader-objdump.cc`:

* In `OnFuncType`, we have:
   ```cpp
   objdump_state_->function_param_counts[index] = param_count;
   ```
   where `index` is the index of the type i.e. we treat the keys of `function_param_counts` as type indices.

* In `BeginFunctionBody`, we have:
   ```cpp
   local_index_ = objdump_state_->function_param_counts[index];
   ```
   where `index` is the index of the function i.e. we treat the keys of `function_param_counts` as function indices.

This seems to be consistent with the output I'm seeing.

It's entirely possible I've misunderstood something, so please feel free to correct me! If my diagnosis is correct, I'm happy to try putting together a fix, but I'm equally happy if it's simpler for someone else to write one.

[module.zip](https://github.com/WebAssembly/wabt/files/11886114/module.zip)

mwilliamson:
Previously, in BinaryReaderObjdumpDisassemble::BeginFunctionBody, we had:

    local_index_ = objdump_state_->function_param_counts[index];

where index is the index of the function i.e. we treat the keys of function_param_counts as function indices.

However, function_param_counts is populated in OnFuncType with:

    objdump_state_->function_param_counts[index] = param_count;

where index is the index of the type i.e. we treat the keys of function_param_counts as type indices.

This discrepancy would cause the locals to be incorrectly numbered in the "Code Disassembly" section.

This fixes the discrepancy by adding a new field, function_types, which maps from function indices to type indices, and is populated in BinaryReaderObjdump::OnFunction. This field is used in BinaryReaderObjdumpDisassemble::BeginFunctionBody to get the type index for the given function, which is then used to get the parameter count.

Fixes #2264.

### DIFF
--- a/test/dump/local-indices.txt
+++ b/test/dump/local-indices.txt
@@ -1,0 +1,19 @@
+;;; TOOL: run-objdump
+(module
+  (func (param i32)
+    (local i32))
+  (func (param i32)
+    (local i32)))
+(;; STDOUT ;;;
+
+local-indices.wasm:	file format wasm 0x1
+
+Code Disassembly:
+
+000018 func[0]:
+ 000019: 01 7f                      | local[1] type=i32
+ 00001b: 0b                         | end
+00001d func[1]:
+ 00001e: 01 7f                      | local[1] type=i32
+ 000020: 0b                         | end
+;;; STDOUT ;;)
--- a/include/wabt/binary-reader-objdump.h
+++ b/include/wabt/binary-reader-objdump.h
@@ -86,6 +86,7 @@ struct ObjdumpState {
   ObjdumpLocalNames local_names;
   std::vector<ObjdumpSymbol> symtab;
   std::map<Index, Index> function_param_counts;
+  std::map<Index, Index> function_types;
 };
 
 Result ReadBinaryObjdump(const uint8_t* data,
--- a/src/binary-reader-objdump.cc
+++ b/src/binary-reader-objdump.cc
@@ -953,7 +953,8 @@ Result BinaryReaderObjdumpDisassemble::BeginFunctionBody(Index index,
   last_opcode_end = 0;
   in_function_body = true;
   current_function_index = index;
-  local_index_ = objdump_state_->function_param_counts[index];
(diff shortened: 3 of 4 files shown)


test/dump/local-indices.txt
```diff
@@ -0,0 +1,60 @@
+;;; TOOL: run-objdump
+;;; ARGS0: -v
```
sbc100, line 2: Do we need `-v` here?  Maybe keep the test smaller by removing it?

sbc100 (commented):
Thanks for fixing this!

lgtm % one comment/net.

sbc100 (approved):

test/dump/local-indices.txt
mwilliamson: Smaller sounds good to me, it's gone!

PR #2267, opened 2023-07-01
wasm2c: Use inline asm to force reads only on explicitly listed archs

### LINKED ISSUE #2266: error: couldn't allocate input reg for constraint 'r' on MIPS with clang when compiling wasm2c-generated source
When compiling Firefox (which uses wasm2c) for MIPS with clang, the following error happens:
```
rlbox.wasm.c:142:1: error: couldn't allocate input reg for constraint 'r'
 DEFINE_LOAD(f32_load, f32, f32, f32)
 ^
 rlbox.wasm.c:128:14: note: expanded from macro 'DEFINE_LOAD'
     wasm_asm("" ::"r"(result));                            \
```

GCC is happy with it, but fundamentally, the asm constraint is wrong (r is for GPRs, and here the result is a floating point value).

Those constraints were added in a7d484ef3d1f64649dd9f48cb1d5434f0fda561b, and the reason given was:
> Preventing load instructions from being optimized away if their value is unused (using inline assembly with an input operand and empty code). This is necessary to f
orce an OOB trap on platforms that use mprotect and the signal handler to detect OOB.

Would using `__attribute__((used))` work instead? Or volatile?

keithw:
Fixes #2266

### DIFF
--- a/wasm2c/examples/fac/fac.c
+++ b/wasm2c/examples/fac/fac.c
@@ -66,11 +66,12 @@ static inline bool func_types_eq(const wasm_rt_func_type_t a,
 #define MEMCHECK(mem, a, t) RANGE_CHECK(mem, a, sizeof(t))
 #endif
 
-#ifdef __GNUC__
-#define wasm_asm __asm__
+#if defined(__GNUC__) && (defined(__x86_64__) || defined(__aarch64__))
+#define FORCE_READ(var) __asm__("" ::"r"(var));
 #else
-#define wasm_asm(X)
+#define FORCE_READ(var)
 #endif
+// TODO: handle MIPS and other architectures
 
 #if WABT_BIG_ENDIAN
 static inline void load_data(void* dest, const void* src, size_t n) {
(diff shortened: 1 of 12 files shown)


wasm2c/examples/fac/fac.c
```diff
@@ -66,11 +66,12 @@ static inline bool func_types_eq(const wasm_rt_func_type_t a,
 #define MEMCHECK(mem, a, t) RANGE_CHECK(mem, a, sizeof(t))
 #endif
 
-#ifdef __GNUC__
-#define wasm_asm __asm__
+#if defined(__GNUC__) && (defined(__x86_64__) || defined(__aarch64__))
```
sbc100, line 69: Should enable this by default and just allow it to be turned off selectively.. ?    Maybe we could #error out to first folks on MIPS to opt out (e.g. `#error no support for strict trapping on this architecture.. you can opt out using -DWASM2C_DISABLED_STRICT_TRAPPING` )

glandium:
This probably needs a rebase.

keithw:
Done in #2274

PR #2271, opened 2023-07-30
c-writer.cc: cleanup handling of functions & tags (NFC)

keithw:
This PR has a bunch of miscellaneous cleanups in c-writer.cc, in preparation for the tail-call PR.

- ~~eliminate redundant import_syms_ SymbolSet~~ (split out to #2275)
- eliminate ResultType struct
- replace ExternalPtr (only used for tags) with TagPtr
- refer to functions with bare name (no need for * or &)
- don't use "return" for functions returning void (fixes warning on MSVC)
- add missing "static" on ExportName
- fix spacing around init_instance_import

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -94,8 +94,9 @@ struct GlobalName {
   const std::string& name;
 };
 
-struct ExternalPtr : GlobalName {
-  using GlobalName::GlobalName;
+struct TagSymbol : GlobalName {
+  explicit TagSymbol(const std::string& name)
+      : GlobalName(ModuleFieldType::Tag, name) {}
 };
 
 struct ExternalRef : GlobalName {
@@ -140,11 +141,6 @@ struct TypeEnum {
 struct SignedType {
   explicit SignedType(Type type) : type(type) {}
   Type type;
-};
-
-struct ResultType {
-  explicit ResultType(const TypeVector& types) : types(types) {}
-  const TypeVector& types;
 };
 
 struct TryCatchLabel {
@@ -267,9 +263,9 @@ class CWriter {
   static std::string Mangle(std::string_view name, bool double_underscores);
   static std::string MangleName(std::string_view);
   static std::string MangleModuleName(std::string_view);
-  std::string ExportName(std::string_view module_name,
-                         std::string_view export_name);
-  std::string ExportName(std::string_view export_name);
+  static std::string ExportName(std::string_view module_name,
+                                std::string_view export_name);
+  std::string ExportName(std::string_view export_name) const;
   std::string ModuleInstanceTypeName() const;
   static std::string ModuleInstanceTypeName(std::string_view module_name);
   void ClaimName(SymbolSet& set,
@@ -277,7 +273,8 @@ class CWriter {
                  char type_suffix,
                  std::string_view wasm_name,
                  const std::string& c_name);
-  std::string FindUniqueName(SymbolSet& set, std::string_view proposed_name);
+  std::string FindUniqueName(SymbolSet& set,
+                             std::string_view proposed_name) const;
   std::string ClaimUniqueName(SymbolSet& set,
                               SymbolMap& map,
                               char type_suffix,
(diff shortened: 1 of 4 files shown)


shravanrn (approved):
LGTM

src/c-writer.cc
```diff
@@ -94,8 +94,9 @@ struct GlobalName {
   const std::string& name;
 };
 
-struct ExternalPtr : GlobalName {
-  using GlobalName::GlobalName;
+struct TagPtr : GlobalName {
+  explicit TagPtr(const std::string& name)
```
sbc100, line 98: is `Ptr` the right name here?

src/c-writer.cc
```diff
@@ -453,6 +450,8 @@ class CWriter {
 
   void PushFuncSection(std::string_view include_condition = "");
 
+  bool IsImport(const std::string& name) const;
```
sbc100, line 453: A good chunk of this PR seems to be adding and using this helper?  Can we maybe land this on its own to make this PR smaller?

src/c-writer.cc
```diff
@@ -473,7 +472,6 @@ class CWriter {
   StackVarSymbolMap stack_var_sym_map_;
   SymbolSet global_syms_;
   SymbolSet local_syms_;
-  SymbolSet import_syms_;
```
sbc100, line 476: Was this simply unused?

src/c-writer.cc
```diff
@@ -3389,10 +3378,9 @@ void CWriter::Write(const ExprList& exprs) {
 
         Write(StackVar(0), " = (wasm_rt_funcref_t){", FuncTypeExpr(func_type),
               ", (wasm_rt_function_ptr_t)",
-              ExternalPtr(ModuleFieldType::Func, func->name), ", ");
+              ExternalRef(ModuleFieldType::Func, func->name), ", ");
```
sbc100, line 3381: Can you explain this line?  What was ExternalPtr previously used and why is ExternalRef now OK to use instead?

sbc100 (commented):
lgtm % comments

src/c-writer.cc
keithw: Welll... yes technically (which is why it was ExternalPtr before -- the tags are represented as a `void*`), but we can probably find a clearer name. `TagSymbol`?

src/c-writer.cc
keithw: Sure, stand by...

src/c-writer.cc
keithw: It was used for finding whether a symbol was imported or not, but it was redundant to `import_module_sym_map_`. This PR creates the `IsImport` method which checks if the name exists in `import_module_sym_map_`, and then we can get rid of `import_syms_`. I'll split this out along with the above.

src/c-writer.cc
keithw: For functions, C lets us use the name of the function (instead of explicitly taking the address of the function) when we need a function pointer, which I think ends up cleaner in the output.

Here are some of the output diffs that show the changes:

- Calling an imported function:
  ```diff
  -  (*w2c_wasi__snapshot__preview1_proc_exit)(instance->w2c_wasi__snapshot__preview1_instance, var_i0);
  +  w2c_wasi__snapshot__preview1_proc_exit(instance->w2c_wasi__snapshot__preview1_instance, var_i0);
  ```

- Representing a funcref element segment:
  ```diff
   static const wasm_elem_segment_expr_t elem_segment_exprs_w2c_test_e0[] = {
  -  {w2c_test_t1, (wasm_rt_function_ptr_t)&w2c_test_f1, 0},
  +  {w2c_test_t1, (wasm_rt_function_ptr_t)w2c_test_f1, 0},
   };
  ```

sbc100: I see.. it wasn't obvious to me that that was the effect of this change, but that sgtm

src/c-writer.cc
keithw: Renamed to `TagSymbol`

src/c-writer.cc
keithw: Split out to #2275

PR #2272, opened 2023-07-30
wasm2c: implement the tail-call proposal

keithw:
This PR implements the tail-call proposal in wasm2c. (Will be needed to update the testsuite -- the exception-handling tests now use tail calls.)

The basic strategy is:
- If a function doesn't make any tail-calls, it's written unchanged.
- If a function does make a tail-call, it is written twice: once as the "normal" version, and once as the "tailcallee" version.
- The normal version has the usual signature, and the tailcallee version has the signature `void fn(void **instance_ptr, void *tail_call_stack, wasm_rt_tailcallee_t *next)`
- When a "tailcallee" function starts, it unpacks its params from the `tail_call_stack` argument, and when it returns, it packs its results back into the `tail_call_stack`
- If a "tailcallee" function returns via a `return_call` or `return_call_indirect`, it sets `next` to point to the new function and `instance_ptr` to point to its originating module instance. If it returns normally, it sets `next` to NULL.
- When a "normal" function does a `return_call` or `return_call_indirect`, it implements a trampoline: it allocates space for the tail_call_stack, calls the tailcallee version of function, and then loops forever until `next` is NULL.

To simplify the generated code:
- when tail-calling a function that's known **not** to make a tail-call itself, the generated code just does a normal call instead. This allows wasm2c to avoid generating a lot of extra "tailcallee" versions of functions that don't need them.
- when importing a function, a weak import is used to supply "default" tail-call behavior (converting it into a normal call). This allows the host to avoid having to implement a "tailcallee" version of every import (e.g. the spectest imports, WASI, etc.) just in case it gets tail-called.


shravanrn:
Sorry for the delay on this. The last 45 days or so has been super busy for me. 
@keithw Is this ready for review now or are you still working on this?

keithw:
No worries -- I think it's ready for review.

src/binary-reader-ir.cc
```diff
@@ -891,12 +891,22 @@ Result BinaryReaderIR::OnCallRefExpr() {
 }
 
 Result BinaryReaderIR::OnReturnCallExpr(Index func_index) {
+  if (current_func_) {
+    // syntactically, a return_call expr can occur in an init expression
+    // (outside a function)
+    current_func_->features_used.tailcall = true;
```
shravanrn, line 897: Could you clarify here/the comment on why this is an indication of tail calls being used. I didn't quite follow the current comment

src/template/wasm2c.declarations.c
```diff
@@ -613,3 +619,10 @@ DEFINE_TABLE_FILL(externref)
 #define FUNC_TYPE_EXTERN_T(x) const char x[]
 #define FUNC_TYPE_T(x) static const char x[]
 #endif
+
+#if (__STDC_VERSION__ >= 201112L) || defined(_Static_assert)
+#define wasm_static_assert(X) _Static_assert(X, "assertion failure")
+#else
+#define wasm_static_assert(X) \
```
shravanrn, line 626: I did a quick test, it looks like MSVC generates a warning for this while clang and gcc seem happy. Unfortunately, I can't test with mingw easily as it doesn't seem to be setup right on Godbolt. We should do a sanity check to make sure all standard C compilers are ok with this construction.

test/wasm2c/add.txt
```diff
@@ -681,8 +687,23 @@ DEFINE_TABLE_FILL(externref)
...
+#if (__STDC_VERSION__ >= 201112L) || defined(_Static_assert)
+#define wasm_static_assert(X) _Static_assert(X, "assertion failure")
+#else
+#define wasm_static_assert(X) \
+  extern int(*wasm2c_assert(void))[!!sizeof(struct { int x : (X) ? 2 : -1; })];
+#endif
+
 static u32 w2c_test_add_0(w2c_test*, u32, u32);
 
+#ifndef wasm_multi_ii
```
shravanrn, line 699: Hmm... is the "wasm_multi_ii" avoidable here? (I haven't got to the part where I dig into what this is used for yet. Will look some more in the next pass)

wasm2c/examples/callback/main.c
```diff
@@ -26,7 +26,8 @@ int main(int argc, char** argv) {
    */
   wasm_rt_func_type_t fn_type =
       wasm2c_callback_get_func_type(1, 0, WASM_RT_I32);
-  wasm_rt_funcref_t fn_ref = {fn_type, (wasm_rt_function_ptr_t)print, &inst};
+  wasm_rt_funcref_t fn_ref = {
+      fn_type, (wasm_rt_function_ptr_t)print, {NULL}, &inst};
```
shravanrn, line 30: It's a bit unfortunate that this leaks to the external types of the Wasm module versus being managed internally. This is probably fine, but would be nice to somehow hide from the host/embedder.

wasm2c/wasm-rt-impl.c
```diff
@@ -414,7 +414,7 @@ const char* wasm_rt_strerror(wasm_rt_trap_t trap) {
     case WASM_RT_TRAP_UNREACHABLE:
       return "Unreachable instruction executed";
     case WASM_RT_TRAP_CALL_INDIRECT:
-      return "Invalid call_indirect";
+      return "Invalid call_indirect or return_call_indirect";
```
shravanrn, line 417: Is there a reason this error code has to be overloaded?

shravanrn (commented):
This is just a partial review for now looking at the non-main files. I'll need to look c-writer in the next pass to get to the core of the impl. I also still need to spend some time understanding why hello.txt (which uses wasi) has changed significantly.

okuoku:
(fwiw: I tried to write "_have you considered `__attribute__((musttail))` to directly enforce tail-call_" here.. But it seems it is not suitable for the purpose as it is too restrictive https://github.com/llvm/llvm-project/issues/54964 and anyway we still need "tailcalee" variant of the function to support importing.)

src/template/wasm2c.declarations.c
keithw: Hmm, my only access to MSVC is via our CI, and it appeared to work there without complaint (https://github.com/WebAssembly/wabt/actions/runs/6152502422/job/16694736954#step:14:2320). What's the warning?

wasm2c/examples/callback/main.c
keithw: Yeah... a challenge here is that our "module-to-host" interface is the same as the "module-to-module" interface. So if we want modules to be able to make tailcalls on imported functions from other modules, the tailcallee has to be part of the funcref type visible at this interface boundary. (And if we want the host to be able to make tailcalls to functions from a Wasm module, it also has to be part of the funcref type...)

wasm2c/wasm-rt-impl.c
keithw: I guess not, but it helps share a lot of code between `call_indirect` and `return_call_indirect`.

src/c-writer.cc
```diff
@@ -1452,6 +1501,15 @@ void CWriter::WriteMultivalueType(const TypeVector& types) {
   Write(CloseBrace(), ";", Newline(), "#endif  /* ", name, " */", Newline());
 }
 
+void CWriter::WriteMultivalueParamTypes() {
+  for (TypeEntry* type : module_->types) {
+    FuncType* func_type = cast<FuncType>(type);
+    if (func_type->GetNumParams() > 1) {
+      WriteMultivalueType(func_type->sig.param_types);
```
sbc100, line 1508: It looks like "multivalue" is referring to any function type that takes more than one param?  Is that right?   If so I think it might be best to choose a different term since "multivalue" has a specific existing meaning in the Wasm world.

sbc100: Also. do we need to write out all these types, even for functions that are not tail callable?

test/wasm2c/add.txt
keithw: Done.

src/c-writer.cc
keithw: First comment was done in #2296. Just pushed a commit to this PR (269e5d83110b5e3bca43f780f13ccc4899dd62d8) for the second. The type structs are now only declared when necessary.

src/binary-reader-ir.cc
shravanrn: ah gotcha

src/template/wasm2c.declarations.c
shravanrn: I am seeing `<source>(9): warning C4116: unnamed type definition in parentheses`
Here is the godbolt link https://godbolt.org/z/ePe66x15

wasm2c/wasm-rt-impl.c
shravanrn: ah, gotcha. Tbh, not a big deal. We can fix it if easy, or leave as is.

shravanrn:
> (fwiw: I tried to write "_have you considered `__attribute__((musttail))` to directly enforce tail-call_" here.. But it seems it is not suitable for the purpose as it is too restrictive [llvm/llvm-project#54964](https://github.com/llvm/llvm-project/issues/54964) and anyway we still need "tailcalee" variant of the function to support importing.)

Interesting! Thanks for sharing this. @keithw would probably know more, but I'll also try to have a look to see if we can use to ensure that the C code being generated would be tail-called when compiled with gcc/clang

wasm2c/wasm-rt.h
```diff
@@ -259,14 +268,19 @@ typedef struct {
...
+  wasm_rt_tailcallee_t func_tailcallee;
   /** A function instance is a closure of the function over an instance
    * of the originating module. The module_instance element will be passed into
    * the function at runtime. */
   void* module_instance;
 } wasm_rt_funcref_t;
 
 /** Default (null) value of a funcref */
-static const wasm_rt_funcref_t wasm_rt_funcref_null_value = {NULL, NULL, NULL};
+static const wasm_rt_funcref_t wasm_rt_funcref_null_value = {NULL,
```
sbc100, line 280: I think you can just drop the assignment here.  Static data defaults to zero initialized so it should be equivalent, as long as the compiler allows it.

wasm2c/wasm-rt.h
```diff
@@ -245,6 +245,15 @@ typedef enum {
  */
 typedef void (*wasm_rt_function_ptr_t)(void);
 
+/**
+ * A pointer to a "tail-callee" function, called by a tail-call
+ * trampoline or by another tail-callee function. (The definition uses a
+ * single-member struct to allow a recursive definition.)
+ */
+typedef struct wasm_rt_tailcallee_t {
+  void (*fn)(void**, void*, struct wasm_rt_tailcallee_t*);
```
sbc100, line 254: Naming the params here might provide more info to the reader?

src/template/wasm2c.declarations.c
sbc100: Is the `wasm_` prefix the correct one to use here?  Why not `wasm_rt_`?

src/c-writer.cc
```diff
@@ -656,6 +679,18 @@ std::string CWriter::ExportName(std::string_view module_name,
          MangleName(export_name);
 }
 
+/* The C symbol for a tail-callee export from this module. */
+std::string CWriter::TailCallExportName(std::string_view export_name) const {
+  return kTailCallSymbolPrefix + ExportName(export_name);
+}
+
+/* The C symbol for a tail-callee export from an arbitrary module. */
+// static
```
sbc100, line 688: Looks strange mixing C and C++ comments directly next each other like this.. maybe its fine since the `// static` is more like the type hint?

src/c-writer.cc
```diff
@@ -1745,6 +1820,67 @@ void CWriter::WriteImports() {
...
+          Newline());
+    WriteTailCallFuncDeclaration(
+        kTailCallFallbackPrefix + module_prefix_ + '_' +
+        ExportName(import->module_name, import->field_name));
+    Write(Newline(), "#else", Newline());
+    Write("__attribute__((weak)) ");
+    WriteTailCallFuncDeclaration(
+        TailCallExportName(import->module_name, import->field_name));
+    Write(Newline(), "#endif", Newline(), OpenBrace());
+    Write("next->fn = NULL;", Newline());
```
sbc100, line 1846: Would it be readable/maintainable to define a macro in the internal header rather than try to stamp out the `if/else/endif` here?

src/c-writer.cc
```diff
@@ -2712,6 +2943,110 @@ void CWriter::Write(const Func& func) {
...
+void CWriter::WriteTailCallee(const Func& func) {
+  func_ = &func;
+  in_tail_callee_ = true;
+  local_syms_.clear();
+  local_sym_map_.clear();
+  stack_var_sym_map_.clear();
+  func_sections_.clear();
+  func_includes_.clear();
+
+  Stream* prev_stream = stream_;
```
sbc100, line 2955: Is it worth extracting these 10 lines into a new helper method?  e.g. `BeginFunction`?

sbc100 (commented):
lgtm % comments

test/wasm2c/tail-calls.txt
```diff
@@ -0,0 +1,858 @@
+;;; TOOL: run-wasm2c
+;;; ARGS0: --debug-names
+;;; ARGS*: --enable-tail-call
+(type $i32_f32 (func (param i32 f32)))
```
shravanrn, line 4: Any chance we can use a test with a return value as well. It would be nice to see the codegen for this part.

test/wasm2c/tail-calls.txt
```diff
@@ -0,0 +1,858 @@
...
+  wasm_static_assert(sizeof(struct wasm_multi_if) <= 1024);
+  CHECK_CALL_INDIRECT(instance->w2c_tab, w2c_test_i32_f32, var_i2);
+  if (!instance->w2c_tab.data[var_i2].func_tailcallee.fn) {
+    CALL_INDIRECT(instance->w2c_tab, void (*)(void*, u32, f32), w2c_test_i32_f32, var_i2, instance->w2c_tab.data[var_i2].module_instance, var_i0, var_f1);
+    goto var_Bfunc;
+  }
+  {
+    void *instance_ptr_storage;
+    void **instance_ptr = &instance_ptr_storage;
+    char tail_call_stack[1024];
```
shravanrn, line 815: I wonder if this is going to put a lot of stack pressure if we have a tail called function whose body starts another tail call chain. Could we declare the tail_call_stack directly as a union of the param types (`struct wasm_multi_if`) and any result type of the callee?

shravanrn (changes requested):
(Still need to look at a part of c-writer)

src/template/wasm2c.declarations.c
keithw: I guess ny thinking was that these static asserts aren't part of the runtime ("_rt_") API, or even something that happens at runtime at all.

wasm2c/wasm-rt.h
keithw: Done.

wasm2c/wasm-rt.h
keithw: Done.

src/c-writer.cc
keithw: Done (or at least attempted -- doing the #pragma inside a #define was a little tricky).

src/c-writer.cc
keithw: Done.

test/wasm2c/hello.txt
```diff
@@ -62,9 +62,11 @@ wasm_rt_func_type_t wasm2c_test_get_func_type(uint32_t param_count, uint32_t res
 
 /* import: 'wasi_snapshot_preview1' 'fd_write' */
 u32 w2c_wasi__snapshot__preview1_fd_write(struct w2c_wasi__snapshot__preview1*, u32, u32, u32, u32);
+void wasm2c_tailcall_w2c_wasi__snapshot__preview1_fd_write(void **instance_ptr, void *tail_call_stack, wasm_rt_tailcallee_t *next);
```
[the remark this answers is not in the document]
keithw: I think it would be tricky to conditionally emit these based on whether it's needed for a particular import -- we'd need to generate it if the import is ever called with `return_call` **or** if it's ever the target of a "ref.func" (including in an elem segment or global initializer) **or** if it's exported. We could add that tracking to the BinaryReaderIR but it would need to happen in a bunch of places. We're not really tracking anything specific to tail calls at that point; it's more like "can we prove that this import is never tail-called by anybody." :-/

If you want, what we could do is expose the Features to CWriter, and then this could be disabled globally behind `--enable-tail-call` (or, probably someday soon, `--disable-tail-call`). Would that work for you?

src/c-writer.cc
```diff
@@ -517,6 +537,9 @@ static constexpr char kLabelSuffix = kParamSuffix + 1;
 static constexpr char kGlobalSymbolPrefix[] = "w2c_";
 static constexpr char kLocalSymbolPrefix[] = "var_";
 static constexpr char kAdminSymbolPrefix[] = "wasm2c_";
```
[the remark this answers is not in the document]
keithw: Good catch! I switched it to "wasm_".

test/wasm2c/tail-calls.txt
```diff
@@ -0,0 +1,858 @@
...
+  u32 var_i0, var_i2;
+  f32 var_f1;
+  var_i0 = 1u;
+  var_f1 = 2;
+  var_i2 = 0u;
+  wasm_static_assert(sizeof(struct wasm_multi_if) <= 1024);
+  CHECK_CALL_INDIRECT(instance->w2c_tab, w2c_test_i32_f32, var_i2);
+  if (!instance->w2c_tab.data[var_i2].func_tailcallee.fn) {
+    CALL_INDIRECT(instance->w2c_tab, void (*)(void*, u32, f32), w2c_test_i32_f32, var_i2, instance->w2c_tab.data[var_i2].module_instance, var_i0, var_f1);
+    goto var_Bfunc;
```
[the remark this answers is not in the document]
keithw: Good idea! Done. Hopefully I didn't screw it up.

test/wasm2c/tail-calls.txt
keithw: Sure, let me work on this.

test/wasm2c/tail-calls.txt
keithw: In a **pure** chain of tail calls, the instance_ptr_storage and tail_call_stack only get declared once. A tail-callee won't have to declare these when making a tail-call. However, you're right that if a tail-callee makes a normal call, and then _that_ function makes a tail-call, then another tail_call_stack will get declared as another 1024-byte array.

Unfortunately I don't think we can be a lot smarter in picking an adaptive size for the tail_call_stack. The problem is that we don't know the union of param types -- e.g. if we tail-call a function in a different module that wants to tail-call another function with a type that we've never seen.

src/template/wasm2c.declarations.c
sbc100: Is `wasm_` the right prefix though?   What else do we put in `wasm_`?   There are not public symbol that exist outside the TU, right?  So perhaps simply `static_assert` with no namespace?

keithw: Unfortunately `static_assert` is a reserved keyword in C23 and C++11. :-(

test/wasm2c/hello.txt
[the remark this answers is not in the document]
shravanrn: > If you want, what we could do is expose the Features to CWriter, and then this could be disabled globally behind --enable-tail-call (or, probably someday soon, --disable-tail-call). Would that work for you?

Sure, this could be a good alternative for us!

test/wasm2c/tail-calls.txt
shravanrn: ah, i see. Ok, closing this as I don't see an immediate way to optimize this even in the single module scenario without a bunch of additional tracking.

src/template/wasm2c.declarations.c
sbc100: Perhaps we can detect that and not define this at all in those cases?  i.e. define `static_assert` if and only if the compiler didn't already.

test/wasm2c/hello.txt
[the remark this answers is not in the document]
keithw: Great, just pushed.

test/wasm2c/tail-calls.txt
keithw: Just pushed -- agreed that this is illustrative (it also caught a missing newline).

src/c-writer.cc
keithw: Agreed that it's a little strange, but it matches the existing style (e.g. on CWriter::ExportName and CWriter::ModuleInstanceTypeName). We could maybe change all these together in a future PR?

src/c-writer.cc
```diff
@@ -2645,16 +2847,38 @@ void CWriter::WriteUnwindTryCatchStack(const Label* label) {
...
+  if (in_tail_callee_) {
+    Write(CloseBrace(), Newline(), "return;", Newline());
+    return;
+  }
+
+  Write("while (next->fn) { next->fn(instance_ptr, tail_call_stack, next); }",
+        Newline());
+  PushTypes(func_->decl.sig.result_types);
+  const Index num_results = func_->decl.sig.result_types.size();
+  if (num_results == 1) {
```
shravanrn, line 2860: Code cleanup question: Can this part of the code where c-writer handle single return vs. multi-value return be factored out into a helper that can be used in the "ExprType::Call"" handling as well as here? It seems a bit messy having to special case this multiple times. (If not, no worries. Not blocking)

src/c-writer.cc
```diff
@@ -1745,6 +1823,57 @@ void CWriter::WriteImports() {
...
+void CWriter::WriteTailCallWeakImports() {
+  for (const Import* import : unique_imports_) {
+    if (import->kind() != ExternalKind::Func) {
+      continue;
+    }
+    const Func& func = cast<FuncImport>(import)->func;
+    Write(Newline(), "/* handler for missing tail-call on import: '",
+          SanitizeForComment(import->module_name), "' '",
+          SanitizeForComment(import->field_name), "' */", Newline());
+    Write("WEAK_FUNC_DECL(",
```
shravanrn, line 1835: We don't need to fix this now: but in the future we may need to have a macro that controls whether the weak func's are a part of compiled output of wasm2c. This will allow use in toolchains that have bad weak symbol support.

```c
# ifndef WASM_NO_TAILCALL_DEFAULTS
WEAK_FUNC_DECL(...);
# endif
```

src/c-writer.cc
```diff
@@ -2712,6 +2940,101 @@ void CWriter::Write(const Func& func) {
...
+  BeginFunction(func);
+  in_tail_callee_ = true;
+  PushFuncSection();
+  WriteTailCallFuncDeclaration(GetTailCallRef(func.name));
+  Write(" ", OpenBrace());
+  WriteTailCallAsserts(func.decl.sig);
+  Write(ModuleInstanceTypeName(), "* instance = *instance_ptr;", Newline());
+
+  std::vector<std::string> index_to_name;
+  MakeTypeBindingReverseMapping(func_->GetNumParamsAndLocals(), func_->bindings,
```
shravanrn, line 2954: The mix of `func` and `func_` in the same function is a bit confusing. From what I can see, they are the same obj? If these are the same could be use either one. If different, could we rename one to be more descriptive?

src/c-writer.cc
```diff
@@ -2712,6 +2940,101 @@ void CWriter::Write(const Func& func) {
...
+          Write(DefineParamName(index_to_name[param_index]));
+          ++count;
+        }
+        ++param_index;
+      }
+      if (count != 0) {
+        Dedent(4);
+        Write(";", Newline());
+      }
+    }
```
shravanrn, line 2984: Could we refactor some of this into its own function for readability?

src/c-writer.cc
```diff
@@ -2712,6 +2940,101 @@ void CWriter::Write(const Func& func) {
...
+      stream_->WriteData(buf->data.data(), buf->data.size());
+    }
+
+    if (i == 0) {
+      WriteStackVarDeclarations();  // these come immediately after section #0
+                                    // (return type/name/params/locals)
+    }
+  }
+
+  Write(CloseBrace(), Newline());
```
shravanrn, line 3033: It would be nice to dedup this code from `CWriter::Write(const Func& func)` if they are the same

shravanrn (changes requested):
Sorry, one more batch of comments. I finally managed to go through the whole PR this time. Biggest concern is one new source of UB here that we need to avoid. 

I think I finally understand the strategy for tailcalls you have chosen and the codegen for this in the example files now makes sense to me. This is nice in that we won't need any C compiler attributes to force a tail-call, since we are doing it by hand in the tail_call_stack.

I can't say I understand 100% of the c-writer code, but it mostly made sense to me. With a little code-dedup and refactoring, I think this is nearly there.

test/pipes.txt
```diff
@@ -2,6 +2,8 @@
 ;;; RUN: bash -c '%(wasm-stats)s <(%(wat2wasm)s /dev/stdin <<<"(module)" -o /dev/stdout)'
```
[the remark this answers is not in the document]
keithw: I think it just never got the tests run on it (with `-r`) since it was committed, so any run of `run-tests.py -r` is going to make these whitespace changes.

test/wasm2c/tail-calls.txt
```diff
@@ -0,0 +1,953 @@
...
+  static_assert(sizeof(struct wasm_multi_id) <= 1024);
+  {
+    void *instance_ptr_storage;
+    void **instance_ptr = &instance_ptr_storage;
+    char tail_call_stack[1024];
+    wasm_rt_tailcallee_t next_storage;
+    wasm_rt_tailcallee_t *next = &next_storage;
+    {
+      struct wasm_multi_id *tmp = (struct wasm_multi_id *)tail_call_stack;
+      tmp->i0 = var_i0;
```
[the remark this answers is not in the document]
keithw: Hmm, I don't _think_ this is undefined behavior -- every object's value is being accessed as the same type as when it was stored-to or modified. And if not, I don't think an explicit `memcpy` would fix it -- it would still be UB afaik.

(Footnote: That's assuming the module is valid, which the command-line wasm2c tool enforces, but CWriter just assumes as a precondition. I think if we're worried about invalid modules, CWriter probably would need bolstering in a lot of places including this one. :-/)

test/wasm2c/tail-calls.txt
```diff
@@ -0,0 +1,953 @@
...
+    next->fn = wasm_tailcall_w2c_test_infiniteloop;
+    while (next->fn) { next->fn(instance_ptr, tail_call_stack, next); }
+    {
+      struct wasm_multi_id *tmp = tail_call_stack;
+      var_i0 = tmp->i0;
+      var_d1 = tmp->d1;
+    }
+  }
+  goto var_Bfunc;
+  var_Bfunc:;
```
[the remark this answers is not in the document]
keithw: It looks like wasm2c has done this since the beginning (https://github.com/WebAssembly/wabt/blob/main/src/c-writer.cc#L1111); I think the reason is to make the code correct even when the label occurs at the very end of a function. (A label must precede a statement but the null statement is okay.)

src/c-writer.cc
keithw: Yeah, agreed. I think we did this partway with Spill/Unspill, but I think if you look at all the places we have an `if (num_params == 1)` and `if (num_results == 1)` in the code, I couldn't find enough commonality to factor out further. :-/

src/c-writer.cc
keithw: Fair enough. These weak symbols are only generated when tail calls are explicitly enabled (with `--enable-tail-call`), so I don't think there's an immediate issue. I do imagine this may become the default soon now that tail calls are at phase 4, so the story for these toolchains might turn into "you'll want to run wasm2c with `--disable-tail-call`".

src/c-writer.cc
keithw: Fixed in 7a08fc17285ff5276896d85475b798aaae4da856 to use "func" consistently (matching the existing `CWriter::Write(const Func&)`).

src/c-writer.cc
keithw: Sure, done in b5e9ad1cfac1343eb127a5494ab05e19ddef889d

src/c-writer.cc
keithw: Done in 934b6695a314d54514f62a3f8989ffafeb3e3874

test/wasm2c/tail-calls.txt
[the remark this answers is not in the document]
shravanrn: ah gotcha! :+1:

test/wasm2c/tail-calls.txt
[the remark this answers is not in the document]
shravanrn: Hmm... We may need to go through this a bit more before I am fully convinced. Here is my reasoning below for why this could be UB. I want to make sure this doesn't apply before we commit. Let me know your thoughts.

Of the top of my head, I believe the high level reason I think this maybe an issue, is that it is in-effect a use of type punning, which I believe in general is UB, even with the property "every object's value is being accessed as the same type as when it was stored-to or modified". This property is sufficient for the safe use of union types, but without union types, I think this may not be enough. 

I believe the heart of this discrepancy is that C can assume that pointers of different types do not alias, so any code of the form 

```c
struct wasm_multi_id* a = (struct wasm_multi_id*) tail_call_stack;
...
struct wasm_multi_ii* b = (struct wasm_multi_ii*) tail_call_stack;
```

is compiled assuming `a` and `b` don't alias. 

Fwiw, i did run this through uban and it didn't complain, so maybe its ok? However, this pattern makes me nervous. I think with local vars + memcpy, there is no possibility of the code being UB as I believe the C standard says this is always safe much more directly. In past projects, I have seen that compiler is definitely smart enough to optimize these extra memcpy's and local vars away, so its unlikely to be a performance hit. Given this, I think this maybe the best course of action here.

In case its useful, I wroteup a small godbolt demo for how to save the contents of different structs in shared char buffer (both how we do this currently and how to do it with local vars without UB)

https://godbolt.org/z/79jWbcWPb

Let me know your thoughts!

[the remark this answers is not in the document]
shravanrn: (Also found a couple of links when searching
https://stackoverflow.com/questions/63422076/why-is-type-punning-considered-ub
https://adriann.github.io/undefined_behavior.html)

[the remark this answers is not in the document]
keithw: Ok, changed it to use memcpy everywhere.

src/template/wasm2c.declarations.c
```diff
@@ -613,3 +619,21 @@ DEFINE_TABLE_FILL(externref)
...
+#define WEAK_FUNC_DECL(func, fallback)                             \
+  __pragma(comment(linker, "/alternatename:" #func "=" #fallback)) \
+                                                                   \
+      void                                                         \
+      fallback(void** instance_ptr, void* tail_call_stack,         \
+               wasm_rt_tailcallee_t* next)
+#else
+#define WEAK_FUNC_DECL(func, fallback)                                        \
+  __attribute__((weak)) void func(void** instance_ptr, void* tail_call_stack, \
+                                  wasm_rt_tailcallee_t* next)
```
sbc100, line 638: This version seems to completely ignore `fallback`.. how does that work?

shravanrn (approved):

src/template/wasm2c.declarations.c
shravanrn: i think the fallback is just an alternate way to implement weak symbols in MSVC? IIUC, they aren't used in the non MSVC environments as weak symbols are available.

sbc100: I guess I don't understand how that fallback function is used on MSVC but a weak symbol is used otherwise.    Do we also declare the fallback symbol on under MSVC?

keithw: Yeah, the "fallback" parameter is only used on MSVC.

On GCC/clang, the "weak import" is implemented by defining a stub version of the import under its real name, and marking it as a weak symbol so it will be overridden by the real import if it exists.

On MSVC, it's implemented by defining the stub version of the import under the "fallback" name, and then telling the linker to use the fallback as a substitute for the real import if the real import doesn't exist.

keithw:
@sbc100 Did you want to weigh in again on this one? My thinking is that after we get this in, we can finally update the testsuite and possibly cut a new release.

sbc100 (approved):

keithw:
Ok, merging now, and then let's try to get the testsuite updated, a release cut, and #2308 merged (not necessarily in that order).

PR #2273, opened 2023-07-30
Fix CI from upgraded cmake version on GitHub Ubuntu runner

keithw:
GitHub recently upgraded the cmake version on its Ubuntu runners, which broke our CI build with some new warnings (errors for us).

One of them is fixable (related to a deprecated Python package), but one of them is in libuv (which is a dependency of the uvwasi submodule) so less easy to fix. I added `-Wno-deprecated` to the cmake command line on GitHub actions to silence the warning for now.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,7 +50,7 @@ jobs:
     - name: mkdir
       run: mkdir -p out
     - name: cmake
-      run: cmake .. -G Ninja -DWITH_WASI=ON -DWERROR=ON -Werror=dev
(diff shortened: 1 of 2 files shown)


sbc100 (approved):

keithw:
I filed a PR to fix the underlying deprecation warning in libuv (https://github.com/libuv/libuv/pull/4111); I think we can go back and remove `-Wno-deprecated` once it (or something like it) is merged, libuv cuts a release, the release is picked up by uvwasi, and we update our uvwasi submodule.

PR #2274, opened 2023-07-30
wasm2c: ensure force read constraints compile for clang on mips

### LINKED ISSUE #2266: error: couldn't allocate input reg for constraint 'r' on MIPS with clang when compiling wasm2c-generated source

shravanrn:
This is a clean slate attempt to fix #2266. To fix this i looked into how clang and gcc handle floating point constraints across a host of platforms.

Godbolt link: https://godbolt.org/z/ddeWc6ees
Results are in table below.

Given this, it seems the simplest course of action that maintains functionality as is today, but fix the mips platform is to special case this. We can separately file a bug upstream (but best case scenario is that will likely get fixed only for future clang versions, so I'm not sure we can rely on just that).

Fyi, @glandium

gcc
---

x86 - both
x86-64 - both
aarch32 - only "r"
aarch64 - only "r"
mips - only "r"
mips64 - both
riscv32 - both
riscv64 -  both
sparc - both
sparc64 - both
powerpc - both
powerpc64 - both

clang
-----
x86 - both
x64 - both
aarch32 - only "r"
aarch64 - only "r"
mips -  only "f"
mips64 - only "f"
riscv32 -  both
riscv64 -  both
powerpc64 - both

### DIFF
--- a/src/template/wasm2c.declarations.c
+++ b/src/template/wasm2c.declarations.c
@@ -48,9 +48,17 @@ static inline bool func_types_eq(const wasm_rt_func_type_t a,
 #endif
 
 #ifdef __GNUC__
-#define wasm_asm __asm__
+#define FORCE_READ_INT(var) __asm__("" ::"r"(var));
+// Clang on Mips requires "f" constraints on floats
+#if defined(__clang__) && \
+    (defined(mips) || defined(__mips__) || defined(__mips))
+#define FORCE_READ_FLOAT(var) __asm__("" ::"f"(var));
 #else
-#define wasm_asm(X)
+#define FORCE_READ_FLOAT(var) __asm__("" ::"r"(var));
+#endif
+#else
+#define FORCE_READ_INT(var)
+#define FORCE_READ_FLOAT(var)
 #endif
 
 #if WABT_BIG_ENDIAN
@@ -72,13 +80,13 @@ static inline void load_data(void* dest, const void* src, size_t n) {
     RANGE_CHECK((&m), m.size - o - s, s);       \
     load_data(&(m.data[m.size - o - s]), i, s); \
   } while (0)
-#define DEFINE_LOAD(name, t1, t2, t3)                                  \
+#define DEFINE_LOAD(name, t1, t2, t3, force_read)                      \
   static inline t3 name(wasm_rt_memory_t* mem, u64 addr) {             \
     MEMCHECK(mem, addr, t1);                                           \
     t1 result;                                                         \
     wasm_rt_memcpy(&result, &mem->data[mem->size - addr - sizeof(t1)], \
                    sizeof(t1));                                        \
-    wasm_asm("" ::"r"(result));                                        \
+    force_read(result);                                                \
     return (t3)(t2)result;                                             \
   }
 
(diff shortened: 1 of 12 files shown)


src/template/wasm2c.declarations.c
```diff
@@ -47,10 +47,18 @@ static inline bool func_types_eq(const wasm_rt_func_type_t a,
 #define MEMCHECK(mem, a, t) RANGE_CHECK(mem, a, sizeof(t))
 #endif
 
-#ifdef __GNUC__
-#define wasm_asm __asm__
+#if !defined(WASM2C_DISABLED_STRICT_TRAPPING) && defined(__GNUC__)
```
keithw, line 50: Given that our fallback is still going to be "don't force the read" (instead of "bomb out with an error"), can we avoid adding the `WASM2C_DISABLED_STRICT_TRAPPING` macro?

keithw (commented):
lgtm % comment. Thank you for doing this whole investigation!

src/template/wasm2c.declarations.c
shravanrn: Fixed!

shravanrn:
> lgtm % comment. Thank you for doing this whole investigation!

Fixed :+1:

keithw (approved):

src/template/wasm2c.declarations.c
```diff
@@ -48,9 +48,17 @@ static inline bool func_types_eq(const wasm_rt_func_type_t a,
 #endif
 
 #ifdef __GNUC__
-#define wasm_asm __asm__
+#define FORCE_READ_INT(var) __asm__("" ::"r"(var));
+// Clang on Mips requires "f" constraints on floats
```
sbc100, line 52: Maybe add link to upstream llvm bug here, so we can perhaps remove this one day?

sbc100 (commented):
I wonder if this could be accompanied by some kind CI test?  Can this be verified by simply compiling some wasm2c output with `clang --target=mips` maybe?

shravanrn:
> I wonder if this could be accompanied by some kind CI test? Can this be verified by simply compiling some wasm2c output with `clang --target=mips` maybe?

My take is that this feels like overkill at the moment, and we can push this until/if we see other mips incompatibilities

src/template/wasm2c.declarations.c
shravanrn: Sure, let me submit a separate PR for this

shravanrn: Link: https://github.com/llvm/llvm-project/issues/64241

shravanrn: https://github.com/WebAssembly/wabt/pull/2276

shravanrn: @keithw @sbc100 Looks like the bug in llvm got fixed. We may have to keep an eye if we need add an llvm version check here to keep things compatible

sbc100: Version check makes sense to me yes

PR #2278, opened 2023-07-31
wat-writer.cc: resolve missing type index in call_indirect

### LINKED ISSUE #2270: desugar: overflow issue for typeidx of call_indirect instruction
wat-desugar (1.0.33) seems to have an overflow issue for the `typeidx` of the `call_indirect` instruction.

e.g.:
This module (located [here](https://github.com/WebAssembly/spec/blob/master/test/core/ref_func.wast)):
``` wat
(module
  (func $f (import "M" "f") (param i32) (result i32))
  (func $g (param $x i32) (result i32)
    (i32.add (local.get $x) (i32.const 1))
  )

  (global funcref (ref.func $f))
  (global funcref (ref.func $g))
  (global $v (mut funcref) (ref.func $f))

  (global funcref (ref.func $gf1))
  (global funcref (ref.func $gf2))
  (func (drop (ref.func $ff1)) (drop (ref.func $ff2)))
  (elem declare func $gf1 $ff1)
  (elem declare funcref (ref.func $gf2) (ref.func $ff2))
  (func $gf1)
  (func $gf2)
  (func $ff1)
  (func $ff2)

  (func (export "is_null-f") (result i32)
    (ref.is_null (ref.func $f))
  )
  (func (export "is_null-g") (result i32)
    (ref.is_null (ref.func $g))
  )
  (func (export "is_null-v") (result i32)
    (ref.is_null (global.get $v))
  )

  (func (export "set-f") (global.set $v (ref.func $f)))
  (func (export "set-g") (global.set $v (ref.func $g)))

  (table $t 1 funcref)
  (elem declare func $f $g)

  (func (export "call-f") (param $x i32) (result i32)
    (table.set $t (i32.const 0) (ref.func $f))
    (call_indirect $t (param i32) (result i32) (local.get $x) (i32.const 0))
  )
  (func (export "call-g") (param $x i32) (result i32)
    (table.set $t (i32.const 0) (ref.func $g))
    (call_indirect $t (param i32) (result i32) (local.get $x) (i32.const 0))
  )
  (func (export "call-v") (param $x i32) (result i32)
    (table.set $t (i32.const 0) (global.get $v))
    (call_indirect $t (param i32) (result i32) (local.get $x) (i32.const 0))
  )
)
```
will look like this when run through `wat-desugar`:
``` wat
(module
  (import "M" "f" (func $f (param i32) (result i32)))
  (func $g (param $x i32) (result i32)
    local.get $x
    i32.const 1
    i32.add)
  (global (;0;) funcref (ref.func $f))
  (global (;1;) funcref (ref.func $g))
  (global $v (mut funcref) (ref.func $f))
  (global (;3;) funcref (ref.func $gf1))
  (global (;4;) funcref (ref.func $gf2))
  (func (;2;)
    ref.func $ff1
    drop
    ref.func $ff2
    drop)
  (elem (;0;) declare func $gf1 $ff1)
  (elem (;1;) declare func $gf2 $ff2)
  (func $gf1)
  (func $gf2)
  (func $ff1)
  (func $ff2)
  (func (;7;) (result i32)
    ref.func $f
    ref.is_null)
  (export "is_null-f" (func 7))
  (func (;8;) (result i32)
    ref.func $g
    ref.is_null)
  (export "is_null-g" (func 8))
  (func (;9;) (result i32)
    global.get $v
    ref.is_null)
  (export "is_null-v" (func 9))
  (func (;10;)
    ref.func $f
    global.set $v)
  (export "set-f" (func 10))
  (func (;11;)
    ref.func $g
    global.set $v)
  (export "set-g" (func 11))
  (table $t 1 funcref)
  (elem (;2;) declare func $f $g)
  (func (;12;) (param $x i32) (result i32)
    i32.const 0
    ref.func $f
    table.set $t
    local.get $x
    i32.const 0
    call_indirect $t (type 4294967295))
  (export "call-f" (func 12))
  (func (;13;) (param $x i32) (result i32)
    i32.const 0
    ref.func $g
    table.set $t
    local.get $x
    i32.const 0
    call_indirect $t (type 4294967295))
  (export "call-g" (func 13))
  (func (;14;) (param $x i32) (result i32)
    i32.const 0
    global.get $v
    table.set $t
    local.get $x
    i32.const 0
    call_indirect $t (type 4294967295))
  (export "call-v" (func 14))
  (type (;0;) (func (param i32) (result i32)))
  (type (;1;) (func))
  (type (;2;) (func (result i32))))
```

Notice that the `typeidx` of `call_indirect` is **4294967295** when it should be **0** (in this case).

keithw:
Fixes #2270

### DIFF
--- a/test/desugar/call_indirect.txt
+++ b/test/desugar/call_indirect.txt
@@ -1,0 +1,24 @@
+;;; TOOL: wat-desugar
+(table 0 funcref)
+(func
+ i32.const 0
+ i32.const 0
+ i32.const 0
+ i32.const 0
+ call_indirect (type 0)
+ call_indirect
+ call_indirect (param i32))
+(;; STDOUT ;;;
+(module
+  (table (;0;) 0 funcref)
+  (func (;0;)
+    i32.const 0
+    i32.const 0
+    i32.const 0
+    i32.const 0
+    call_indirect (type 0)
+    call_indirect (type 0)
+    call_indirect (type 1))
+  (type (;0;) (func))
+  (type (;1;) (func (param i32))))
+;;; STDOUT ;;)
--- a/src/wat-writer.cc
+++ b/src/wat-writer.cc
@@ -671,7 +671,11 @@ Result WatWriter::ExprVisitorDelegate::OnCallIndirectExpr(
   writer_->WritePutsSpace(Opcode::CallIndirect_Opcode.GetName());
   writer_->WriteVarUnlessZero(expr->table, NextChar::Space);
   writer_->WriteOpenSpace("type");
-  writer_->WriteVar(expr->decl.type_var, NextChar::Newline);
+  const auto type_var =
+      expr->decl.has_func_type
+          ? expr->decl.type_var
+          : Var{writer_->module.GetFuncTypeIndex(expr->decl), expr->loc};
+  writer_->WriteVar(type_var, NextChar::Newline);
   writer_->WriteCloseNewline();
   return Result::Ok;
 }
@@ -911,7 +915,11 @@ Result WatWriter::ExprVisitorDelegate::OnReturnCallIndirectExpr(
     ReturnCallIndirectExpr* expr) {
   writer_->WritePutsSpace(Opcode::ReturnCallIndirect_Opcode.GetName());
   writer_->WriteOpenSpace("type");
-  writer_->WriteVar(expr->decl.type_var, NextChar::Space);
+  const auto type_var =
+      expr->decl.has_func_type
+          ? expr->decl.type_var
+          : Var{writer_->module.GetFuncTypeIndex(expr->decl), expr->loc};
+  writer_->WriteVar(type_var, NextChar::Space);
   writer_->WriteCloseNewline();
   return Result::Ok;
 }


test/desugar/call_indirect.txt
```diff
@@ -0,0 +1,24 @@
+;;; TOOL: wat-desugar
```
sbc100, line 1: Does this issue only effect wat-desugar?  From the change it looks like it would also effect wasm2wat?

sbc100 (approved):

test/desugar/call_indirect.txt
keithw: I think the BinaryReaderIR is never going to make a CallIndirect Expr where decl.has_func_type is false, because the binary format always has a sig_index right there physically. So I think the only way this can happen is if the Module originated as WAT and is being written out again as WAT. (Which afaik only happens in wat-desugar.) FWIW the BinaryWriter always does this lookup unconditionally.

sbc100: I wonder if we should just drop `wat-desugar` completely.. and have folks use `wat2wasm + wasm2wat` to get this effect?    Is this tool worth maintaining on its own?

keithw: No objection from me to getting rid of wat-desugar. I doubt it gets a lot of use. But we should probably fix this WatWriter issue either way. :-)

sbc100: Sure this lgtm either way.

PR #2282, opened 2023-08-12
wasm-strip: add `-R/--remove-section` option

dzfrias:
This adds an option to the `wasm-strip` tool. When using it, I had cases where I'd want to only cut out a specific section, like `__wasm_bindgen_unstable` while still keeping other information, like DWARF sections.

This change is just a suggestion, but I believe it would be nice to have at least as an option for the `wasm-strip` interface.

### DIFF
--- a/src/tools/wasm-strip.cc
+++ b/src/tools/wasm-strip.cc
@@ -64,11 +64,11 @@ static void ParseOptions(int argc, char** argv) {
 class BinaryReaderStrip : public BinaryReaderNop {
  public:
   explicit BinaryReaderStrip(std::set<std::string_view> sections_to_keep,
-                             std::set<std::string_view> sections_to_exclude,
+                             std::set<std::string_view> sections_to_remove,
                              Errors* errors)
       : errors_(errors),
         sections_to_keep_(sections_to_keep),
-        sections_to_exclude_(sections_to_exclude),
+        sections_to_remove_(sections_to_remove),
         section_start_(0) {
     stream_.WriteU32(WABT_BINARY_MAGIC, "WASM_BINARY_MAGIC");
     stream_.WriteU32(WABT_BINARY_VERSION, "WASM_BINARY_VERSION");
@@ -99,16 +99,16 @@ class BinaryReaderStrip : public BinaryReaderNop {
   Result BeginCustomSection(Index section_index,
                             Offset size,
                             std::string_view section_name) override {
-    if (!sections_to_exclude_.empty() &&
-        sections_to_exclude_.count(section_name) > 0) {
-      return Result::Ok;
-    } else if (!sections_to_exclude_.empty()) {
-      WriteCustom(size);
+    if (!sections_to_remove_.empty() &&
+        sections_to_remove_.count(section_name) > 0) {
       return Result::Ok;
     }
 
-    if (sections_to_keep_.count(section_name) > 0) {
-      WriteCustom(size);
+    if (sections_to_keep_.count(section_name) > 0 ||
+        !sections_to_remove_.empty()) {
+      stream_.WriteU8Enum(BinarySection::Custom, "section code");
+      WriteU32Leb128(&stream_, size, "section size");
+      stream_.WriteData(state->data + section_start_, size, "section data");
     }
     return Result::Ok;
   }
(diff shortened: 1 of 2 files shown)


sbc100 (commented):
I think it might be better to follow the `strip` command here and use `-R/--remove-section`.

From the man page for `strip`:

```
       -R sectionname
       --remove-section=sectionname
           Remove any section named sectionname from the output file, in addition to whatever
           sections would otherwise be removed.  This option may be given more than once.  Note that
           using this option inappropriately may make the output file unusable.  The wildcard
           character * may be given at the end of sectionname.  If so, then any section starting with
           sectionname will be removed.

           If the first character of sectionpattern is the exclamation point (!) then matching
           sections will not be removed even if an earlier use of --remove-section on the same
           command line would otherwise remove it.  For example:

                     --remove-section=.text.* --remove-section=!.text.foo

           will remove all sections matching the pattern '.text.*', but will not remove the section
           '.text.foo'.
```

src/tools/wasm-strip.cc
```diff
@@ -28,6 +28,7 @@ using namespace wabt;
 static std::string s_filename;
 static std::string s_outfile;
 static std::set<std::string_view> v_sections_to_keep{};
+static std::set<std::string_view> v_sections_to_exclude{};
```
sbc100, line 31: I would avoid using the term exclude here since it could be interpreted as "exclude from the stripping" i.e. "keep".

I think `remove` is more clear.

sbc100:
Also, did you know that `llvm-strip` can already do this?

src/tools/wasm-strip.cc
dzfrias: I agree - I'll rename it!

dzfrias:
`-R` is definitely better, I was not sure about the naming but I think that's a lot better. I wasn't aware that `llvm-strip` could do this! I can close this PR, if you don't think this is a worth-it change to this tool.

sbc100:
If you have access to llvm-strip and it works for you then maybe we can just close this.

On the other hand this looks pretty much ready to land, and I think its a reasonable feature.  All we would want is a test for it.

dzfrias:
While it wouldn't be too hard to get `llvm-strip`, I personally think this would make `wasm-strip` more complete, so this feature would be nice. I also think it makes sense to have since `wabt` decided to include an object stripping tool in the first place.

I can write a test for it in the near future.

dzfrias:
I wrote a test! Let me know if there's anything else I should do

src/tools/wasm-strip.cc
```diff
@@ -52,15 +53,22 @@ static void ParseOptions(int argc, char** argv) {
...
 }
 
 class BinaryReaderStrip : public BinaryReaderNop {
  public:
   explicit BinaryReaderStrip(std::set<std::string_view> sections_to_keep,
+                             std::set<std::string_view> sections_to_exclude,
                              Errors* errors)
       : errors_(errors),
         sections_to_keep_(sections_to_keep),
+        sections_to_exclude_(sections_to_exclude),
```
sbc100, line 71: Can you replace `exclude` with `remove` here?

src/tools/wasm-strip.cc
```diff
@@ -91,10 +99,16 @@ class BinaryReaderStrip : public BinaryReaderNop {
...
                             std::string_view section_name) override {
+    if (!sections_to_exclude_.empty() &&
+        sections_to_exclude_.count(section_name) > 0) {
+      return Result::Ok;
+    } else if (!sections_to_exclude_.empty()) {
+      WriteCustom(size);
+      return Result::Ok;
+    }
+
     if (sections_to_keep_.count(section_name) > 0) {
```
sbc100, line 110: Maybe combine these two conditions.. avoiding the need for the new method:

```
if (sections_to_keep_.count(section_name) > 0 || !sections_to_exclude_.empty()) {
   ...
}
```

(should make the diff smaller?)

sbc100 (approved):
lgtm % some nits

src/tools/wasm-strip.cc
```diff
@@ -91,7 +99,13 @@ class BinaryReaderStrip : public BinaryReaderNop {
   Result BeginCustomSection(Index section_index,
                             Offset size,
                             std::string_view section_name) override {
-    if (sections_to_keep_.count(section_name) > 0) {
+    if (!sections_to_remove_.empty() &&
```
sbc100, line 102: I don't think you need the first half of this condition.  Just `if (sections_to_remove_.count(section_name) > 0) {` should be fine.

sbc100 (approved):

src/tools/wasm-strip.cc
dzfrias: Ah yeah... not sure what got me to put that there in the first place...

sbc100 (approved):

Issue #1376, opened 2020-03-25
Custom sections wasm2wat / wat2wasm


syrusakbary:
We are researching ways to support custom sections in the text file format (wat) so we can transform it easily to wasm and viceversa.

One of the ideas we came up with is to automatically base64 encode the custom section when decoding from a wasm file, and automatically base64 decode when passing from wat to wasm.

Here's the proposal:

### wasm2wat

When doing wasm2wat, we can base64 encode the contents of each of the binary sections.

Proposal for wat:
```wasm
(module
  (section sectionName "base64:aGVsbG8=") ;; this is the base64 encoding of "hello"
  (func (export "addTwo") (param i32 i32) (result i32)
    local.get 0
    local.get 1
    i32.add))
```

### wat2wasm

When encoding the text format to wasm, we could just read the section value, see if it's using an encoding we support, and if that's the case we can just transform it into its binary format and append the section binary content to the wasm.

Are there other ways we could accomplish it?

Thoughts on this proposal?

binji:
This is one of the motivating use cases for the [annotations proposal](https://github.com/WebAssembly/annotations/blob/master/proposals/annotations/Overview.md#examples). In particular, [this gist](https://gist.github.com/binji/d1cfff7faaebb2aa4f8b1c995234e5a0) goes into some detail about how this could be supported.

syrusakbary:
Thanks for the prompt response, that makes sense.

Do you think we could support the following via annotations in wabt?

```
$ wasm2wat wasmwithcustomsections.wasm > wasmwithcustomsections.wat
```

`wasmwithcustomsections.wat` will have the following contents:

```wasm
(module
  (@section "sectionName" "base64:aGVsbG8=") ;; this is the base64 encoding of "hello"
  (func (export "addTwo") (param i32 i32) (result i32)
    local.get 0
    local.get 1
    i32.add))
```

------

```
$ wat2wasm wasmwithcustomsections.wat -o wasmwithcustomsections.wasm
```

`wasmwithcustomsections.wasm` will have the "sectionName" custom section with the content "hello"

binji:
I guess I don't see why this is preferable to the syntax proposed in the gist above?

```wast
(module
  (@custom "sectionName" (after function) "hello")
  ...
)
```

the `wat` format already has a way to specify non-printable bytes, if necessary: `"\xx"` where "x" is a hex digit.

syrusakbary:
[...]

PR #2284, opened 2023-08-21
Implement custom section reading/writing

dzfrias:
Draft PR for #1376.

Not fully finished yet, still needs a few things:

- wat2wasm support
- Handling of special custom sections that `BinaryReader` has callbacks for:
  - name
  - dylink0
  - dylink
  - reloc
  - target_features
  - linking
  - code_metadata
- Tests
- Review for consistency with rest of codebase style

I'm not super familiar with all the quirks/idioms of C++, and so I generally followed along with the rest of the codebase. That being said, I could've overlooked some important nuances in this initial draft.

Continuing my conversation with @keithw, I'm still not sure about the best way to handle the custom sections `BinaryReader` has callbacks for. Implementation wise, the backtracking method would be much easier. I go more in-depth on the topic in the second section of [this comment](https://github.com/WebAssembly/wabt/issues/1376#issuecomment-1686707242).


dzfrias:
I used the backtracking method for reading the special custom sections handled by `BinaryReader` already. Let me know if there are any problems with this implementation.

dzfrias:
I had to update [one of the test inputs](https://github.com/dzfrias/wabt/blob/d965ceb4a8c7119d9081d8498408d5250c1bbafa/test/parse/annotations.txt) to conform with the syntax of the `@custom` annotation.

Let me know if I missed something in this PR!

keithw:
This is looking really nice! Could you please add some roundtrip tests (including for custom sections that use a "place")? I'd also love to see a binary test that shows a custom "name" section for binary names that can't be represented as IDs in the text format, a la https://github.com/WebAssembly/spec/issues/617

It's sort of unfortunate that the annotations proposal itself is never going to supply us with tests for @custom annotations (for reasons discussed at https://github.com/WebAssembly/annotations/issues/15). And while I **wish** this were enough to let us round-trip an LLVM-generated relocatable object from binary to text and back to binary with the linking and relocation sections still valid, BinaryWriter doesn't necessarily write the binary exactly the same (e.g. it may drop the DataCount section). :-(

dzfrias:
> Could you please add some roundtrip tests (including for custom sections that use a "place")?

I'll get working on this! I might have to change a few things, because `WatWriter` writes each custom section at the end of the module, so a completely lossless roundtrip isn't possible with my current implementation.

> I'd also love to see a binary test that shows a custom "name" section for binary names that can't be represented as IDs in the text format, a la https://github.com/WebAssembly/spec/issues/617

Sounds good!

src/wast-parser.cc
```diff
@@ -762,6 +762,13 @@ Result WastParser::ErrorIfLpar(const std::vector<std::string>& expected,
 
 bool WastParser::ParseBindVarOpt(std::string* name) {
   WABT_TRACE(ParseBindVarOpt);
+  if (PeekIsAnnotation("name")) {
```
dzfrias, line 765: Is it okay that this is in `ParseBindVarOpt`? I'm not sure if the `@name` annotation is valid in every place that it's used.

keithw:
> On a second look at the other roundtrip tests, is it necessary to have a completely lossless conversion into the `Module`? 

I don't think this is required -- I just meant something like the other roundtrip tests, where it tracks the expected output (so we can tell if it changes in a future commit).

dzfrias:
I wrote a roundtrip test for custom sections. I also roundtripped the `@name` annotation. Let me know if there's anything else I should do!

dzfrias:
No problem; I reverted those commits. I'll open a PR for the `@name` stuff once this one is merged!

src/wast-parser.cc
```diff
@@ -1222,9 +1224,68 @@ Result WastParser::ParseScript(std::unique_ptr<Script>* out_script) {
...
+  Custom custom = Custom(loc, section_name, data);
+  module->customs.push_back(custom);
+
+  return Result::Ok;
+}
+
+bool WastParser::PeekIsCustom() {
+  // If IsLparAnn succeeds, tokens_.front() must have text, as it is an LparAnn
+  // token.
+  return IsLparAnn(PeekPair()) && tokens_.front().text() == "custom";
```
keithw, line 1279: Can we throw a `options_->features.annotations_enabled() &&` into this test for safety (and consistency with `PeekMatchRefType`)?

keithw (approved):
lgtm % comments

@sbc100 do you want to weigh in?

keithw:
Okay, merging -- thank you for contributing this! I'm already enjoying seeing the custom sections from LLVM-produced modules in the wasm2wat output.

PR #2285, opened 2023-08-23
Validator: normalize handling of Select in ExprVisitor Delegate

### LINKED ISSUE #2283: `wasm-validate` gives false positive errors that occur after a true positive.
Say we have the file repro.wat with an $invalid function, which doesn't match it's return signature, and a following valid function
with a select instruction:
```
(module
(func $invalid (result i32))
(func $normal (local i32)
i32.const 1
i32.const 2
i32.const 3
select
local.set 0
)
)
``` 
Trying to build it...
```
$ wat2wasm repro.wat
repro.wat:2:16: error: type mismatch in implicit return, expected [i32] but got []
(func $invalid (result i32))
               ^
repro.wat:8:1: error: type mismatch at end of function, expected [] but got [i32]
local.set 0
^^^^^^^^^
```
...gives an error in the valid function. Let's force the build with:
```
$ wat2wasm repro.wat --no-check
```
Now we have an invalid repro.wasm, and validating it...
```
$ wasm-validate repro.wasm
repro.wasm:000001b: error: type mismatch in implicit return, expected [i32] but got []
repro.wasm:000002a: error: type mismatch at end of function, expected [] but got [i32]
```
again gives an error in the valid function.
If you comment out the $invalid function:
```
(;(func $invalid (result i32));)
```
The remaining $normal function builds and validates successfully.

The errors seem to show up after `select` instructions, at the end of the function in version 1.0.33. In 1.0.27 it's a bit different, the errors seem to show up for the instruction immediately after the `select`. Keep in mind this bug that may account for some of the off-by-ones https://github.com/WebAssembly/wabt/issues/1217

Ideally, wasm-validate would be able to re-synchronise itself after errors like this, or abort immediately after an unrecoverable error (with a message saying as such and where validation stopped), so users don't think there are more issues than there actually are.

keithw:
Fixes #2283

Previously, the OnSelectExpr delegate would terminate validation if the SharedValidator found an error in the expression, or if the Validator had previously found an error at any point in validating the module. This commit normalizes the behavior to match how the Validator handles other expression types.

I'd be more comfortable with this if I had been able to replicate the bug that's hinted at by the TODO we're removing (https://github.com/WebAssembly/wabt/blob/main/src/validator.cc#L527-L532), but I haven't been able to. :-( @binji any ideas? (cc: @majaha)

### DIFF
--- a/src/validator.cc
+++ b/src/validator.cc
@@ -524,12 +524,7 @@ Result Validator::OnReturnCallIndirectExpr(ReturnCallIndirectExpr* expr) {
 Result Validator::OnSelectExpr(SelectExpr* expr) {
   result_ |= validator_.OnSelect(expr->loc, expr->result_type.size(),
                                  expr->result_type.data());
-  // TODO: Existing behavior fails when select fails.
-#if 0
   return Result::Ok;
-#else
-  return result_;
-#endif
 }
 
 Result Validator::OnStoreExpr(StoreExpr* expr) {
--- a/test/parse/expr/bad-select-multi.txt
+++ b/test/parse/expr/bad-select-multi.txt
@@ -14,7 +14,4 @@
 out/test/parse/expr/bad-select-multi.txt:10:5: error: invalid arity in select instruction: 2.
     select (result i32 i32)
     ^^^^^^
-out/test/parse/expr/bad-select-multi.txt:11:5: error: type mismatch at end of function, expected [] but got [... i32, i32, i32, i32]
-    unreachable
-    ^^^^^^^^^^^
 ;;; STDERR ;;)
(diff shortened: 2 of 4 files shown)


sbc100 (approved):
I think if the tests pass we can just land this.

binji:
> I'd be more comfortable with this if I had been able to replicate the bug that's hinted at by the TODO we're removing (https://github.com/WebAssembly/wabt/blob/main/src/validator.cc#L527-L532), but I haven't been able to. :-( @binji any ideas?

It looks like I was trying to keep the existing behavior during some code cleanup? I probably intended to fix it later but never did :-/

Anyway, I think what you've done here is correct

PR #2286, opened 2023-08-25
wasm2c: consolidate pack/unpack of tags, multi-value results (NFC)

keithw:
Split out from #2272 (tail calls).

This PR consolidates the machinery to spill and unspill groups of params and results to/from a struct, and uses this to handle tags and multi-value results in the same way. In the tail-call PR, we'll use the same machinery to pack/unpack the params and results from a tail-callee function.

This PR also pulls out a function to unwind the try-catch stack to a given label. We currently use this in one place (writing a GotoLabel). In the tail-call PR, we'll also use it in the return_call and return_call_indirect instructions.

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -259,7 +259,6 @@ class CWriter {
   static constexpr char MangleType(Type);
   static constexpr char MangleField(ModuleFieldType);
   static std::string MangleMultivalueTypes(const TypeVector&);
-  static std::string MangleTagTypes(const TypeVector&);
   static std::string Mangle(std::string_view name, bool double_underscores);
   static std::string MangleName(std::string_view);
   static std::string MangleModuleName(std::string_view);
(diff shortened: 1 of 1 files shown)


src/c-writer.cc
```diff
@@ -2614,6 +2596,43 @@ bool CWriter::IsImport(const std::string& name) const {
   return import_module_sym_map_.count(name);
 }
 
+template <typename sources>
+void CWriter::Spill(const TypeVector& types, bool ptr, const sources& src) {
+  for (Index i = 0; i < types.size(); ++i) {
+    Write("tmp", ptr ? "->" : ".");
+    Writef("%c%d = ", MangleType(types.at(i)), i);
+    Write(src(i), ";", Newline());
```
sbc100, line 2604: So these always spill and unspill to a struct `tmp` that is assumed that have the correct members.

The mangling of the type seem unnecessary but maybe makes things easier to debug/read?   i.e  just is the index `i` should be enough to make each field unique?

sbc100 (approved):
What do you think about using `wasm2c:` as the PR title prefix rather than `c-writer.cc:`?  (IMHO that make it more obvious what this PR is about).

lgtm, seems like to nice refactoring

src/c-writer.cc
keithw: Hmm, good point. This is copied from the current code for multi-value returns and tags, but I think you're right that we could avoid the mangled types and call the fields just "val0" "val1" "val2" or something. (We need **some** alpha prefix since I don't think we can name a field literally just "0" or "1".) Anyway I guess let's keep this as-is for now and we can strip it down in a future change if anybody wants.

PR #2287, opened 2023-08-26
Update testsuite

### LINKED ISSUE #2201: Is `global.get` allowed in element segments?
As far as I can tell from the webassembly spec, it should be valid to use the `global.get` instruction in an element segment.

However I am unable to use `wat2wasm` on this module

```wat
(module
  (import "m" "f" (global funcref))
  (table 0 10 funcref)
  (elem 0 (offset (i32.const 0)) funcref (global.get 0)))
```

If I try, I get this error:

```
foo.wat:4:43: error: invalid elem expression expression; must be either ref.null or ref.func.
  (elem 0 (offset (i32.const 0)) funcref (global.get 0)))
                                          ^^^^^^^^^^
```

If I use the `--no-check` flag I am able to override the check and generate a resulting `.wasm` file.

Looking at the generated file in a hexdump, it looks exactly like I would expect it to. However I am unable to get `wasm2wat` to accept this generated module, even if I use the `--no-check` flag.

Am I misreading the spec, or is this a bug in wabt?

keithw:
Sequenced behind #2272.

This updates the testsuite. The major changes:

- The exception-handling tests now require tail-calls (tbd in #2272).
- There's a new test (memory_copy1 from multi-memory), which required #2294 (now merged).
- The elem tests now use global.get in an elem expr, which required #2288 (now merged).

(The only missing test is relaxed-simd/relaxed_dot_product.txt, which it looks like we believe is going away per #2115.)

### DIFF
--- a/test/spec/binary-leb128.txt
+++ b/test/spec/binary-leb128.txt
@@ -83,37 +83,39 @@ out/test/spec/binary-leb128.wast:718: assert_malformed passed:
   0000014: error: unable to read u32 leb128: function body count
 out/test/spec/binary-leb128.wast:731: assert_malformed passed:
   0000022: error: unable to read u32 leb128: load offset
-out/test/spec/binary-leb128.wast:750: assert_malformed passed:
+out/test/spec/binary-leb128.wast:751: assert_malformed passed:
   0000022: error: unable to read u32 leb128: load offset
-out/test/spec/binary-leb128.wast:769: assert_malformed passed:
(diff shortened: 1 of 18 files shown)


keithw:
Reverting to draft as I think there may be a cleaner way to do the global.get change, and in any event I'll try to split this one up.

sbc100 (approved):
Rubberstamp lgtm

PR #2288, opened 2023-08-30
Share reading/validation code between elem exprs & other const exprs

### LINKED ISSUE #2201: Is `global.get` allowed in element segments?

keithw:
This continues the work from #1783 and reduces special handling of elem exprs, by treating them the same as other const expressions (init expressions).

Fixes #2201 (now allows global.get in an elem expr)

This is a prerequisite for updating the testsuite (spun out from #2287). It seemed cleaner to do it this way than to add a `OnElemSegmentElemExpr_GlobalGet` to all the readers and validators, given that the validation rules are the same as when global.get appears in any other const expression.


src/binary-reader-objdump.cc
```diff
@@ -1711,14 +1704,22 @@ Result BinaryReaderObjdump::BeginGlobal(Index index, Type type, bool mutable_) {
...
-                                        bool as_unsigned) {
-  assert(expr.insts.size() > 0);
+                                        bool as_unsigned,
+                                        bool with_prefix) {
+  if (with_prefix) {
+    PrintDetails(" - init ");
+  }
+
+  if (expr.insts.empty()) {
+    PrintDetails("malformed\n");
```
sbc100, line 1714: Do we do this kind of things elsewhere in this file?   How about "<MALFORMED>" to be extra clear?

sbc100 (approved):
Nice!  

I've long thought this would be good idea but wan't sure how to make it happen.  Thanks.

src/binary-reader-objdump.cc
keithw: Sounds good. I went with `<EMPTY>` to be safe -- on further review I don't think an empty const expression is malformed, just invalid.

PR #2292, opened 2023-09-08
[wasm2c] Use wasm_rt_memcpy everywhere in generated code

sbc100:


matthias-blume (approved):

matthias-blume (approved):

keithw (approved):
lgtm

(In general, my understanding is that GCC and clang already replace `memcpy` with `__builtin_memcpy` unless `-fno-builtins` is given on the command line, so I'm not totally sure if `wasm_rt_memcpy` is necessary -- maybe in case the embedder wants to use a different memcpy implementation for some reason? But agreed we should be consistent.)

sbc100:
> lgtm
> 
> (In general, my understanding is that GCC and clang already replace `memcpy` with `__builtin_memcpy` unless `-fno-builtins` is given on the command line, so I'm not totally sure if `wasm_rt_memcpy` is necessary -- maybe in case the embedder wants to use a different memcpy implementation for some reason? But agreed we should be consistent.)

Indeed, this desire for consistency is coming from https://github.com/AndroidWasm/wasm_ndk.  Seems fairly reasonable.

sbc100:
The CI failures I can repro locally, when I use gcc 12.2.0 on my local machine.  Not sure why they are not showing up on main?   I don't think they relate to this PR.. but trying to figure out why we don't see them on all PRs

keithw:
Looks like GCC is seeing the `__builtin_memcpy` that would OOB the data segment array, but not able to see that it traps in the previous line and the bad `__buildin_memcpy` is unreachable. (And I guess with plain `memcpy`, it doesn't have as much visibility to see that it's an OOB array access.)

These looks similar:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56456

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104165

We may have to add `-Wno-error=array-bounds` to the `args` array in `run-spec-wasm2c.py`...

PR #2293, opened 2023-09-08
[wasm2c] Add '-Wno-array-bounds' when building wasm2c output

sbc100:
We are seeing some (spurious?) warning from gcc 12.2.

I've been seeing them locally, but they started to show up in CI as part of #2292.


keithw (approved):
lgtm

PR #2296, opened 2023-09-15
c-writer.cc: clean up some handling of typevector structs (NFC)

keithw:
Split out from #2272 per https://github.com/WebAssembly/wabt/pull/2272#discussion_r1326166696. We had been using "Multivalue" for all the typevector structs (both true "multi-value" result types, but also tag types, and then #2272 had used this term to refer to structs that hold the params to a tail-called function).

This PR makes it so we only use "multivalue" to refer to a multi-value result type, not a tag type.

It also deduplicates the typevector struct declarations in the output.

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -258,7 +258,7 @@ class CWriter {
 
   static constexpr char MangleType(Type);
   static constexpr char MangleField(ModuleFieldType);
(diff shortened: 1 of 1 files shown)


sbc100 (approved):
Nice!  Thanks Keith.

PR #2297, opened 2023-09-16
write invalid names using `@name` annotation

dzfrias:
Following up #2284, a few commits were reverted related to the `@name` annotation so that the PR would focus on a single addition. This PR makes use of the `@name` annotation when writing names containing invalid characters in the text format (see WebAssembly/spec#617).

Let me know if there's anything else I should add!

### DIFF
--- a/src/wast-parser.cc
+++ b/src/wast-parser.cc
@@ -608,7 +608,7 @@ TokenType WastParser::Peek(size_t n) {
       }
       if ((options_->features.code_metadata_enabled() &&
            cur.text().find("metadata.code.") == 0) ||
-          cur.text() == "custom") {
+          cur.text() == "custom" || cur.text() == "name") {
         tokens_.push_back(cur);
         continue;
       }
@@ -762,6 +762,13 @@ Result WastParser::ErrorIfLpar(const std::vector<std::string>& expected,
 
 bool WastParser::ParseBindVarOpt(std::string* name) {
   WABT_TRACE(ParseBindVarOpt);
+  if (PeekIsAnnotation("name")) {
+    // Should be the @name
+    Consume();
+    CHECK_RESULT(ParseQuotedText(name));
+    EXPECT(Rpar);
+    return true;
+  }
   if (!PeekMatch(TokenType::Var)) {
     return false;
   }
@@ -1174,7 +1181,7 @@ Result WastParser::ParseModule(std::unique_ptr<Module>* out_module) {
       auto module_command = cast<ScriptModuleCommand>(std::move(command));
       *module = std::move(module_command->module);
     }
-  } else if (IsModuleField(PeekPair()) || PeekIsCustom()) {
+  } else if (IsModuleField(PeekPair()) || PeekIsAnnotation("custom")) {
     // Parse an inline module (i.e. one with no surrounding (module)).
     CHECK_RESULT(ParseModuleFieldList(module.get()));
   } else if (PeekMatch(TokenType::Eof)) {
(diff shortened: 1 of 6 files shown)


src/wast-parser.cc
```diff
@@ -608,7 +608,7 @@ TokenType WastParser::Peek(size_t n) {
       }
       if ((options_->features.code_metadata_enabled() &&
```
keithw, line 609: Seems like maybe there's enough conditions here that they would benefit by getting broken out into a separate function at this point?

test/roundtrip/invalid-name.txt
```diff
@@ -0,0 +1,13 @@
+;;; TOOL: run-roundtrip
+;;; ARGS: --stdout --no-check --enable-annotations --debug-names
```
keithw, line 2: Why is `--no-check` used here -- what makes the module invalid?

test/roundtrip/invalid-name.txt
```diff
@@ -0,0 +1,13 @@
...
+(module
+  (type (;0;) (func (result i32)))
+  (func (@name "$hi hello hey") (type 0) (result i32)
+    i32.const 42))
+(;; STDOUT ;;;
+(module
+  (type (;0;) (func (result i32)))
+  (func (@name "$hi hello hey") (type 0) (result i32)
+    i32.const 42)
+  (@custom "name" "\01\0f\01\00\0chi hello hey\02\03\01\00\00"))
```
keithw, line 12: Do we want to be printing both the `@name` annotation and also the `@custom "name"` annotation with the same info? (Not 100% sure what the right behavior is, but we do suppress the `@custom` annotation for most other cases that are already handled elsewhere...)

keithw (commented):
Looks pretty straightforward. Could you please also add a few more tests, e.g. a roundtrip test for the example given in https://github.com/WebAssembly/annotations/blob/main/proposals/annotations/Overview.md (under "Expressing names")?

Also: what can be named here? Can this name a branch (something that WABT doesn't normally support)? I'm wondering if this can trick wasm2c into having a branch named "Bfunc", which would be a problem.

test/roundtrip/invalid-name.txt
dzfrias: Hmm, this is the current behavior of names in general. Even without the `@name` annotation, both are included. I'll work on suppressing this, though.

test/roundtrip/invalid-name.txt
dzfrias: Ah, I think I must've left this in when looking at another roundtrip test for reference. Thanks for the catch!

src/wast-parser.cc
dzfrias: Agreed. Honestly, I was a bit surprised to see this logic in the `Peek` method, too.  Perhaps it would be better to put this all in it's own method anyway, with a more straightforward way of parsing/skipping certain annotations.

dzfrias:
> Could you please also add a few more tests, e.g. a roundtrip test for the example given in https://github.com/WebAssembly/annotations/blob/main/proposals/annotations/Overview.md (under "Expressing names")?

Looking at this test, it seems I misinterpreted the use of `@name`. In the proposal example, parts of the WebAssembly module have two name: one for use in the WAT representation, and one for use the binary. This initial implementation just assumed that there would be one name for both. I'll work on implementing the intended version.

dzfrias:
Upon working on the implementation of the work in my above comment, it seems like there are many possible places one could use the `@name` annotation. Is there a well-defined collection of all these places? For example, would it work in module imports?

keithw:
You raise some good questions but I definitely don't have a view on what the behavior is supposed to be. I guess we should just match other implementations (if there are other implementations -- maybe this is too soon?). It's a little unfortunate that these annotation specifications aren't going to get tests. :-(

dzfrias:
I'll wait until WebAssembly/annotations#21 is resolved. If the `$"name"` syntax is approved, then most of these problems shouldn't be relevant anymore.

PR #2299, opened 2023-09-17
Remove intrusive_list (NFC)

keithw:
This PR replaces the `intrusive_list` class with a std::list of unique_ptr's, trying to keep everything else mostly the same.


include/wabt/intrusive-list.h
```diff
@@ -1,631 +0,0 @@
...
-// features:
-//
-// * Add "extract_" functions that remove an element from the list and return
-//   it.
-// * Only supports move-only operations
-// * No allocator support
-// * No initializer lists
-// * Asserts instead of exceptions
-// * Some functions are not implemented (merge, remove, remove_if, reverse,
-//   unique, sort, non-member comparison operators)
```
sbc100, line 34: Do you know the rationale for creating this class in the first place?  Does the reasoning for it no longer hold?

keithw: It's not super clear (https://github.com/WebAssembly/wabt/pull/542), but generally the reason to prefer an intrusive list vs. `std::list` is about performance (an intrusive_list only needs one allocation to make a list node with both the contained object and the prev/next pointers). And indeed, I just benchmarked this PR (with `wasm-validate` [clang.wasm](https://github.com/turbolent/w2c2/blob/main/examples/clang/clang.wasm)) and `std::list` is a significant slowdown for large modules. :-(

main branch:
```
time ./bin/wasm-validate ~/w2c2/examples/clang/clang.wasm

real	0m1.887s
user	0m1.451s
sys	0m0.436s
```

this PR:
```
time ./bin/wasm-validate ~/w2c2/examples/clang/clang.wasm

real	0m2.967s
user	0m2.531s
sys	0m0.436s
```

I wasn't expecting it to be this bad -- let's hold off on this for now until I find a better way.

binji: I believe I originally switched from a vector to an intrusive list, since I thought we'd want to be able to do insertions into the list. I'm not sure that was a good decision, tbh. It may make sense to see whether switching it back to a vector would work.

keithw: ModuleFieldList looks very easy to switch back to a vector (and it's a meaningful performance improvement -- down to 2.5 seconds total), but ExprList is going to be more involved since the WastParser currently wants to splice it. :-/

keithw: Well, here's a version with std::vector for the ExprList and std::deque for the ModuleFieldList (since we sometimes iterate over the ModuleFieldList while also appending to it...). Still unfortunately a performance regression:

```
$ time ./bin/wasm-validate ~/w2c2/examples/clang/clang.wasm 

real	0m2.456s
user	0m1.972s
sys	0m0.484s
```

PR #2300, opened 2023-09-17
Remove CircularArray (NFC)

keithw:
This PR removes the CircularArray class (only used in one place) and its unit test, in favor of a small replacement inside the WastParser.

I suspect we could also just use a std::queue with probably only a small performance penalty -- maybe that's cleaner/simpler?


include/wabt/wast-parser.h
```diff
@@ -262,7 +262,28 @@ class WastParser {
   Errors* errors_;
   WastParseOptions* options_;
 
-  CircularArray<Token, 2> tokens_;
+  // two-element queue of upcoming tokens
```
sbc100, line 265: I'd still rather give this data type a name and define it out-of-line, if only to make the containing class easier to read.

include/wabt/wast-parser.h
```diff
@@ -262,7 +262,28 @@ class WastParser {
...
+    }
+    void pop_front() {
+      assert(tokens[i]);
+      tokens[i].reset();
+      i = !i;
+    }
+    const Token& at(bool n) const { return tokens[i ^ n].value(); }
+    const Token& front() const { return at(0); }
+    bool empty() const { return !tokens[i]; }
+    size_t size() const { return empty() ? 0 : 1 + tokens[!i].has_value(); }
```
sbc100, line 285: Are all these methods actually used in this file?

sbc100 (approved):

include/wabt/wast-parser.h
sbc100: Maybe `TokenQueue`?

keithw: Done.

include/wabt/wast-parser.h
keithw: No... they're just used in wast-parser.cc. Moved the definitions to the .cc file.

sbc100 (approved):

PR #2308, opened 2023-10-02
wasm2c: atomic and shared mem operations using c11

shravanrn:
Implement wasm2c atomic ops and shared memory ops with C11 atomics as discussed in https://github.com/WebAssembly/wabt/pull/2268#issuecomment-1662696781

This is the first of several PRs to fully support the various primitives in the threads/shared_memory/atomics proposal.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -423,9 +423,10 @@ if (WABT_INSTALL_RULES)
 endif ()
 
 if (HAVE_SETJMP_H)
-  set(WASM_RT_FILES "wasm2c/wasm-rt-impl.h" "wasm2c/wasm-rt-impl.c" "wasm2c/wasm-rt-exceptions-impl.c")
+  set(WASM_RT_FILES "wasm2c/wasm-rt-impl.h" "wasm2c/wasm-rt-impl.c" "wasm2c/wasm-rt-exceptions-impl.c" "wasm2c/wasm-rt-mem-impl.c")
 
   add_library(wasm-rt-impl STATIC ${WASM_RT_FILES})
+  target_link_libraries(wasm-rt-impl ${CMAKE_THREAD_LIBS_INIT})
   add_library(wabt::wasm-rt-impl ALIAS wasm-rt-impl)
   if (WABT_BIG_ENDIAN)
     target_compile_definitions(wasm-rt-impl PUBLIC WABT_BIG_ENDIAN=1)
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -388,8 +388,8 @@ class CWriter {
   void WriteGlobal(const Global&, const std::string&);
   void WriteGlobalPtr(const Global&, const std::string&);
   void WriteMemories();
-  void WriteMemory(const std::string&);
-  void WriteMemoryPtr(const std::string&);
+  void WriteMemory(const std::string&, const Memory& memory);
+  void WriteMemoryPtr(const std::string&, const Memory& memory);
   void WriteTables();
   void WriteTable(const std::string&, const wabt::Type&);
   void WriteTablePtr(const std::string&, const Table&);
@@ -1328,6 +1328,15 @@ void CWriter::WriteGetFuncTypeDecl() {
         Newline());
 }
 
+static std::string GetMemoryTypeString(const Memory& memory) {
+  return memory.page_limits.is_shared ? "wasm_rt_shared_memory_t"
+                                      : "wasm_rt_memory_t";
+}
+
+static std::string GetMemoryAPIString(const Memory& memory, std::string api) {
+  return memory.page_limits.is_shared ? (api + "_shared") : api;
+}
+
 void CWriter::WriteInitExpr(const ExprList& expr_list) {
   if (expr_list.empty()) {
     WABT_UNREACHABLE;
@@ -1736,7 +1745,7 @@ void CWriter::BeginInstance() {
       }
 
       case ExternalKind::Memory: {
-        Write("wasm_rt_memory_t");
+        Write(GetMemoryTypeString(cast<MemoryImport>(import)->memory));
         break;
       }
 
@@ -1779,7 +1788,8 @@ void CWriter::BeginInstance() {
 
       case ExternalKind::Memory:
         WriteMemory(std::string("*") +
-                    ExportName(import->module_name, import->field_name));
+                        ExportName(import->module_name, import->field_name),
+                    cast<MemoryImport>(import)->memory);
         break;
 
       case ExternalKind::Table: {
@@ -2027,19 +2037,20 @@ void CWriter::WriteMemories() {
     bool is_import = memory_index < module_->num_memory_imports;
     if (!is_import) {
       WriteMemory(
-          DefineInstanceMemberName(ModuleFieldType::Memory, memory->name));
+          DefineInstanceMemberName(ModuleFieldType::Memory, memory->name),
+          *memory);
       Write(Newline());
     }
     ++memory_index;
   }
 }
 
-void CWriter::WriteMemory(const std::string& name) {
-  Write("wasm_rt_memory_t ", name, ";");
+void CWriter::WriteMemory(const std::string& name, const Memory& memory) {
+  Write(GetMemoryTypeString(memory), " ", name, ";");
 }
 
-void CWriter::WriteMemoryPtr(const std::string& name) {
-  Write("wasm_rt_memory_t* ", name, "(", ModuleInstanceTypeName(),
+void CWriter::WriteMemoryPtr(const std::string& name, const Memory& memory) {
+  Write(GetMemoryTypeString(memory), "* ", name, "(", ModuleInstanceTypeName(),
         "* instance)");
 }
 
@@ -2169,7 +2180,8 @@ void CWriter::WriteDataInitializers() {
         max = memory->page_limits.is_64 ? (static_cast<uint64_t>(1) << 48)
                                         : 65536;
       }
-      Write("wasm_rt_allocate_memory(",
+      std::string func = GetMemoryAPIString(*memory, "wasm_rt_allocate_memory");
+      Write(func, "(",
             ExternalInstancePtr(ModuleFieldType::Memory, memory->name), ", ",
             memory->page_limits.initial, ", ", max, ", ",
             memory->page_limits.is_64, ");", Newline());
@@ -2444,7 +2456,7 @@ void CWriter::WriteExports(CWriterPhase kind) {
       case ExternalKind::Memory: {
         const Memory* memory = module_->GetMemory(export_->var);
         internal_name = memory->name;
-        WriteMemoryPtr(mangled_name);
+        WriteMemoryPtr(mangled_name, *memory);
         break;
       }
 
@@ -2754,7 +2766,8 @@ void CWriter::WriteFree() {
     for (const Memory* memory : module_->memories) {
       bool is_import = memory_index < module_->num_memory_imports;
       if (!is_import) {
-        Write("wasm_rt_free_memory(",
+        std::string func = GetMemoryAPIString(*memory, "wasm_rt_free_memory");
+        Write(func, "(",
               ExternalInstancePtr(ModuleFieldType::Memory, memory->name), ");",
               Newline());
       }
@@ -3706,7 +3719,8 @@ void CWriter::Write(const ExprList& exprs) {
         Memory* memory = module_->memories[module_->GetMemoryIndex(
             cast<MemoryGrowExpr>(&expr)->memidx)];
 
-        Write(StackVar(0), " = wasm_rt_grow_memory(",
+        std::string func = GetMemoryAPIString(*memory, "wasm_rt_grow_memory");
+        Write(StackVar(0), " = ", func, "(",
               ExternalInstancePtr(ModuleFieldType::Memory, memory->name), ", ",
               StackVar(0), ");", Newline());
         break;
@@ -4923,7 +4937,7 @@ void CWriter::Write(const ConvertExpr& expr) {
 }
 
 void CWriter::Write(const LoadExpr& expr) {
-  const char* func = nullptr;
+  std::string func;
   // clang-format off
   switch (expr.opcode) {
     case Opcode::I32Load: func = "i32_load"; break;
@@ -4954,6 +4968,7 @@ void CWriter::Write(const LoadExpr& expr) {
   // clang-format on
 
   Memory* memory = module_->memories[module_->GetMemoryIndex(expr.memidx)];
+  func = GetMemoryAPIString(*memory, func);
 
   Type result_type = expr.opcode.GetResultType();
   Write(StackVar(0, result_type), " = ", func, "(",
@@ -4967,7 +4982,7 @@ void CWriter::Write(const LoadExpr& expr) {
 }
 
 void CWriter::Write(const StoreExpr& expr) {
-  const char* func = nullptr;
+  std::string func;
   // clang-format off
   switch (expr.opcode) {
     case Opcode::I32Store: func = "i32_store"; break;
@@ -4987,6 +5002,7 @@ void CWriter::Write(const StoreExpr& expr) {
   // clang-format on
 
   Memory* memory = module_->memories[module_->GetMemoryIndex(expr.memidx)];
+  func = GetMemoryAPIString(*memory, func);
 
   Write(func, "(", ExternalInstancePtr(ModuleFieldType::Memory, memory->name),
         ", (u64)(", StackVar(1), ")");
@@ -5560,7 +5576,7 @@ void CWriter::Write(const LoadZeroExpr& expr) {
 }
 
 void CWriter::Write(const AtomicLoadExpr& expr) {
-  const char* func = nullptr;
+  std::string func;
   // clang-format off
   switch (expr.opcode) {
     case Opcode::I32AtomicLoad: func = "i32_atomic_load"; break;
@@ -5577,6 +5593,7 @@ void CWriter::Write(const AtomicLoadExpr& expr) {
   // clang-format on
 
   Memory* memory = module_->memories[module_->GetMemoryIndex(expr.memidx)];
+  func = GetMemoryAPIString(*memory, func);
 
   Type result_type = expr.opcode.GetResultType();
   Write(StackVar(0, result_type), " = ", func, "(",
@@ -5590,7 +5607,7 @@ void CWriter::Write(const AtomicLoadExpr& expr) {
 }
 
 void CWriter::Write(const AtomicStoreExpr& expr) {
-  const char* func = nullptr;
+  std::string func;
   // clang-format off
   switch (expr.opcode) {
     case Opcode::I32AtomicStore: func = "i32_atomic_store"; break;
@@ -5607,6 +5624,7 @@ void CWriter::Write(const AtomicStoreExpr& expr) {
   // clang-format on
 
   Memory* memory = module_->memories[module_->GetMemoryIndex(expr.memidx)];
+  func = GetMemoryAPIString(*memory, func);
 
   Write(func, "(", ExternalInstancePtr(ModuleFieldType::Memory, memory->name),
         ", (u64)(", StackVar(1), ")");
@@ -5617,7 +5635,7 @@ void CWriter::Write(const AtomicStoreExpr& expr) {
 }
 
 void CWriter::Write(const AtomicRmwExpr& expr) {
-  const char* func = nullptr;
+  std::string func;
   // clang-format off
   switch (expr.opcode) {
     case Opcode::I32AtomicRmwAdd: func = "i32_atomic_rmw_add"; break;
@@ -5668,6 +5686,8 @@ void CWriter::Write(const AtomicRmwExpr& expr) {
   // clang-format on
 
   Memory* memory = module_->memories[module_->GetMemoryIndex(expr.memidx)];
+  func = GetMemoryAPIString(*memory, func);
+
   Type result_type = expr.opcode.GetResultType();
 
   Write(StackVar(1, result_type), " = ", func, "(",
@@ -5681,7 +5701,7 @@ void CWriter::Write(const AtomicRmwExpr& expr) {
 }
 
 void CWriter::Write(const AtomicRmwCmpxchgExpr& expr) {
-  const char* func = nullptr;
+  std::string func;
   // clang-format off
   switch(expr.opcode) {
     case Opcode::I32AtomicRmwCmpxchg: func = "i32_atomic_rmw_cmpxchg"; break;
@@ -5697,6 +5717,8 @@ void CWriter::Write(const AtomicRmwCmpxchgExpr& expr) {
   // clang-format on
 
   Memory* memory = module_->memories[module_->GetMemoryIndex(expr.memidx)];
+  func = GetMemoryAPIString(*memory, func);
+
   Type result_type = expr.opcode.GetResultType();
 
   Write(StackVar(2, result_type), " = ", func, "(",
--- a/src/template/wasm2c_atomicops.declarations.c
+++ b/src/template/wasm2c_atomicops.declarations.c
@@ -1,123 +1,11 @@
-#if defined(_MSC_VER)
+#include <stdatomic.h>
 
-#include <intrin.h>
+#if WABT_BIG_ENDIAN
+#error "wasm2c atomics not supported on big endian"
+#endif
 
-// Use MSVC intrinsics
-
-// For loads and stores, its not clear if we can rely on register width loads
-// and stores to be atomic as reported here
-// https://learn.microsoft.com/en-us/windows/win32/sync/interlocked-variable-access?redirectedfrom=MSDN
-// or if we have to reuse other instrinsics
-// https://stackoverflow.com/questions/42660091/atomic-load-in-c-with-msvc
-// We reuse other intrinsics to be cautious
-#define atomic_load_u8(a) _InterlockedOr8(a, 0)
-#define atomic_load_u16(a) _InterlockedOr16(a, 0)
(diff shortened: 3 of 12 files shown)


shravanrn:
@keithw @sbc100 any chance you could have a look at this? This is blocking for the next patch for wait/notify

keithw:
Sure, can we finish tail calls and then land this?

shravanrn:
> Sure, can we finish tail calls and then land this?

@keithw I'm hoping we can avoid serializing on the PRs as these two features shouldn't affect each other. This PR is mostly straightforward (replace compiler intrinsics with c11 primitives and introduce a new type for shared memory).

I suspect there won't really be any serious merge conflicts introduced, but if you have any concerns here, I think we can still start with the review process and figure out the landing separately?

Does this seem reasonable?

wasm2c/wasm-rt-impl-mem.c
```diff
@@ -0,0 +1,165 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// This file is not meant to be used directly, and is only used by including it
+// in wasm-rt-impl.c. This file may be included multiple times to implement
+// memory operations on different memory tyeps --- wasm_rt_memory_t and
+// wasm_rt_memory_shared_t depending on the value of WASM_RT_SHARED_MEM_OPS.
+
+#include <wasm-rt-impl.h>
```
keithw, line 22: I think with `#include "wasm-rt-impl.h"` (quotes instead of angle brackets) we can avoid having to add the `target_include_directories` line to CMakeLists.txt.

wasm2c/wasm-rt-impl-mem.c
```diff
@@ -0,0 +1,165 @@
...
+
+#ifndef WASM_RT_SHARED_MEM_OPS
```
keithw, line 24: I like how this shares code between the non-shared and shared versions of allocate/grow/free memory, but the multiply-included `.c` file seems a bit confusing. How do you feel about having the .c file use the preprocessor macros directly to define the non-shared operations and then conditionally the shared ones, i.e. the same strategy that wasm-rt-impl.c uses for `DEFINE_TABLE_OPS(funcref)` and `DEFINE_TABLE_OPS(externref)`?

wasm2c/wasm-rt.h
```diff
@@ -403,6 +432,26 @@ uint64_t wasm_rt_grow_memory(wasm_rt_memory_t*, uint64_t pages);
  */
 void wasm_rt_free_memory(wasm_rt_memory_t*);
 
+#ifdef WASM_RT_C11_AVAILABLE
+/**
+ * Shared memory version of wasm_rt_allocate_memory
+ */
+void wasm_rt_allocate_shared_memory(wasm_rt_shared_memory_t*,
```
keithw, line 439: Document the new runtime API functions in wasm2c/README.md?

keithw (commented):
I'm definitely no shared-memory expert, but this looks pretty reasonable % comments.

wasm2c/wasm-rt-impl-mem.c
shravanrn: At a high level I don't mind either approach. In practice, the challenge with pre-processor macros is that they aren't easy to work with while debugging --- single step is generally broken when dealing with pre-processor macros. The nice bit about this double include approach is that single step debugging works flawlessly without an issues or extra setup. Given this my first inclination would be to leave it as is, but I am not wedded to the idea. Let me know your thoughts.

test/wasm2c/spec/threads/atomic.txt
```diff
@@ -1,6 +1,8 @@
 ;;; TOOL: run-spec-wasm2c
 ;;; ARGS*: --enable-threads
-;; atomic operations --- This is a subset of the full test suite, that is currently supported by wasm2c
+;; atomic operations --- This is a subset of the full test suite taken from the
+;;   threads proposal repo, from the upstream-rebuild branch (as the test suite
+;;   repo is out of date).
```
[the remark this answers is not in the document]
shravanrn: Unfortunately not yet. The upstream main branch currently has broken results for these new tests and is not ready for use yet. We can switch this over once upstream-rebuild is merged into main. I believe @conrad-watt is working on this merge.

test/run-spec-wasm2c.py
```diff
@@ -466,7 +466,7 @@ def Compile(cc, c_filename, out_dir, *cflags):
                  '-Wno-infinite-recursion',
                  '-fno-optimize-sibling-calls',
                  '-frounding-math', '-fsignaling-nans',
-                 '-std=c99', '-D_DEFAULT_SOURCE']
+                 '-std=c11', '-D_DEFAULT_SOURCE']
```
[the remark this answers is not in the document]
shravanrn: On thinking further, rather than a complete switch to c11, maybe we could keep the code working for c99 in general, but use c11 in tests that need threads/atomics.

 @keithw I think for this, my plan forward is to try to update the test infrastructure to use c11 only when specified as a flag in the test. Does this make sense as path forward?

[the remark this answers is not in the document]
shravanrn: Does this + the new docs address this sufficiently?

wasm2c/wasm-rt.h
shravanrn: Done

wasm2c/README.md
```diff
@@ -327,6 +349,23 @@ arguments and returning `void` . e.g.
 
 `wasm_rt_free_memory` frees the memory instance.
 
+`wasm_rt_allocate_memory_shared` initializes a memory instance that can be
+shared by different Wasm instances, and allocates at least enough space for the
```
keithw, line 353: Should this be "shared by different Wasm threads" (or "agents") rather than "instances"? Even a normal memory can be shared by different module instances.

wasm2c/README.md
```diff
@@ -327,6 +349,23 @@ arguments and returning `void` . e.g.
...
+given number of initial pages. The memory must be cleared to zero. The `is64`
+parameter indicates if the memory is indexed with an i32 or i64 address.
+
+`wasm_rt_grow_memory_shared` must grow the given shared memory instance by the
```
keithw, line 357: Can this reference the (very similar) description of `wasm_rt_grow_memory` instead of duplicating the text?

wasm2c/wasm-rt-impl-mem.c
keithw: I guess my preference would be to keep the style consistent (e.g. with what we do for tables). I guess if you want this single-steppable in a debugger (but to reuse the shared part of these routines in the human-written source code), we could think about using a code generator or something?

src/template/wasm2c_atomicops.declarations.c
```diff
@@ -1,139 +1,77 @@
...
-#define atomic_compare_exchange_u16(a, expected_ptr, desired) __atomic_compare_exchange_helper((u16*)(a), expected_ptr, desired)
-#define atomic_compare_exchange_u32(a, expected_ptr, desired) __atomic_compare_exchange_helper((u32*)(a), expected_ptr, desired)
-#define atomic_compare_exchange_u64(a, expected_ptr, desired) __atomic_compare_exchange_helper((u64*)(a), expected_ptr, desired)
-// clang-format on
-
-#define atomic_fence() __atomic_thread_fence(__ATOMIC_SEQ_CST)
+#include <stdatomic.h>
 
+#if WABT_BIG_ENDIAN
+#error "wasm2c atomics not supported on big endian"
```
SoniEx2, line 4: oh, really? :(

src/template/wasm2c_atomicops.declarations.c
```diff
@@ -160,73 +105,115 @@ DEFINE_ATOMIC_STORE(i64_atomic_store8, u8, u64)
...
-    return (t2)ret;                                                  \
+#define DEFINE_ATOMIC_RMW(name, opname, op, t1, t2)                            \
+  static inline t2 name(wasm_rt_memory_t* mem, u64 addr, t2 value) {           \
+    MEMCHECK(mem, addr, t1);                                                   \
+    ATOMIC_ALIGNMENT_CHECK(addr, t1);                                          \
+    t1 wrapped = (t1)value;                                                    \
+    t1 ret;                                                                    \
+    wasm_rt_memcpy(&ret, &mem->data[addr], sizeof(t1));                        \
+    ret = ret op wrapped;                                                      \
+    return (t2)ret;                                                            \
```
SoniEx2, line 116: hmm, shouldn't this write back to the memory...?

wasm2c/wasm-rt-impl-mem.c
```diff
@@ -0,0 +1,157 @@
...
+  uint64_t new_size = new_pages * PAGE_SIZE;
+  uint64_t delta_size = delta * PAGE_SIZE;
+#if WASM_RT_USE_MMAP
+  MEMORY_DATA_TYPE new_data = memory->data;
+  int ret = os_mprotect((void*)(new_data + old_size), delta_size);
+  if (ret != 0) {
+    return (uint64_t)-1;
+  }
+#else
+  MEMORY_DATA_TYPE new_data = realloc((void*)memory->data, new_size);
```
SoniEx2, line 115: realloc is potentially unsound for shared memories, if multiple threads are active, the allocation could be moved while other threads are accessing it.

src/c-writer.cc
```diff
@@ -2026,7 +2037,8 @@ void CWriter::WriteDataInitializers() {
         max = memory->page_limits.is_64 ? (static_cast<uint64_t>(1) << 48)
                                         : 65536;
       }
-      Write("wasm_rt_allocate_memory(",
+      std::string func = GetMemoryAPIString(memory, "wasm_rt_allocate_memory");
+      Write(func, "(",
```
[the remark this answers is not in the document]
shravanrn: I think i personally find the current version more readable. I'll leave it like this for now, but please reopen if you feel strongly

src/template/wasm2c_atomicops.declarations.c
shravanrn: It's more that I have no way to test these. If you have a BIG_ENDIAN machine, and you can run the atomic tests, you could remove this error in a separate patch after this PR :)

src/template/wasm2c_atomicops.declarations.c
shravanrn: ooh, good catch! will fix.

wasm2c/wasm-rt-impl-mem.c
shravanrn: yup! I was initially planning on making these changes in a new PR along with wait/notify runtime support. However, on re-thinking it may make sense to ensure this PR is self contained. Will fix!

wasm2c/wasm-rt-impl-mem.c
shravanrn: Ok. I will work on modifying this to use a code-gen style. I'll probably try to follow a similar format as used in the `src/template` folder.

shravanrn:
@keithw @sbc100 Since we are waiting for the discussion on the other PRs, can we try landing this? If this looks good, could one of you sign off on this?

sbc100:
Sorry I didn't get to looking at this yet.  I will try to take a look in the morning (PST)

shravanrn:
@keithw @sbc100 Happy new year all! :) Just a ping to have a look when you get the chance

shravanrn:
> @keithw @sbc100 Happy new year all! :) Just a ping to have a look when you get the chance

@keithw @sbc100 Bumping this up

sbc100:
Apologies for taking so long to look at this one.   Will try to get to it soon/today.

wasm2c/wasm-rt.h
```diff
@@ -326,6 +354,31 @@ typedef struct {
   bool is64;
 } wasm_rt_memory_t;
 
+#ifdef WASM_RT_C11_AVAILABLE
+/** A shared Memory object. */
+typedef struct {
+  /** The linear memory data, with a byte length of `size`. The memory is marked
```
sbc100, line 360: Newline after `/**` for multiline comments (at least thats the style I assume we use here).

wasm2c/wasm-rt.h
```diff
@@ -326,6 +354,31 @@ typedef struct {
...
+   * Unfortunately, the C standard does not state what happens if there are
+   * overlaps in two memory accesses which have a memory order, e.g., an
+   * atomic32 being read from the same location an atomic64 is read. One way to
+   * prevent optimizations from assuming non-overlapping behavior as typically
+   * done in C is to mark the memory as volatile. Thus the memory is atomic and
+   * volatile.*/
+  _Atomic volatile uint8_t* data;
+  /** The current and maximum page count for this Memory object. If there is no
+   * maximum, `max_pages` is 0xffffffffu (i.e. UINT32_MAX). */
+  uint64_t pages, max_pages;
```
sbc100, line 372: Are there other places we declare more than once var in a single decl like this?    I would tend to avoid this, but maybe there is precedent in wabt?

wasm2c/template/wasm-rt-mem-impl.tmpl.c
```diff
@@ -0,0 +1,286 @@
...
+    abort();
+  }
+  memory->data = addr;
+#else
+  memory->data = calloc(byte_length, 1);
+#endif
+}
+
+static uint64_t grow_memory_impl(wasm_rt_TEMPLATE_REPLACE_memory_t* memory,
+                                 uint64_t delta) {
```
sbc100, line 218: It looks like only a few function in this file actually use `_TEMPLATE_REPLACE_`  .. I wonder if its worth leaving everything else in the non-template file?

Also, I wonder if we could simply use the pre-processor here rather than cmake processing?

How about something like this where we just re-include the same file for each type?

```
#define MEMORY_TYPE wasm_rt_memory_t
#include <wasm_rt_memfuncs.inc>
#define MEMORY_TYPE wasm_rt_shared_memory_t
#include <wasm_rt_memfuncs.inc>
```

This could avoid the cmake stuff and the new checked-in prebuilt?

shravanrn: Yup, I had this earlier, but @keithw  suggested that the template approach is preferred. Personally, I have no preference and I am happy to go with the concensus

wasm2c/wasm-rt.h
shravanrn: Yup, this is to be consistent with the style in the `wasm_rt_memory_t` definition just before this. I'm not opposed to avoiding this, but we can follow up in a separate PR since I have to modify `wasm_rt_memory_t` as well when we change this.

CMakeLists.txt
```diff
@@ -422,10 +422,34 @@ if (WABT_INSTALL_RULES)
...
 endif ()
 
+set(THREADS_PREFER_PTHREAD_FLAG ON)
+find_package(Threads)
+
+add_custom_command(
+  OUTPUT gen-wasm2c-rt-prebuilt
+
+  COMMAND ${CMAKE_COMMAND} -D out="${WABT_SOURCE_DIR}/wasm2c/prebuilt/wasm-rt-mem-impl.c" -D in="${WABT_SOURCE_DIR}/wasm2c/template/wasm-rt-mem-impl.tmpl.c" -D replace="" -P ${WABT_SOURCE_DIR}/scripts/gen-wasm2c-rt-templates.cmake
+  COMMAND ${CMAKE_COMMAND} -D out="${WABT_SOURCE_DIR}/wasm2c/prebuilt/wasm-rt-mem-shared-impl.c" -D in="${WABT_SOURCE_DIR}/wasm2c/template/wasm-rt-mem-impl.tmpl.c" -D replace="_shared" -P ${WABT_SOURCE_DIR}/scripts/gen-wasm2c-rt-templates.cmake
```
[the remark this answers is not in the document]
shravanrn: In general, the typical reasons for checking in the prebuilt file are
1. We can see the output which helps code reviews
2. Someone who builds the code doesn't have to install dependencies that do code generation

In wasm2c's use cases however, point 1 isn't really useful as the template files are more readable than the output; and point 2 only uses cmake and python which are both already required for the build. I am not opposed to making this part of the build directory, but this is something we can do in a separate PR if needed since it has to modify other template builds unrelated to this change.

wasm2c/template/wasm-rt-mem-impl.tmpl.c
sbc100: @keithw WDYT?  I think it we can use the simple c pre-processor approach for this kind of thing thats better than cmake magic.

CMakeLists.txt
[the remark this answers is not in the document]
sbc100: I think (1) applies to the other prebuilts, but perhaps not so much to this one (where input and output look basically identical).    So if we do go with templates here (which I would prefer that we simply avoid) then maybe we don't need to check in the prebuilts in this case.

(BTW I don't think python is hard dependency at build time, is it?)

wasm2c/wasm-rt.h
shravanrn: Happy to make this change, but the wasm-rt.h header doesn't seem to have much consistency on this. Again a broader cleanup maybe needed as a part of a separate PR.

sbc100: Sure thing.  Happy to see cleanup here or as separate PRs either before or after this lands.

CMakeLists.txt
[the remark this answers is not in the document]
shravanrn: Not sure. There is this line in the CMake
https://github.com/WebAssembly/wabt/blob/main/CMakeLists.txt#L666

But I don't know what the "COMPONENTS Interpreter" part does... From context, it would seem python is required for the build if you build the interpreter in wabt. 

If it isn't a hard dependency at the moment, any build time codegen (vs. the current checkin time codegen) will make python a hard build dependency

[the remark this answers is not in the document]
sbc100: But we don't use python to generate any of the codegen stuff do we?   IIUC its just used for testing.   (Sorry we might be getting off topic here).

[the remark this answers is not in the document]
shravanrn: oh my mistake, we don't use python for this template generation. This only requires CMake. Given this, would you prefer to change this to generate in the build folder?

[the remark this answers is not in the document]
shravanrn: also @keithw thoughts on this?

wasm2c/wasm-rt.h
shravanrn: Made a separate branch locally to fix this. Will follow up in a separate PR.

shravanrn:
@keithw Just bumping this in your inbox. Please have a look at above questions when you have a chance.

test/run-spec-wasm2c.py
```diff
@@ -445,29 +445,32 @@ def _Action(self, command):
...
+        cstd_flag = ['-std=c11'] if use_c11 else ['-std=c99']
+        args += cstd_flag + ['-c', c_filename, '-o', o_filename, '-O2',
+                             '-Wall', '-Werror', '-Wno-unused',
+                             '-Wno-array-bounds',
+                             '-Wno-ignored-optimization-argument',
+                             '-Wno-tautological-constant-out-of-range-compare',
+                             '-Wno-infinite-recursion',
+                             '-fno-optimize-sibling-calls',
+                             '-frounding-math', '-fsignaling-nans',
+                             '-D_DEFAULT_SOURCE']
```
sbc100, line 473: Nit: How about doing this below which avoid the reindenting and also avoids the single line `if/`else` which I'm not a fan of myself.

```
if use_c11:
  args.append('-std=c99')
else:
  args.append('-std=c11')
```

test/run-spec-wasm2c.py
```diff
@@ -602,6 +605,10 @@ def main(args):
                 return SKIPPED
             cflags.append('-DSUPPORT_MEMORY64=1')
 
+        use_c11 = False
+        if options.enable_threads:
+            use_c11 = True
```
sbc100, line 610: How about just `use_c11 = options.enable_threads`

test/run-spec-wasm2c.py
```diff
@@ -616,24 +623,28 @@ def main(args):
...
         if options.compile:
-            # Compile wasm-rt-impl.
-            wasm_rt_impl_c = os.path.join(options.wasmrt_dir, 'wasm-rt-impl.c')
-            o_filenames.append(Compile(cc, wasm_rt_impl_c, out_dir, *cflags))
+            # Compile runtime code
+            source_files = [
+                main_filename,
+                os.path.join(options.wasmrt_dir, 'wasm-rt-impl.c'),
+                os.path.join(options.wasmrt_dir, 'wasm-rt-exceptions-impl.c'),
+                os.path.join(options.wasmrt_dir, 'prebuilt', 'wasm-rt-mem-impl.c'),
```
sbc100, line 639: Assuming @keithw doesn't object I'd rather drop the `prebuilt` here and just use the pre-processor to make this work.

wasm2c/template/wasm-rt-mem-impl.tmpl.c
keithw: My only real preference would be to try to keep the approach consistent between "what we do for tables" (funcref + externref) vs. "what we do for memories" (normal + shared). I don't care a ton about what the approach is and don't want to apply stop energy, but it would be nice to avoid a flowering of different mechanisms. If you want to use a multiply included `.inc` file for normal+shared memories, fine with me, but then let's do the same thing for funcref+externref tables.

test/run-spec-wasm2c.py
keithw: No objection from me; just the mild preference make this and whatever we do for tables match each other (https://github.com/WebAssembly/wabt/pull/2308#discussion_r1465921700).

wasm2c/template/wasm-rt-mem-impl.tmpl.c
sbc100: Are you saying that we have already take on of these approaches for (funcref + externref) ?     Can you point me to that?

keithw: Was thinking of https://github.com/WebAssembly/wabt/blob/main/src/template/wasm2c.declarations.c#L554-L604

test/run-spec-wasm2c.py
shravanrn: will fix

test/run-spec-wasm2c.py
shravanrn: will fix

test/run-spec-wasm2c.py
shravanrn: I'll follow with the change to tables in a separate PR to keep this one clean

wasm2c/wasm-rt-mem-impl.c
```diff
@@ -0,0 +1,176 @@
...
+    abort();                                    \
+  }
+#define PTHREAD_MEMORY_LOCK_AQUIRE(name)      \
+  if (pthread_mutex_lock(&(name)) != 0) {     \
+    fprintf(stderr, "Lock acquire failed\n"); \
+    abort();                                  \
+  }
+#define PTHREAD_MEMORY_LOCK_RELEASE(name)     \
+  if (pthread_mutex_unlock(&(name)) != 0) {   \
+    fprintf(stderr, "Lock release failed\n"); \
```
sbc100, line 62: Should we have another macro here for LOG_FAILURE, that doesn't nothing in NDEBUG builder?

CMakeLists.txt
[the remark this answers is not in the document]
shravanrn: fixed

wasm2c/template/wasm-rt-mem-impl.tmpl.c
shravanrn: Will fix this in a follow up PR

wasm2c/wasm-rt-mem-impl-helper.inc
```diff
@@ -0,0 +1,166 @@
...
+}
+
+#undef MEMORY_LOCK_RELEASE
+#undef MEMORY_LOCK_AQUIRE
+#undef MEMORY_LOCK_VAR_INIT
+#undef MEMORY_CELL_TYPE
+#undef MEMORY_API_NAME
+#undef MEMORY_TYPE
+#undef WASM_RT_MEM_OPS
+#undef WASM_RT_MEM_OPS_SHARED
```
sbc100, line 164: In cases where I've used/seen this pattern I think its normally up to the includer to undefine these last two after the `#include`.

shravanrn:
@sbc100 @keithw Have made all the changes, could you please sign off on the review

sbc100 (approved):
Thanks!  I like this method much more, personally

wasm2c/wasm-rt-mem-impl.c
```diff
@@ -0,0 +1,176 @@
+/*
```
sbc100, line 1: How about just calling calling these `wasm-rt-impl-mem.c`  (so they share a common prefix with `wasm-rt-impl.c`?

wasm2c/wasm-rt-mem-impl.c
shravanrn: I think for cases that cause a abort/crash, I am generally in favor of having the message always print. So my preference would be to keep this as is unless you feel strongly.

sbc100: I don't feel strongly.  If someone really cares about code size they could remove them..

wasm2c/wasm-rt-mem-impl.c
shravanrn: Hmm... this would make it inconsistent with exceptions which is in `wasm-rt-exceptions-impl.c`. Happy to go with either, but we can maybe do this rename in a follow up PR to since we have to change multiple things

wasm2c/wasm-rt-mem-impl-helper.inc
shravanrn: will fix

PR #2309, opened 2023-10-02
feat: treat non-seekable files as pipes

sethp:
Previously, attempting to read from a pipe would result in an error: 'not a regular file', disallowing use of files like /dev/stdin or /dev/fd/3, named fifos, sockets, etc.

The tools already understand how to (try to) read from non-regular files, so this change attempts to do so when the input is not seek-able (the "regular file" capability that's in use here).

Additionally, this adds a test for the new behavior using a bash herestring and process substitution (the latter of which shows up in argv as something like `/dev/fd/NN`). Since bash isn't commonly installed on Windows, this change also introduces a new capability to filter tests to specific platforms (sorry).

### DIFF
--- a/src/common.cc
+++ b/src/common.cc
@@ -59,14 +59,16 @@ const char* g_reloc_type_name[] = {
 };
 WABT_STATIC_ASSERT(WABT_ARRAY_SIZE(g_reloc_type_name) == kRelocTypeCount);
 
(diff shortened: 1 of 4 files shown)


sbc100 (approved):
LGTM!

Thanks for working on this.

sethp:
Of course, thanks for the quick response!

PR #2314, opened 2023-10-26
rename typeof

walkingeyerobot:
While typeof is not a standard c++ keyword, it is in the gnu dialect (i.e. -std=gnu++17). Simply renaming typeof to typeoffunc and TypeOf to TypeOfFunc lets wabt be built with either dialect.

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -431,8 +431,8 @@ class CWriter {
   void Spill(const TypeVector&);
   void Unspill(const TypeVector&);
 
-  template <typename Vars, typename TypeOf, typename ToDo>
-  void WriteVarsByType(const Vars&, const TypeOf&, const ToDo&);
+  template <typename Vars, typename TypeOfFunc, typename ToDo>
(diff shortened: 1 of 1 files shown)


src/c-writer.cc
```diff
@@ -431,8 +431,8 @@ class CWriter {
   void Spill(const TypeVector&);
   void Unspill(const TypeVector&);
 
-  template <typename Vars, typename TypeOf, typename ToDo>
-  void WriteVarsByType(const Vars&, const TypeOf&, const ToDo&);
+  template <typename Vars, typename TypeOfFunc, typename ToDo>
+  void WriteVarsByType(const Vars&, const TypeOfFunc&, const ToDo&);
```
sbc100, line 435: I don't think there is any need to replace the CamelCase versions of this string, just the argument name down on line 2960, right?

walkingeyerobot: Yes, but I was trying to be consistent. Happy to switch back if you prefer that.

keithw (approved):
lgtm

PR #2319, opened 2023-10-31
Initialize const variable with user provided default constructor

itf:
Fixes error:
default initialization of an object of const type 'const wasm_rt_funcref_t' without a user-provided default constructor


keithw:
Can you give some more details about how to replicate the issue you're seeing, e.g. which compiler/version/configuration? The CI tests with gcc 11, clang 14, and Visual Studio 17 on Linux, Mac, and Windows.

itf:
I was mistakenly building this as C++ instead of C code, which resulted in the error.

PR #2321, opened 2023-11-05
wasm2c: The big-endian port (with CI)

SoniEx2:
This uh, is definitely a bit of a mess. Maybe needs splitting up into 2 or 3 PRs? (Tests for the test harness could be useful maybe?)

This includes #2161 and whatnot.

Marking as a draft pending CI and preliminary feedback. i386 is failing, unsure how to fix, maybe just turn it off altogether for the time being unless someone wants to work on that?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -191,6 +191,7 @@ jobs:
       run: docker run --rm -t wabt cmake --build build --target run-unittests
 
     - name: C API tests
+      if: ${{ matrix.arch != 's390x' }}
       run: docker run --rm -t wabt cmake --build build --target run-c-api-tests
 
     - name: Tests
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -167,3 +167,31 @@ jobs:
       run: cmake --build out --target run-c-api-tests
     - name: tests
       run: cmake --build out --target run-tests
+
+  build-arch:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: [s390x, i386]
+    steps:
+    - uses: actions/checkout@v1
+      with:
+        submodules: true
+
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v2
+      with:
+        platforms: ${{matrix.arch}}
+
+    - name: Build image
+      run: docker build --platform linux/${{matrix.arch}} -t wabt -f scripts/Dockerfile .
+
+    - name: Unit tests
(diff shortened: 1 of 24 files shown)


sbc100:
Can you mention in the PR title that this is for wasm2c?

keithw:
Replicating some feedback from Discord:

Thank you for doing all this work!

We should probably find a way to split this up. Definitely making it s390x-only (skipping i386) seems like a good first step.
I think it's going to be a challenge to merge a PR that adds a CI workflow that lasts 70+ minutes. Maaybe it's more tolerable if it's a non-required workflow, but still...

A small split-off PR could be "fix the format-style bugs in the existing SIMD test harness."

And then another one could be "abstract those simde calls" (and explain why you need to do that).

It seems like it would be nicer if we were not using distcc or the "run a whole Docker container in qemu" stuff. I'd hope that we can just explicitly invoke qemu when needed. But maybe this is harder than I realized.

sbc100:
> Replicating some feedback from Discord:
> 
> Thank you for doing all this work!
> 
> We should probably find a way to split this up. Definitely making it s390x-only (skipping i386) seems like a good first step. I think it's going to be a challenge to merge a PR that adds a CI workflow that lasts 70+ minutes. Maaybe it's more tolerable if it's a non-required workflow, but still...

We can also run longer CI jobs overnight, or optionally rather than making them the default.

Honestly do think its reasonably to consider platforms like s390 and i386 as second tier / best effort supported if only due to the small number of users on those platforms.


> 
> A small split-off PR could be "fix the format-style bugs in the existing SIMD test harness."
> 
> And then another one could be "abstract those simde calls" (and explain why you need to do that).
> 
> It seems like it would be nicer if we were not using distcc or the "run a whole Docker container in qemu" stuff. I'd hope that we can just explicitly invoke qemu when needed. But maybe this is harder than I realized.

SoniEx2:
> It seems like it would be nicer if we were not using distcc or the "run a whole Docker container in qemu" stuff. I'd hope that we can just explicitly invoke qemu when needed. But maybe this is harder than I realized.

We tried to keep it maintainable. This setup allows easily adding more archs, as well as avoiding the complexities of finding/configuring the cross-compiler (especially system libraries etc), and of setting up the chroot for running the tests in (we'd need the chroot either way, since it needs to find the appropriate architecture system libraries, at least in the case of dynamic linking).

Docker gives us the "setup a cross-arch chroot". distcc gives us the "find a faster cross-compiler". 70 minutes is blazing fast compared to how long this was taking before distcc (think around 4 hours)!

Issue #2304, opened 2023-09-26
wasm-decompile: Assertion `is_name()' failed.


nepx:
wasm-decompile crashes when trying to process [this WAT file](https://github.com/WebAssembly/wabt/files/12731218/test.zip). Interestingly enough, it works after being processed with wasm-opt. 

The full text of the error is: 

```
wabt-main/include/wabt/ir.h:62: const string& wabt::Var::name() const: Assertion `is_name()' failed.
```

keithw:
I think the root issue here is that wasm-decompile doesn't know how to handle a `br` to the implicit function label, which is basically equivalent to a `return`. My guess is that wasm-opt is converting this to a `return`, which wasm-decompile does know how to handle.

Here's a minimal test case that also triggers the same assertion failure:
```wasm
(func br 0)
```

Handling this cleanly is probably pretty doable but requires some mucking with the decompiler.

PR #2324, opened 2023-11-10
Fix implicit br return

dzfrias:
An in-progress fix for #2304. For now, I believe it's working for any `br` depth. However, I'm not really sure how to handle the `br_if` case. Should it be replaced with an `if` and then a `return`? If so, I couldn't really find a way to create a new `IfExpr` and compile that on the fly. Maybe someone with more experience with the decompiler source code has a good way to do it.

### DIFF
--- a/include/wabt/decompiler-ast.h
+++ b/include/wabt/decompiler-ast.h
@@ -217,8 +217,12 @@ struct AST {
         return;
       }
       case ExprType::Br: {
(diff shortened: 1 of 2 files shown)


keithw:
I think this PR probably makes you the leading active wasm-decompiler person we have. :-) Your strategy sounds good but I don't think I have any better ideas...

dzfrias:
Hmm ok. I'll work on finding a solution that works, then!

PR #2326, opened 2023-11-11
Clean up WABT_BIG_ENDIAN handling in tests

SoniEx2:
Split from #2321


keithw:
lgtm.

Longer-term, I wonder whether the wasm2c generated code should require the user to manually add `-DWABT_BIG_ENDIAN` on the compiler command line when targeting big-endian (as currently), or if it can autodetect the platform endianness in the preprocessor (e.g. via `#include <endian.h>` or http://esr.ibiblio.org/?p=5095).

keithw (approved):

PR #2327, opened 2023-11-11
wasm2c: Fix test harness UB with SIMD

SoniEx2:
This was segfaulting/crashing/other issues. Split from #2321

### DIFF
--- a/test/run-spec-wasm2c.py
+++ b/test/run-spec-wasm2c.py
@@ -408,7 +408,7 @@ class CWriter(object):
         return ', '.join(self._Constant({'type': const['lane_type'], 'value': val}) for val in const['value'])
 
     def _SIMDFound(self, num, lane_type, lane_count):
-        return 'simde_wasm_%sx%d_extract_lane(actual, %d)' % (lane_type, lane_count, num)
+        return 'v128_%sx%d_extract_lane(actual, %d)' % (lane_type, lane_count, num)
 
     def _SIMDFoundList(self, lane_type, lane_count):
         return ', '.join(self._SIMDFound(num, lane_type, lane_count) for num in range(lane_count))
@@ -644,7 +644,10 @@ def main(args):
 
             # Run the resulting binary
             if options.run:
-                utils.Executable(main_exe, forward_stdout=True).RunWithArgs()
+                error = utils.Executable(main_exe).RunWithArgsForError()
+                if error:
+                    print(error, file=sys.stderr)
+                    return 1
 
     return 0
 
--- a/test/harness-windows/wasm2c/simd_formatting.txt
+++ b/test/harness-windows/wasm2c/simd_formatting.txt
@@ -1,0 +1,18 @@
+;;; TOOL: run-spec-wasm2c
+;;; ERROR: 1
+;;; PLATFORMS: Windows
+(module
+  (func (export "x") (param $x v128) (result v128) (local.get $x))
+)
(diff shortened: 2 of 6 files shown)


SoniEx2:
...ugh windows.

nvm, we do not know how to do this.

keithw:
Thanks for finding and fixing this! I respect all the testing effort, but given that we don't have any other "test error messages" tests, I'm tempted to say it might be easier/tolerable to take the two-line fix by itself and omit the rest (including the Windows and non-Windows test and the machinery necessary to support that). :-/ Thoughts?

SoniEx2:
two-line fix? the lane extractors need to be tweaked too, or else stuff happens (formatting an i16 as an u16 is wrong and produces weird/confusing results on some platforms).

~~besides maybe this can encourage more contributions in this area. at least that's what we're hoping for...~~

SoniEx2:
also, what are these?

https://github.com/WebAssembly/wabt/blob/main/test/pipes.txt
https://github.com/WebAssembly/wabt/blob/main/test/spectest-interp-assert-failure.txt
https://github.com/WebAssembly/wabt/blob/main/test/spectest-interp-error-count.txt
https://github.com/WebAssembly/wabt/blob/main/test/spectest-interp-invalid-literal.txt
https://github.com/WebAssembly/wabt/blob/main/test/too-many-arguments.txt
https://github.com/WebAssembly/wabt/blob/main/test/two-commands.txt

test/run-spec-wasm2c.py
```diff
@@ -644,7 +644,10 @@ def main(args):
 
             # Run the resulting binary
             if options.run:
-                utils.Executable(main_exe, forward_stdout=True).RunWithArgs()
+                error = utils.Executable(main_exe, forward_stdout=True).RunWithArgsAndError()
```
keithw, line 647: Given that we're printing `error` out now (which includes the stdout from the process, if captured), can we remove the `forward_stdout=True` from this line?

That will prettify the new test's output a little; instead of `None` (for the stdout that went directly to the terminal), it'll say `0/1 tests passed.` (since stdout is now being captured in the Error).

keithw:
> two-line fix? the lane extractors need to be tweaked too, or else stuff happens (formatting an i16 as an u16 is wrong and produces weird/confusing results on some platforms).

Fair enough, let me try to understand the rest of it then. Just gave a small review; everything else lgtm.

test/run-spec-wasm2c.py
SoniEx2: yeah, we don't like how this test runner is set up. unfortunately we do not have the capacity to re-engineer it to do things properly.

because in this case, we were trying to keep the stdout in stdout but also get the underlying status code into stderr. (since this runs in the context of `run-spec-wasm2c.py` we can't match the status code with the usual status code matching directive `ERROR`, but nothing stops us from shoving it in stderr.)

keithw: Fair enough on the larger point, but if we get rid of the `, forward_stdout=True` from this line, the output from the new test becomes a little prettier, none of the old tests change, and stdout still goes to stdout on successful runs. Everything else lgtm.

keithw:
:+1: One more request before merging: please `run-tests.py -r simd_formatting` so those newlines don't have to be added the next time somebody does a rebase across all the tests.

SoniEx2:
oh. uh, we're guessing those also apply to the windows one? (no easy access to windows sorry)

keithw (approved):

PR #2328, opened 2023-11-13
wasm2c: Improve address abstraction for BE support

SoniEx2:
This doesn't fix much on BE (fixes atomics, doesn't fix SIMD) but it cleans things up and makes it more maintainable.

### DIFF
--- a/src/template/wasm2c.declarations.c
+++ b/src/template/wasm2c.declarations.c
@@ -1,4 +1,25 @@
 
+// Computes a pointer to an object of the given size in a little-endian memory.
+//
+// On a little-endian host, this is just &mem->data[addr] - the object's size is
+// unused. On a big-endian host, it's &mem->data[mem->size - addr - n], where n
+// is the object's size.
+//
+// Note that mem may be evaluated multiple times.
+//
+// Parameters:
+// mem - The memory.
+// addr - The address.
+// n - The size of the object.
+//
+// Result:
+// A pointer for an object of size n.
+#if WABT_BIG_ENDIAN
+#define MEM_ADDR(mem, addr, n) &(mem)->data[(mem)->size - (addr) - (n)]
+#else
+#define MEM_ADDR(mem, addr, n) &(mem)->data[addr]
+#endif
+
 #define TRAP(x) (wasm_rt_trap(WASM_RT_TRAP_##x), 0)
 
 #if WASM_RT_USE_STACK_DEPTH_COUNT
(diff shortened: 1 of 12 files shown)


src/template/wasm2c.declarations.c
```diff
@@ -1,4 +1,10 @@
 
+#if WABT_BIG_ENDIAN
+#define LEADDR(mem, addr, n) (mem)->data[(mem)->size - (addr) - (n)]
```
keithw, line 3: Can we name this something a little more friendly looking (how do you feel about `MEM_ADDR`), and maybe with a comment about its purpose and the meaning of the arguments?

src/template/wasm2c.declarations.c
```diff
@@ -67,70 +73,42 @@ static inline bool func_types_eq(const wasm_rt_func_type_t a,
...
-}
-#define LOAD_DATA(m, o, i, s)      \
-  do {                             \
-    RANGE_CHECK((&m), o, s);       \
-    load_data(&(m.data[o]), i, s); \
+
+#define LOAD_DATA(m, o, i, s)           \
+  do {                                  \
+    RANGE_CHECK((&m), o, s);            \
+    load_data(&LEADDR(&m, o, s), i, s); \
```
keithw, line 94: Do all the callsites have it as `&LEADDR`? (And if so, can the macro return the address?)

keithw (commented):
lgtm at least for the main declarations and SIMD one.

@shravanrn, this touches the atomic ops declarations which I know you're in the process of revising -- do you have a view on this?

src/template/wasm2c.declarations.c
SoniEx2: We would insist on keeping it as `LEADDR`. In the hypothesis wasm ever adds a big-endian memory type (and an accompanying `v128be`), we can easily introduce `BEADDR` and have a mix of little-endian and big-endian memories in the same module. :sweat_smile:

shravanrn:
> lgtm at least for the main declarations and SIMD one.
> 
> @shravanrn, this touches the atomic ops declarations which I know you're in the process of revising -- do you have a view on this?

@keithw It's fine by me, I can rebase my change, as this seems straightforward to adopt.

That said, I do have some points about big endian support in general. Right now, the security audits we have done for Firefox's use of wasm2c does not cover wasm2c's big endian support. In fact, I think I saw some details that looked weird/wrong/insecure, but didn't investigate as this is not used in the Firefox use case. Longer term, we probably want to do a more thorough evaluation if we want to claim we support this OR include a compiler warning saying this is not tested completely OR just remove support. I don't want to derail this thread though, so we can follow up about this separately.

src/template/wasm2c.declarations.c
shravanrn: Please correct me if I'm wrong, but I think there is no proposal about this sort of change at the moment. I think, we can always refactor this to LEADDR, if that happens. For the current wasm support, however, I think @keithw 's suggestion of the MEM_ADDR name will keep the code more readable.

src/template/wasm2c.declarations.c
```diff
@@ -67,70 +88,42 @@ static inline bool func_types_eq(const wasm_rt_func_type_t a,
...
+#endif
 }
-#define LOAD_DATA(m, o, i, s)                   \
-  do {                                          \
-    RANGE_CHECK((&m), m.size - o - s, s);       \
-    load_data(&(m.data[m.size - o - s]), i, s); \
+
+#define LOAD_DATA(m, o, i, s)          \
+  do {                                 \
+    RANGE_CHECK((&m), o, s);           \
```
shravanrn, line 108: I think this changes the exact range check being performed in big endian mode. While I haven't investigated thoroughly, I suspect this is probably the right check vs. what was already in place (or maybe both checks do the same thing?) 

@keithw @sbc100 @kripken Any chance y'all could have a look here and see if this makes sense to you?

shravanrn (commented):
I think we can wait to confirm the one range check we spotted and for the naming fix before merging.

src/template/wasm2c.declarations.c
SoniEx2: if `RANGE_CHECK` had an `o < 0` check, they would be equivalent. we believe this was changed at some point. indeed, the bug seems to trace back to https://github.com/WebAssembly/wabt/commit/407c5f65d172bba3d8b62579468d9c241556ef4b

abstracting out `LEADDR` should make these regressions less likely.

src/template/wasm2c.declarations.c
keithw: Hmm, yeah, I don't want the tail to end up wagging the dog here. We care about big-endian support (at least insofar as we have users who care), but not to the degree that it ends up making the codebase confusing or imposing to new contributors. There's nothing inherently little-endian about an expression like `&(mem)->data[addr]`.

How would you feel about `MEMREF` ? Or anything else you want to propose?

src/template/wasm2c.declarations.c
keithw: The change lgtm given that it makes the tests pass on a BE architecture. My earlier effort to match the behavior was definitely best-effort and untested (in my defense, before that PR, there was *no* range check on active data loads at all...).

keithw:
> Longer term, we probably want to do a more thorough evaluation if we want to claim we support this OR include a compiler warning saying this is not tested completely OR just remove support.

Seems reasonable; maybe we can add something to the README about the architectures we test on or feel most confident in.

keithw (approved):

PR #2330, opened 2023-11-15
wasm2c: initialize wasm_rt_funcref_null_value to NULL

wrv:
This PR fixes a build issue in C++ projects that include the wasm2c output header by adding a user-provided default constructor to `wasm_rt_funcref_null_value`.

### DIFF
--- a/wasm2c/wasm-rt.h
+++ b/wasm2c/wasm-rt.h
@@ -279,13 +279,13 @@ typedef struct {
 } wasm_rt_funcref_t;
 
 /** Default (null) value of a funcref */
-static const wasm_rt_funcref_t wasm_rt_funcref_null_value;
+#define wasm_rt_funcref_null_value ((wasm_rt_funcref_t){NULL, NULL, {NULL}, NULL})
 
 /** The type of an external reference (opaque to WebAssembly). */
 typedef void* wasm_rt_externref_t;
 
 /** Default (null) value of an externref */
-static const wasm_rt_externref_t wasm_rt_externref_null_value = NULL;
+#define wasm_rt_externref_null_value ((wasm_rt_externref_t){NULL})
 
 /** A Memory object. */
 typedef struct {


sbc100:
Is this header supposed to be included in any files except the `.c` files that were generated by wasm2c?    (Sorry I actually can't remember if its useful outside of the generated code?). @keithw

sbc100:
> This PR fixes a build issue in C++ projects that include the wasm2c output header by adding a user-provided default constructor to `wasm_rt_funcref_null_value`.

What happens without this patch?  Do you get a compiler error or just a warning?

wrv:
Without this patch you get the following compiler error:

```
wasm-rt.h:282:32: error: uninitialized 'const wasm_rt_funcref_null_value' [-fpermissive]
  282 | static const wasm_rt_funcref_t wasm_rt_funcref_null_value;
      |                                ^~~~~~~~~~~~~~~~~~~~~~~~~~
wasm-rt.h:267:16: note: 'const struct wasm_rt_funcref_t' has no user-provided default constructor
  267 | typedef struct {
      |                ^
wasm-rt.h:269:23: note: and the implicitly-defined constructor does not initialize 'const char* wasm_rt_funcref_t::func_type'
  269 |   wasm_rt_func_type_t func_type;
```

Here is an example when compiling RLBox: https://github.com/wrv/rlbox_wasm2c_sandbox/actions/runs/6859462708/job/18651814085

shravanrn:
> Is this header supposed to be included in any files except the `.c` files that were generated by wasm2c? (Sorry I actually can't remember if its useful outside of the generated code?). @keithw

Beyond the internal use, the header is needed for the rest of the system to interact with the Wasm module. For example, the header is consumed by RLBox (which is in C++) to provide an safe & ABI compatible interface for the host application to use the Wasm module

shravanrn:
> > > Is this header supposed to be included in any files except the `.c` files that were generated by wasm2c? (Sorry I actually can't remember if its useful outside of the generated code?). @keithw
> > 
> > 
> > @sbc100 Beyond the internal use, the header is needed for the rest of the system to interact with the Wasm module. For example, the header is consumed by RLBox (which is in C++) to provide a safe & ABI compatible interface of the Wasm module to the host application
> 
> Is defining this statically in each compilation unit the right thing to do? Should we instead declare it as extern here? and define it in the wasm-rt.c file?

I think declaring it as extern, and defining it in the .c file would be the normal practice. 
While the current approach having this static const included in the header means it will show up in all  c/c++ files that use this header, the compiler probably optimizes this away, so this is probably fine in practice as well.

I don't have a strong preference either way. @keithw thoughts?

sbc100:
> > Is defining this statically in each compilation unit the right thing to do? Should we instead declare it as extern here? and define it in the wasm-rt.c file?
> 
> I think declaring it as extern, and defining it in the .c file would be the normal approach. While the current approach having this static const included in the header means it will show up in all c/c++ files that use this header, the compiler probably optimizes this away, so this is probably fine in practice as well.
> 
> I don't have a strong preference either way. @keithw thoughts?

As it stands I think each object file that references this will get a different address for this struct.. but then again I think its most likely compared by value so that address shouldn't matter I guess.

keithw:
In an ideal world these would be `constexpr`, but that doesn't work in C. How about this, which seems to work in both C and C++ and avoids the object having to live in each translation unit? (Also we should probably add a C++ CI check if we want to avoid this happening again...)

```c++
/** Default (null) value of a funcref */
#define wasm_rt_funcref_null_value ((wasm_rt_funcref_t){})

[...]

/** Default (null) value of an externref */
#define wasm_rt_externref_null_value ((wasm_rt_externref_t){})
```

wrv:
>Also we should probably add a C++ CI check if we want to avoid this happening again...

I've been testing this by changing `wasm2c/examples/fac/main.c` to `wasm2c/examples/fac/main.cc`, but I don't see if the examples are built anywhere in the CI.

keithw (approved):

PR #2332, opened 2023-11-19
wasm2c runtime: add per-thread initialization/free entrypoint

### LINKED ISSUE #2046: wasm2c use of globals and statics introduce race conditions in concurrent environments

### LINKED ISSUE #2189: wasm2c runtime: add per-thread init/free, and make wasm_rt_init/free return and take a `void*`?
Trying to get #2046 closed off (use of globals preventing the wasm2c runtime from being thread-safe).

It looks like the only remaining blocker is the `static char* g_alt_stack` variable, which currently gets malloc'ed once per process via `wasm_rt_init()` which calls `os_install_signal_handler()`.

The challenge is that: if we want the runtime to be thread-safe, every thread will need a **different** allocated region for the alt stack, so that each thread can (in theory) be handling stack exhaustion by running the SIGSEGV signal handler at the same time. Each thread will have to install its own alt stack separately by calling `sigaltstack()`. (The signal handler itself only has to be installed once per process.)

I think my preferred way to fix this would be:
- we eliminate the global `g_alt_stack` variable
- we add a new API pair that looks like:
   ```c
   void* wasm_rt_thread_init();
   wasm_rt_thread_free(void*);
   ```
   `wasm_rt_thread_init()` is responsible for (a) malloc'ing the altstack (sadly this has to be done at runtime because SIGSTKSZ is only known at runtime), (b) installing it via `sigaltstack()`, and (c) returning it via an opaque `void*` that the caller has to keep around. `wasm_rt_thread_free()` does the reverse (de-installing the altstack **if** it's still the current altstack, and freeing the memory either way)
   
   If the host wants to use wasm2c-generated code across multiple threads, it needs to call `wasm_rt_thread_init()` in each thread before calling in to a wasm2c-generated module.
- we don't require every user to call these functions (especially not in single-threaded code), but we change the signature of `wasm_rt_init()` and `wasm_rt_free()` to be:
    ```c
   void* wasm_rt_init();
   wasm_rt_free(void*);
   ```
   so that they can internally call `wasm_rt_thread_init()` and `wasm_rt_thread_free()` for whatever thread they are run on.

I think I like this design better than the gymnastics of trying to make `g_alt_stack` a thread-local global and worrying about managing it on a per-thread basis (e.g., what happens if the user calls `wasm_rt_thread_init()` on the same thread that they called `wasm_rt_free()` on), but it does increase the burden for the host of having to keep this `void*` around.

@shravanrn @sbc100 Do you have thoughts?

keithw:
This lets wasm2c-generated modules be used in a multithreaded context (on POSIX) by giving each thread a runtime API entrypoint to initialize and free the altstack, which is used to catch segfaults from stack exhaustion. On Windows, explicit counting of stack depth is used to control stack exhaustion.

Closes #2189. Closes #2046, in the sense that this finishes getting rid of global state in our distributed runtime that doesn't have a thread-safe way of using it.


sbc100 (approved):

sbc100:
Maybe wait for @keithw who might have more feedback..

keithw:
@shravanrn, any concerns before this goes in? I know you were the one initially concerned about thread safety in the runtime via #2046.

The semantics are basically:

- `wasm_rt_init()` initializes the runtime and is basically unchanged.
  - By default, it installs the signal handler, allocates an altstack in g_alt_stack, and installs it
  - But if WASM_RT_SKIP_SIGNAL_RECOVERY was set, it doesn't do anything.
- wasm_rt_init_thread() initializes any subsequent thread.
  - By default, it allocates an altstack, installs it, and returns its location.
  - But if WASM_RT_SKIP_SIGNAL_RECOVERY was set, it doesn't do anything.
- Same as before, `wasm_rt_is_initialized()` gets asserted by the generated code in module instantiation (so it's called unless NDEBUG is set)
- If signal handling is being used to catch stack exhaustion (i.e. if WASM_RT_USE_STACK_DEPTH_COUNT isn't set), `wasm_rt_is_initialized()` will now check to make sure that the **current _thread_ has been initialized** by checking for the existence of an altstack (whether it was the wasm2c runtime that installed it, or the embedder).

So what's new here is that wasm_rt_is_initialized() is going to start enforcing that there's an altstack in place if we're using signals to catch stack exhaustion, and people who run wasm2c-generated code with asserts on will hit that. Hopefully that's acceptable to Firefox and anybody else using WASM_RT_SKIP_SIGNAL_RECOVERY...?

shravanrn:
Looking today. Sorry for the delay

wasm2c/wasm-rt-impl.c
```diff
@@ -211,28 +257,30 @@ static void os_cleanup_signal_handler(void) {
...
 
 void wasm_rt_init(void) {
 #if WASM_RT_INSTALL_SIGNAL_HANDLER
-  if (!g_signal_handler_installed) {
-    g_signal_handler_installed = true;
-    os_install_signal_handler();
-  }
+#if !WASM_RT_USE_STACK_DEPTH_COUNT
+  assert(!g_alt_stack);
+  g_alt_stack = os_allocate_and_install_altstack();
```
shravanrn, line 269: Nit: It may make sense to call `wasm_rt_thread_init` within here, as logically what we are saying is that the main thread gets has any thread-specific instantiation performed by default. Same with `wasm_rt_free`

shravanrn (changes requested):

shravanrn:
Looks reasonable % comments. I am still thinking through the design choice of "main thread is special and the user doesn't have to call `wasm_rt_thread_init`. I expect that on balance, this makes sense from a backward compatibility standpoint, even though it makes the internal code a bit messier.

wasm2c/wasm-rt-impl.c
```diff
@@ -43,9 +43,11 @@ static bool g_signal_handler_installed = false;
 #ifdef _WIN32
 static void* g_sig_handler_handle = 0;
 #else
+#if !WASM_RT_USE_STACK_DEPTH_COUNT
```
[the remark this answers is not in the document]
keithw: Done.

wasm2c/wasm-rt-impl.c
```diff
@@ -172,26 +174,47 @@ static void os_signal_handler(int sig, siginfo_t* si, void* unused) {
...
+static bool os_has_altstack_installed() {
+  /* check for altstack already in place */
+  stack_t ss;
+  if (sigaltstack(NULL, &ss) != 0) {
+    perror("sigaltstack failed");
+    abort();
+  }
+
+  return !(ss.ss_flags & SS_DISABLE);
+}
```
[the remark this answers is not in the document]
keithw: Well, we do CI on Mac (with asserts enabled, at least I'm pretty sure)... AFAIK this is sigaltstack's only way of indicating a lack of alternate stack. But that's all the non-Linux testing I've done. I think if there are other POSIXy platforms that care, it would be great to get them in the CI (not necessarily using GitHub Actions to do it).

wasm2c/wasm-rt-impl.c
```diff
@@ -202,6 +225,29 @@ static void os_install_signal_handler(void) {
...
 }
 
+static void os_disable_and_deallocate_altstack(void* alt_stack) {
+  /* verify altstack was still in place */
+  stack_t ss;
+  if (sigaltstack(NULL, &ss) != 0) {
+    perror("sigaltstack failed");
+    abort();
+  }
+  if ((ss.ss_flags & SS_DISABLE) || (ss.ss_sp != alt_stack) ||
```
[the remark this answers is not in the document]
keithw: Done.

wasm2c/wasm-rt-impl.c
keithw: Yeah... I think in a world where `wasm_rt_init` returns the pointer (and `wasm_rt_free` takes it), that makes total sense. But if `g_alt_stack` only exists if certain preprocessor macros are set a certain way, then `wasm_rt_init` would have to condition the call to `wasm_rt_thread_init` similarly, and that feels weird -- why aren't we doing a thread init if `WASM_RT_USE_STACK_DEPTH_COUNT` is set?

So I think as long as `g_alt_stack` exists, `wasm_rt_init` has to take responsibility for knowing the details itself; it can't take advantage of the abstraction barrier.

keithw:
> Looks reasonable % comments. I am still thinking through the design choice of "main thread is special and the user doesn't have to call `wasm_rt_thread_init`. I expect that on balance, this makes sense from a backward compatibility standpoint, even though it makes the internal code a bit messier.

My own druthers would have been to get rid of `g_alt_stack`, but there was some opposition to changing the API to wasm_rt_init/wasm_rt_free for non-multi-threaded-users (https://github.com/WebAssembly/wabt/issues/2189). This seemed to be the best compromise that preserves the current API.

wasm2c/wasm-rt-impl.c
```diff
@@ -241,9 +289,34 @@ bool wasm_rt_is_initialized(void) {
...
+  g_alt_stack = NULL;
+#endif
+#endif
+}
+
+void* wasm_rt_init_thread(void) {
+#if WASM_RT_USE_STACK_DEPTH_COUNT || (!WASM_RT_INSTALL_SIGNAL_HANDLER)
+  return NULL;
+#else
+  return os_allocate_and_install_altstack();
```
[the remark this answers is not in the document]
keithw: I think we can always do this when we have more fields to add; it's an opaque `void*` either way.

wasm2c/wasm-rt-impl.c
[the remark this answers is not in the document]
keithw: On further review... I just got rid of `g_alt_stack` without changing the public API.

wasm2c/wasm-rt-impl.c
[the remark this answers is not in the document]
keithw: On further review... I just made `wasm_rt_init_thread` return `void` for now. The OS is already keeping track of the altstack.

wasm2c/wasm-rt-impl.c
keithw: On further review... I got rid of g_alt_stack and just let the OS keep track of it.

wasm2c/wasm-rt-impl.c
```diff
@@ -202,6 +221,29 @@ static void os_install_signal_handler(void) {
...
+    if (sigaltstack(&ss, NULL) != 0) {
+      perror("sigaltstack failed");
+      abort();
+    }
+    assert(!os_has_altstack_installed());
+    free(alt_stack);
+    return;
+  }
+
+  fprintf(stderr, "Warning: wasm2c altstack was modified unexpectedly\n");
```
sbc100, line 244: Can you invert the check above and put this error in an early return.

Also, do we want this error string in release builds?

wasm2c/wasm-rt-impl.c
```diff
@@ -42,8 +42,6 @@
 static bool g_signal_handler_installed = false;
 #ifdef _WIN32
 static void* g_sig_handler_handle = 0;
-#else
-static char* g_alt_stack = 0;
 #endif
```
sbc100, line 45: Maybe too much of a pain but removing g_alt_stack now seems like a separable change, along with `SS_DISABLE` stuff?

keithw: Okay, let me try to split this out into a new PR... (And maybe we should keep track of it in a thread-local just for safety anyway.)

wasm2c/wasm-rt-impl.c
keithw: On further further view... I think I found a way to do this to satisfy everybody. wasm_rt_init() calls wasm_rt_init_thread() as you propose. `g_alt_stack` is still there but it's a thread-local mostly just used for safety (it gets cross-checked against the actual altstack).

wasm2c/wasm-rt-impl.c
keithw: Done.

wasm2c/wasm-rt-impl.c
keithw: Done.

keithw:
This is ready for review again and is a lot smaller now that #2338 has been merged.

sbc100 (approved):
I love how much smaller this change is now!  Thanks for splitting it up.

lgtm from my POV but maybe wait for @shravanrn too

shravanrn (approved):

PR #2338, opened 2023-11-28
wasm2c runtime: refactor handling of alternate stack (NFC)

keithw:
Split out from #2332.

This makes the `g_alt_stack` variable thread-local and splits out the internal routines that handle allocation/installation. This will make it a smaller change to add the per-thread initialization API in #2332.

### DIFF
--- a/wasm2c/wasm-rt-impl.c
+++ b/wasm2c/wasm-rt-impl.c
@@ -251,10 +251,9 @@ static void os_disable_and_deallocate_altstack(void) {
 static void os_install_signal_handler(void) {
   struct sigaction sa;
   memset(&sa, '\0', sizeof(sa));
-#if WASM_RT_USE_STACK_DEPTH_COUNT
   sa.sa_flags = SA_SIGINFO;
-#else
-  sa.sa_flags = SA_SIGINFO | SA_ONSTACK;
+#if !WASM_RT_USE_STACK_DEPTH_COUNT
+  sa.sa_flags |= SA_ONSTACK;
 #endif
   sigemptyset(&sa.sa_mask);
   sa.sa_sigaction = os_signal_handler;
--- a/wasm2c/wasm-rt-impl.c
+++ b/wasm2c/wasm-rt-impl.c
@@ -172,6 +172,8 @@ static void os_signal_handler(int sig, siginfo_t* si, void* unused) {
   }
 }
 
+#if !WASM_RT_USE_STACK_DEPTH_COUNT
+/* These routines set up an altstack to handle SIGSEGV from stack overflow. */
 static bool os_has_altstack_installed() {
   /* check for altstack already in place */
   stack_t ss;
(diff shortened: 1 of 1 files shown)


wasm2c/wasm-rt-impl.c
```diff
@@ -172,14 +172,41 @@ static void os_signal_handler(int sig, siginfo_t* si, void* unused) {
...
+/* Use alt stack to handle SIGSEGV from stack overflow */
+static void os_allocate_and_install_altstack(void) {
+  /* verify altstack not already allocated */
+#ifndef NDEBUG
+  if (g_alt_stack) {
+    fprintf(
+        stderr,
+        "wasm-rt error: tried to re-allocate thread-local alternate stack\n");
+    abort();
+  }
```
sbc100, line 195: I'm not sure what the convention is within this file, but would it make sense to write these 6 lines as `assert(!g_alt_stack && "wasm-rt error: tried to re-allocate thread-local alternate stack")`?

wasm2c/wasm-rt-impl.c
```diff
@@ -202,6 +235,41 @@ static void os_install_signal_handler(void) {
...
+    perror("sigaltstack failed");
+    abort();
+  }
+
+  if ((!g_alt_stack) || (ss.ss_flags & SS_DISABLE) ||
+      (ss.ss_sp != g_alt_stack) || (ss.ss_size != SIGSTKSZ)) {
+#ifndef NDEBUG
+    fprintf(stderr,
+            "wasm-rt warning: alternate stack was modified unexpectedly\n");
+#endif
```
sbc100, line 259: And this could just be `assert(false && ..`?

wasm2c/wasm-rt-impl.c
```diff
@@ -241,8 +311,15 @@ bool wasm_rt_is_initialized(void) {
 }
 
 void wasm_rt_free(void) {
+  if (!wasm_rt_is_initialized()) {
+    fprintf(stderr, "wasm_rt_free: wasm runtime not initialized.\n");
+    return;
+  }
```
sbc100, line 317: Maybe just `assert(wasm_rt_is_initialized() && ...` (no need for early return since its basically undefined behaviour to do this?)

wasm2c/wasm-rt-impl.c
```diff
@@ -172,14 +172,41 @@ static void os_signal_handler(int sig, siginfo_t* si, void* unused) {
...
+  if (sigaltstack(NULL, &ss) != 0) {
+    perror("sigaltstack failed");
+    abort();
+  }
+
+  return !(ss.ss_flags & SS_DISABLE);
+}
+
+/* Use alt stack to handle SIGSEGV from stack overflow */
+static void os_allocate_and_install_altstack(void) {
```
sbc100, line 187: Wrap all these static functions related to altstack in `if !WASM_RT_USE_STACK_DEPTH_COUNT`?

wasm2c/wasm-rt-impl.c
keithw: Sure, done.

wasm2c/wasm-rt-impl.c
keithw: Done.

wasm2c/wasm-rt-impl.c
keithw: Done.

wasm2c/wasm-rt-impl.c
keithw: I'd rather not make this one abort() on failure, since it seems plausible the application would install its own altstack at some point, and that shouldn't be a fatal error...

wasm2c/wasm-rt-impl.c
```diff
@@ -188,10 +210,52 @@ static void os_install_signal_handler(void) {
...
+#endif
 
+static void os_install_signal_handler(void) {
   struct sigaction sa;
   memset(&sa, '\0', sizeof(sa));
+#if WASM_RT_USE_STACK_DEPTH_COUNT
+  sa.sa_flags = SA_SIGINFO;
+#else
   sa.sa_flags = SA_SIGINFO | SA_ONSTACK;
+#endif
```
sbc100, line 258: How about:

```
sa.sa_flags = SA_SIGINFO;
#if !WASM_RT_USE_STACK_DEPTH_COUNT
sa.sa_flags |=  SA_ONSTACK;
#endif
```

sbc100 (approved):
lgtm % final nit/comment

wasm2c/wasm-rt-impl.c
keithw: Sure, done.

PR #2339, opened 2023-11-30
Fix wasm-rt-impl defines nesting

wrv:
This PR fixes the nesting order of `WASM_RT_INSTALL_SIGNAL_HANDLER` and `WASM_RT_USE_STACK_DEPTH_COUNT` in `wasm_rt_is_intialized` to match the rest of the file.

Before this PR, we were getting an `undefined reference` error:
```
rlbox_wasm2c_sandbox/build/_deps/mod_wasm2c-src/wasm2c/wasm-rt-impl.c:297:8: warning: implicit declaration of function ‘os_has_altstack_installed’ [-Wimplicit-function-declaration]
  297 |   if (!os_has_altstack_installed()) {
      |        ^~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/wasm2c/wasm-rt-impl.c
+++ b/wasm2c/wasm-rt-impl.c
@@ -293,12 +293,12 @@ void wasm_rt_init(void) {
 }
 
 bool wasm_rt_is_initialized(void) {
+#if WASM_RT_INSTALL_SIGNAL_HANDLER
 #if !WASM_RT_USE_STACK_DEPTH_COUNT
   if (!os_has_altstack_installed()) {
     return false;
   }
 #endif
-#if WASM_RT_INSTALL_SIGNAL_HANDLER
   return g_signal_handler_installed;
 #else
   return true;


sbc100 (approved):
I guess we are missing test configuration here?

keithw:
Hmm, I don't love the idea of making this change -- the altstack is required for correct operation if WASM_RT_USE_STACK_DEPTH_COUNT is false, so it seems appropriate for `wasm_rt_is_initialized()` to verify that (https://github.com/WebAssembly/wabt/pull/2332#issuecomment-1823608262).

Separately, though I don't understand the configuration that gives rise to the problem. The `os_has_altstack_installed()` function is declared and defined inside an `#if !WASM_RT_USE_STACK_DEPTH_COUNT`. And the line pointed to (that uses this function) is within the same `#if` test. So... how does the error arise, and can you help us replicate it or contribute a CI workflow that captures this situation to avoid this happening next time?

wrv:
`os_has_altstack_installed()` is declared and defined if `WASM_RT_INSTALL_SIGNAL_HANDLER` is defined and `WASM_RT_USE_STACK_DEPTH_COUNT` is not defined: https://github.com/WebAssembly/wabt/blob/main/wasm2c/wasm-rt-impl.c#L166-L177C39 

In our case, neither `WASM_RT_INSTALL_SIGNAL_HANDLER` nor `WASM_RT_USE_STACK_DEPTH_COUNT` are defined, so we get the implicit declaration error.

I've been working on a CI configuration here: https://github.com/wrv/wabt/actions/runs/7042089655/workflow#L126-L132 but I'm getting some ASAN errors now when `WASM_RT_USE_STACK_DEPTH_COUNT` is not defined.

wrv:
@keithw in the paragraph you referenced, you say (emphasis added)
> wasm_rt_is_initialized() is going to start enforcing that there's an altstack in place *if we're using signals to catch stack exhaustion*

so would it not make sense to move the altstack code within `WASM_RT_INSTALL_SIGNAL_HANDLER `, as this PR proposes, and the rest of the code abides by?

On the CI side, this is the relevant combination that [RLBox uses](https://github.com/PLSysSec/rlbox_wasm2c_sandbox/blob/master/CMakeLists.txt#L224C1-L227C78):
```
WASM_RT_USE_MMAP=1
WASM_RT_SKIP_SIGNAL_RECOVERY=1
WASM_RT_USE_STACK_DEPTH_COUNT=0
```

In attempting to test the configurations, I'm stuck on the best way to fix the failing tests. All the failing tests are an `ASSERT_EXHAUSTION` , but with no stack depth count, the `WASM_RT_TRAP_EXHAUSTION` is never thrown. Here is a recent run: https://github.com/wrv/wabt/actions/runs/7051406861/job/19194215328 . Can y'all provide insight on how to best address this?

PR #2340, opened 2023-11-30
wasm2c: Implement big-endian SIMD

SoniEx2:
~~This PR does not include the necessary test harness changes, we'll add those in another PR along with some regression tests too.~~

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -1297,9 +1297,11 @@ void CWriter::Write(const Const& const_) {
       break;
     }
     case Type::V128: {
-      Writef("simde_wasm_i32x4_const(0x%08x, 0x%08x, 0x%08x, 0x%08x)",
-             const_.vec128().u32(0), const_.vec128().u32(1),
-             const_.vec128().u32(2), const_.vec128().u32(3));
+      Writef("v128_const(0x%02x", const_.vec128().u8(0));
+      for (int i = 1; i < 16; i++) {
+        Writef(", 0x%02x", const_.vec128().u8(i));
+      }
+      Write(")");
       break;
     }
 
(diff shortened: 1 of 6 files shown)


keithw:
Looks fine. Do you think it would be reasonable and a good idea to split out the big-endian stuff into a separate template and conditionally include the result? (Sort of an out-of-sight/out-of-mind that might make this codebase a little less intimidating?)

SoniEx2:
Not really. The generated code (wasm2c output) should be the same whether the host is little-endian or big-endian, so you can run wasm2c on a little-endian host and build the result on a big-endian host and vice-versa.

keithw:
Oh, definitely agreed -- but I mean, do you think we would make these files more approachable if we separated out the big-endian implementations into a new template (just like we did with SIMD & atomic operations).

SoniEx2:
SIMD and atomic are conditionally emitted, while big-endian... is not. intentionally so.

and there's no good way to split them since they're wrapped in a big `#if`/`#else`/`#endif`.

keithw:
Yeah, I meant -- is it cleaner / less imposing to move the big-endian stuff into a new template and have c-writer emit the prebuilt version surrounded by an `#ifdef WABT_BIG_ENDIAN` / `#endif` around the whole thing? "No, not cleaner" is a reasonable answer, but I don't think I'm asking about something mechanistically impossible.

SoniEx2:
we suppose... it's only really relevant for SIMD, because SIMD is a bit weird in that it acts like a 128-bit slice of memory. the implementation would likely be much simpler upstream, but we're not sure if there's anything we can do about that...

(for context, interpreter doesn't use simde and has a much simpler big-endian implementation.)

keithw (approved):
Ok, lgtm.

SoniEx2:
we no longer think it would be necessarily better to push this onto simde and would rather just have it in wabt.

sbc100:
> we no longer think it would be necessarily better to push this onto simde and would rather just have it in wabt.

Who is the we in this context?

sbc100:
> just us (Soni).
> 
> simde "supports" wasm simd128 on big-endian platforms in the sense that their tests pass on big-endian platforms, but... it doesn't support endian-sensitive operations on those simd128 (it has no tests for endian-sensitive operations). in practice their implementation will follow platform endianness, so they effectively have a "big-endian simd128" and our patch (this PR) just makes sure wabt can get the desired little-endian semantics out of it.
> 
> in some cases, simde has optimized implementations of these operations and we can still take advantage of those optimized implementations when doing the necessary adaptations, while our initial attempt to "upstream" these changes would've effectively eliminated those optimized implementations. the work of "upstreaming" these would require a fairly extensive refactoring of simde that we personally feel is not worth the trouble. it would be harder for simde to maintain than it is for wabt.

I see, thanks for the explanation.

@keithw does this change look good to land now?

keithw:
Yes, still good to land (I approved back in December). Will merge shortly if no further objection?

SoniEx2:
poke?

keithw:
No objection heard, merging.

PR #2356, opened 2023-12-15
wasm2c: Add macro and tests to allow disabling stack exhaustion checks

shravanrn:
(Continued from https://github.com/WebAssembly/wabt/pull/2354)

This PR adds an macro `WASM_RT_NONCONFORMING_STACK_UNCHECKED` to allow the embedder to unsafely remove all stack checks from wasm2c. This would mean the embedder is not conformant with the Wasm spec. This configuration is useful for embedders like RLBox which are unlikely to be affected by stack exhaustion, and don't want to pay performance penalties for this. The wasm2c build in CI that uses rlbox flags has been correspondingly updated to capture this.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -172,7 +172,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       USE_NINJA: "1"
-      WASM2C_CFLAGS: "-DWASM_RT_USE_MMAP=1 -DWASM_RT_SKIP_SIGNAL_RECOVERY=1 -DWASM2C_TEST_EMBEDDER_SIGNAL_HANDLING"
+      WASM2C_CFLAGS: "-DWASM_RT_USE_MMAP=1 -DWASM_RT_SKIP_SIGNAL_RECOVERY=1 -DWASM_RT_NONCONFORMING_UNCHECKED_STACK_EXHAUSTION=1 -DWASM2C_TEST_EMBEDDER_SIGNAL_HANDLING"
     steps:
     - uses: actions/setup-python@v1
       with:
@@ -182,5 +182,5 @@ jobs:
         submodules: true
     - run: sudo apt-get install ninja-build
     - run: make clang-debug
-    - name: tests (excluding memory64)
-      run: ./test/run-tests.py --exclude-dir memory64
+    - name: tests (wasm2c tests excluding memory64)
+      run: ./test/run-tests.py wasm2c --exclude-dir memory64
--- a/wasm2c/wasm-rt.h
+++ b/wasm2c/wasm-rt.h
@@ -166,11 +166,20 @@ extern "C" {
 #define WASM_RT_INSTALL_SIGNAL_HANDLER 0
 #endif
 
+/* This macro, if defined, allows the embedder to disable all stack exhaustion
+ * checks. This a non conformant configuration, i.e., this does not respect
+ * Wasm's specification, and may compromise security. Use with caution.
+ */
+#ifndef WASM_RT_NONCONFORMING_UNCHECKED_STACK_EXHAUSTION
+#define WASM_RT_NONCONFORMING_UNCHECKED_STACK_EXHAUSTION 0
+#endif
+
 /* We need to detect and trap stack overflows. If we use a signal handler on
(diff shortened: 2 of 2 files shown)


test/run-tests.py
```diff
@@ -928,10 +929,10 @@ def main(args):
                         action='store_true')
     parser.add_argument('patterns', metavar='pattern', nargs='*',
                         help='test patterns.')
-    parser.add_argument('--exclude-dir', action='append', default=[],
-                        help='directory to exclude.')
+    parser.add_argument('--exclude', action='append', default=[],
+                        help='directory names or file names to exclude.')
```
sbc100, line 933: not: How about "file or directory names to exclude"

sbc100 (commented):
Seems reasonable to me.   IIUC this approach is what you and @keithw disccussed in #2354?

shravanrn:
> Seems reasonable to me. IIUC this approach is what you and @keithw disccussed in #2354?

Yup, I believe so. 
@keithw is this what you had in mind?

test/spec-wasm2c-prefix.c
```diff
@@ -373,6 +373,7 @@ static void posix_signal_handler(int sig, siginfo_t* si, void* unused) {
 }
 
 static void posix_install_signal_handler(void) {
+#if !WASM_RT_NONCONFORMING_STACK_UNCHECKED
```
keithw, line 376: Do we need these changes to spec-wasm2c-prefix.c?

shravanrn: No strong opinion. This was done so that the rlbox wasm2c build in the CI could run tests in an environment using a setup that is as close to the real use as possible. Since the real use does not apply a stack exhaustion handler, we expect tests relying on stack-exhaustion-traps to fail, and this captures the expected behavior.

If you prefer not to do this, I can remove this (and also re-add the stack exhaustion tests in the rlbox wasm2c build in the CI)

keithw: I can see both sides, but I guess on balance my preference (not that strong either) would be to avoid adding branches to the spec-test harness to accommodate a nonconforming use case; it doesn't feel like the complexity is worth it. No preference on whether you add those tests or not though -- dwyw.

keithw:
lgtm % comment. Could we call it `WASM_RT_NONCONFORMING_UNCHECKED_STACK_EXHAUSTION` or something to be very clear about what's "unchecked" about this?

shravanrn:
@keithw Done. Could you sign off on the change when you have a chance?

.github/workflows/build.yml
```diff
@@ -182,5 +182,5 @@ jobs:
         submodules: true
     - run: sudo apt-get install ninja-build
     - run: make clang-debug
-    - name: tests (excluding memory64)
-      run: ./test/run-tests.py --exclude-dir memory64
+    - name: tests (wasm2c tests excluding memory64)
+      run: ./test/run-tests.py *wasm2c* --exclude-dir memory64
```
keithw, line 186: I don't *think* it needs the stars here (which might get interpreted by the shell) -- I think just `./test/run-tests.py wasm2c --exclude-dir memory64` is enough.

keithw (approved):
lgtm % comment. I like how minimal this is...

keithw:
Just saw @sunfishcode's comment on the other thread -- let's hold off on merging (especially since this one is more about sandboxing) to try to get consensus.

shravanrn:
> Just saw @sunfishcode's comment on the other thread -- let's hold off on merging (especially since this one is more about sandboxing) to try to get consensus.

Alright. Fwiw, if we end up not wanting to expose the non-conforming approach on this change, the best path forward for us is to go back to the original commit I had here --- where we add a macro of the form `WASM_RT_STACK_EXHAUSTION_CHECKED_BY_EMBEDDER` which basically say its the host's responsibility to take care of this (by installing signal handlers etc.). We can do what we need to from the rlbox side while wasm2c can say it doesn't promote explicit non-conformance

shravanrn:
@keithw Given the discussion in #2357, I think we can now retry landing this. Please let me know your thoughts

keithw:
Go for it!

PR #2359, opened 2024-01-03
Handle zero local count in disassembly

rvagg:
Previously: `local[0..4294967295]`
Current: `local[]`


rvagg:
An alternative here is to just not print them - perhaps that would be the more sensible approach?

test/binary/function-local-count-zero-dump.txt
```diff
@@ -0,0 +1,47 @@
+;;; TOOL: run-objdump-gen-wasm
+;;; ARGS: -x
```
sbc100, line 2: Maybe put this file under `test/dump/`

sbc100 (commented):
Yes, I think simply not printing them makes sense.

test/binary/function-local-count-zero-dump.txt
rvagg: ok, done, but this file is very similar `test/binary/function-local-count-zero.txt` that already exists in there, but it uses `TOOL: run-gen-wasm`.

rvagg:
> not printing them

Done

A case against not printing them is data hiding - they were declared in there for _some reason_ and hiding them from someone trying to inspect binaries could be misleading, maybe it's best to have maximum information about the bytes in there?

sbc100:
> > not printing them
> 
> Done
> 
> A case against not printing them is data hiding - they were declared in there for _some reason_ and hiding them from someone trying to inspect binaries could be misleading, maybe it's best to have maximum information about the bytes in there?

I'm not sure I agree.  wasm-objdump hides all kind of information and basically just prints a nice human readable summary.    I can't think of any reason why one would even deliberately declare zero locals of a certain type.. can you? 

However, I don't feel at all strongly about this so whichever way you prefer seems fine with me.

test/binary/function-local-count-zero-dump.txt
sbc100: Hmm... I see.   Maybe just update that test to use `run-objdump-gen-wasm`?  Would be loosing any coverage by doing that?

rvagg: Yeah, I think so, it'd be great if they could both be done in one file but in lieu of that a separate file seemed sensible.

What would be lost is coverage of the ability to convert this:

```
  func {
    local_decls[4]
    locals[1] i64
    locals[0] i32
    locals[2] f64
    locals[0] f32
  }
```

to this:

```
  (func (;1;) (type 0)
    (local i64 f64 f64)))
```

i.e. ignoring the zero-locals when generating the text format.

So it's a related problem, but more strict bounds since it would be invalid to have them in the text format even if they exist in the binary.

rvagg:
> I can't think of any reason why one would even deliberately declare zero locals of a certain type.. can you?

I've been trying to think up some hypotheticals for this, given that it's even legal to declare zero locals, which seems odd. The best version I can come up with is a compile tool that always declares all types in a fixed order, regardless of whether they're used or not, and just sets the count when they are used. There might be compile pipeline reasons that makes it easier to keep local index references by doing that. :shrug:

My preference would be to show as much information as is practical, it's particularly useful to compiler builders to see what they're doing, not so much for programmers looking for optimisations. Would you mind if I take off the latest commit and restore to the previous state?

There's also another cases I came across of data being obscured in objdump disassemble—printing the full `0xfc` and `0xfd` opcode strings, but I'll raise that in another issue, I can't work out whether it's intentional or not.

rvagg:
Removed latest commit, back to original which shows the zero locals and leaves the test file in the same directory as its friend.

Also for completeness I'll add that the core spec does have a test that asserts that zero locals is permissible and doesn't error: https://github.com/WebAssembly/spec/blob/48771711fabb3604fccf236f2bdb33afb0ad2aca/test/core/binary.wast#L384-L397

sbc100:
> Removed latest commit, back to original which shows the zero locals and leaves the test file in the same directory as its friend.
> 
> Also for completeness I'll add that the core spec does have a test that asserts that zero locals is permissible and doesn't error: https://github.com/WebAssembly/spec/blob/48771711fabb3604fccf236f2bdb33afb0ad2aca/test/core/binary.wast#L384-L397

Given that we run the entire test suite I wonder if that means we can now remove the downstream custom test we have (the existing one).

test/binary/function-local-count-zero-dump.txt
sbc100: It looks like you can add `;;; TOOL: run-objdump-gen-wasm` to the existing test and perhaps get the result you want without creating a new test file (tests seem to allow multiple `TOOL` lines)

rvagg: Oh, good call, I didn't think this was possible. Added now and it works.

rvagg:
> Given that we run the entire test suite

But not a disassembly dump of it; at least it's not shown up here in test failures. Maybe that would be a good addition though, it's something I've been doing with my own little toy project and it's been highlighting all sorts of edge cases.

sbc100 (approved):

PR #2365, opened 2024-01-12
Fix Features of wabt.post.js can't work in strict mode

Changqing-JING:
`for ([f, v] of Object.entries(FEATURES))` leads to `ReferenceError: f is not defined` in `use strict`. 
Fix it

### DIFF
--- a/src/wabt.post.js
+++ b/src/wabt.post.js
@@ -95,7 +95,7 @@ function allocateCString(s) {
 /// Features
 function Features(obj) {
   this.addr = Module._wabt_new_features();
-  for ([f, v] of Object.entries(FEATURES)) {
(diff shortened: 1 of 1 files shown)


sbc100 (commented):
Should we add `-sSTRICT` (or just `-sSTRICT_JS`) to the build flags to wabt.js maybe?

Changqing-JING:
@sbc100 
Thank you very much for review. `-sSTRICT_JS` can't be set, because:
The current CMakeLists has already set `-s MODULARIZE=1`, which is conflict with `-sSTRICT_JS`.
If `-sSTRICT_JS` is set, it will lead to error:
`em++: error: STRICT_JS doesn't work with MODULARIZE or EXPORT_ES6`

sbc100 (approved):

PR #2367, opened 2024-01-17
Fix emscripten build warning and add missing export

Changqing-JING:
1. Fix wabtjs bug that missing `writeAsciiToMemory` built by new version emcc
2. Refine the code quality and CI of wabtjs
2.1: Add warning as error in emcc build
2.2 Port the test cases from https://github.com/AssemblyScript/wabt.js/tree/main/tests

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -78,7 +78,7 @@ jobs:
       run: |
         docker run -di --name emscripten -v $(pwd):/src emscripten/emsdk:latest bash
         docker exec emscripten emcc -v
-        docker exec emscripten emcmake cmake -B emscripten -DWERROR=ON -DBUILD_TESTS=OFF
+        docker exec emscripten emcmake cmake -B emscripten -DWERROR=ON
         docker exec -w /src/emscripten emscripten make -j $(nproc)
         
   wasi:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -79,7 +79,7 @@ jobs:
         docker run -di --name emscripten -v $(pwd):/src emscripten/emsdk:latest bash
         docker exec emscripten emcc -v
         docker exec emscripten emcmake cmake -B emscripten -DWERROR=ON -DBUILD_TESTS=OFF
-        docker exec emscripten bash -c "cd /src/emscripten && make -j $(nproc)"
+        docker exec -w /src/emscripten emscripten make -j $(nproc)
         
   wasi:
     name: wasi
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -78,8 +78,9 @@ jobs:
       run: |
         docker run -di --name emscripten -v $(pwd):/src emscripten/emsdk:latest bash
         docker exec emscripten emcc -v
-        docker exec emscripten emcmake cmake .
-        docker exec emscripten make -j 2 VERBOSE=1
+        docker exec emscripten emcmake cmake -B emscripten -DWERROR=ON -DBUILD_TESTS=OFF
+        docker exec emscripten bash -c "cd /src/emscripten && make -j $(nproc)"
+        
   wasi:
     name: wasi
     runs-on: ubuntu-latest
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -823,6 +823,7 @@ if (EMSCRIPTEN)
     -sMODULARIZE=1
     -sEXPORT_NAME=WabtModule
     -sWASM=0
+    -sEXPORTED_RUNTIME_METHODS=writeAsciiToMemory
     -Oz
   )
   string(REPLACE ";" " " LIBWABT_LINK_FLAGS_STR "${LIBWABT_LINK_FLAGS}")
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -1633,7 +1633,7 @@ void CWriter::SerializeFuncType(const FuncType& func_type,
     *next_byte++ = MangleType(func_type.GetResultType(i));
   }
 
-  assert(next_byte - mangled_signature == len);
+  assert(next_byte - mangled_signature == static_cast<ptrdiff_t>(len));
 
   // step 4: SHA-256 the whole string
   sha256({mangled_signature, len}, serialized_type);
--- a/wasm2c/wasm-rt-impl.c
+++ b/wasm2c/wasm-rt-impl.c
@@ -148,6 +148,7 @@ static void os_cleanup_signal_handler(void) {
 #endif
 
 #else
+#if WASM_RT_USE_MMAP
 static void* os_mmap(size_t size) {
   int map_prot = PROT_NONE;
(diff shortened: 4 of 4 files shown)


Changqing-JING:
@sbc100 Sorry for creating another PR in such short time regarding wabtjs.
I just found another bug which caused by updating new version emsdk.
Beside the bug itself, recent bugs leads me to reflect the code quality of the emcc build. I think there are at least two problems:
1. Warning as error is not enabled in CI
2. No test run on the libwabt.js, so that many bug can't be detected by CI. 

I ported the testcases for Assemblyscript/wabt.js to here, then some errors in wabt.post.js should can be found by CI.

CMakeLists.txt
```diff
@@ -823,6 +823,7 @@ if (EMSCRIPTEN)
     -s MODULARIZE=1
     -s EXPORT_NAME=\"'WabtModule'\"
     -s WASM=0
+    -s EXPORTED_RUNTIME_METHODS="[writeAsciiToMemory]"
```
sbc100, line 826: You can just do `-s EXPORTED_RUNTIME_METHODS=writeAsciiToMemory`

src/c-writer.cc
```diff
@@ -1633,7 +1633,7 @@ void CWriter::SerializeFuncType(const FuncType& func_type,
     *next_byte++ = MangleType(func_type.GetResultType(i));
   }
 
-  assert(next_byte - mangled_signature == len);
+  assert(next_byte - mangled_signature == static_cast<ptrdiff_t>(len));
```
sbc100, line 1636: What is the error/warning that you see without this?

.github/workflows/build.yml
```diff
@@ -78,8 +78,18 @@ jobs:
       run: |
         docker run -di --name emscripten -v $(pwd):/src emscripten/emsdk:latest bash
         docker exec emscripten emcc -v
-        docker exec emscripten emcmake cmake .
-        docker exec emscripten make -j 2 VERBOSE=1
+        docker exec emscripten emcmake cmake -B emscripten -DWERROR=ON -DBUILD_TESTS=OFF
+        docker exec emscripten bash -c "cd /src/emscripten && emmake make -j $(nproc)"
```
sbc100, line 82: Why is this `cd` needed now but not before?

You don't need to run emmake there..  all this emscripten stuff should already be handled by `emcmake` above and embedded into the resulting `Makefile`.

sbc100 (commented):
Thanks for looking into this.

I'm not sure about adding this new test suite, especially since its comes from another project.  Perhaps you could move that part into a separate PR?

src/c-writer.cc
Changqing-JING: `/src/src/c-writer.cc:1636:40: error: comparison of integers of different signs: 'long' and 'unsigned int' [-Werror,-Wsign-compare]`

because the result of `ptr-ptr` is `ptrdiff_t`

.github/workflows/build.yml
Changqing-JING: Because before cmake was out at project root, I moved the cmake out to ./emscripten folder, which is planned for store build file for emcc(according to gitignore file).

Changqing-JING:
@sbc100 
Thank you for review, I fixed your comments and split the test files out of this PR

src/c-writer.cc
sbc100: Hmm.. I wonder why this is the only platform on which we see this?  Perhaps we don't build enough debug configurations?  I will investigate.

Changqing-JING: Because the under layer type of ptrdiff_t is implementation defined. And in most GCC and Clang implementation, it's defined as `long` or `long int`.
When compiler got following code,
```
const char * a;
const char* b;
unsigned int x;
(a-b)==x;
```
It will do implicit type conversion
```
const char * a;
const char* b;
unsigned int x;
(a-b)==static_cast<long>(x);
```

And the key point there is, clang wasm build in our CI is 32bit platform, other builds are x86_64 or arm64.
In GCC/clang, `long` is 32 bit on 32bit platform and 64 bit on 64 bit platform.
Then on 64 bit platform, cast unsigned int to long is safe, but on 32 bit platform not. 
It's the reason why only emcc pipeline throw this error.

sbc100: Thanks.  I'm working on adding a debug i686 (32-bit x86) build so that this kind of thing won't happen again: #2369

Changqing-JING:
@sbc100 Can this PR be merged? Anything else to fix?

.github/workflows/build.yml
```diff
@@ -78,8 +78,9 @@ jobs:
       run: |
         docker run -di --name emscripten -v $(pwd):/src emscripten/emsdk:latest bash
         docker exec emscripten emcc -v
-        docker exec emscripten emcmake cmake .
-        docker exec emscripten make -j 2 VERBOSE=1
+        docker exec emscripten emcmake cmake -B emscripten -DWERROR=ON -DBUILD_TESTS=OFF
+        docker exec -w /src/emscripten emscripten make -j $(nproc)
```
sbc100, line 82: Why drop the `VERBOSE=1` here?

.github/workflows/build.yml
```diff
@@ -78,8 +78,9 @@ jobs:
       run: |
         docker run -di --name emscripten -v $(pwd):/src emscripten/emsdk:latest bash
         docker exec emscripten emcc -v
-        docker exec emscripten emcmake cmake .
-        docker exec emscripten make -j 2 VERBOSE=1
+        docker exec emscripten emcmake cmake -B emscripten -DWERROR=ON -DBUILD_TESTS=OFF
```
sbc100, line 81: Why not build the tests?

wasm2c/wasm-rt-impl.c
```diff
@@ -168,6 +169,7 @@ static int os_mprotect(void* addr, size_t size) {
 static void os_print_last_error(const char* msg) {
   perror(msg);
 }
+#endif
```
sbc100, line 172: @keithw @shravanrn  WDYT about this change?

shravanrn: Sure, this should be fine assuming the existing tests pass.

.github/workflows/build.yml
Changqing-JING: Just VERBOSE level trace is not quite necessary in most cases, because build warning and error show by default. VERBOSE level trace some time makes warning and error message not quite visible inside a large mount of text.
But it's not a big issue, I can also add it back. What do you think?

.github/workflows/build.yml
Changqing-JING: Fixed, added it back

.github/workflows/build.yml
sbc100: In think as long as we have `-DWERROR=ON` when its probably find to keep verbose build output when building during CI

sbc100 (approved):

PR #2368, opened 2024-01-17
Simplify emscripten command line flags. NFC

sbc100:


keithw (approved):
Def. not an expert on these flags, but lgtm

PR #2370, opened 2024-01-19
wasm2c: Cleanup formatting of wasm-rt.h

shravanrn:
Minor formatting fixes for the wasm-rt.h header. Can be landed only after #2308 to avoid merge conflicts. (See only the last commit for the correct diff)


shravanrn:
@sbc100 This should now be ready to review and land

sbc100 (approved):

PR #2371, opened 2024-01-22
Add test files for wabtjs

Changqing-JING:
Since some bugs are caused by not enough test in wabt.post.js, such as
https://github.com/WebAssembly/wabt/pull/2367
https://github.com/WebAssembly/wabt/pull/2365

Some test cases of wabt.post.js need to be added to grantee code quality. 
One proposal is to migrate the test cases of wabt.js. Because wabt.js is just a npm wrapper of wabt, a lot of user also use wabt from wabt.js by npm install.

But if maintainer have other ideas to test wabt.post.js, I can also try them.


sbc100:
Can you remind me what downstream project this code is coming from?  Can you mention that in the descirption and link to the downstream code?

If we go ahead with this is its probably work working with downstream to migrate these tests such that they can then be removed from downstream (no point testing the same thing in multiple places).

Changqing-JING:
Thank you.
The test are ported from https://github.com/AssemblyScript/wabt.js/blob/main/tests/index.js
Yes, you are right. I can created a PR to remove the duplicate test in wabt.js if this PR is merged

PR #2378, opened 2024-01-31
Remove unnecessary restriction

SoniEx2:
Thanks to MEM_ADDR we don't need to think about big-endian vs little-endian for any scalar(!) operations.

### DIFF
--- a/src/prebuilt/wasm2c_atomicops_source_declarations.cc
+++ b/src/prebuilt/wasm2c_atomicops_source_declarations.cc
@@ -1,13 +1,6 @@
 const char* s_atomicops_source_declarations = R"w2c_template(#include <stdatomic.h>
 )w2c_template"
 R"w2c_template(
-#if WABT_BIG_ENDIAN
-)w2c_template"
-R"w2c_template(#error "wasm2c atomics not supported on big endian"
-)w2c_template"
-R"w2c_template(#endif
-)w2c_template"
-R"w2c_template(
 #ifndef WASM_RT_C11_AVAILABLE
 )w2c_template"
 R"w2c_template(#error "C11 is required for Wasm threads and shared memory support"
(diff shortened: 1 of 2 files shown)


SoniEx2:
we'll see if this passes in CI, we did not test it locally.

shravanrn (approved):
Oops, I probably should have made the change to `DEFINE_ATOMIC_XCHG` in the first place :) 

We can eliminate this "unsupported" restriction if there is some way to test this. If the CI tests this, happy to remove the restriction

sbc100:
> Oops, I probably should have made the change to `DEFINE_ATOMIC_XCHG` in the first place :)
> 
> We can eliminate this "unsupported" restriction if there is some way to test this. If the CI tests this, happy to remove the restriction
> 
> @sbc100 Does the CI test the big endian case?

I don't think we do test it no.  IIUC it is supported on best effort basis.  If @SoniEx2 has tested locally I think that is enough.

shravanrn:
Hmm... I think @SoniEx2 mentioned a local test wasn't run... Not sure how to proceed given this

SoniEx2:
currently assuming `MEM_ADDR` takes care of it, which is correct in the context of scalar ops (including scalar atomic ops). (SIMD is known to be broken on big-endian.)

shravanrn:
I think the only minor concern is that we don't necessarily want to remove the "# error" as this our way of telling end users that "we haven't tested this is BIG_ENDIAN, so don't use it production". 

But if @sbc100 is ok with this, I also don't have an objection to put it out there and seeing if any users complain about broken functionality. I'll leave this to @sbc100

SoniEx2:
honestly if `#warning` wasn't C23 we'd totally use it here.

anyway we're sure we can figure something out for testing this stuff... we think...

shravanrn:
Sounds good. Let's merge

SoniEx2:
@sbc100 sorry, needed to rebase ^^

PR #2380, opened 2024-02-01
Add s390x GHA workflow (cross-compile)

SoniEx2:
This adds a cross-compile workflow for s390x. Heavily based on #1971 , but runs significantly faster.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -185,3 +185,50 @@ jobs:
     - run: make clang-debug
     - name: tests (wasm2c tests excluding memory64)
       run: ./test/run-tests.py wasm2c --exclude-dir memory64
+
+  build-cross:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: [s390x]
+    services: # still faster on debian...
+      distcc:
+        image: debian:latest
+        options: --health-cmd distccmon-text --health-interval 5s --health-start-period 5m debian:latest bash -c "apt-get update && apt-get install -y g++-s390x-linux-gnu distcc && distccd --daemon --no-detach"
+        ports:
+          - 3632:3632
+    env:
+      QEMU_LD_PREFIX: /usr/${{matrix.arch}}-linux-gnu/
+    steps:
+    - uses: actions/setup-python@v1
+      with:
+        python-version: '3.x'
+    - uses: actions/checkout@v1
+      with:
+        submodules: true
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v2
+      with:
+        platforms: ${{matrix.arch}}
+        image: "tonistiigi/binfmt:master"
+    - name: install ninja
+      run: sudo apt-get install ninja-build
+    - name: install the toolchain
+      run: sudo apt-get install g++-${{matrix.arch}}-linux-gnu
+    - name: install distcc
+      run: sudo apt-get install distcc
+    - name: mkdir distcc symlinks
+      run: sudo mkdir -p /opt/bin/distcc_symlinks
+    - name: distcc symlink
+      run: sudo ln -s /usr/bin/distcc /opt/bin/distcc_symlinks/${{matrix.arch}}-linux-gnu-gcc # only CC is needed
+    - name: cmake
+      run: cmake -S . -B out -G Ninja -DCMAKE_TOOLCHAIN_FILE=../scripts/TC-${{matrix.arch}}.cmake -DWITH_WASI=ON -DWERROR=OFF -Werror=dev -Wno-deprecated
+    - name: build
+      run: cmake --build out
+    - name: check if generated files are up-to-date
+      run: python ./scripts/check_clean.py
+    - name: unittests
+      run: cmake --build out --target run-unittests
+    - name: tests
+      run: cmake --build out --target run-tests
--- a/scripts/TC-s390x.cmake
+++ b/scripts/TC-s390x.cmake
@@ -1,0 +1,11 @@
+set(CMAKE_SYSTEM_NAME Linux)
+
+set(CMAKE_C_COMPILER /opt/bin/distcc_symlinks/s390x-linux-gnu-gcc)
+set(CMAKE_CXX_COMPILER s390x-linux-gnu-g++)
+
+set(CMAKE_FIND_ROOT_PATH /usr/s390x-linux-gnu)
+
+set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
+
+set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
+set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)


sbc100 (commented):
Change itself looks good!

If the step take too long we could run it post-commit I support.. or make it optional to wait for it to complete.  Lets see how long it takes..

SoniEx2:
just a *bit* faster =^-^=

also, cc @alexreinking ?

.github/workflows/build.yml
```diff
@@ -185,3 +185,62 @@ jobs:
...
+      run: sudo apt-get install distcc
+    - name: mkdir distcc symlinks
+      run: sudo mkdir -p /opt/bin/distcc_symlinks
+    - name: distcc symlink
+      run: sudo ln -s /usr/bin/distcc /opt/bin/distcc_symlinks/${{matrix.arch}}-linux-gnu-gcc # only CC is needed
+    - name: mkdir
+      run: mkdir -p out
+    - name: cmake
+      run: cmake -DCMAKE_TOOLCHAIN_FILE=../scripts/TC-${{matrix.arch}}.cmake .. -G Ninja -DWITH_WASI=ON -DWERROR=OFF -Werror=dev -Wno-deprecated
+      working-directory: out
```
alexreinking, line 228: This can be simplified a little:
```suggestion
    - name: cmake
      run: cmake -G Ninja -S . -B out -DCMAKE_TOOLCHAIN_FILE=./scripts/TC-${{matrix.arch}}.cmake -DWITH_WASI=ON -DWERROR=OFF -Werror=dev -Wno-deprecated
```

.github/workflows/build.yml
```diff
@@ -185,3 +185,62 @@ jobs:
...
+      fail-fast: false
+      matrix:
+        arch: [s390x]
+    services: # still faster on debian...
+      distcc:
+        image: debian:latest
+        options: --health-cmd distccmon-text --health-interval 5s --health-start-period 5m debian:latest bash -c "apt-get update && apt-get install -y g++-s390x-linux-gnu distcc && distccd --daemon --no-detach"
+        ports:
+          - 3632:3632
+    steps:
```
alexreinking, line 201: Another simplification:

```suggestion
    env:
      CMAKE_GENERATOR: Ninja
      QEMU_LD_PREFIX: /usr/${{matrix.arch}}-linux-gnu/
    steps:
```

Then get rid of `-G Ninja` and the other `QEMU_LD_PREFIX` settings below.

alexreinking:
Super happy to see this effort picked back up -- and super happy to see it improved! I commented a couple easy ways to simplify the workflow steps.

I no longer work at Google, though, so I'd appreciate it if you would change the co-authored-by email to `alex.reinking@gmail.com`.

.github/workflows/build.yml
sbc100: We use `-G Ninja` elsewhere in this file though... so maybe leave that one as is for consistency?

alexreinking: @sbc100 - totally non-critical, just a suggestion!  I think `QEMU_LD_PREFIX` does lower the risk for forgetting it if the workflow has to change.

.github/workflows/build.yml
SoniEx2: we would rather keep this consistent with the `build` job, aside from the toolchain file (and, currently, the `-DWERROR=OFF`).

alexreinking: This isn't consistent everywhere in this file, though... look at line 161 for instance. I'd sooner move the file towards having fewer unnecessary steps (which just add noise to logs)

.github/workflows/build.yml
SoniEx2: that works...? well, let's find out! (github actions documentation is not exactly the easiest thing to read...)

.github/workflows/build.yml
SoniEx2: hmm... @sbc100 thoughts?

sbc100: Maybe make separate PR to convert to using `-S + -B` everywhere before we lang this change?

alexreinking: The `-S` and `-B` flags were introduced in CMake 3.13 while the minimum version for wabt is CMake 3.16 and has been for a while. No hazard with using them here.

SoniEx2: hmm even the places that use `-S + -B` aren't consistent about order. would you prefer `-S . -B out -G Ninja` or `-G Ninja -S . -B out`?

sbc100: No opinion there..

sbc100:
lgtm in general, but lets see how long it takes (and make sure it passes :)

SoniEx2:
it'll pass after #2382 and #2340 .

SoniEx2:
the only blocker now is #2340 , it'd be really nice to get this merged...

SoniEx2:
there we go!

SoniEx2:
does this need anything else?

sbc100:
> does this need anything else?

How long does it take compared the current slowest CI step?

keithw:
The previous slowest step was the CIFuzz (10 minutes); this one takes about 17-18 minutes.

.github/workflows/build.yml
```diff
@@ -185,3 +185,55 @@ jobs:
...
+    - name: build
+      run: cmake --build out
+    - name: check if generated files are up-to-date
+      run: python ./scripts/check_clean.py
+    - name: unittests
+      run: cmake --build out --target run-unittests
+      if: ${{ matrix.arch != 's390x' }} # currently broken...
+    - name: c-api-tests
+      run: cmake --build out --target run-c-api-tests
+      if: ${{ matrix.arch != 's390x' }}
```
sbc100, line 237: Give that s390x is the only arch these blocks are now dead code aren't they?

.github/workflows/build.yml
```diff
@@ -185,3 +185,55 @@ jobs:
     - run: make clang-debug
     - name: tests (wasm2c tests excluding memory64)
       run: ./test/run-tests.py wasm2c --exclude-dir memory64
+
+  build-arch:
```
sbc100, line 189: How about calling this `build-s390x`?   Maybe `build-arch` harks of Arch linux?

sbc100 (approved):
lgtm % nits

.github/workflows/build.yml
SoniEx2: this is explicitly intended to make it easy to add more archs (arm64 anyone?) but you're right that there's a slight readability tradeoff here.

.github/workflows/build.yml
SoniEx2: hmm, maybe `build-cross`?

sbc100: Sure! sgtm

.github/workflows/build.yml
sbc100: I guess I'd rather see that added later once we do actually want more than one cross arch tested here.

But I don't feel too strongly, since its not much code and its already written.

.github/workflows/build.yml
SoniEx2: done!

SoniEx2:
btw now that we have #2386 we re-enabled run-unittests!

(that just leaves run-c-api-tests as dead code)

SoniEx2:
btw qemu does not emulate stack overflows correctly, would have to run the entire thing in machine emulation for that. maybe something to do as future work? (not really relevant when not using signal handler tho)

SoniEx2:
is there anything else missing for this? (we took out the c-api-tests since c-api is unsupported on big-endian... maybe component model will provide a better bridge between big-endian and wasm...)

SoniEx2:
big endian*

we really care about old macs, amigas, the wii, the sega genesis. retro computing really. not so much the s390x, the s390x is a bit out of reach for us. maybe one day it'll become retro computing tho! that would be kinda cool.

sbc100:
> big endian*
> 
> we really care about old macs, amigas, the wii, the sega genesis. retro computing really. not so much the s390x, the s390x is a bit out of reach for us. maybe one day it'll become retro computing tho! that would be kinda cool.

Very interesting.  And you are wanting to run wasm on actual hardware using these old devices?   Or are you wanting to run wasm inside emulators of these devices?

SoniEx2:
yep! it'd be cool to run it on real hardware =^-^=

PR #2382, opened 2024-02-01
Do not use guard pages on big-endian

SoniEx2:


SoniEx2:
honestly we don't feel like going through the effort of supporting guard pages on big-endian when we can use the more portable bounds checking. it would be nice to have, sure, but it shouldn't be a blocker.

sbc100 (approved):

PR #2385, opened 2024-02-06
fallback to pthreads if threads aren't available

walkingeyerobot:
glibc was slow to implement threads, whereas pthreads are much more readily available. I'm on a platform where my glibc is old enough that it doesn't have threads.h and I'm unable to upgrade, but everything works find if I just use pthreads.

### DIFF
--- a/wasm2c/wasm-rt.h
+++ b/wasm2c/wasm-rt.h
@@ -58,23 +58,19 @@ extern "C" {
 #endif
 
 /**
- * Apple and Windows devices don't implement the C11 threads.h. We use pthreads
- * on Apple devices, and CriticalSection APIs for Windows.
+ * Many devices don't implement the C11 threads.h. We use CriticalSection APIs
+ * for Windows and pthreads everywhere else.
  */
 #ifdef WASM_RT_C11_AVAILABLE
 
-#ifdef __APPLE__
-#include <pthread.h>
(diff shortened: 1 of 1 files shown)


wasm2c/wasm-rt.h
```diff
@@ -58,19 +58,19 @@ extern "C" {
...
-#ifdef __APPLE__
-#include <pthread.h>
-#define WASM_RT_MUTEX pthread_mutex_t
-#define WASM_RT_USE_PTHREADS 1
-#elif defined(_WIN32)
+#if defined(_WIN32)
 #include <windows.h>
 #define WASM_RT_MUTEX CRITICAL_SECTION
 #define WASM_RT_USE_CRITICALSECTION 1
+#elif defined(__STDC_NO_THREADS__)
```
shravanrn, line 70: Can we change this to `#elif defined(__APPLE__) || defined(__STDC_NO_THREADS__)`

I haven't checked if there are non-conforming libc's in older Apple devices, so it seems safer.

shravanrn (changes requested):

wasm2c/wasm-rt.h
walkingeyerobot: sure. done.

shravanrn (approved):

PR #2386, opened 2024-02-08
Fix rot13 test on big-endian platforms

SoniEx2:
arguably more like "implement big-endian support in rot13 test" rather than a "fix" but anyway.

### DIFF
--- a/src/test-interp.cc
+++ b/src/test-interp.cc
@@ -505,8 +505,13 @@ TEST_F(InterpTest, Rot13) {
 
     EXPECT_LT(ptr + size, memory->ByteSize());
 
+#if WABT_BIG_ENDIAN
+    std::copy(string_data.rbegin(), string_data.rbegin() + size,
+              memory->UnsafeData() + memory->ByteSize() - ptr - size);
+#else
     std::copy(string_data.begin(), string_data.begin() + size,
               memory->UnsafeData() + ptr);
+#endif
 
     results[0].Set(size);
     return Result::Ok;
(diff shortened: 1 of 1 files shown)


src/test-interp.cc
```diff
@@ -505,8 +505,13 @@ TEST_F(InterpTest, Rot13) {
 
     EXPECT_LT(ptr + size, memory->ByteSize());
 
+#if WABT_BIG_ENDIAN
+    std::copy(string_data.rbegin(), string_data.rbegin() + size,
+              memory->UnsafeData() + memory->ByteSize() - ptr - size);
+#else
     std::copy(string_data.begin(), string_data.begin() + size,
               memory->UnsafeData() + ptr);
+#endif
```
sbc100, line 514: It it worth extracting a named function here?

sbc100 (commented):
lgtm % comment

src/test-interp.cc
SoniEx2: it's the only test that needs this (*currently) so probably not yet?

sbc100: What I mean is that this logic seems to be duplicated with lines 535 to 542 below.   Would a "CopyUserMemory" helper function make sense?

SoniEx2: not in this context, these are slightly different operations (copy to memory vs copy from memory) and we can't think of a good way to abstract over both of them.

the APIs between host and wasm are rather lacking in pretty much every standalone wasm engine we've tried. allegedly interface types were supposed to help with this but uh... wabt doesn't really have those yet. (and probably won't for a while...)

sbc100 (approved):

SoniEx2:
@sbc100 sorry, try again, it needed to be rebased

PR #2387, opened 2024-02-08
clang-format-diff: fix failing CI for newly pushed branches

SoniEx2:
Don't try to find upstream in detached HEAD

HEAD is detached for new branches.

- - -

This is only really relevant when creating a new branch (e.g. in a fork, before submitting a PR) but it's been bothering us *forever* so we decided to fix it.

Before: https://github.com/SoniEx2/wasm2kotlin/actions/runs/7830157694/job/21363568601

After: https://github.com/SoniEx2/wasm2kotlin/actions/runs/7830660258/job/21365229032

### DIFF
--- a/scripts/clang-format-diff.sh
+++ b/scripts/clang-format-diff.sh
@@ -5,12 +5,23 @@ set -o pipefail
 
 if [ -n "$1" ]; then
   BRANCH="$1"
+elif [ "$CI" != "true" ]; then
+  echo "Please specify a base branch in the command line"
+  exit 1
 elif [ -n "$GITHUB_EVENT_BEFORE" ] && [ "push" = "$GITHUB_EVENT_NAME" ]; then
   BRANCH="$GITHUB_EVENT_BEFORE"
 elif [ -n "$GITHUB_BASE_REF" ]; then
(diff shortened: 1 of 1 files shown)


scripts/clang-format-diff.sh
```diff
@@ -9,8 +9,10 @@ elif [ -n "$GITHUB_EVENT_BEFORE" ] && [ "push" = "$GITHUB_EVENT_NAME" ]; then
   BRANCH="$GITHUB_EVENT_BEFORE"
 elif [ -n "$GITHUB_BASE_REF" ]; then
   BRANCH="origin/$GITHUB_BASE_REF"
-else
+elif git symbolic-ref -q HEAD; then
```
sbc100, line 12: Maybe add a comment here?   "Default to `origin/main` if we are in a detached head"?

Can you update to PR title to be something like that too?

Is the defaulting to `origin/main` always make sense?  What if the someone is using `upstream` as the name of their remote?

SoniEx2: honestly we don't have enough context for how this script came to be so we're just gonna restrict it to CI unless passed a branch manually. let us know if that breaks anyone's workflow.

sbc100 (approved):

PR #2389, opened 2024-02-10
c-writer.cc: represent float consts as integer literals

### LINKED ISSUE #2388: [wasm2c] MSVC misscompiles for certain fp constants
Given WASM
```wasm
0000024: 44                                        ; f64.const
0000025: 0000 0000 0000 e0c1                       ; f64 literal

(module
  (func (export "addTwo") (result f64)
    f64.const -0x1p+31
  )
)
```
and wasm2c conversion output
```c
//wasm2c test.wasm
f64 w2c__addTwo_0(w2c_* instance) {
  FUNC_PROLOGUE;
  f64 var_d0;
  var_d0 = -2147483648;
  FUNC_EPILOGUE;
  return var_d0;
}
```
there is slight discrepancy in output of different compilers.
Clang & GCC correctly output `-2147483648.000000`
MSVC throws warning (error with /sdl) and outputs `2147483648.000000` (notice absence of negative mark)
https://godbolt.org/z/MrjMzdeKE
image

While the conversion output is likely correct according to C spec (at least Clang and GCC are not screaming with -Wall), it's very inconvenient for Windows users to do these fixups manually.

Proposal is to use alternative notation
```c
  var_d0 = -2147483648.0; //Notice additional decimal
```
which makes all tested compilers happy and correct.
https://godbolt.org/z/5sE4adbYd
image

keithw:
Fixes #2388

WIP PR -- should add a test


SoniEx2:
it would be nice to get this merged...

sbc100 (approved):

SoniEx2:
presumably this can be closed now

PR #2351, opened 2023-12-11
Silence compile warning

jwatt:
warning: comparison of integers of different signs: 'uint64_t' (aka 'unsigned long long') and 'int' [-Wsign-compare]

Issue #25, opened 2016-01-24
Incorrect code emitted involving unreachable code.


JSStats:
The following code emits a single `unreachable` operator which does not appear correct. When $i1 is zero this function would return 1.

```
(module
 (func (param $i1 i32) (result i32)
   (if_else (get_local $i1)
     (block (unreachable))
     (i32.const 1))))
```

```
0000004: 02                                         ; WASM_SECTION_FUNCTIONS
0000005: 01                                         ; num functions
; function 0
0000006: 00                                         ; func flags
0000007: 0000                                       ; func signature index
0000009: 0000                                       ; func body size
000000b: 01                                         ; OPCODE_BLOCK
000000c: 01                                         ; num expressions
000000d: 15                                         ; OPCODE_UNREACHABLE
0000009: 0300                                       ; FIXUP func body size
000000e: 06                                         ; WASM_SECTION_END
```

Some discussion at https://github.com/WebAssembly/design/issues/527

binji:
Just tested this and got the following output:
```
0000000: 01                                         ; WASM_SECTION_SIGNATURES
0000001: 01                                         ; num signatures
; signature 0
0000002: 01                                         ; num params
0000003: 01                                         ; result_type
0000004: 01                                         ; param type
0000005: 02                                         ; WASM_SECTION_FUNCTIONS
0000006: 01                                         ; num functions
; function 0
0000007: 00                                         ; func flags
0000008: 0000                                       ; func signature index
000000a: 0000                                       ; func body size
000000c: 04                                         ; OPCODE_IF_THEN
000000d: 0e                                         ; OPCODE_GET_LOCAL
000000e: 00                                         ; remapped local index
000000f: 01                                         ; OPCODE_BLOCK
0000010: 01                                         ; num expressions
0000011: 15                                         ; OPCODE_UNREACHABLE
0000012: 09                                         ; OPCODE_I8_CONST
0000013: 01                                         ; u8 literal
000000a: 0800                                       ; FIXUP func body size
0000014: 06                                         ; WASM_SECTION_END
```
Looks right to me.

JSStats:
Sorry for the noise, my mix-up.

PR #2400, opened 2024-03-14
wasm2c: Fix warning on signed/unsigned comparison

shravanrn:


sbc100 (approved):

keithw (approved):
lgtm -- dupe of #2351 (I have the same question I had there but I assume the answer is basically "it happens on a newer compiler")

shravanrn:
> lgtm -- dupe of #2351 (I have the same question I had there but I assume the answer is basically "it happens on a newer compiler")

I think it's more the case that this code is executed only if there is a custom "grow_failed" handler. So, we don't see the warning in this CI, but it shows up on the Firefox build. I'll update the RLBox wasm2c test to integrate a custom grow failed handler when I get the chance to make sure this code is exercised.


Separately, any idea if there is something wrong with the CI runner? I'm trying to track down the CI crash but it seems like the crashes aren't related to this commit?

sbc100:
Interesting, binaryen just ran into some santizer-only failures that were unrelated to any changes to binaryen itself. @kripken does this looks like the same issue?

kripken:
Hmm, it could be! Same symptoms, which are the ASan build just exits with an error, no message, nothing else. We also saw a TSan error which actually did contain an error message.

https://github.com/WebAssembly/binaryen/pull/6396 has my investigation and fixes in Binaryen. Basically my best guess is that the github actions bots changed something, and recent LLVM versions are needed to handle it. Clang 17 is enough for ASan, but clang 18 is needed to also fix TSan if you want that as well.

keithw:
We're having this problem on a lot of our students' machines too. The recent update to Ubuntu 23.10 (to kernel 6.5.0-25-generic #25-Ubuntu) seems to have broken a lot of the ASan output from g++ 13.2 and clang++ 16.0.6. Apparently there's some ASLR-related change? Haven't searched for more yet. :-(

shravanrn:
Hmm... not sure what is the best way to proceed. Would it be reasonable to disable the failing CI builds for now? Happy to wait to land if someone is working on fixing the CI already (I would help but I unfortunately don't have too many cycles at the moment).

shravanrn:
Testing a fix per
https://github.com/actions/runner/issues/3207#issuecomment-2000066889

shravanrn:
Looks like that fix worked. Will spin this out into its own PR

PR #2401, opened 2024-03-17
wasm2c: Fix periodic ASLR related crashes in sanitizer builds

shravanrn:
Fix random sanitizer crashes encountered in #2400. Fix is from 
https://github.com/actions/runner/issues/3207#issuecomment-2000066889

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -116,6 +116,8 @@ jobs:
       with:
         submodules: true
     - run: sudo apt-get install ninja-build
+    - name: workaround for ASLR+ASAN compatibility # See https://github.com/actions/runner/issues/3207
+      run: sudo sysctl -w vm.mmap_rnd_bits=28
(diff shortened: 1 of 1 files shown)


sbc100:
Oh interesting.    Do you think this will be permanently required fix?   I wonder if you can link to some kind of upstream bug which, when resolved, would mean we could remove this?

I assume this would have been an alternative way to fix binaryen in https://github.com/WebAssembly/binaryen/pull/6396

.github/workflows/build.yml
```diff
@@ -116,6 +116,8 @@ jobs:
       with:
         submodules: true
     - run: sudo apt-get install ninja-build
+    - name: workaround for ASLR+ASAN compatibility
```
sbc100, line 119: Can you add a link to https://github.com/actions/runner/issues/3207 here in comment ?

shravanrn:
@sbc100 Added a link in comments

sbc100 (approved):

PR #2404, opened 2024-03-25
Suppress pass-failed warnings/errors

### LINKED ISSUE #2403: Error while running testsuite (simd_lane, simd_load) "loop not vectorized"
- wabt 1.0.34
- clang 17.0.6

Extracted lines:
```
out/test/wasm2c/spec/simd_lane/simd_lane.0.c:967:6: error: loop not vectorized: the optimizer was unable to perform the requested transformation; the transformation might be disabled or specified as part of an unsupported transformation ordering [-Werror,-Wpass-failed=transform-warning]
out/test/wasm2c/spec/simd_load/simd_load.12.c:803:6: error: loop not vectorized: the optimizer was unable to perform the requested transformation; the transformation might be disabled or specified as part of an unsupported transformation ordering [-Werror,-Wpass-failed=transform-warning]
```

Full build log: [wabt-1.0.34_broken_simd_tests:20240319-155457.log](https://github.com/WebAssembly/wabt/files/14653567/wabt-1.0.34_broken_simd_tests.20240319-155457.log)

SoniEx2:
Fixes #2403 

simde obviously requests loop vectorization, but we shouldn't fail for missed vectorizations as we're more focused on testing for correctness.

### DIFF
--- a/test/run-spec-wasm2c.py
+++ b/test/run-spec-wasm2c.py
@@ -471,6 +471,9 @@ def Compile(cc, c_filename, out_dir, use_c11, *cflags):
                  '-Wno-ignored-optimization-argument',
                  '-Wno-tautological-constant-out-of-range-compare',
                  '-Wno-infinite-recursion',
+                 # simde calls for pragma clang loop vectorize(enable),
+                 # which can generate warnings if the optimization pass fails.
+                 '-Wno-pass-failed',
                  '-fno-optimize-sibling-calls',
                  '-frounding-math', '-fsignaling-nans',
                  '-D_DEFAULT_SOURCE']


sbc100:
Can you give an example of the failure messages you are seeing?

Why would llvm be generating warning about loop vectorization failure unless we are explicitly asking to run that pass?  (are we doing that?)

SoniEx2:
see also simd-everywhere/simde#1151

sbc100:
But if we aren't seeing these warning when we build/run our tests why suppress the warning here?    If `test/run-spec-wasm2c.py` is generating warning why don't we see them in our CI?

SoniEx2:
because our clang is too old (it's using clang 14)

SoniEx2:
the macos-14 runner reproduces the warning, but it also fails due to the other thing. (unimplemented simd read intrinsics or something?)

we could also wait a few months for CI to start failing.

sbc100:
This sounds like maybe a bug in clang, no?   Why would it tell the user about some internal pass failure, unless the user somehow opts into that pass specifically... are we somehow doing that in the generated code?

SoniEx2:
> are we somehow doing that in the generated code?

Yes. simde expands `pragma clang loop vectorize(enable)` which requests that pass specifically:

https://github.com/simd-everywhere/simde/blob/cf68aaf07105a71db9aa7dc076e9b75a701bf58c/simde/simde-common.h#L379

sbc100:
> > are we somehow doing that in the generated code?
> 
> Yes. simde expands `pragma clang loop vectorize(enable)` which requests that pass specifically:
> 
> https://github.com/simd-everywhere/simde/blob/cf68aaf07105a71db9aa7dc076e9b75a701bf58c/simde/simde-common.h#L379

Thanks for context. I think that was what I was missing.

How about adding a comment here such as: `# simde explicitly enables xxx via pragma yyy, which means clang can generate warnings when this pass fails.  `

sbc100 (approved):

PR #2411, opened 2024-05-03
Harden against invalid alignment

SoniEx2:
Chasing https://crbug.com/oss-fuzz/51561

Let's see if this breaks anything...

The issue only triggers if both conditions are true:

- Multi-memory is enabled
- Alignment (masked) is between 32 and 63

### DIFF
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -122,6 +122,8 @@ class BinaryReader {
   [[nodiscard]] Result ReadIndex(Index* index, const char* desc);
   [[nodiscard]] Result ReadOffset(Offset* offset, const char* desc);
   [[nodiscard]] Result ReadAlignment(Address* align_log2, const char* desc);
+  [[nodiscard]] Result CheckAlignment(Address* align_log2, const char* desc);
+  [[nodiscard]] Result TakeHasMemidx(Address* align_log2, bool* has_memidx);
   [[nodiscard]] Result ReadMemidx(Index* memidx, const char* desc);
   [[nodiscard]] Result ReadMemLocation(Address* alignment_log2,
                                        Index* memidx,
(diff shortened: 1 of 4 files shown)


SoniEx2:
cc @keithw because oss-fuzz

src/binary-reader.cc
```diff
@@ -432,8 +432,7 @@ Result BinaryReader::ReadOffset(Offset* offset, const char* desc) {
 Result BinaryReader::ReadAlignment(Address* alignment_log2, const char* desc) {
   uint32_t value;
   CHECK_RESULT(ReadU32Leb128(&value, desc));
-  if (value >= 128 ||
-      (value >= 32 && !options_.features.multi_memory_enabled())) {
+  if (value >= 128 || (value & ((1 << 6) - 1)) >= 32) {
```
keithw, line 435: This is a bit gnarly... is there any way to simplify/factor this?

test/regress/undefined-shifts.txt
```diff
@@ -0,0 +1,20 @@
+;;; TOOL: run-gen-wasm-bad
```
keithw, line 1: Looks great but can we contribute this test upstream?

keithw (commented):
Thank you for fixing this!

test/regress/undefined-shifts.txt
SoniEx2: we're not sure. the fix/bug is only relevant to multi-memory (so it would go into the multi-memory proposal), but the test we used as a reference (test/binary/bad-alignment.txt) doesn't appear to be in upstream? (at least we couldn't find it in third_party/testsuite. not that we can even update that at this time...)

src/binary-reader.cc
SoniEx2: this is the simplest approach we could find without duplicating the entire `ReadAlignment`. but um... yeah we don't like it either.

`ReadAlignment` is currently used in `ReadMemLocation` and in `ReadLinkingSection` (for `SegmentInfo`), and as far as we can tell the latter doesn't want multi-memory flag stuff. so its use is a bit overloaded... we could split them, we guess? thoughts?

test/regress/undefined-shifts.txt
SoniEx2: (we don't actually know how to upstream this test. run-gen-wasm-bad / gen-wasm.py seems to be a custom tool, we guess we could run it manually to generate the required invalid binary wasm...?)

keithw: That sounds right -- it'll have to be a `(module binary...)` in the script language.

SoniEx2:
ping @keithw @sbc100 (tests have also been upstreamed, tho wabt is currently a bit out of date on spec tests)

keithw:
Thanks, and thanks for the ping. Do you think it would be better to consolidate the three `ReadAlignment`, `TakeHasMemidx`, and `CheckAlignment` functions into a single `ReadAlignment(Address* align_log2, bool* has_memidx, const char* desc);` that just does everything? That seems safer than requiring the caller to remember to call `CheckAlignment` each time (if `ReadAlignment` just always does it).

keithw:
E.g. https://github.com/keithw/wabt/commit/b09d4e082a3edadde8945777d75eb476b5f7c368

SoniEx2:
we figured having the check in the same function is what caused the issue in the first place. further, different contexts may give different meanings to the alignment flags, so we would be against calling it "has memidx" in that case.

would you have another way to enforce that CheckAlignment is called?

SoniEx2:
how about this?

keithw:
Let's just go back to the way you had it and call it done.

SoniEx2:
oh ;-; okay

keithw (approved):

PR #2420, opened 2024-05-14
Fix bug in wasm2c's tail-call optimization code generation

squk:


sbc100 (commented):
I guess that means we don't have any test coverage for this?  @keithw does that make sense?  Is it hard to write tests for this?

squk:
@sbc100 I updated the failing test case. But it looks like we should also have test cases to ensure the golden files can compile as well which would've prevented this.

sbc100:
> @sbc100 I updated the failing test case. But it looks like we should also have test cases to ensure the golden files can compile as well, which would've prevented this.

Indeed, no need to block this change on that though.

sbc100 (approved):

keithw:
Hmm, I realize this means there are zero upstream tail-call tests that use imports at all, much less attempt to tail-call them. There probably should be some. :-/

PR #2424, opened 2024-05-20
wasm2c: Use wabt::StripExtension

### LINKED ISSUE #2423: wasm2c compiling minimal example issues
When I try to compile this module

```
(module $empty)
```

with
```
wasm2wat --version

wat2wasm empty.wat
wasm2c empty.wasm -o empty
```
I get the following error message:
```
1.0.35
terminate called after throwing an instance of 'std::out_of_range'
  what():  basic_string_view::substr: __pos (which is 18446744073709551615) > __size (which is 5)
./do.sh: line 4: 392733 Aborted                 (core dumped) wasm2c empty.wasm -o empty
```

SoniEx2:
Fixes #2423

### DIFF
--- a/src/tools/wasm2c.cc
+++ b/src/tools/wasm2c.cc
@@ -117,16 +117,6 @@ static void ParseOptions(int argc, char** argv) {
   }
 }
 
-// TODO(binji): copied from binary-writer-spec.cc, probably should share.
-static std::string_view strip_extension(std::string_view s) {
-  std::string_view ext = s.substr(s.find_last_of('.'));
-  std::string_view result = s;
-
-  if (ext == ".c")
-    result.remove_suffix(ext.length());
-  return result;
-}
-
 Result Wasm2cMain(Errors& errors) {
   if (s_num_outputs < 1) {
     fprintf(stderr, "Number of output files must be positive.\n");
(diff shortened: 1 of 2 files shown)


src/tools/wasm2c.cc
```diff
@@ -119,12 +119,19 @@ static void ParseOptions(int argc, char** argv) {
...
+    std::string_view ext = s.substr(pos);
+    std::string_view result = s;
+
+    if (ext == ".c") {
+      result.remove_suffix(ext.length());
+      return result;
+    }
+  }
+  fprintf(stderr, "output filename must end in .c");
+  exit(1);
```
sbc100, line 133: How about if we instead just return early if not `.` is found?

That would seem to fix better with the name of the extension, and avoid having to create this new error condition.

sbc100: In fact I think you can maybe just remove this function completely and use `wabt::StripExtension`

SoniEx2: (...what if we rename it to "base name" and make it a breaking change?)

sbc100: IIUC `basename` normally means stripping the directory component, not the extension component.

SoniEx2: we mean the output flag, `-o`, we could make it take a "base name": `-o empty.c` -> `-o empty`. which makes somewhat more sense since it also generates a `.h`.

sbc100: I think if we just strip the extension that is fine.   I think it makes sense to write `-o foo.c` (at least to me).

sbc100: (i.e no need to do a breaking change here I think).

sbc100 (approved):

test/regress/regress-2423.txt
```diff
@@ -0,0 +1,3 @@
+;;; TOOL: run-wasm2c
```
sbc100, line 1: I tend to prefer to give tests a useful name rather than just regress-XX where possible.   Something like `test/wasm2c/output_no_extension.wast`?

sbc100: But happy to see this land either way.

SoniEx2: sure

PR #2425, opened 2024-05-20
[wasm-objdump] Fix reporting of imported 64-bit tables

sbc100:


test/run-tests.py
```diff
@@ -56,7 +56,7 @@
         ('ARGS', ['%(in_file)s']),
     ],
     'run-objdump': [
-        ('RUN', '%(wat2wasm)s %(in_file)s -o %(temp_file)s.wasm'),
```
SoniEx2, line 59: hmm, wouldn't it be better to use `ARGS0: --enable-memory64` instead of changing this here?

sbc100: I makes sense to me that we should enable all features when translating the input wat file in preparation of wasm-objdump.   The `run-objdump` is all about testing of objdump and we don't need to be limiting the set of input features we use.

sbc100: Also, `run-objdump` test is generic type of test so it wouldn't make sense (to me) to single out just one feature to enable (even though it would be enough to make this test pass).

SoniEx2: no the `ARGS0: --enable-memory64` goes on the test case not here.

(side note: we really need more tests for `--disable-*` ngl...)

sbc100: I think `run-objdump` should default to enabling all features when it runs wat2wasm since the wat2wasm command is not part of the test here.. its the `wasm-objdump` command that we are trying to test, so we might as well default to all features for the `wat2wasm` part.

sbc100: I reverted this part for now so we don't need to make this change as part of this bugfix.

Thanks for the reminder about `ARGS0:`

dschuff (approved):

PR #2426, opened 2024-05-21
Correct wasm2c example Makefiles

Rexicon226:
- Passing "-lm" into the prereq isn't the correct way add the flag. This correctly adds it to the command.

- The "rot13" example incorrectly assumed that the "rot13.h" file would be generated by the time that "main.c" was being compiled, however there is no rule supporting this and it would fail. I've also added "rot13.h" to the clean rule.


wasm2c/examples/rot13/Makefile
```diff
@@ -4,9 +4,12 @@ CFLAGS=-I../..
...
 
 clean:
-	rm -rf rot13 rot13.wasm rot13.c *.o
+	rm -rf rot13 rot13.wasm rot13.c rot13.h *.o
 
-rot13: main.o rot13.o ../../wasm-rt-impl.o ../../wasm-rt-mem-impl.o -lm
+rot13: main.o rot13.o ../../wasm-rt-impl.o ../../wasm-rt-mem-impl.o
+	$(CC) $(LDFLAGS) -o $@ $^ -lm
+
+main.o: rot13.c
```
sbc100, line 12: Would it make more sense to put rot13.h on the right here?  I guess that only makes sense if its easy to also add `rot13.h` to the left hand side of the `rot13.c` rules?

Rexicon226: I've done that, what do you think?

wasm2c/examples/rot13/Makefile
```diff
@@ -4,14 +4,17 @@ CFLAGS=-I../..
 all: rot13
 
 clean:
-	rm -rf rot13 rot13.wasm rot13.c *.o
+	rm -rf rot13 rot13.wasm rot13.c rot13.h *.o
 
-rot13: main.o rot13.o ../../wasm-rt-impl.o ../../wasm-rt-mem-impl.o -lm
+rot13: main.o rot13.o ../../wasm-rt-impl.o ../../wasm-rt-mem-impl.o
+	$(CC) $(LDFLAGS) -o $@ $^ -lm
```
sbc100, line 10: Would it be enough to set `LDFLAGS=-lm` above?  (i.e. can we keep the implicit rule if we do it this way?)

sbc100: LDFLAGS seems like it it gets added to the implicit link command:

```
$ touch Makefile
$ touch test.c
$ make LDFLAGS=-lxxx test
cc   -lxxx  test.c   -o test
/usr/bin/ld: cannot find -lxxx: No such file or directory
```

Rexicon226: Interesting. When I test it, I get different results:
```make
# Use implicit rules for compiling C files.
CFLAGS=-I../..
LDFLAGS=-lm

...

rot13: main.o rot13.o ../../wasm-rt-impl.o ../../wasm-rt-mem-impl.o
	$(CC) -o $@ $^
```

```
$ make                                                                                                                                      (base)
../../../bin/wat2wasm rot13.wat -o rot13.wasm
../../../bin/wasm2c rot13.wasm -o rot13.c --disable-simd
cc -I../..   -c -o main.o main.c
cc -I../..   -c -o rot13.o rot13.c
cc -o rot13 main.o rot13.o ../../wasm-rt-impl.o ../../wasm-rt-mem-impl.o
/usr/bin/ld: rot13.o: in function `wasm_floor':
rot13.c:(.text+0x16e): undefined reference to `floor'
/usr/bin/ld: rot13.o: in function `wasm_floorf':
rot13.c:(.text+0x1ab): undefined reference to `floorf'
/usr/bin/ld: rot13.o: in function `wasm_ceil':
rot13.c:(.text+0x1ed): undefined reference to `ceil'
/usr/bin/ld: rot13.o: in function `wasm_ceilf':
rot13.c:(.text+0x22a): undefined reference to `ceilf'
/usr/bin/ld: rot13.o: in function `wasm_trunc':
rot13.c:(.text+0x26c): undefined reference to `trunc'
/usr/bin/ld: rot13.o: in function `wasm_truncf':
rot13.c:(.text+0x2a9): undefined reference to `truncf'
/usr/bin/ld: rot13.o: in function `wasm_nearbyintf':
rot13.c:(.text+0x2e6): undefined reference to `nearbyintf'
/usr/bin/ld: rot13.o: in function `wasm_nearbyint':
rot13.c:(.text+0x328): undefined reference to `nearbyint'
/usr/bin/ld: rot13.o: in function `wasm_sqrt':
rot13.c:(.text+0x41c): undefined reference to `sqrt'
/usr/bin/ld: rot13.o: in function `wasm_sqrtf':
rot13.c:(.text+0x459): undefined reference to `sqrtf'
collect2: error: ld returned 1 exit status
make: *** [Makefile:11: rot13] Error 1
```

The flag doesn't get passed into the `cc` call. Maybe I'm missing something :smile:

sbc100: You would also need to remove the action itself.  If you include the explicit link command you also need to include the explicit LDFLAGS.   Hopefully you can remove it and let the command be implict

Rexicon226: @sbc100 Sorry for disappearing for a couple of weeks, I had exams.

I looked back into this and I'm running into an issue I can't get past.
Using:
```make
# Use implicit rules for compiling C files.
CFLAGS=-I../..
LDFLAGS=-lm

all: fac

clean:
	rm -rf fac fac.wasm fac.c *.o

fac: main.o fac.o ../../wasm-rt-impl.o

fac.wasm: fac.wat ../../../bin/wat2wasm
	../../../bin/wat2wasm $< -o $@

fac.c: fac.wasm ../../../bin/wasm2c
	../../../bin/wasm2c $< -o $@ --disable-simd

.PHONY: all clean

```
without the action, I get the almost correct:
```
cc -lm  fac.o main.o ../../wasm-rt-impl.o   -o fac
```

however this still leads to undefined references because `-lm` is passed before the file arguments.

Any idea on how to solve this?

sbc100: Looks like you can use LDLIBS: https://www.gnu.org/software/make/manual/html_node/Catalogue-of-Rules.html#index-linking_002c-predefined-rule-for

Rexicon226: Good call, that works!

wasm2c/examples/fac/fac.c
```diff
@@ -1,702 +0,0 @@
-/* Automatically generated by wasm2c */
```
sbc100, line 1: Can you restore this file?   I like to have these files checked in so that when we change the wasm2c generator we can see the effect it has on the output (as part of the diff).

Rexicon226: Done!

PR #2433, opened 2024-06-21
Replace table.full with table.fill

mjbshaw:
Presumably this was a typo and the `TableFill` op code (text instruction: `table.fill`) token should use the word "fill" instead of "full."


keithw:
Thanks, nice catch! I think this string is only used to print the error message in a case like

```wasm
(func v128.const table.fill 0 0 drop)
```

which currently generates:

> error: Unexpected type at start of simd constant. Expected one of: i8x16, i16x8, i32x4, i64x2, f32x4, f64x2. Found "table.full".

keithw (approved):

PR #2434, opened 2024-06-21
Fix 2nd argument's type definition of some f32/f64 opcodes in opcde.def

tianstcht:
The 2nd argument's type ​​​​of these opcodes are different from the wasm spec


sbc100:
It looks like its the second argument type you are changing here, not the return type?   Other than updating the PR description to reflect that this lgtm!   Thanks.    Kind of shame that this isn't pragmatically verified somehow/somewhere in wabt.

sbc100 (approved):

tianstcht:
> It looks like its the second argument type you are changing here, not the return type? Other than updating the PR description to reflect that this lgtm! Thanks. Kind of shame that this isn't pragmatically verified somehow/somewhere in wabt.

Thank you, it's my typo, now it has been changed.

PR #2435, opened 2024-06-21
Avoid creating temporary vector copies when checking signature

mjbshaw:
I was doing some local bench marking and found a significant amount of wasm-validate's time came from copying (and then destroying) temporary vectors here. [Currently `CheckSignature()` is calling `PrintStackIfFailed()`]((https://github.com/WebAssembly/wabt/blob/65b61e0a3fde8c4fc82cee53bd5cdd0959b6ffa6/src/type-checker.cc#L257)) and passing `sig` (which is already a `TypeVector), whic creates a copy. This change eliminates the unnecessary vector copy and adds a static assertion that should prevent this from happening again.

### DIFF
--- a/include/wabt/type-checker.h
+++ b/include/wabt/type-checker.h
@@ -18,6 +18,7 @@
 #define WABT_TYPE_CHECKER_H_
 
 #include <functional>
+#include <type_traits>
 #include <vector>
 
 #include "wabt/common.h"
@@ -182,6 +183,9 @@ class TypeChecker {
 
   template <typename... Args>
   void PrintStackIfFailed(Result result, const char* desc, Args... args) {
+    // Assert all args are Type or Type::Enum. If it's a TypeVector then
+    // PrintStackIfFailedV() should be used instead.
+    static_assert((std::is_constructible_v<Type, Args> && ...));
     // Minor optimization, check result before constructing the vector to pass
     // to the other overload of PrintStackIfFailed.
     if (Failed(result)) {
--- a/src/type-checker.cc
+++ b/src/type-checker.cc
@@ -254,7 +254,7 @@ Result TypeChecker::CheckSignature(const TypeVector& sig, const char* desc) {
   for (size_t i = 0; i < sig.size(); ++i) {
     result |= PeekAndCheckType(sig.size() - i - 1, sig[i]);
   }
-  PrintStackIfFailed(result, desc, sig);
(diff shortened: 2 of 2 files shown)


sbc100:
Under what circumstances do these args end up being vectord?  Multi-value?

mjbshaw:
I linked to where it happens in the comment. [`CheckSignature() is passing its `sig` vector`](https://github.com/WebAssembly/wabt/blob/65b61e0a3fde8c4fc82cee53bd5cdd0959b6ffa6/src/type-checker.cc#L257) to `PrintStackIfFailed()`. I don't see any other occurrances, though, so an easy alternative would be to change line 257 to use `PrintStackIfFailedV()` instead.

sbc100:
> I linked to where it happens in the comment. [`CheckSignature()` is passing its `sig` vector](https://github.com/WebAssembly/wabt/blob/65b61e0a3fde8c4fc82cee53bd5cdd0959b6ffa6/src/type-checker.cc#L257) to `PrintStackIfFailed()`. I don't see any other occurrances, though, so an easy alternative would be to change line 257 to use `PrintStackIfFailedV()` instead.

Hmm.. yes it does looks like `PrintStackIfFailedV` would be the correct thing to call here.  Otherwise are we not making a vector of a vector?    It seems like `PrintStackIfFailed` should be called only with individual types as arguments.

mjbshaw:
> yes it does looks like `PrintStackIfFailedV` would be the correct thing to call here.

Okay, I've updated the PR to instead directly call `PrintStackIfFailedV`.

> Otherwise are we not making a vector of a vector?

There's only a single argument in the variadic template, so `{args...}` desugars to `{arg}`, where `arg` is a vector derived from the `sig` argument. This form of braced initialization in this situation does not try to create a vector-of-vectors.

> It seems like `PrintStackIfFailed` should be called only with individual types as arguments.

Yeah that seems to be the intention.

sbc100 (approved):
Can you update the PR description and title?

I wonder if we could add some type checking to PrintStackIfFailed such that it verifies that each its arguments is indeed a Type?  (and not come more complex type like this)?

mjbshaw:
> Can you update the PR description and title?

Done.

> I wonder if we could add some type checking to PrintStackIfFailed such that it verifies that each its arguments is indeed a Type? (and not come more complex type like this)?

I've added a `static_assert` that enforces this (and confirmed it works as intended).

sbc100 (approved):

Issue #2431, opened 2024-06-19
Invalid Read Memory in wabt::interp


Messi-Q:
## Version
commit [4beb525](https://github.com/WebAssembly/wabt/commit/4beb52501f863f1cbd92e0ce4879f04f473935b5)
Author: Peng Qian messi.qp711@gmail.com
Date: Wen Jun 10:48:12 2024

## Compile
```
cd wabt
mkdir build
cd build 
cmake -DCMAKE_CXX_FLAGS="-fsanitize=address -g" -DCMAKE_C_FLAGS="-fsanitize=address -g" -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address" ..
make
```

## Reproduce
```
./wasm-interp id:000008,sig:11,src:003840,op:havoc,rep:2
```

## ASAN Log
```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==1084839==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000008 (pc 0x562656e0b4a1 bp 0x7ffc49ce8f90 sp 0x7ffc49ce8f60 T0)
==1084839==The signal is caused by a READ memory access.
==1084839==Hint: address points to the zero page.
    #0 0x562656e0b4a0 in wabt::interp::DataSegment::IsValidRange(unsigned long, unsigned long) const /home/peng/Documents/all_wasm_vm/new_version_test/wabt/src/interp/interp.cc:734
    #1 0x562656e098aa in wabt::interp::Memory::Init(unsigned long, wabt::interp::DataSegment const&, unsigned long, unsigned long) /home/peng/Documents/all_wasm_vm/new_version_test/wabt/src/interp/interp.cc:617
    #2 0x562656e1e4fd in wabt::interp::Thread::DoMemoryInit(wabt::interp::Instr, wabt::interp::RefPtr<wabt::interp::Trap>*) /home/peng/Documents/all_wasm_vm/new_version_test/wabt/src/interp/interp.cc:2091
    #3 0x562656e179d8 in wabt::interp::Thread::StepInternal(wabt::interp::RefPtr<wabt::interp::Trap>*) /home/peng/Documents/all_wasm_vm/new_version_test/wabt/src/interp/interp.cc:1526
[...]

Issue #2311, opened 2023-10-08
Out-of-Bound Memory Write in DataSegment::Drop()


mobsceneZ:
### Environment

```shell
OS               : Linux 5.10.16.3-microsoft-standard-WSL2 #1 SMP Fri Apr 2 22:23:49 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
Commit           : 0e78c24fd231d5ee67ccd271bfa317faa963281c
Version          : 1.0.33 (git~1.0.33-35-gdddc03d3)
Clang Verison    : 12.0.1
Build            : mkdir build && cd build && export CC=clang CXX=clang++ CFLAGS="-fsanitize=address -g" CXXFLAGS="-fsanitize=address -g" && cmake .. && cmake --build .
Affected Tool    : wasm-interp
Enabled Features : None
Impact           : Out-of-Bound Memory Write Access
```

### Proof of Concept
[poc-wasm-interp-02.zip](https://github.com/WebAssembly/wabt/files/12840051/poc-wasm-interp-02.zip)

### Stack Trace Provide By AddressSanitizer
```shell
$  ~/wabt_asan/bin/wasm-interp poc.wasm
AddressSanitizer:DEADLYSIGNAL
=================================================================
==3641==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000008 (pc 0x00000067f426 bp 0x7ffd04e28310 sp 0x7ffd04e28300 T0)
==3641==The signal is caused by a WRITE memory access.
==3641==Hint: address points to the zero page.
    #0 0x67f426 in wabt::interp::DataSegment::Drop() /home/lain/wabt_asan/include/wabt/interp/interp-inl.h:906:9
    #1 0x6670be in wabt::interp::Thread::DoDataDrop(wabt::interp::Instr) /home/lain/wabt_asan/src/interp/interp.cc:2081:33
    #2 0x65b29a in wabt::interp::Thread::StepInternal(wabt::interp::RefPtr<wabt::interp::Trap>*) /home/lain/wabt_asan/src/interp/interp.cc:1511:32
    #3 0x65352b in wabt::interp::Thread::Run(int, wabt::interp::RefPtr<wabt::interp::Trap>*) /home/lain/wabt_asan/src/interp/interp.cc:1086:19
[...]

Issue #2310, opened 2023-10-08
Out-of-Bound Memory Read in DataSegment::IsValidRange()


mobsceneZ:
### Environment

```shell
OS               : Linux 5.10.16.3-microsoft-standard-WSL2 #1 SMP Fri Apr 2 22:23:49 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
Commit           : 0e78c24fd231d5ee67ccd271bfa317faa963281c
Version          : 1.0.33 (git~1.0.33-35-gdddc03d3)
Clang Verison    : 12.0.1
Build            : mkdir build && cd build && export CC=clang CXX=clang++ CFLAGS="-fsanitize=address -g" CXXFLAGS="-fsanitize=address -g" && cmake .. && cmake --build .
Affected Tool    : wasm-interp
Enabled Features : None
Impact           : Out-of-Bound Memory Read Access
```

### Proof of Concept
[poc-wasm-interp-01.zip](https://github.com/WebAssembly/wabt/files/12840027/poc-wasm-interp-01.zip)

### Stack Trace Provide By AddressSanitizer
```shell
$  ~/wabt_asan/bin/wasm-interp poc.wasm
AddressSanitizer:DEADLYSIGNAL
=================================================================
==3549==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000008 (pc 0x00000064a0fe bp 0x7ffcceb61670 sp 0x7ffcceb61640 T0)
==3549==The signal is caused by a READ memory access.
==3549==Hint: address points to the zero page.
    #0 0x64a0fe in wabt::interp::DataSegment::IsValidRange(unsigned long, unsigned long) const /home/lain/wabt_asan/src/interp/interp.cc:734:19
    #1 0x649cd7 in wabt::interp::Memory::Init(unsigned long, wabt::interp::DataSegment const&, unsigned long, unsigned long) /home/lain/wabt_asan/src/interp/interp.cc:617:11
    #2 0x666cb4 in wabt::interp::Thread::DoMemoryInit(wabt::interp::Instr, wabt::interp::RefPtr<wabt::interp::Trap>*) /home/lain/wabt_asan/src/interp/interp.cc:2075:3
    #3 0x65b199 in wabt::interp::Thread::StepInternal(wabt::interp::RefPtr<wabt::interp::Trap>*) /home/lain/wabt_asan/src/interp/interp.cc:1510:32
[...]

PR #2436, opened 2024-06-27
fix: null-defer about data and elem in wasm-interp

tianstcht:
fix some null-defer case about data and elem.

- issue: [2431](https://github.com/WebAssembly/wabt/issues/2431)

### DIFF
--- a/src/interp/interp.cc
+++ b/src/interp/interp.cc
@@ -2085,10 +2085,9 @@ RunResult Thread::DoReinterpret() {
 RunResult Thread::DoMemoryInit(Instr instr, Trap::Ptr* out_trap) {
   Memory::Ptr memory{store_, inst_->memories()[instr.imm_u32x2.fst]};
   TRAP_IF(inst_->datas().size() <= instr.imm_u32x2.snd,
-      StringPrintf("out of bounds data access: access at %u "
-                    ">= max value %" PRIu64,
-                    instr.imm_u32x2.snd,
-                    inst_->datas().size()));
+          StringPrintf("out of bounds data access: access at %u "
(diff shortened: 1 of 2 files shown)


src/interp/interp.cc
```diff
@@ -2084,6 +2084,11 @@ RunResult Thread::DoReinterpret() {
 
 RunResult Thread::DoMemoryInit(Instr instr, Trap::Ptr* out_trap) {
   Memory::Ptr memory{store_, inst_->memories()[instr.imm_u32x2.fst]};
+  TRAP_IF(inst_->datas().size() <= instr.imm_u32x2.snd,
+      StringPrintf("out of bounds data access: access at %u "
+                    ">= max value %" PRIu64,
+                    instr.imm_u32x2.snd,
+                    inst_->datas().size()));
```
sbc100, line 2091: Shouldn't validation make this impossible?

sbc100: I'm not sure the interpreter should be enforcing constraints that the validator should have already enforced... otherwise we will end up duplicating all the validation rules here.

tianstcht: There are some formatting issue in the CI log, so I have chosen to use `%zu` to format vector's size correctly.

tianstcht:
The patch also fix issue [2310](https://github.com/WebAssembly/wabt/issues/2310) and [2311](https://github.com/WebAssembly/wabt/issues/2311) with the same null-defer bug

SoniEx2:
checking the sizes is okay but it's the wrong approach: the module in #2431 should fail validation, it's missing a data section

PR #2437, opened 2024-07-02
Fix compiling wasm2c's runtime library with Android NDK(macro definition conflicts with sysroot)

squk:
Building wasm2c'c wasm_rt with the Android NDK results in the following error
```
third_party/wabt/wasm2c/wasm-rt-mem-impl.c:28:9: error: 'PAGE_SIZE' macro redefined [-Werror,-Wmacro-redefined]
#define PAGE_SIZE 65536
        ^
third_party/android/ndk/stable/toolchains/llvm/prebuilt/linux-x86_64/bin/../real-bin/../sysroot/usr/include/sys/user.h:37:9: note: previous definition is here
#define PAGE_SIZE 4096
        ^
1 error generated.
```

This PR fixes that.

### DIFF
--- a/wasm2c/wasm-rt-mem-impl.c
+++ b/wasm2c/wasm-rt-mem-impl.c
@@ -25,7 +25,7 @@
 #include <sys/mman.h>
 #endif
 
-#define PAGE_SIZE 65536
+#define WASM_PAGE_SIZE 65536
 
 #ifdef WASM_RT_GROW_FAILED_HANDLER
 extern void WASM_RT_GROW_FAILED_HANDLER();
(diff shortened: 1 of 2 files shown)


wasm2c/wasm-rt-mem-impl.c
```diff
@@ -25,7 +25,7 @@
 #include <sys/mman.h>
 #endif
 
-#define PAGE_SIZE 65536
+#define WASM_RT_PAGE_SIZE 65536
```
sbc100, line 28: How about WASM_PAGE_SIZE.. since otherwise it might be confused with the host (runtime) page size?

sbc100:
@mathiasblume did you run into this?

squk:
@sbc100 IIRC @matthias-blume  doesn't use the wasm runtime in his experiments and has forked the c-writer

wasm2c/wasm-rt-mem-impl.c
squk: Done.

sbc100 (approved):

sbc100:
Looks like there is a separate cmake issue in CI that we should fix before landing this.

sbc100:
Rebasing on top of the main branch should fix the current CI failure.

squk:
@sbc100 done :)

squk:
Thanks!

PR #2440, opened 2024-07-10
Fix deprecated FetchContent_Populate warning third_party/uvwasi cmake build

Changqing-JING:
See https://github.com/nodejs/uvwasi/issues/277


CMakeLists.txt
```diff
@@ -573,7 +573,12 @@ if (BUILD_TOOLS)
   )
 
   if(WITH_WASI)
+    set(CMAKE_WARN_DEPRECATED_BACK ${CMAKE_WARN_DEPRECATED})
+    set(CMAKE_WARN_DEPRECATED OFF)
+    cmake_policy(SET CMP0169 OLD)
     add_subdirectory("third_party/uvwasi" EXCLUDE_FROM_ALL)
+    cmake_policy(SET CMP0169 NEW)
+    set(CMAKE_WARN_DEPRECATED ${CMAKE_WARN_DEPRECATED_BACK})
```
sbc100, line 581: I wonder if we should instead update uvwasi?   maybe there fixed on on their end?

Changqing-JING: No, I check the latest uvwasi, it didn't fix this. And the build of wabt should not rely on the code quality of thirdparty submodule. The warning/waring-as-error setting in thirdparty submodules are uncontrollable. We would better suppress warnings in submodule and only apply the warning-as-error to self code

Changqing-JING:
CI passed with the fix

CMakeLists.txt
```diff
@@ -573,7 +573,10 @@ if (BUILD_TOOLS)
   )
 
   if(WITH_WASI)
+    set(CMAKE_POLICY_DEFAULT_CMP0169_BACK ${CMAKE_POLICY_DEFAULT_CMP0169})
```
sbc100, line 576: Can you add a comment here? Something like: "uvwasi uses the deprecated FetchContent_Populate function" ?

Issue #2454, opened 2024-08-20
Missing support for the (somewhat) recently added relocation types


SingleAccretion:
See https://github.com/llvm/llvm-project/blob/e56ad22b4a41e65984d6997b2c2496a20f906d1d/llvm/include/llvm/BinaryFormat/WasmRelocs.def#L27-L31 and https://github.com/WebAssembly/wabt/blob/1edeaf2948b4afd711e50ba2007ee44503cd80fe/src/common.cc#L46-L59

I hit it specifically with `R_WASM_FUNCTION_INDEX_I32`. This affects e. g. `wasm-objdump -x -j linking`.

sbc100:
Would you have time to send a PR? 

BTW are you using the annotation proposal?  Or, how are you generated binaries that contain this relocation type?

SingleAccretion:
> Would you have time to send a PR?

Can do that. It'll take a certain amount of setup, since one needs to get the build building, tests written and running, etc.

>  Or, how are you generated binaries that contain this relocation type?

Custom object writer: https://github.com/dotnet/runtimelab/blob/feature/NativeAOT-LLVM/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/WasmObjectWriter.cs. It's the same use case (stack traces) as we've discussed in another issue.

PR #2457, opened 2024-08-29
Add missing relocation types (and delete a non-existent one)

SingleAccretion:
Follows the template of https://github.com/WebAssembly/wabt/commit/96001693fb1694318630675ea69e832412b624c4.

Fixes https://github.com/WebAssembly/wabt/issues/2454.


SingleAccretion:
@sbc100

.gitignore
```diff
@@ -1,5 +1,7 @@
 /bin
 /build
+/build-debug
+/build-release
```
sbc100, line 4: Can you revert his file?   The wabt Makefile uses subdirectories or `out/..` , I you use a different place for your output you can always add these to your `$HOME/.gitignore` so it effects all your projects.

sbc100 (approved):
lgtm % comment.

.gitignore
SingleAccretion: Reverted.

SoniEx2: can use `$GIT_DIR/info/exclude` too, see https://git-scm.com/docs/gitignore (`man gitignore`)

PR #2458, opened 2024-09-06
wasm2c: Add segue option to make exclusive use of the segment register

shravanrn:
The 3rd of the 3 PRs needed for production Linux Segue support (i.e., use x86 segment registers to refer to the Wasm heap for single-memory Wasm modules). This PR adds an option for applications to tell wasm2c that it has exclusive use of the gs register for this purpose and need not save and restore it on each use --- something that is particularly expensive on pre-Ivybridge CPUs which requires syscalls to set segment registers.


shravanrn:
@sbc100 Please have a look when you have the chance. This should be the last segue commit prior to us deploying in Firefox.

test/wasm2c/minimal.txt
```diff
@@ -119,10 +119,21 @@ static inline void* wasm_rt_segue_read_base() {
   }
 }
 static inline void wasm_rt_segue_write_base(void* base) {
-  if (wasm_rt_fsgsbase_inst_supported) {
-    __builtin_ia32_wrgsbase64((uintptr_t)base);
+  bool should_set = true;
+#if WASM_RT_SEGUE_FREE_SEGMENT
+  if (wasm_rt_last_segment_val == base) {
+    should_set = false;
```
sbc100, line 125: Can we replace this line with `return;` and the avoid the `should_set` variable completely?

shravanrn: Fixed! You're right, that's way better :)

.github/workflows/build.yml
```diff
@@ -203,7 +203,7 @@ jobs:
     env:
       USE_NINJA: "1"
       WASM2C_CC: "clang"
-      WASM2C_CFLAGS: "-DWASM_RT_USE_MMAP=1 -DWASM_RT_SKIP_SIGNAL_RECOVERY=1 -DWASM_RT_NONCONFORMING_UNCHECKED_STACK_EXHAUSTION=1 -DWASM2C_TEST_EMBEDDER_SIGNAL_HANDLING -DWASM_RT_ALLOW_SEGUE=1 -mfsgsbase -DWASM_RT_SANITY_CHECKS=1 -Wno-pass-failed"
+      WASM2C_CFLAGS: "-DWASM_RT_USE_MMAP=1 -DWASM_RT_SKIP_SIGNAL_RECOVERY=1 -DWASM_RT_NONCONFORMING_UNCHECKED_STACK_EXHAUSTION=1 -DWASM2C_TEST_EMBEDDER_SIGNAL_HANDLING -DWASM_RT_ALLOW_SEGUE=1 -DWASM_RT_SEGUE_FREE_SEGMENT=1 -mfsgsbase -DWASM_RT_SANITY_CHECKS=1 -Wno-pass-failed"
```
sbc100, line 206: On the name of this new macro, would it make sense to call it something like `WASM_RT_SEGUE_PERSISTENT` ?

wasm2c/wasm-rt-impl.c
```diff
@@ -62,6 +62,10 @@ bool wasm_rt_fsgsbase_inst_supported = false;
 #include <unistd.h>       // For syscall
 #endif
 
+#if WASM_RT_SEGUE_FREE_SEGMENT
+WASM_RT_THREAD_LOCAL void* wasm_rt_last_segment_val = 0;
```
sbc100, line 66: NULL?

sbc100 (approved):
lgtm % minor comments.

.github/workflows/build.yml
sbc100: Having read the change again, I'm not sure my name is any better.  lgtm either way.

shravanrn: Ah ok, if no preference, I'll leave it as is.

SoniEx2: (hmm, what about `SEGUE_SEGMENT_EXCLUSIVE`?)

Issue #2445, opened 2024-07-23
SEGV in wabt::Module::operator=


renatahodovan:
GIT revision: ed1ce97e
Built with `USE_ASAN=ON`
Run: `./build/wat2wasm test.wat`
Test: 
```wat
(module quote "\7c")
```

Backtrace:
```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==1100246==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7f5a434be881 bp 0x7ffcc43aba80 sp 0x7ffcc43ab238 T0)
==1100246==The signal is caused by a READ memory access.
==1100246==Hint: address points to the zero page.
    #0 0x7f5a434be881 in memcpy string/../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:222
    #1 0x55ab4a3b93c4 in __asan_memcpy (wabt/bin/wat2wasm+0x1823c4) (BuildId: a5f593cb6819aa12ce15795bdae38dd1e63c168d)
    #2 0x55ab4a5467b5 in wabt::Module::operator=(wabt::Module&&) (wabt/bin/wat2wasm+0x30f7b5) (BuildId: a5f593cb6819aa12ce15795bdae38dd1e63c168d)
    #3 0x55ab4a508c3c in wabt::WastParser::ParseModuleCommand(wabt::Script*, std::unique_ptr<wabt::Command, std::default_delete<wabt::Command>>*) (wabt/bin/wat2wasm+0x2d1c3c) (BuildId: a5f593cb6819aa12ce15795bdae38dd1e63c168d)
    #4 0x55ab4a506fdc in wabt::WastParser::ParseModule(std::unique_ptr<wabt::Module, std::default_delete<wabt::Module>>*) (wabt/bin/wat2wasm+0x2cffdc) (BuildId: a5f593cb6819aa12ce15795bdae38dd1e63c168d)
    #5 0x55ab4a53d611 in wabt::ParseWatModule(wabt::WastLexer*, std::unique_ptr<wabt::Module, std::default_delete<wabt::Module>>*, std::vector<wabt::Error, std::allocator<wabt::Error>>*, wabt::WastParseOptions*) (wabt/bin/wat2wasm+0x306611) (BuildId: a5f593cb6819aa12ce15795bdae38dd1e63c168d)
[...]

PR #2461, opened 2024-09-11
Fix issue#2445 Null Pointer Deref

haruki3hhh:
Fix issue 2445:
https://github.com/WebAssembly/wabt/issues/2445

### DIFF
--- a/src/wast-parser.cc
+++ b/src/wast-parser.cc
@@ -3439,7 +3439,11 @@ Result WastParser::ParseModuleCommand(Script* script, CommandPtr* out_command) {
                 error.loc.offset, error.message.c_str());
         }
       }
-      *module = std::move(*m.get());
+      if (m) {
+        *module = std::move(*m.get());
+      } else {
+        return Result::Error; // Handle the null case appropriately
+      }
       *out_command = std::move(command);
       break;
   }


src/wast-parser.cc
```diff
@@ -3439,7 +3439,11 @@ Result WastParser::ParseModuleCommand(Script* script, CommandPtr* out_command) {
                 error.loc.offset, error.message.c_str());
         }
       }
-      *module = std::move(*m.get());
+      if (m) {
+        *module = std::move(*m.get());
```
sbc100, line 3443: Is it possible for `ParseWatModule` above to succeed without `*m` being set?

Would this work instead:

```
if (errors.length()) { 
  return Result::Error;
}
assert(*m);
```

haruki3hhh:
Hi,
After adding this patch, execute the poc, it shows:
```
# ./wat2wasm ../@POC@
../@POC@:1:2: error: error in quoted module: @0x100000001: unexpected token "|", expected a module field or a module.
(module quote "\7c")
 ^^^^^^
```
And ASAN is not triggered.

Do you think we should make some further improvements about it? 
I'm glad to do it, if needed.

sbc100:
It would be good to add that as a test case I think.   Seems like the kind of test that could also be upstreamed to the wasm spec test suite too (once it lands here).

haruki3hhh:
If I want to add it into test case, should I directly add a .txt file to https://github.com/WebAssembly/wabt/tree/main/test/parse?

src/wast-parser.cc
SoniEx2: doesn't `ParseWatModule` also return a `Result`? probably a good idea to check that instead.

haruki3hhh:
@sbc100 @SoniEx2 Hi folks, I added a testcase, could you also please take a look at it?

https://github.com/WebAssembly/wabt/pull/2461/files#diff-198b0c14be90b6dbd98ff0d8d3289a686f4daa9733c748db3dadfaa5c55635fd

SoniEx2:
the test looks great.

haruki3hhh:
Hi folks, @SoniEx2 @sbc100 

Do you think I need to do some further improvements for this patch? Or we can merge it?

I'd like to help, if further improvements are needed!

SoniEx2:
ParseWatModule returns a Result, it would be a good idea to check that Result.

src/wast-parser.cc
sbc100: Yes that sounds good.    You can do something like this

```
if (Failed(ParseWatModule(lexer.get(), &m, &errors, options_)) { 
   for (const auto& error : errors) {
      ...
   }
   return Result::Error;
}
assert(errors.length == 0);
```

Issue #2398, opened 2024-03-04
Invalid Memory Read in FreeList<wabt::interp::Object*>::IsUsed()


mobsceneZ:
### Environment

```shell
OS               : Linux 5.15.133.1-microsoft-standard-WSL2 #1 SMP Thu Oct 5 21:02:42 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
Commit           : 9fdd024249b6b181d98a4164700ca6ee09f970d9 - 1471dffee8bf9939044b80d34256956a28138e96
Version          : 1.0.33 (git~1.0.33-35-gdddc03d3)
Clang Verison    : 13.0.0
Build            : mkdir build && cd build && export CC=clang CXX=clang++ CFLAGS="-fsanitize=address -g" CXXFLAGS="-fsanitize=address -g" && cmake .. && cmake --build .
Affected Tool    : wasm-interp
Enabled Features : None
Impact           : Invalid Memory Read
```

### Proof of Concept
[wasm-poc-03.zip](https://github.com/WebAssembly/wabt/files/14482639/wasm-poc-03.zip)

### Stack Trace Provide By AddressSanitizer
```shell
AddressSanitizer:DEADLYSIGNAL
=================================================================
==2716==ERROR: AddressSanitizer: SEGV on unknown address 0x611000081078 (pc 0x00000053636f bp 0x7fff7f323910 sp 0x7fff7f3238f0 T0)
==2716==The signal is caused by a READ memory access.
    #0 0x53636f in wabt::interp::FreeList<wabt::interp::Object*>::IsUsed(unsigned long) const /home/lain/wabt_asan/include/wabt/interp/interp-inl.h:180:39
    #1 0x62e4d6 in wabt::interp::Store::IsValid(wabt::interp::Ref) const /home/lain/wabt_asan/include/wabt/interp/interp-inl.h:482:19
    #2 0x5fdd8f in wabt::interp::Store::HasValueType(wabt::interp::Ref, wabt::Type) const /home/lain/wabt_asan/src/interp/interp.cc:213:8
    #3 0x603cb3 in wabt::interp::Table::Grow(wabt::interp::Store&, unsigned int, wabt::interp::Ref) /home/lain/wabt_asan/src/interp/interp.cc:505:3
[...]

PR #2462, opened 2024-09-11
Fix issue#2398

haruki3hhh:
fix issue: https://github.com/WebAssembly/wabt/issues/2398

### DIFF
--- a/include/wabt/interp/interp-inl.h
+++ b/include/wabt/interp/interp-inl.h
@@ -177,7 +177,7 @@ inline void FreeList<Ref>::Delete(Index index) {
 
 template <typename T>
 bool FreeList<T>::IsUsed(Index index) const {
-  return (reinterpret_cast<uintptr_t>(list_[index]) & ptrFreeBit) == 0;
+  return index < list_.size() && (reinterpret_cast<uintptr_t>(list_[index]) & ptrFreeBit) == 0;
 }
 
 template <typename T>


sbc100:
I wonder if we could figure out a test case for this.   Maybe there is a missing test case upstream?

haruki3hhh:
Hi @sbc100, I'm not quite sure about the test case, I use the POC to trigger it in a related issue. I'm glad to help if needed.

sbc100:
Ah, yes, I guess we can use the POC to craft a new test.    I don't have time to do that right now but I can maybe help out next week with that.

haruki3hhh:
I uploaded poc.wasm and poc.wat all in poc.zip
[poc.zip](https://github.com/user-attachments/files/16969857/poc.zip)
It may help you!

Also, do you think this patch is OK at this time? I'm glad to make it better, if needed

sbc100:
The patch seems OK but I'd like to see how the POC causes the index here to be out of range.  There might be a better fix elsewhere in the codebase that avoid this being out-of-range in the first place (and perhaps we can instead assert here that the index is in-range?)

SoniEx2:
this one's a tricky one, we've managed to minify it to this and still reproduce the original issue:

```wat
(module
  (type (;3;) $reti32 (func (result i32)))
  (type (;4;) $externref (func (param i64 f64 f64 i64) (result externref)))
  ;;(type (;4;) $externref (func (result externref)))
  (type (;5;) $basic_func (func))
  (func (;4;) $start (type $basic_func)
    call $funny_recursion_1
    drop
  )
  (func (;6;) $grow_funcs_by_256 (type $reti32) (result i32)
    ref.null func
    i32.const 256
    table.grow 0
    )
  (func (;8;) $funny_recursion_1 (type $reti32) (result i32)
    i32.const 1
    ref.func $funny_recursion_2
    table.set 0
    i64.const -432345564227567366
    f64.const nan (;=nan;)
    f64.const inf (;=inf;)
    i64.const 65535
    i32.const 1
    call_indirect (type $externref)
    i32.const 0
    table.grow 1
    )
  (func (;9;) $funny_recursion_2 (type $externref) (param i64 f64 f64 i64) (result externref)
    call $grow_funcs_by_256
    drop
    call $funny_recursion_1
    unreachable
    )
  (table (;0;) 16 funcref)
  (table (;1;) 16 externref)
  (start $start)
  (elem (;0;) declare func $grow_funcs_by_256 $funny_recursion_1 $funny_recursion_2)
  )
```

but we are struggling to minify it further...

SoniEx2:
correct fix:

```diff
diff --git a/src/interp/interp.cc b/src/interp/interp.cc
index 74f33eb8..cb5623d8 100644
--- a/src/interp/interp.cc
+++ b/src/interp/interp.cc
@@ -1989,7 +1989,7 @@ RunResult Thread::DoCall(const Func::Ptr& func, Trap::Ptr* out_trap) {
     PushValues(func_type.results, results);
   } else {
     if (PushCall(*cast<DefinedFunc>(func.get()), out_trap) == RunResult::Trap) {
-      return RunResult::Ok;
+      return RunResult::Trap;
     }
   }
   return RunResult::Ok;
```

SoniEx2:
(half tempted to go putting a `#define RETRAP(x) if ((x) == RunResult::Trap) { return RunResult::Trap; }` all over the place...)

PR #2465, opened 2024-09-17
wasm2c: Use wrappers for function references

SoniEx2:
Clang 17(?) tightened UBSAN checks, so that you now get this:

```
- test/wasm2c/spec/call_indirect.txt
  expected error code 0, got 1.
  STDERR MISMATCH:
  --- expected
  +++ actual
  @@ -0,0 +1,3 @@
  +out/test/wasm2c/spec/call_indirect/call_indirect.0.c:2144:12: runtime error: call to function w2c_call__indirect__0__wasm_f0 through pointer to incorrect function type 'unsigned int (*)(void *)'
  +/home/runner/work/wabt/wabt/out/test/wasm2c/spec/call_indirect/call_indirect.0.c:1925: note: w2c_call__indirect__0__wasm_f0 defined here
  +SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior out/test/wasm2c/spec/call_indirect/call_indirect.0.c:2144:12 
  STDOUT MISMATCH:
  --- expected
  +++ actual
  @@ -1 +0,0 @@
  -134/134 tests passed.
```

This happens because emitted functions use a typed module instance, while function references use a `void*` instance. It is UB in C to call the former with the latter, so clang is correct here.

We had to pick one of two ways to fix this: either emit `void*` wrapper functions that do the appropriate downcasting for any module functions that go into a table (potentially including imported functions), or the approach that takes significantly less effort of changing everything to `void*` and downcasting internally. ~~We obviously chose the latter.~~ We eventually started emitting wrapper functions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -99,7 +99,7 @@ jobs:
 
   sanitize:
     name: sanitize
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     env:
       USE_NINJA: "1"
       CC: "clang"
--- a/include/wabt/ir.h
+++ b/include/wabt/ir.h
@@ -1282,7 +1282,9 @@ struct Module {
     bool threads = false;
   } features_used;
 
-  // wasm2c uses this information to limit its output.
+  // The BinaryReaderIR tracks function references used by the module, whether
+  // in element segment initializers, global initializers, or functions. wasm2c
+  // uses this information to limit its output in WriteFuncRefWrappers.
   std::set<Index> used_func_refs;
 };
 
--- a/include/wabt/ir.h
+++ b/include/wabt/ir.h
@@ -25,6 +25,7 @@
 #include <string_view>
 #include <type_traits>
 #include <vector>
+#include <set>
 
 #include "wabt/binding-hash.h"
 #include "wabt/common.h"
@@ -1280,6 +1281,9 @@ struct Module {
     bool exceptions = false;
     bool threads = false;
   } features_used;
+
+  // wasm2c uses this information to limit its output.
+  std::set<Index> used_func_refs;
 };
 
 enum class ScriptModuleType {
--- a/include/wabt/ir.h
+++ b/include/wabt/ir.h
@@ -1284,7 +1284,8 @@ struct Module {
 
   // The BinaryReaderIR tracks function references used by the module, whether
   // in element segment initializers, global initializers, or functions. wasm2c
-  // uses this information to limit its output in WriteFuncRefWrappers.
+  // needs to emit wrappers for any functions that might get used as function
+  // references, and uses this information to limit its output.
   std::set<Index> used_func_refs;
 };
 
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -552,7 +552,7 @@ static constexpr char kParamSuffix =
 static constexpr char kLabelSuffix = kParamSuffix + 1;
 
 static constexpr char kGlobalSymbolPrefix[] = "w2c_";
-static constexpr char kWrapperSymbolPrefix[] = "w2cw_";
+static constexpr char kWrapperSymbolPrefix[] = "wrap_";
 static constexpr char kLocalSymbolPrefix[] = "var_";
 static constexpr char kAdminSymbolPrefix[] = "wasm2c_";
 static constexpr char kTailCallSymbolPrefix[] = "wasm_tailcall_";
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -103,6 +103,11 @@ struct ExternalRef : GlobalName {
   using GlobalName::GlobalName;
 };
 
+struct WrapperRef : GlobalName {
+  explicit WrapperRef(const std::string& name)
+      : GlobalName(ModuleFieldType::Func, name) {}
+};
+
 struct TailCallRef : GlobalName {
   explicit TailCallRef(const std::string& name)
       : GlobalName(ModuleFieldType::Func, name) {}
@@ -342,6 +347,7 @@ class CWriter {
   void Write(const GlobalName&);
   void Write(const TagSymbol&);
   void Write(const ExternalRef&);
+  void Write(const WrapperRef&);
   void Write(const TailCallRef&);
   void Write(const ExternalInstancePtr&);
   void Write(const ExternalInstanceRef&);
@@ -379,6 +385,7 @@ class CWriter {
   void WriteImportFuncDeclaration(const FuncDeclaration&,
                                   const std::string& module_name,
                                   const std::string&);
+  void WriteFuncRefWrapper(const Func* func);
   void WriteCallIndirectFuncDeclaration(const FuncDeclaration&,
                                         const std::string&);
   void ComputeSimdScope();
@@ -402,6 +409,7 @@ class CWriter {
   void WriteDataInitializers();
   void WriteElemInitializerDecls();
   void WriteElemInitializers();
+  void WriteFuncRefWrappers();
   void WriteElemTableInit(bool, const ElemSegment*, const Table*);
   bool IsSingleUnsharedMemory();
   void InstallSegueBase(Memory* memory, bool save_old_value);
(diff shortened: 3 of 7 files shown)


test/spec-wasm2c-prefix.c
```diff
@@ -342,31 +342,31 @@ static w2c_spectest spectest_instance;
...
   printf("spectest.print_i32_f32(%d %g)\n", i, f);
 }
 
-void w2c_spectest_print_f64(w2c_spectest* instance, double d) {
+void w2c_spectest_print_f64(void* instance, double d) {
   printf("spectest.print_f64(%g)\n", d);
 }
 
-void w2c_spectest_print_f64_f64(w2c_spectest* instance, double d1, double d2) {
+void w2c_spectest_print_f64_f64(void* instance, double d1, double d2) {
```
sbc100, line 369: Using `void*` everwhere seems strictly worse.  Could you explain (perhaps in the PR description) why this is needed?

SoniEx2: It's not *strictly* needed, the alternative is to emit wrapper functions for anything that goes into a table, tho this way was *significantly* less effort.

SoniEx2: done(?)

sbc100:
Perhaps we can use some kind of annotation to tell UBSAN to ignore these call_indirect mismatches?

SoniEx2:
This appears to be relevant reading material: https://github.com/python/cpython/issues/111178

If we are understanding this right, this kind of thing is important for "control-flow integrity", a hardware feature(?) to prevent exploits. It would seem ill-advised to ignore it in wasm2c.

keithw:
Is it feasible to instead change the CALL_INDIRECT macro to generate the correct type signature (instead of making everything void*)?

SoniEx2:
No, tables can change at runtime, even with dynamic linking (think dlopen()ing wasm2c'd modules). We don't know the type of the module instance, and we *can't* know all of the possible module instance types.

It's either this, or wrapper functions.

shravanrn:
> cc @shravanrn (can you take a look if this doesn't break anything with the segue stuff?)

This should have no interaction with segue. Also the CI already includes a segue test, so it should (hopefully) be enough if we pass CI

> If we are understanding this right, this kind of thing is important for "control-flow integrity", a hardware feature(?) to prevent exploits. It would seem ill-advised to ignore it in wasm2c.

So I assume you are referring to clang's  software-enforced control-flow integrity (CFI). It's actually entirely fine to ignore clang CFI in wasm2c, as Wasm in effect has its own CFI built-in (the table's function type checks in Wasm are basically the same as Clang's CFI). Arguably, clang's CFI is slightly more fine-grained, but imho not sufficiently beneficial to make a significant change over.

To be clear this is different from whether we should worry about UB or not. In general, we should really try to avoid UB for reasons that have nothing to do with CFI support.


** On the current PR **: As a general rule, I am not a huge fan of regressing any strong typing guarantees we have in the code base. This is one of the few things preventing us from accidentally adding bugs as we add features. If anything, I think we should move to more string types and less voids (or general types) where possible.

Given all this, I think the right approach for this PR is probably one of the other options suggested earlier
- Either modify CALL_INDIRECT to call a wrapper function which does add the cast
- Modify CALL_INDIRECT to add the cast (assuming this is possible)

SoniEx2:
rebased and rewritten

include/wabt/ir.h
```diff
@@ -1280,6 +1281,9 @@ struct Module {
     bool exceptions = false;
     bool threads = false;
   } features_used;
+
+  // wasm2c uses this information to limit its output.
```
sbc100, line 1285: I wonder if we should be more specific here.   How about `funcs_in_table_elems`?   Or `required_func_wrappers`?

And perhaps a more specific comment such as "// wasm2c needs to generate small wrapper functions for any functions used in table elements.   We keeps track of them here so we can emit them in WriteFuncRefWrappers"

SoniEx2: they can be used in table, global, elem initializers or (if used in any of the above, or apparently if exported?) function code. so it's not just table elements that we need to worry about.

we don't like the idea of putting `required_func_wrappers` in `ir.h` since `ir.h` is more general than wasm2c.

sbc100: Oh I see.  In that case I guess just update the comment mentioning why wasm2c needs to track this info.

SoniEx2: done

SoniEx2:
@sbc100 ping?

test/wasm2c/check-imports.txt
```diff
@@ -802,11 +802,15 @@ FUNC_TYPE_T(w2c_test_t0) = "\x07\x80\x96\x7a\x42\xf7\x3e\xe6\x70\x5c\x2f\xac\x83
 FUNC_TYPE_T(w2c_test_t1) = "\x72\xab\x00\xdf\x20\x3d\xce\xa1\xf2\x29\xc7\x9d\x13\x40\x7e\x98\xac\x7d\x41\x4a\x53\x2e\x42\x42\x61\x55\x2e\xaa\xeb\xbe\xc6\x35";
 FUNC_TYPE_T(w2c_test_t2) = "\x92\xfb\x6a\xdf\x49\x07\x0a\x83\xbe\x08\x02\x68\xcd\xf6\x95\x27\x4a\xc2\xf3\xe5\xe4\x7d\x29\x49\xe8\xed\x42\x92\x6a\x9d\xda\xf0";
 
+static u32 w2cw_w2c_test_f1(void *instance) {
```
sbc100, line 805: Can we drip the second `w2c_` prefix here?   `w2cw_test_f1` should suffice?   

Also, how about using a more meaningful prefix?  `wrapper_w2c_test_f1`?    `void_wrapper_w2c_test_f1`?   I don't think it really matters so if you prefer `w2cw` thats fine with me too.

sbc100: I don't believe it can conflict.  All user-controlled variables have prefixes, so we are free to use whatever prefix we like.

SoniEx2: alright, we'll change it to `wrap_`. also rebase it we guess.

src/c-writer.cc
```diff
@@ -2286,6 +2298,53 @@ void CWriter::WriteElemInitializerDecls() {
...
+  Write("static ", func->decl.sig.result_types, " ", WrapperRef(func->name));
+  Write("(void *instance");
+  if (func->GetNumParams() != 0) {
+    Indent(4);
+    for (Index i = 0; i < func->GetNumParams(); ++i) {
+      Write(", ");
+      if (i != 0 && (i % 8) == 0) {
+        Write(Newline());
+      }
+      Write(func->GetParamType(i), " ", kLocalSymbolPrefix);
```
sbc100, line 2311: Can we just use `arg_` here instead of `kLocalSymbolPrefix`?

.github/workflows/build.yml
```diff
@@ -99,7 +99,7 @@ jobs:
 
   sanitize:
     name: sanitize
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
sbc100, line 102: Why this?

SoniEx2: pulls in newer clang. (17)

SoniEx2: see also the relevant CI run: https://github.com/WebAssembly/wabt/actions/runs/10896075346/job/30235292696

src/c-writer.cc
SoniEx2: `kLocalSymbolPrefix` is `var_`...?

we mean, we guess? we were going for consistency.

.github/workflows/build.yml
sbc100: I see.  SGTM..  I wonder wonder why `ubuntu-latest` is not already 24.04?

SoniEx2: :shrug: github takes a while to roll out these things. we should set it back when `ubuntu-latest` becomes `24.04`.

src/c-writer.cc
sbc100: I was thinking that for these functions `arg_` would make more sense since they are all arguments to the function.   But I guess we use `var_` for them elsewhere so LGTM.

sbc100 (approved):

PR #2466, opened 2024-09-17
Fix bug in wasm2c runtime's alternate stack deallocation

squk:
This makes `wasm_rt_init()` properly reentrant. By setting `g_alt_stack = NULL`, we allow execution to continue beyond [wasm-rt-impl.c L171 ](https://github.com/WebAssembly/wabt/blob/main/wasm2c/wasm-rt-impl.c#L171) in `os_allocate_and_install_altstack`. 

Applies to debug builds only due to `assert`.


sbc100:
Seems reasonable to me.  I guess technically the resetting could also be in `ifndef NDEBUG` but perhaps thats overkill?

squk:
Yeah I dont think this should be in `ifndef NDEBUG`, as assigning the pointer to null is the proper cleanup behavior. Regardless of if the failure is triggered in debug or prod builds, the runtime state should be properly maintained in both.

sbc100:
> Yeah I dont think this should be in `ifndef NDEBUG`, as assigning the pointer to null is the proper cleanup behavior. Regardless of if the failure is triggered in debug or prod builds, the runtime state should be properly maintained in both.

I think "assigning the pointer to null is the proper cleanup behavior" is debatable, but I also don't think its super important in this case.

Taken to the limit I think that would mean that all C++ destructors should zero all their fields.   For performance reasons we don't tend to do this kind of thing in release builds.

shravanrn:
Agreed, that this would be needed in this case due to re-entrancy, i.e., calling `os_allocate_and_install_altstack`,  `os_disable_and_deallocate_altstack`, `os_allocate_and_install_altstack` in sequence should succeed.

shravanrn (approved):

squk:
Thanks for the reviews yall!

Issue #2469, opened 2024-09-22
wasm2c: UB in exception-handling


SoniEx2:
test case:

```wasm
(module
  (tag $e0)
  (func $longjmp-bait (throw $e0))
  (func (export "setjmp-bait") (param $return-early i32) (result i32)
    (local $value i32)
    (try $try
      (do
        (br_if $try (local.get $return-early))
        (local.set $value (i32.const 1))
        (call $longjmp-bait)
      )
      (catch $e0)
    )
    (local.get $value)
  )
)

(assert_return (invoke "setjmp-bait" (i32.const 1)) (i32.const 0))
(assert_return (invoke "setjmp-bait" (i32.const 0)) (i32.const 1))
```

`(invoke "setjmp-bait" (i32.const 0))` returns 1 in debug, 0 in release.

unfortunately, fixing this might take a huge perf hit. (`volatile` locals.)

(also no, ubsan doesn't catch this yet.)

PR #2470, opened 2024-09-22
binary/wat: Implement EHv4

SoniEx2:
This pull request implements EHv4. Binary is mostly untested until interp is working.


SoniEx2:
marking this as ready for review even if it's not fully complete, would love to hear feedback.

sbc100:
Adding @aheejin too

src/binary-reader-ir.cc
```diff
@@ -1125,6 +1143,7 @@ Result BinaryReaderIR::OnRefFuncExpr(Index func_index) {
 }
 
 Result BinaryReaderIR::OnRefNullExpr(Type type) {
+  module_->features_used.exceptions |= (type == Type::ExnRef);
```
sbc100, line 1146: Is this needed?  Wouldn't the Type have to be declared in the type section in order to produce this?   (i.e. isn't' this already handled the by the type section parsing?

src/binary-reader-objdump.cc
```diff
@@ -903,6 +904,50 @@ Result BinaryReaderObjdumpDisassemble::OnOpcodeType(Type type) {
...
+    if (catch_.kind == CatchKind::Catch || catch_.kind == CatchKind::CatchRef) {
+      buffer.append(std::to_string(catch_.tag));
+    }
+    buffer.append(" ").append(std::to_string(catch_.depth)).append(" ");
+  }
+
+  if (!buffer.empty()) {
+    // remove trailing space
+    buffer.pop_back();
+  }
```
sbc100, line 944: I wonder if there is some way to refactor this such to avoid adding and removing the trailing whitespace like this?

src/interp/binary-reader-interp.cc
```diff
@@ -856,6 +851,16 @@ Result BinaryReaderInterp::EndFunctionBody(Index index) {
...
+  // FIXME(Soni): does the value of `values` even matter here? it used to be
+  // always 0 when this call was in BeginFunctionBody.
+  // NOTE: we don't count the parameters, as they're not part of the frame.
+  func_->handlers.push_back(HandlerDesc{HandlerKind::Catch,
+                                        istream_.end(),
+                                        Istream::kInvalidOffset,
+                                        {},
+                                        {Istream::kInvalidOffset},
+                                        static_cast<u32>(local_decl_count_),
+                                        0});
```
sbc100, line 863: Why did this code move?

sbc100 (commented):
Wow, thanks for all the work on this.  Very impressive!

Looks good so far.

src/binary-reader-ir.cc
SoniEx2: if we're not wrong, our understanding is that `ref.null` calls for a `reftype`, not a type section `typeidx`.

e.g. this is valid

```
(module
  (func
    ref.null exn
    drop
  )
)
```

so anything that produces `exnref` must set `features_used.exceptions`, that way `wasm2c` knows to include their definition. (meanwhile something like `throw_ref` doesn't need to set it.)

src/interp/binary-reader-interp.cc
SoniEx2: fix for #2476 (you may want us to split this out, along with the fix for #2469 ?). we implemented the fix while trying to figure out how to work the interpreter (unfortunately we still don't know how to implement EHv4 in the interpreter)

sbc100: Yup, lets split out whatever we can.

src/binary-reader-ir.cc
sbc100: Sounds reasonable.

SoniEx2:
@sbc100 how do you fix the lint task?

sbc100:
I think you want to add `// clang-format off`` /  `// clang-format on` around those switch statements.

SoniEx2:
there we go. good enough. smh.

sigh...

SoniEx2:
we can bump the testsuite later, so we're splitting that out as well...

SoniEx2:
there are a lot of TODOs related to reftypes in the interpreter...

SoniEx2:
we think we figured it out, just need to get #2483 and #2484 and then we can get all tests to pass (not that we have enough tests for EHv4 but anyway)

SoniEx2:
we believe that's all we really need for EHv4, @sbc100 let us know if there's something we missed (or if you have any ideas for how to split this up, because we don't...)

SoniEx2:
@sbc100 when do you think you can get around to reviewing this?

sbc100:
Do you want to rebase this change and we can see how much of it remains after the split out changes.

include/wabt/binary-reader.h
```diff
@@ -59,6 +59,13 @@ struct TypeMut {
 };
 using TypeMutVector = std::vector<TypeMut>;
 
+struct RawCatch {
+  CatchKind kind;
+  Index tag;
+  Index depth;
+};
+using RawCatchVector = std::vector<RawCatch>;
```
sbc100, line 67: Why is "Raw" in the name here?

src/c-writer.cc
```diff
@@ -2284,7 +2289,8 @@ void CWriter::WriteElemInitializerDecls() {
       continue;
     }
 
-    if (elem_segment->elem_type == Type::ExternRef) {
+    if (elem_segment->elem_type == Type::ExternRef ||
+        elem_segment->elem_type == Type::ExnRef) {
       // no need to store externref elem initializers because only
       // ref.null is possible
```
sbc100, line 2295: Update this comment?

src/c-writer.cc
```diff
@@ -2284,7 +2289,8 @@ void CWriter::WriteElemInitializerDecls() {
       continue;
     }
 
-    if (elem_segment->elem_type == Type::ExternRef) {
+    if (elem_segment->elem_type == Type::ExternRef ||
+        elem_segment->elem_type == Type::ExnRef) {
```
sbc100, line 2293: Should we have some kind of isRef() helper here?

src/c-writer.cc
```diff
@@ -2356,7 +2362,8 @@ void CWriter::WriteElemInitializers() {
...
       // no need to store externref elem initializers because only
       // ref.null is possible
```
sbc100, line 2368: Update comment.

src/c-writer.cc
```diff
@@ -3252,7 +3260,8 @@ void CWriter::WriteLocals(const std::vector<std::string>& index_to_name) {
       func_->local_types, [](auto x) { return x; },
       [&](Index local_index, Type local_type) {
         Write(DefineParamName(index_to_name[num_params + local_index]), " = ");
-        if (local_type == Type::FuncRef || local_type == Type::ExternRef) {
+        if (local_type == Type::FuncRef || local_type == Type::ExternRef ||
+            local_type == Type::ExnRef) {
```
sbc100, line 3264: Some kind of `isRef()` helper here too?

src/interp/binary-reader-interp.cc
```diff
@@ -647,7 +649,7 @@ Result BinaryReaderInterp::BeginInitExpr(FuncDesc* func) {
   label_stack_.clear();
   func_ = func;
   func_->code_offset = istream_.end();
-  Type type = func->type.results[0];
+  Type type = func_->type.results[0];
```
sbc100, line 652: This change looks unrelated?

src/interp/binary-reader-interp.cc
```diff
@@ -1066,6 +1068,11 @@ Result BinaryReaderInterp::OnEndExpr() {
     HandlerDesc& desc = func_->handlers[local_label->handler_desc_index];
     desc.try_end_offset = istream_.end();
     assert(desc.catches.size() == 0);
+  } else if (label_type == LabelType::TryTable) {
+    // TryTable blocks need a try_end_offset
+    Label* local_label = TopLabel();
+    HandlerDesc& desc = func_->handlers[local_label->handler_desc_index];
+    desc.try_end_offset = istream_.end();
```
sbc100, line 1075: Is this different to above block?  Looks like maybe only in the assertion?

wasm2c/wasm-rt-impl.c
```diff
@@ -340,6 +340,8 @@ const char* wasm_rt_strerror(wasm_rt_trap_t trap) {
       return "Uncaught exception";
     case WASM_RT_TRAP_UNALIGNED:
       return "Unaligned atomic memory access";
+    case WASM_RT_TRAP_NULL_REF:
+      return "Null reference";
```
sbc100, line 344: Is this actually possible?  Doesn't the type system / validation ensure no null refrences can exist at the point of use?   I guess it depends if excref is nullable?

sbc100 (commented):
Thanks!

This PR is still pretty huge.  I manged to glance over it and it looks good, but I didn't dive into the interp stuff or the wasm2c stuff.

@keithw @aheejin do either of you have time to help review this?

I think maybe landing the wasm2c change after everything else might be one more logical place to split this?

include/wabt/binary-reader.h
SoniEx2: they match what's in the binary, vs the representation used in the wabt IR.

src/interp/binary-reader-interp.cc
SoniEx2: should we merge them? what do we do about the assert?

wasm2c/wasm-rt-impl.c
SoniEx2: yes, exnref can be null, tho the tests never throw a null exnref as far as we can tell.

aheejin:
Thanks for the work! I'll be OOO for a week from tomorrow; I'm not an expert of Wabt but I can take a look once I come back. (You don't need to wait for me in case others think the PR is good to land, for sure)

But first I'd also like to suggest to split the PR. Smaller PRs are easier to review. I think this can probably be splitted into text parsing / binary parsing / type checking / interpreter / wasm2c.

SoniEx2:
split out the wasm2c stuff

aheejin:
It looks this still contains many stuff other than the interpreter. I think parsing and printing can go first before the interpreter and split to another PR.

SoniEx2:
parsing and printing are tricky because we don't run roundtrip against spec 🙃

SoniEx2:
yeeted interp, why not

SoniEx2:
@sbc100 ping?

include/wabt/binary-reader.h
sbc100: But we do we the term "Raw" to mean "in the binary" elsewhere in this file?   i.e. what do we do for other types that have binary and wabt IR representations?

test/roundtrip/fold-unreachable.txt
```diff
@@ -26,7 +26,12 @@
   (func (result i32)
     i32.const 0
     return
-    i32.eqz))
+    i32.eqz)
+
+  (func
+    ref.null exn
+    throw_ref
+  ))
```
sbc100, line 34: Why this addition to this test?  Was the idea to put some unreachable instruction here after the throw?

sbc100 (approved):

test/roundtrip/fold-unreachable.txt
SoniEx2: oh, we probably should indeed. but we were just checking if throw_ref folds correctly.

include/wabt/binary-reader.h
SoniEx2: there are... none. the only other type that's slightly out of place is TypeMut. we'll see if we can come up with a better name...

SoniEx2:
@sbc100 how is this?

sbc100 (approved):

PR #2471, opened 2024-09-23
Fix error message for ref.is_null

### LINKED ISSUE #2453: Module fails to compile without a proper error message
The following module fails to compile:

```wasm
(module
    (func (param externref) (result i32)
        (ref.is_null)
    )
)
```

Running `wat2wasm` on this leads to an exit code if 1, but no proper error message. Debugging with GDB shows that the typechecker throws an error here:
https://github.com/WebAssembly/wabt/blob/1edeaf2948b4afd711e50ba2007ee44503cd80fe/src/type-checker.cc#L814

The problem is, that even though the `result` is now set to an error, no error message is printed, as the `PeekType` function doesn't do so. This code is obviously not valid WASM (its missing a `local.get 0` in front of `ref.is_null`), but I still would expect a proper error message.

<details><summary>Example with `i32.add`</summary>

```wasm
(module
    (func (param i32) (result i32)
        (i32.add)
    )
)
```

Gives a proper error message:
```
simple.wat:3:10: error: type mismatch in i32.add, expected [i32, i32] but got []
        (i32.add)
         ^^^^^^^
```
</details>

(NOTE: After looking at the at the following the `PeekType` call, I assume the condition should have been `type == Type::Any || !type.IsRef()`)

SoniEx2:
Fixes #2453 in a bit of a silly way. (Conveniently, we already have tests for this, but nobody noticed they were broken.)

### DIFF
--- a/test/spec/ref_is_null.txt
+++ b/test/spec/ref_is_null.txt
@@ -4,10 +4,10 @@
 init(externref:1) =>
 deinit() =>
 out/test/spec/ref_is_null.wast:52: assert_invalid passed:
-  out/test/spec/ref_is_null/ref_is_null.1.wasm:000001b: error: type mismatch in ref.is_null, expected [(ref -1)] but got [i32]
+  out/test/spec/ref_is_null/ref_is_null.1.wasm:000001b: error: type mismatch in ref.is_null, expected [reference] but got [i32]
   000001b: error: OnRefIsNullExpr callback failed
 out/test/spec/ref_is_null.wast:56: assert_invalid passed:
-  out/test/spec/ref_is_null/ref_is_null.2.wasm:0000018: error: type mismatch in ref.is_null, expected [(ref -1)] but got []
+  out/test/spec/ref_is_null/ref_is_null.2.wasm:0000018: error: type mismatch in ref.is_null, expected [reference] but got []
   0000018: error: OnRefIsNullExpr callback failed
 16/16 tests passed.
 ;;; STDOUT ;;)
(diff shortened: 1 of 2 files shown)


SoniEx2:
cc @Janrupf

test/spec/ref_is_null.txt
```diff
@@ -4,9 +4,10 @@
 init(externref:1) =>
 deinit() =>
 out/test/spec/ref_is_null.wast:52: assert_invalid passed:
-  out/test/spec/ref_is_null/ref_is_null.1.wasm:000001b: error: type mismatch in ref.is_null, expected reference but got [i32]
+  out/test/spec/ref_is_null/ref_is_null.1.wasm:000001b: error: type mismatch in ref.is_null, expected [(ref -1)] but got [i32]
```
sbc100, line 7: What does `[(ref -1)]` mean here?

Not a big deal but isn't the old error message a little more readable?

SoniEx2: It's a "reference" of an invalid type (`kInvalidIndex`). We don't know which proposal defines `Type::Reference` (typed references?) and we're misusing it a little (or arguably a lot), but it should be fairly safe since it's just for the error message.

sbc100: How was it the `reference` to printed before?  Can we find a way to keep it that way?   `expected [(ref -1)]` is not very clear is it?

SoniEx2: The old code printed its own message, without going through `PopAndCheck1Type`/`PrintStackIfFailed` like everything else. We believe the consistency is important.

We can try to turn it into `expected [reference]`, but it'll require some care.

sbc100 (approved):

PR #2472, opened 2024-09-23
Fix call_ref on empty stack

SoniEx2:
Same issue as #2471 but for `call_ref`.

We don't believe there's a prior issue for this.

### DIFF
--- a/src/type-checker.cc
+++ b/src/type-checker.cc
@@ -522,21 +522,14 @@ Result TypeChecker::OnCallIndirect(const TypeVector& param_types,
 
 Result TypeChecker::OnIndexedFuncRef(Index* out_index) {
   Type type;
-  CHECK_RESULT(PeekType(0, &type));
-  Result result = Result::Ok;
-  if (!(type == Type::Any || type.IsReferenceWithIndex())) {
-    TypeVector actual;
-    actual.push_back(type);
-    std::string message =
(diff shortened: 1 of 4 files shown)


SoniEx2:
@sbc100

src/type-checker.cc
```diff
@@ -522,21 +522,14 @@ Result TypeChecker::OnCallIndirect(const TypeVector& param_types,
...
-  if (!(type == Type::Any || type.IsReferenceWithIndex())) {
-    TypeVector actual;
-    actual.push_back(type);
-    std::string message =
-        "type mismatch in call_ref, expected reference but got " +
-        TypesToString(actual);
-    PrintError("%s", message.c_str());
-    result = Result::Error;
+  Result result = PeekType(0, &type);
+  if (!type.IsReferenceWithIndex()) {
```
sbc100, line 526: Is it OK to use `type` here even if PeekType fails?  i.e. should we return early if PeekType fails?

SoniEx2: yeah this is fine. PeekType will set Type to Any

sbc100 (approved):

SoniEx2:
(you'll have to rebase it)

SoniEx2:
(actually we'll rebase it)

PR #2474, opened 2024-09-24
Refactor binary reader to make use of section order

SoniEx2:
We wanted to see if we could rewrite binary reader to make better use of wasm's section ordering guarantees. This is a first pass attempt.

It's not entirely equivalent but we tried to preserve error messages as much as possible. Let us know what you think.


SoniEx2:
it appears that the spec folks plan to relax section order in the future, so we should probably not do this. oh well.

PR #2478, opened 2024-10-01
wasm-interp: Fix catch handlers' value stack sizes

### LINKED ISSUE #2476: weird bug with exception handling (interpreter)
hey @takikawa (noticed you wrote exception handling support for the interpreter) would you have any idea why this test: https://github.com/WebAssembly/exception-handling/pull/332

```wast
(module
  (tag $e0)
  (func $longjmp-bait (throw $e0))
  (func (export "setjmp-bait") (param $return-early i32) (result i32)
    (local $value i32)
    (try $try
      (do
        (br_if $try (local.get $return-early))
        (local.set $value (i32.const 1))
        (call $longjmp-bait)
      )
      (catch $e0)
    )
    (local.get $value)
  )
)

(assert_return (invoke "setjmp-bait" (i32.const 0)) (i32.const 1))
```

causes an assertion failure?

```
$ ./out/spectest-interp --trace ./playground/thing.json --enable-exceptions
#0.   12: V:1  | alloca 1
#0.   20: V:2  | local.get $2
#0.   28: V:3  | br_unless @44, 0
#0.   44: V:2  | i32.const 1
#0.   52: V:3  | local.set $2, 1
#0.   60: V:2  | call $0
#1.    0: V:2  | throw $0
#0.   76: V:1  | catch_drop 1
#0.   84: V:1  | local.get $1
#0.   92: V:2  | drop_keep $2 $1
#0.  104: V:0  | return
spectest-interp: /home/soniex2/git/github/wabt/src/interp/interp.cc:1078: void wabt::interp::Thread::PopValues(const ValueTypes &, Values *): Assertion `values_.size() >= types.size()' failed.
Aborted (core dumped)
```

(we noticed funny stuff at the `throw $0`, it appears the call to `DoThrow` is popping more values than it should... but we can't figure this out at all)

SoniEx2:
Fixes the value stack size of the catch handler. There were two (related) issues here:

- The previous code used `func_->locals.size()` as soon as the function was available, but it hadn't processed the function's locals yet, so it was always empty. (This might not matter in practice, as it's only used by the "function-wide catch handler", which just rethrows.)
- The previous code didn't take the function's locals into account when computing the value stack height (relative to the function frame) for a try-catch block. So, it would drop the locals when catching an exception.

Closes #2476 

(Split from #2470 )

### DIFF
--- a/src/interp/binary-reader-interp.cc
+++ b/src/interp/binary-reader-interp.cc
@@ -849,8 +849,9 @@ Result BinaryReaderInterp::EndFunctionBody(Index index) {
 Result BinaryReaderInterp::OnLocalDeclCount(Index count) {
   local_decl_count_ = count;
   local_count_ = 0;
-  // FIXME(Soni): does the value of `values` even matter here? it used to be
-  // always 0 when this call was in BeginFunctionBody.
+  // Continuation of the implicit func label, used for exception handling. (See
+  // BeginFunctionBody.)
+  // We need the local count for this, so we must do it here.
   // NOTE: we don't count the parameters, as they're not part of the frame.
   func_->handlers.push_back(HandlerDesc{HandlerKind::Catch,
                                         istream_.end(),
--- a/src/interp/binary-reader-interp.cc
+++ b/src/interp/binary-reader-interp.cc
@@ -831,13 +831,6 @@ Result BinaryReaderInterp::BeginFunctionBody(Index index, Offset size) {
   // try-delegate instruction.
   PushLabel(LabelKind::Try, Istream::kInvalidOffset, Istream::kInvalidOffset,
             func_->handlers.size());
-  func_->handlers.push_back(HandlerDesc{HandlerKind::Catch,
-                                        istream_.end(),
-                                        Istream::kInvalidOffset,
-                                        {},
-                                        {Istream::kInvalidOffset},
-                                        static_cast<u32>(func_->locals.size()),
-                                        0});
   return Result::Ok;
 }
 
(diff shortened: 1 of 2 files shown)


SoniEx2:
@sbc100

src/interp/binary-reader-interp.cc
```diff
@@ -856,6 +849,16 @@ Result BinaryReaderInterp::EndFunctionBody(Index index) {
...
   local_count_ = 0;
+  // FIXME(Soni): does the value of `values` even matter here? it used to be
+  // always 0 when this call was in BeginFunctionBody.
+  // NOTE: we don't count the parameters, as they're not part of the frame.
+  func_->handlers.push_back(HandlerDesc{HandlerKind::Catch,
+                                        istream_.end(),
+                                        Istream::kInvalidOffset,
+                                        {},
+                                        {Istream::kInvalidOffset},
+                                        static_cast<u32>(local_decl_count_),
```
sbc100, line 860: The previous code uses `func_->locals.size()` here.  How is that different?   Perhaps you could mention in the PR description how the old code is wrong?

SoniEx2: done

src/interp/binary-reader-interp.cc
```diff
@@ -856,6 +849,16 @@ Result BinaryReaderInterp::EndFunctionBody(Index index) {
 Result BinaryReaderInterp::OnLocalDeclCount(Index count) {
   local_decl_count_ = count;
   local_count_ = 0;
+  // FIXME(Soni): does the value of `values` even matter here? it used to be
+  // always 0 when this call was in BeginFunctionBody.
```
sbc100, line 853: I'm not sure this FIXME comment is useful, maybe just remove it?

sbc100 (approved):

src/interp/binary-reader-interp.cc
SoniEx2: yeah that's fair.

... ugh, the more we look at all this code the more *questionable* stuff we find, like the `func_->handlers.size()` in `BeginFunctionBody` (which, presumably, always evaluates to `0`?). but that's not the fix we came here for.

SoniEx2:
@sbc100 ping?

sbc100 (approved):

PR #2479, opened 2024-10-01
wasm2c: Fix handling of locals in setjmp targets

### LINKED ISSUE #2469: wasm2c: UB in exception-handling

SoniEx2:
It is UB to read local variables after a call to `setjmp` returns, if those variables have been modified between `setjmp` and `longjmp`, unless they're marked as `volatile`. This marks them as `volatile`.

Closes #2469

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -426,11 +426,12 @@ class CWriter {
   void WriteImportProperties(CWriterPhase);
   void WriteFuncs();
   void BeginFunction(const Func&);
-  void FinishFunction();
+  void FinishFunction(size_t);
   void Write(const Func&);
   void WriteTailCallee(const Func&);
   void WriteParamsAndLocals();
-  void WriteParams(const std::vector<std::string>& index_to_name);
+  void WriteParams(const std::vector<std::string>& index_to_name,
+                   bool setjmp_safe = false);
   void WriteParamSymbols(const std::vector<std::string>& index_to_name);
   void WriteParamTypes(const FuncDeclaration& decl);
   void WriteLocals(const std::vector<std::string>& index_to_name);
@@ -444,7 +445,7 @@ class CWriter {
   void Unspill(const TypeVector&);
 
   template <typename Vars, typename TypeOf, typename ToDo>
-  void WriteVarsByType(const Vars&, const TypeOf&, const ToDo&);
+  void WriteVarsByType(const Vars&, const TypeOf&, const ToDo&, bool);
 
   template <typename sources>
   void Spill(const TypeVector&, const sources& src);
@@ -3043,7 +3044,7 @@ void CWriter::BeginFunction(const Func& func) {
   Write(Newline());
 }
 
-void CWriter::FinishFunction() {
+void CWriter::FinishFunction(size_t stack_var_section) {
   for (size_t i = 0; i < func_sections_.size(); ++i) {
     auto& [condition, stream] = func_sections_.at(i);
     std::unique_ptr<OutputBuffer> buf = stream.ReleaseOutputBuffer();
@@ -3051,7 +3052,7 @@ void CWriter::FinishFunction() {
       stream_->WriteData(buf->data.data(), buf->data.size());
     }
 
-    if (i == 0) {
+    if (i == stack_var_section) {
       WriteStackVarDeclarations();  // these come immediately after section #0
                                     // (return type/name/params/locals)
     }
@@ -3071,6 +3072,7 @@ void CWriter::Write(const Func& func) {
   WriteParamsAndLocals();
   Write("FUNC_PROLOGUE;", Newline());
 
+  size_t stack_vars_section = func_sections_.size() - 1;
   PushFuncSection();
 
   std::string label = DefineLabelName(kImplicitFuncLabel);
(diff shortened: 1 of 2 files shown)


SoniEx2:
@sbc100

SoniEx2:
(the tests would catch this, despite UBSAN not having a setjmp sanitizer, because CI runs tests on both debug and release builds)

src/c-writer.cc
```diff
@@ -3109,6 +3112,11 @@ void CWriter::WriteVarsByType(const Vars& vars,
       if (typeoffunc(var) == type) {
         if (count == 0) {
           Write(type, " ");
+          if (setjmp_safe) {
+            PushFuncSection("exceptions");
+            Write("volatile ");
+            PushFuncSection();
```
sbc100, line 3118: I assume this means we only output this volatile when exceptions are enabled?

sbc100:
> (the tests would catch this, despite UBSAN not having a setjmp sanitizer, because CI runs tests on both debug and release builds)

I'm not quite sure what you are saying here.  Do you mean the new test you are adding would fail without the fix? 

> 
> (upstreaming is proving difficult... [WebAssembly/exception-handling#332](https://github.com/WebAssembly/exception-handling/pull/332) )

src/c-writer.cc
SoniEx2: yep!

SoniEx2:
> I'm not quite sure what you are saying here. Do you mean the new test you are adding would fail without the fix?

the test would fail without the fix, yes. like so: (note the `-O0` vs `-O3`)

```
$ cd /home/soniex2/git/github/wabt && /usr/bin/cmake -E env WASM2C_CC=/usr/bin/clang WASM2C_CFLAGS=\ -fsanitize=undefined\ -fno-sanitize-recover\ -fsanitize-blacklist=/home/soniex2/git/github/wabt/ubsan.blacklist\ -g\ -O0 /usr/bin/python3.12 /home/soniex2/git/github/wabt/test/run-tests.py --bindir /home/soniex2/git/github/wabt/out test/regress/wasm2c-ehv3-setjmp-volatile.txt 
[+1|-0|%100] (0.55s) test/regress/wasm2c-ehv3-setjmp-volatile.txt
$ cd /home/soniex2/git/github/wabt && /usr/bin/cmake -E env WASM2C_CC=/usr/bin/clang WASM2C_CFLAGS=\ -fsanitize=undefined\ -fno-sanitize-recover\ -fsanitize-blacklist=/home/soniex2/git/github/wabt/ubsan.blacklist\ -g\ -O3 /usr/bin/python3.12 /home/soniex2/git/github/wabt/test/run-tests.py --bindir /home/soniex2/git/github/wabt/out test/regress/wasm2c-ehv3-setjmp-volatile.txt 
- test/regress/wasm2c-ehv3-setjmp-volatile.txt
  expected error code 0, got 1.
  STDERR MISMATCH:
  --- expected
  +++ actual
  @@ -0,0 +1 @@
  +wasm2c-ehv3-setjmp-volatile.txt:22: assertion failed: in w2c_wasm2c__ehv3__setjmp__volatile__0__wasm_setjmp0x2Dbait(&wasm2c__ehv3__setjmp__volatile__0__wasm_instance, 0u): expected 1, got 0.
  STDOUT MISMATCH:
  --- expected
  +++ actual
  @@ -1 +1 @@
  -2/2 tests passed.
  +1/2 tests passed.


**** FAILED ******************************************************************
- test/regress/wasm2c-ehv3-setjmp-volatile.txt
    /usr/bin/python3.12 /home/soniex2/git/github/wabt/test/run-spec-wasm2c.py out/test/regress/wasm2c-ehv3-setjmp-volatile.txt --bindir=/home/soniex2/git/github/wabt/out --no-error-cmdline -o out/test/regress/wasm2c-ehv3-setjmp-volatile --enable-exceptions
```

src/c-writer.cc
SoniEx2: also, only when a try block is used.

src/c-writer.cc
```diff
@@ -3043,15 +3044,15 @@ void CWriter::BeginFunction(const Func& func) {
...
+void CWriter::FinishFunction(size_t stack_var_section) {
   for (size_t i = 0; i < func_sections_.size(); ++i) {
     auto& [condition, stream] = func_sections_.at(i);
     std::unique_ptr<OutputBuffer> buf = stream.ReleaseOutputBuffer();
     if (condition.empty() || func_includes_.count(condition)) {
       stream_->WriteData(buf->data.data(), buf->data.size());
     }
 
-    if (i == 0) {
+    if (i == stack_var_section) {
```
sbc100, line 3055: Is the stack_var_section not always the first one in a given function?

SoniEx2: not with these changes. we need to write `volatile` or not based on whether the function uses `try` or not, and we do that by pushing more func sections where relevant. in particular, we save the `stack_var_section` *after* writing params and locals (both of which may contain `volatile`), so we can put the stack vars in the correct place. (also, stack vars don't need `volatile` as any reused stack slots are written over after `longjmp`.)

sbc100 (approved):

Issue #2477, opened 2024-09-29
wasm-decompile: expose NameOpts::None as arg


python273:
It's quite inconvenient to find functions by indexes

https://github.com/WebAssembly/wabt/blob/790bc0472552d80671bdf99ba7652876c463d526/src/tools/wasm-decompile.cc#L89

sbc100:
wasm-decompile is sadly unmaintained right now.  Any contributions would be most welcome.

PR #2482, opened 2024-10-04
wasm-decompile: add function index comments

python273:


src/decompiler.cc
```diff
@@ -826,9 +826,9 @@ struct Decompiler {
         }
       }
       if (is_import) {
-        s += ";";
+        s += cat("; // f", std::to_string(func_index));
```
sbc100, line 829: How about `index:` instead of `f`?

sbc100 (commented):
I assume you find this feature useful?   Can you explain a little more about why its useful?   I assumed the point of the decompiler was so that the user would not have to worry about things like the internal index spaces that live in the binary format?

python273:
You still have to cross-reference with other tools, all of which show function indexes: Chromium debugger, .wat, [Cetus](https://github.com/Qwokka/Cetus). So it's easier to search for functions in text editors when there's a numeric index.

src/decompiler.cc
python273: I think there are other types that are also indexed? So a unique prefix seems better, can change to `func` if it makes more sense.

sbc100: Yes, I've just never seen `f0`, `f1` used to refer to function indexes, so its not obvious to me as the reader what it means.

sbc100:
> You still have to cross-reference with other tools, all of which show function indexes: Chromium debugger, .wat, [Cetus](https://github.com/Qwokka/Cetus). So it's easier to search for functions in text editors when there's a numeric index.

But doesn't the chromium debugger and the wat format both use function names (from the name section)?

Also, I assume in the case when there is no name in the name section the decompiler will use the function index as the function name?

python273:
I'm new to WASM, so not sure in which cases names are available. Here's how most functions look in a random WASM game (Unity, Emscripten):

Chromium / Firefox devtools:
`(func $func704 (param $var0 i32) (param $var1 i32)`

.wat
`(func (;704;) (type 4) (param i32 i32)`

wasm-decompile:
`function f_caa(a:int, b:int) {`

The names are only shown for imported/exported functions:
`(func $stackRestore (;415;) (export "stackRestore") (param $var0 i32)`
`(func $env._glClientWaitSync (;411;) (import "env" "_glClientWaitSync") (param i32 i32 i32 i32) (result i32))`

So in the end you need numerical indexes to navigate.

sbc100:
> I'm new to WASM, so not sure in which cases names are available. Here's how most functions look in a random WASM game (Unity, Emscripten):
> 
> Chromium / Firefox devtools: `(func $func704 (param $var0 i32) (param $var1 i32)`
> 
> .wat `(func (;704;) (type 4) (param i32 i32)`

If you build you module with a name section I think you will see the actual names here.

> 
> wasm-decompile: `function f_caa(a:int, b:int) {`

Ah... where is `caa` coming from here?   It seems like this should be f_703 in the absence of a better/real name.  Perhaps we should fix that?

> 
> The names are only shown for imported/exported functions: `(func $stackRestore (;415;) (export "stackRestore") (param $var0 i32)` `(func $env._glClientWaitSync (;411;) (import "env" "_glClientWaitSync") (param i32 i32 i32 i32) (result i32))`
> 
> So in the end you need numerical indexes to navigate.

import/exported names are just one way to give a function a name.  The better way to to use the names section.

sbc100:
BTW, this change LGTM, with comments resolved.

sbc100 (approved):

python273:
> If you build you module with a name section I think you will see the actual names here.

It's not my module and I don't have the source code :)

> Ah... where is `caa` coming from here? It seems like this should be f_703 in the absence of a better/real name. Perhaps we should fix that?

So I tried adding a cmd arg for `NameOpts::None` (#2477), but it turns it off for other things like variables or globals, making the output way less readable, as it becomes a mess of numbers. Maybe adding an option to turn it off only for function names might be nice, though alphabetic names seem better than 5 digit numbers, so not sure, opted for comments.

sbc100:
> > If you build you module with a name section I think you will see the actual names here.
> 
> It's not my module and I don't have the source code :)
> 

In those cases I would hope the decompiler would put the function index into the name, not only in a comment.


> > Ah... where is `caa` coming from here? It seems like this should be f_703 in the absence of a better/real name. Perhaps we should fix that?
> 
> So I tried adding a cmd arg for `NameOpts::None` (#2477), but it turns it off for other things like variables or globals, making the output way less readable, as it becomes a mess of numbers. 

So is `caa` just a made up string?   Is that somehow less messy than using a number?

> Maybe adding an option to turn it off only for function names might be nice, though alphabetic names seem better than 5 digit numbers, so not sure, opted for comments.

python273:
> So is `caa` just a made up string? 

It's generated from an index, so 704 = caa, 25 = z, 26 = aa, 27 = ba:

https://github.com/WebAssembly/wabt/blob/b15a13ef84b5a12aadeb1200521a89a2c95dce19/include/wabt/generate-names.h#L33-L42

> Is that somehow less messy than using a number?

Just slightly easier to read/remember `f_ficb` instead of `f37419`. But yes, it'd be a good option to add, my main problem was finding the functions having an index from debugger, so a comment works well.

sbc100:
> It's generated from an index, so 704 = caa, 25 = z, 26 = aa, 27 = ba:

Interesting!  That seems like pointless obfuscation to me... why not put the useful information in the name?

PR #2483, opened 2024-10-05
wasm-interp: Fix catch handlers correctly

SoniEx2:
oops :sweat_smile: (local decl count != local count)


SoniEx2:
@sbc100

test/regress/interp-try-catch-multi-locals.txt
```diff
@@ -0,0 +1,22 @@
+;;; TOOL: run-interp-spec
+;;; ARGS*: --enable-exceptions
+;;; NOTE: ref: issue-2476
+(module
+  (tag $e0)
+  (func (export "test") (result i32)
+    (local $a i32)
+    (local $b i32)
```
sbc100, line 8: Can we just update an existing test instead maybe ?

sbc100 (approved):

test/regress/interp-try-catch-multi-locals.txt
SoniEx2: like so?

sbc100 (approved):
Looks reasonable to me

PR #2485, opened 2024-10-08
Raise parse error on NaN in i32 and i64 literals

sjamesr:
Previously, the parser would return result::Error, but would not populate an error message.


sbc100 (approved):
Nice! Thanks for finding that.

PR #2486, opened 2024-10-08
wasm-interp: Fix off-by-one in DoThrow

SoniEx2:


sbc100 (approved):
No test output changed as the result of this?  Would writing a test for this be tricky?  

I don't really know this code myself but the change looks reasonable.

SoniEx2:
(yep:

```
  +out/test/regress/interp-throw-before-try.txt:15: expected an exception to be thrown
  +1/2 tests passed.
```

)

sbc100 (approved):

SoniEx2:
@sbc100

PR #2488, opened 2024-10-15
wasm2c: Fix WASM_RT_THREAD_LOCAL to check for __thread and add an error path

shravanrn:
wasm2c's thread_local needs two fixes

1. `WASM_RT_THREAD_LOCAL` should use `__thread` compiler intrinsic where possible (gcc, clang, icc) to avoid making C11 a hard dependency to compile the wasm2c runtime.
2. The `WASM_RT_THREAD_LOCAL` fallback when there is no thread_local implementation either from compiler intrinsics or C11 is to `#define WASM_RT_THREAD_LOCAL`, i.e., leave it empty. This is pretty unsafe, as this fallback would mean wasm2c generated code would have race conditions if two instances are run in separate threads. This PR adds an extra macro (`WASM_RT_ASSUME_SINGLE_THREADED_APP`) for users to declare their app is single threaded, at which point this fallback is allowed.


wasm2c/wasm-rt.h
```diff
@@ -81,6 +81,10 @@ extern "C" {
 
 #ifdef _MSC_VER
 #define WASM_RT_THREAD_LOCAL __declspec(thread)
+// We use __thread on POSIXy systems. This is disabled on Apple systems right
+// now due to sporadic test failures.
+#elif (defined(__GNUC__) || defined(__clang__)) && !defined(__APPLE__)
+#define WASM_RT_THREAD_LOCAL __thread
 #elif defined(WASM_RT_C11_AVAILABLE)
 #define WASM_RT_THREAD_LOCAL _Thread_local
```
sbc100, line 89: Should we prefer the standards based `_Thread_local` when it is available?  (i.e. put it first in this chain?)

sbc100 (approved):

wasm2c/wasm-rt.h
shravanrn: Yup, makes sense. Will make this change.

PR #2490, opened 2024-10-16
wasm2c: Reset the segment register after call_indirect as the register may have changed

shravanrn:


shravanrn:
@sbc100 Please have a look when you have a chance

sbc100 (approved):

keithw:
Would it be possible to share more context on when/why this is necessary? E.g., why reset it after a call_indirect but not an ordinary call?

shravanrn:
> Would it be possible to share more context on when/why this is necessary? E.g., why reset it after a call_indirect but not an ordinary call?

Sure thing. The call indirect has the potential to invoke arbitrary host functions (because function table could be imported and thus have entries set by the host), while a direct call is mostly reserved for other wasm2c generated code. For the call indirect, the code we have today works fine if the invoked host functions either (1) don't modify %gs or (2) modify and restore %gs. But if the invoked host function clobbers %gs, this poses an issue. There are potential edge cases where this could happen --- for instance in some debug/sanitizer flags are used. Resetting the values after a call indirect is the straightforward fix.

PR #2491, opened 2024-10-18
Fix bug in wasm2c's tail call optimization codegen

squk:
The default wasm-rt implementation defines `wasm_rt_memcpy` as `memcpy` which expects void*

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -1895,7 +1895,7 @@ void CWriter::WriteTailCallWeakImports() {
     Index num_results = func.GetNumResults();
     if (num_params >= 1) {
       Write(func.decl.sig.param_types, " params;", Newline());
-      Write("wasm_rt_memcpy(params, tail_call_stack, sizeof(params));",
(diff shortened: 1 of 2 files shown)


sbc100 (approved):
Does the old output code even compile?

keithw:
I'm a little confused -- we pass the tail-call tests, so... this suggests we need some additional testing here. How is this code-path not hit by the current tests?

shravanrn (approved):

keithw:
I guess there are no spec tests for tail-calls that use multi-value? If so that seems like an omission we should fix. @squk , are you up for either:

- adding a test to this repo (of using tail-call with multi-value) that this PR will make green, and/or
- submitting one upstream to the tail-call proposal repo?

If you don't have cycles for this or don't want to, I think we could probably take it from here too, but I'd love to have a test somewhere.

keithw:
We definitely want this fix and thank you for finding and fixing the bug! If you don't have the cycles to improve the testing, as I said we can probably take it from here.

PR #2432, opened 2024-06-19
Fix handling of data count without data section

SoniEx2:
Closes #2436
Fixes #2310
Fixes #2311
Fixes #2431

PR #2495, opened 2024-10-29
Update testsuite

keithw:
The memory64 `table.wast` test has started to depend on function-references and gc (which WABT doesn't support yet), so vendor an older version of the test.


sbc100:
> The memory64 `table.wast` test has started to depend on function-references and gc (which WABT doesn't support yet), so vendor an older version of the test.

BTW, the reason for this is that memory64 is based on the wasm-3.0 branch of the spec repo where those proposals are already merged.

src/binary-reader.cc
```diff
@@ -3041,7 +3041,8 @@ Result BinaryReader::ReadModule(const ReadModuleOptions& options) {
                "function signature count != function body count");
   // This is checked in ReadDataSection, but it must be checked at the end too,
   // in case the data section was omitted.
-  ERROR_IF(num_data_segments_ == 0 && data_count_ != kInvalidIndex,
+  ERROR_IF(num_data_segments_ == 0 && data_count_ != kInvalidIndex &&
+               data_count_ != 0,
```
sbc100, line 3045: Hmm.. I wonder how we missed this case in #2432?  @SoniEx2 ?

test/wasm2c/spec/memory64/table.txt
```diff
@@ -1,6 +1,82 @@
 ;;; TOOL: run-spec-wasm2c
-;;; STDIN_FILE: third_party/testsuite/proposals/memory64/table.wast
 ;;; ARGS*: --enable-memory64
+;; memory64 table.wast from last commit before wasm-3.0 merge
```
sbc100, line 3: Aren't these files automatically generated by `test/update-spec-tests.py`?    What happens when somebody runs that script again?   Should the script be able to handle this kind of thing?

sbc100 (commented):
Great!   Thanks @keithw 

lgtm % comments

keithw:
> > The memory64 `table.wast` test has started to depend on function-references and gc (which WABT doesn't support yet), so vendor an older version of the test.
> 
> BTW, the reason for this is that memory64 is based on the wasm-3.0 branch of the spec repo where those proposals are already merged.

Makes sense -- I think this is going to become more painful for WABT over time as more proposals go this direction (at least until WABT implements function-references+gc or is refactored on top of wasm-tools or binaryen).

It's caused some issues in the https://github.com/WebAssembly/testsuite/ repo; e.g. memory64 doesn't change `if.wast` or `br_if.wast` or `data.wast` relative to wasm-3.0., but `update-testsuite.sh` still created https://github.com/WebAssembly/testsuite/blob/main/proposals/memory64/br_if.wast (etc.) because those files differ from the main branch. So it's a little more difficult for us to know what .txt files to create with this revision; `test/update-spec-tests.py` wants to make a lot of extra .txt files and we can't run most of them because they depend on wasm-3.0.

src/binary-reader.cc
keithw: This test (asserting that a zero data count section without a data section is valid) wasn't part of that PR and test (https://github.com/WebAssembly/spec/pull/1801). It was added as a direct commit right afterwards (https://github.com/WebAssembly/spec/commit/7ff01f577b3cb0695bddd9083b7d09a289a3ea21#diff-63134a480c82fef0f49d50176aedf7cb823b5f0c187a4a3663165e2affbf8aedR486). So I don't think we tested for that situation at the time.

test/wasm2c/spec/memory64/table.txt
keithw: Fortunately the script won't edit a file that already exists (but it will try to create a bunch of other files, e.g. test/wasm2c/spec/memory64/if.txt because all the wasm-3.0 tests are in the memory64 part of the testsuite repo).

src/binary-reader.cc
SoniEx2: ah yeah, we made a mistake when trying to "simplify" this condition! guess there wasn't a test for it.

sbc100:
> > > The memory64 `table.wast` test has started to depend on function-references and gc (which WABT doesn't support yet), so vendor an older version of the test.
> > 
> > 
> > BTW, the reason for this is that memory64 is based on the wasm-3.0 branch of the spec repo where those proposals are already merged.
> 
> Makes sense -- I think this is going to become more painful for WABT over time as more proposals go this direction (at least until WABT implements function-references+gc or is refactored on top of wasm-tools or binaryen).
> 
> It's caused some issues in the https://github.com/WebAssembly/testsuite/ repo; e.g. memory64 doesn't change `if.wast` or `br_if.wast` or `data.wast` relative to wasm-3.0., but `update-testsuite.sh` still created https://github.com/WebAssembly/testsuite/blob/main/proposals/memory64/br_if.wast (etc.) because those files differ from the main branch. So it's a little more difficult for us to know what .txt files to create with this revision; `test/update-spec-tests.py` wants to make a lot of extra .txt files and we can't run most of them because they depend on wasm-3.0.

Ah yes, I guess we should modify update-testsuite.sh in the test suite repo so that it is aware that not all proposals forked from main.

sbc100 (approved):

sbc100:
Can we remove `test/regress/data-count-without-data-section.txt` now that its coverted by the upstream test suite?  @SoniEx2 ?

SoniEx2:
sure!

keithw:
> sure!

Done.

PR #2496, opened 2024-10-29
binary-reader-ir: Track usage of exception handling in features_used

SoniEx2:


SoniEx2:
@sbc100

sbc100 (approved):
How about "binary-reader-ir: track usage of exception handling in features_used"?

SoniEx2:
sure!

keithw:
lgtm, but shouldn't this be part of #2470 ?

SoniEx2:
we're splitting that into like 100 tiny commits because it makes review easier apparently

PR #2500, opened 2024-11-03
wasm-decompile: Fix unescaped characters in data output.

fedosgad:
Characters `"` and `\` which have special meaning in data representations are not escaped by wasm-decompile and are passed to output as is.
This PR fixes such incorrect behavior.
All tests still pass (although no cases are added).


sbc100 (approved):

sbc100:
I general, a test case would be nice for this kind of change.

PR #2501, opened 2024-11-07
wat-writer.cc: update text serialization of data memuse

keithw:
wat-writer.cc currently prints active data segments (targeting a non-0 memory) using the Wasm 1.0 format:
```wasm
(data 𝑥:memidx_𝐼 ...)
```

... but the multi-memory feature really requires the current format:
```wasm
(data id? 𝑥:memuse_𝐼 ...)
memuse_𝐼 ::= (memory 𝑥:memidx_𝐼)
```

This PR adjusts wat-writer.cc to print data segments (that target a non-0 memory) with the post-1.0 syntax. This only affects modules that use multi-memory feature so I don't think it will break anything.

Before this PR, WABT wasn't able to roundtrip this module with debug-names enabled:
```wasm
(memory 0)
(memory $k i64 (data))
```

... because after round-tripping it would be printed like this (which is invalid because it's parsed so that the data segment itself has id $k but the memuse is memory 0 which is indexed by i32):
```wasm
(module
  (memory (;0;) 0)
  (memory $k i64 0 0)
  (data (;0;) $k (i64.const 0) ""))
```

After this PR, it gets round-tripped like this:
```wasm
(module
  (memory (;0;) 0)
  (memory $k i64 0 0)
  (data (;0;) (memory $k) (i64.const 0) ""))
```

### DIFF
--- a/src/wat-writer.cc
+++ b/src/wat-writer.cc
@@ -1504,7 +1504,11 @@ void WatWriter::WriteDataSegment(const DataSegment& segment) {
   WriteOpenSpace("data");
   WriteNameOrIndex(segment.name, data_segment_index_, NextChar::Space);
   if (segment.kind != SegmentKind::Passive) {
-    WriteMemoryVarUnlessZero(segment.memory_var, NextChar::Space);
+    if (module.GetMemoryIndex(segment.memory_var) != 0) {
+      WriteOpenSpace("memory");
+      WriteVar(segment.memory_var, NextChar::Space);
+      WriteCloseSpace();
+    }
     WriteInitExpr(segment.offset);
   }
   WriteQuotedData(segment.data.data(), segment.data.size());
--- a/test/regress/write-memuse.txt
+++ b/test/regress/write-memuse.txt
@@ -1,0 +1,12 @@
+;;; TOOL: run-roundtrip
+;;; ARGS*: --stdout --enable-multi-memory --enable-memory64 --debug-names
+(module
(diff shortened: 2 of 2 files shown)


sbc100 (approved):

sbc100 (commented):
Are there not spec tests that cover this case ?

keithw:
> Are there not spec tests that cover this case ?

Unfortunately I don't think there's a spec tests that would make us parse the Wasm 1.0 format when using multi-memory by itself. We could have caught this with a roundtrip test, but I don't think we roundtrip every spec test -- only the actual roundtrip tests. And I think it would have had to be a roundtrip test that uses multi-memory and has `--debug-names` enabled on the test to catch this (and **not** `--stdout` because then we don't do the full roundtrip comparison).

SoniEx2:
we probably should roundtrip every test - it's what the spec does, as far as we can tell. but that's a whole undertaking...

keithw:
I agree that would be great (at least for valid modules) -- if it doesn't blow up testing time too much. We could make it an optional test chosen by an environment variable and run it in CI.

If we want to catch everything, ideally we'd be roundtripping with and without folding, with and without debug-names, with and without generate-names, with and without a roundtrip through the spec interpreter's (or wasm-tools's) text-to-binary and binary-to-text pipelines (since we sometimes have an out-of-date binary encoding of something that we don't catch because WABT's writer and reader are consistent).

PR #2504, opened 2024-11-08
wasm2c: minor code cleanup for OS checks for segue

shravanrn:
Define the fsgabase capability lookup to
HWCAP2_FSGSBASE as per
https://www.kernel.org/doc/html/latest/arch/x86/x86_64/fsgs.html#fsgsbase-instructions-enablement


keithw (approved):
Maybe this can link to https://www.kernel.org/doc/html/latest/arch/x86/x86_64/fsgs.html#fsgsbase-instructions-enablement (even just in the commit message) to show where this is documented?

Fair warning, my view is that we should probably work towards some sort of system of hooks (preprocessor macros that wasm2c can insert with the right command-line argument) that allows seque to live in its own file or directory that you can edit without needing a full review on each change.

PR #2506, opened 2024-11-11
wasm2c: update memory/table operations to use u64 + harmonize checks

keithw:
The PR updates the bulk memory operations (memory.fill, memory.copy, table.fill, etc.) to support 64-bit addresses and counts, and standardizes on a 64-bit version of RANGE_CHECK everywhere.

Previously we were only taking u32's for these arguments, even with memory64 enabled. (I don't think the memory64 tests check the ability to use memory.copy or the other operations beyond the first 4 GiB of a memory -- I wonder if there would be a way to add this as an "intensive" test if people don't mind having to allocate >4 GiB to run the test.)

This is a stepping-stone to being able to mix software-bounds-checked i64 memories and "guard-page-checked" i32 memories in the same module (#2507) and supporting custom-page-sizes (#2508).

### DIFF
--- a/src/template/wasm2c.declarations.c
+++ b/src/template/wasm2c.declarations.c
@@ -89,20 +89,22 @@ static inline bool func_types_eq(const wasm_rt_func_type_t a,
   (CHECK_CALL_INDIRECT(table, ft, x),       \
    DO_CALL_INDIRECT(table, t, x, __VA_ARGS__))
 
-#ifdef SUPPORT_MEMORY64
+#if __has_builtin(__builtin_add_overflow)
+#define add_overflow(a, b, resptr) __builtin_add_overflow(a, b, resptr)
+#else /* MSVC */
+static inline bool add_overflow(uint64_t a, uint64_t b, uint64_t* resptr) {
+  return _addcarry_u64(0, a, b, resptr);
+}
+#endif
+
 #define RANGE_CHECK(mem, offset, len)              \
   do {                                             \
     uint64_t res;                                  \
-    if (__builtin_add_overflow(offset, len, &res)) \
+    if (UNLIKELY(add_overflow(offset, len, &res))) \
       TRAP(OOB);                                   \
     if (UNLIKELY(res > mem->size))                 \
       TRAP(OOB);                                   \
   } while (0);
-#else
-#define RANGE_CHECK(mem, offset, len)               \
-  if (UNLIKELY(offset + (uint64_t)len > mem->size)) \
-    TRAP(OOB);
-#endif
 
 #if WASM_RT_USE_SEGUE_FOR_THIS_MODULE && WASM_RT_SANITY_CHECKS
 #include <stdio.h>
@@ -523,16 +525,16 @@ static float wasm_sqrtf(float x) {
   return sqrtf(x);
 }
 
-static inline void memory_fill(wasm_rt_memory_t* mem, u32 d, u32 val, u32 n) {
+static inline void memory_fill(wasm_rt_memory_t* mem, u64 d, u32 val, u64 n) {
   RANGE_CHECK(mem, d, n);
   memset(MEM_ADDR(mem, d, n), val, n);
 }
 
 static inline void memory_copy(wasm_rt_memory_t* dest,
                                const wasm_rt_memory_t* src,
-                               u32 dest_addr,
-                               u32 src_addr,
-                               u32 n) {
+                               u64 dest_addr,
+                               u64 src_addr,
+                               u64 n) {
   RANGE_CHECK(dest, dest_addr, n);
   RANGE_CHECK(src, src_addr, n);
   memmove(MEM_ADDR(dest, dest_addr, n), MEM_ADDR(src, src_addr, n), n);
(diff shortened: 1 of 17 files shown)


src/template/wasm2c.declarations.c
```diff
@@ -89,20 +89,22 @@ static inline bool func_types_eq(const wasm_rt_func_type_t a,
   (CHECK_CALL_INDIRECT(table, ft, x),       \
    DO_CALL_INDIRECT(table, t, x, __VA_ARGS__))
 
-#ifdef SUPPORT_MEMORY64
+#if __has_builtin(__builtin_add_overflow)
+#define add_overflow(a, b, resptr) __builtin_add_overflow(a, b, resptr)
+#else /* MSVC */
```
sbc100, line 94: Seems a little fragile to assume all compilers without `__builtin_add_overflow` must be MSVC.

How about `#else if defined(MSVC) ... #else error`

sbc100 (approved):

src/template/wasm2c.declarations.c
shravanrn: I wonder if we should also consider a simple C-only fallback, although it maybe fine to separate this change.

src/template/wasm2c.declarations.c
```diff
@@ -89,20 +89,22 @@ static inline bool func_types_eq(const wasm_rt_func_type_t a,
...
+static inline bool add_overflow(uint64_t a, uint64_t b, uint64_t* resptr) {
+  return _addcarry_u64(0, a, b, resptr);
+}
+#endif
+
 #define RANGE_CHECK(mem, offset, len)              \
```
shravanrn, line 100: Unfortunately defaulting to a 64-bit check will slow down the performance of bounds checked wasm2c for 32-bit memories. Firefox uses the bounds checked wasm2c for 32-bit devices, and so it is perf sensitive to this. I don't know if this is addressed in a future PR, but this particular PR would be a perf problem from the Firefox use case.

Also, I'll note that defaulting to the 64-bit RANGE check is fine for memcpy, tables, etc. as I don't think it will affect performance, the only concern is the 64-bit RANGE on accesses to 32-bit linear memories

shravanrn (changes requested):

src/template/wasm2c.declarations.c
keithw: Good idea... done.

shravanrn:
> I wonder if there would be a way to add this as an "intensive" test if people don't mind having to allocate >4 GiB to run the test.

Is the concern that the test-suite doesn't run on small machines?

One thought I had to run at least simple tests due to lazy memory allocation on Linux-like OSes? If tests are of the form such as below:

```
create wasm memory of 8 GB
memcpy 100 bytes at index 100
memcpy 100 bytes at index 4GB + 100
```

this should end up allocating only two physical pages for the heap. So apart from the large virtual memory footprint, the test should run fine even on small machines?

src/template/wasm2c.declarations.c
```diff
@@ -89,20 +89,24 @@ static inline bool func_types_eq(const wasm_rt_func_type_t a,
   (CHECK_CALL_INDIRECT(table, ft, x),       \
    DO_CALL_INDIRECT(table, t, x, __VA_ARGS__))
 
-#ifdef SUPPORT_MEMORY64
+#if __has_builtin(__builtin_add_overflow)
+#define add_overflow(a, b, resptr) __builtin_add_overflow(a, b, resptr)
+#elif defined(_MSC_VER)
+static inline bool add_overflow(uint64_t a, uint64_t b, uint64_t* resptr) {
+  return _addcarry_u64(0, a, b, resptr);
+}
```
sbc100, line 97: Maybe used `#define` here for consistency with above?   Or use `static inline` function above?

src/template/wasm2c.declarations.c
keithw: Thanks for these thoughtful (and well-taken) comments. I believe #2507 will nail this for you (by preserving the current RANGE_CHECK on 32-bit, default-page-size memories), so, how about we wait to get alignment on both #2506 and #2507 and then land them at the same time.

I should say that even the current RANGE_CHECK uses 64-bit arithmetic:

```
#define RANGE_CHECK(mem, offset, len)               \
  if (UNLIKELY(offset + (uint64_t)len > mem->size)) \
    TRAP(OOB);
```

... but the difference is that RANGE_CHECK64 does an explicit check for 64-bit overflow. I wish I had the benchmarking infrastructure to promise you it won't affect performance on 32-bit x86 but... safer to wait for #2507 which lets you keep the same code.

keithw:
> > I wonder if there would be a way to add this as an "intensive" test if people don't mind having to allocate >4 GiB to run the test.
> 
> Is the concern that the test-suite may not run on small machines?

Yeah, and also that a test could be really slow to run (e.g. memory.fill with an "n" greater than UINT32_MAX).

src/template/wasm2c.declarations.c

keithw:
I merged this with #2507 per the requests to avoid a regression in between; please speak up if anybody wants to re-review.

keithw:
@shravanrn Could you please release your "request changes" unless you still want something on this one?

shravanrn (approved):

keithw:
Okay, with nothing heard, will merge and move on to #2508.

PR #2507, opened 2024-11-11
wasm2c: parametrize memory bounds checks on a per-memory basis

keithw:
(Sequenced behind #2506)

This PR allows "software-bounds-checked" memories and "guard-page-checked" memories to coexist in the same module. 

It creates two versions of every memory operation: an unrestricted version (that works with any memory) and a `_default32` version (for memories with default page size and i32 indexing).

The unrestricted version calls `MEMCHECK_GENERAL`, which does a 64-bit software `RANGE_CHECK` to check that the operation reads/writes within the bounds of the memory.

The `_default32` version calls `MEMCHECK_DEFAULT32`, which is the same as the old `MEMCHECK`: if the runtime declares `WASM_RT_MEMCHECK_GUARD_PAGES`, it will do nothing. Otherwise it will do a 32-bit software `RANGE_CHECK` (which seems to be one less instruction than the 64-bit `RANGE_CHECK`).

This is a stepping stone to supporting custom-page-sizes (which will need to be software bounds-checked) (#2508).

### DIFF
--- a/src/template/wasm2c.declarations.c
+++ b/src/template/wasm2c.declarations.c
@@ -119,13 +119,23 @@ static inline bool add_overflow(uint64_t a, uint64_t b, uint64_t* resptr) {
 #define WASM_RT_CHECK_BASE(mem)
 #endif
 
+// MEMCHECK_DEFAULT32 is an "accelerated" MEMCHECK used only for
+// default-page-size, 32-bit memories. It may do nothing at all
+// (if hardware bounds-checking is enabled via guard pages)
+// or it may do a slightly faster RANGE_CHECK.
 #if WASM_RT_MEMCHECK_GUARD_PAGES
-#define MEMCHECK(mem, a, t) WASM_RT_CHECK_BASE(mem);
+#define MEMCHECK_DEFAULT32(mem, a, t) WASM_RT_CHECK_BASE(mem);
 #else
-#define MEMCHECK(mem, a, t) \
-  WASM_RT_CHECK_BASE(mem);  \
-  RANGE_CHECK(mem, a, sizeof(t))
+#define MEMCHECK_DEFAULT32(mem, a, t)                \
+  WASM_RT_CHECK_BASE(mem);                           \
+  if (UNLIKELY(a + (uint64_t)sizeof(t) > mem->size)) \
+    TRAP(OOB);
 #endif
+
+// MEMCHECK_GENERAL can be used for any memory
+#define MEMCHECK_GENERAL(mem, a, t) \
+  WASM_RT_CHECK_BASE(mem);          \
+  RANGE_CHECK(mem, a, sizeof(t));
 
 #ifdef __GNUC__
 #define FORCE_READ_INT(var) __asm__("" ::"r"(var));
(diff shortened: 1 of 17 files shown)


src/template/wasm2c.declarations.c
```diff
@@ -161,23 +171,62 @@ static inline void load_data(void* dest, const void* src, size_t n) {
...
+                   sizeof(t1));                                        \
+    force_read(result);                                                \
+    return (t3)(t2)result;                                             \
+  }                                                                    \
+  DEF_MEM_CHECKS0(name, _, t1, return, t3)
 
 #define DEFINE_STORE(name, t1, t2)                                     \
-  static inline void name(wasm_rt_memory_t* mem, u64 addr, t2 value) { \
-    MEMCHECK(mem, addr, t1);                                           \
+  static inline void name##_unchecked(wasm_rt_memory_t* mem, u64 addr, \
```
shravanrn, line 223: Not a strong opinion, but would it be simpler to have something like?

```
#define DEFINE_STORE(name, t1, t2, MEMCHK, suffix)                                  \
  static inline void name##_##suffix(wasm_rt_memory_t* mem, u64 addr,  t2 value) {  \
    MEMCHK(mem, addr, t1);                                                          \
    t1 wrapped = (t1)value;                                                         \
    wasm_rt_memcpy(MEM_ADDR_MEMOP(mem, addr, sizeof(t1)), &wrapped, sizeof(t1));    \
  }

DEFINE_STORE(i32_store, u32, u32, MEMCHECK_DEFAULT32, _default32);
DEFINE_STORE(i32_store, u32, u32, MEMCHECK_GENERAL, _general);
```

wasm2c/wasm-rt-mem-impl-helper.inc
```diff
@@ -74,23 +74,25 @@ void MEMORY_API_NAME(wasm_rt_allocate_memory)(MEMORY_TYPE* memory,
...
-  }
-  int ret = os_mprotect(addr, byte_length);
-  if (ret != 0) {
-    os_print_last_error("os_mprotect failed.");
-    abort();
-  }
-  memory->data = addr;
-#else
-  memory->data = calloc(byte_length, 1);
+  if (WASM_RT_USE_MMAP && !is64) {
```
shravanrn, line 77: Nit: Might be cleaner to have

```
// Check if we should reallocate using mmap
#if WASM_RT_USE_MMAP
if(!is64) {
  ... // original code
  return;
}
#endif

// Fallback to malloc
memory->data = calloc(byte_length, 1);
```

shravanrn (approved):

src/template/wasm2c.declarations.c
keithw: I guess I was trying to avoid the churn of duplicating every occurrence of `DEFINE_LOAD`/`DEFINE_STORE`/`DEFINE_SIMD_LOAD_FUNC`/`DEFINE_SIMD_LOAD_LANE`/`DEFINE_SHARED_LOAD`/etc. ... I think on balance the status quo is probably the less-bad but I don't feel that strongly either.

keithw:
@sbc100 Did you want to review this one? @shravanrn would prefer to land #2506 and #2507 together, so I was planning to wait until everybody is comfortable with this one before merging either.

keithw:
@sbc100 I'm going to merge this with #2506; please let me know if you want to re-review.

PR #2509, opened 2024-11-11
wast-parser.cc: Fix a crash from failing (module quote ...)

keithw:
`ParseWatModule` leaves the module as a nullptr on failure. The attached test could cause a segfault in `wast2json` before this PR.

### DIFF
--- a/test/regress/regress-quote.txt
+++ b/test/regress/regress-quote.txt
@@ -1,0 +1,8 @@
+;;; TOOL: wast2json
+;;; ERROR: 1
+(module quote "invalid")
+(;; STDERR ;;;
+out/test/regress/regress-quote.txt:3:2: error: error in quoted module: @0x100000001: unexpected token "invalid", expected a module field or a module.
+(module quote "invalid")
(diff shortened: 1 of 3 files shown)


test/regress/regress-quote.txt
```diff
@@ -0,0 +1,8 @@
+;;; TOOL: wast2json
+;;; ERROR: 1
+(module quote "invalid")
+(;; STDERR ;;;
+out/test/regress/regress-quote.txt:3:2: error: error in quoted module: @0x100000001: unexpected token "invalid", expected a module field or a module.
+(module quote "invalid")
+ ^^^^^^
```
sbc100, line 7: Maybe put this under `test/wast2json` and call it something more meaningful such as `test-invalid-quoted-modules.txt`?

sbc100 (approved):

sbc100 (approved):

test/regress/regress-quote.txt
keithw: Sounds good, done.

PR #2513, opened 2024-11-20
wasm2c: Implement EHv4

SoniEx2:
Continuation of #2470 / #2512


SoniEx2:
@sbc100 (this might still have some nits? can't remember. something about updating comments?)

SoniEx2:
@sbc100 should be good now

SoniEx2:
hello @shravanrn @keithw @sbc100 can anyone take a look at this?

wasm2c/wasm-rt-impl-tableops.inc
```diff
@@ -18,23 +18,32 @@
 // funcref or externref. For this, the file must be included after defining
 // either WASM_RT_TABLE_OPS_FUNCREF or WASM_RT_TABLE_OPS_EXTERNREF
 
-#if defined(WASM_RT_TABLE_OPS_FUNCREF) && defined(WASM_RT_TABLE_OPS_EXTERNREF)
+#if defined(WASM_RT_TABLE_OPS_FUNCREF) +       \
+        defined(WASM_RT_TABLE_OPS_EXTERNREF) + \
+        defined(WASM_RT_TABLE_OPS_EXNREF) >    \
+    1
 #error \
     "Expected only one of { WASM_RT_TABLE_OPS_FUNCREF, WASM_RT_TABLE_OPS_EXTERNREF } to be defined"
```
sbc100, line 26: This error should now include WASM_RT_TABLE_OPS_EXNREF?

wasm2c/wasm-rt-exceptions.h
```diff
@@ -70,6 +70,69 @@ uint32_t wasm_rt_exception_size(void);
  */
 void* wasm_rt_exception(void);
 
+/**
+ * The maximum size of an exception.
+ */
+#define WASM_EXN_MAX_SIZE 256
```
sbc100, line 76: In practice I think this will normally just be a 4-byte pointer.  So maybe 32 is enough for now?

src/c-writer.cc
```diff
@@ -635,6 +638,19 @@ void CWriter::PopLabel() {
   label_stack_.pop_back();
 }
 
+// static
+constexpr bool CWriter::HasNonNullInitializers(Type type) {
```
sbc100, line 642: Since this is always using in the negative how about inverting it?  i.e. "NullOnlyInitializers"?

sbc100 (approved):
Nice!

lgtm % comments

wasm2c/wasm-rt-exceptions.h
SoniEx2: 256 is what we use today, so we just used that. we would prefer not to break compatibility with existing wasm2c exception handling users (assuming we have some).

sbc100 (approved):

PR #2515, opened 2024-11-25
Issue a nicer error message on wasm components.

sunfishcode:
Decode just enough of the component binary format to recognize when the input is a component, and issue a dedicated error message for it.

Before:

0000008: error: bad wasm file version: 0x1000d (expected 0x1)

After:

0000008: error: wasm components are not yet supported in this tool


SoniEx2 (approved):

SoniEx2:
this should probably have a dedicated test for it fwiw

include/wabt/binary.h
```diff
@@ -21,6 +21,8 @@
 
 #define WABT_BINARY_MAGIC 0x6d736100
 #define WABT_BINARY_VERSION 1
+#define WABT_MODULE_LAYER 0
+#define WABT_COMPONENT_LAYER 1
```
sbc100, line 25: These should probably have `WASM_BINARY_` prefix like everything else here.

How about `WASM_BINARY_LAYER_MODULE` and `WASM_BINARY_LAYER_COMPONENT`?

sunfishcode: Done.

sunfishcode:
@sbc100 I've addressed the review feedback; would you mind taking a look? Thanks!

sbc100 (approved):

sunfishcode:
Rebased on main to pick up the CI fixes.

PR #2517, opened 2024-12-04
CI: Use latest macOS images

SoniEx2:
macos-12 is no longer available, this changes it to macos-13 (latest available x64 macos)

macos-14 is changed to macos-latest (arm64)


SoniEx2:
hmm we guess that doesn't work...

oh well

SoniEx2:
looks like we may have to drop intel mac support?

SoniEx2:
@sbc100

sbc100 (approved):

PR #2531, opened 2025-01-18
Update wasm-rt-impl.h

whirlicote:
It appears to me that the wasm_rt_impl_try macro should have been defined in the `wasm-rt-exceptions.h` header.

This pull request tries to improve the documentation on the current situation in order to improve the user experience.


sbc100:
@SoniEx2

SoniEx2:
uh, actually we believe this is a bug! think this got overlooked during the implementation of EHv3 (cc @keithw )

further, we believe this code technically triggers UB, ref.: https://www.iso-9899.info/n1570.html#7.13.1.1

PR #2535, opened 2025-01-26
optimize in for low memory device

primstar-cool:
+ add step OnSkipFunctionBodyExpr when read with skip_function_bodies
+ add virtual op code OpCodeRaw, when read with skip_function_bodies, the virtual struct will save byte code in a uint8 array, and will write directly on output
+ add macro wabt_expr_make_unique,  for override operator new
+ add MemoryAllocStream can output byte code in an allocated buffer

### DIFF
--- a/include/wabt/ir.h
+++ b/include/wabt/ir.h
@@ -430,7 +430,7 @@ enum class ExprType {
 
   First = AtomicLoad,
   Last = Unreachable,
-  OpCodeRaw = Last + 1, // virual type
+  OpCodeRaw = Last + 1,  // virual type
 };
 
 const char* GetExprTypeName(ExprType type);
@@ -842,9 +842,13 @@ class AtomicFenceExpr : public ExprMixin<ExprType::AtomicFence> {
 
 class OpcodeRawExpr : public ExprMixin<ExprType::OpCodeRaw> {
  public:
-  explicit OpcodeRawExpr(std::vector<uint8_t>&& in_opcode_buffer, FuncSignature& in_func_sig, const Location& loc = Location())
-  : ExprMixin<ExprType::OpCodeRaw>(loc), opcode_buffer(std::move(in_opcode_buffer)), is_extracted(false), func_sig(&in_func_sig) {
-  }
+  explicit OpcodeRawExpr(std::vector<uint8_t>&& in_opcode_buffer,
+                         FuncSignature& in_func_sig,
+                         const Location& loc = Location())
+      : ExprMixin<ExprType::OpCodeRaw>(loc),
+        opcode_buffer(std::move(in_opcode_buffer)),
+        is_extracted(false),
+        func_sig(&in_func_sig) {}
 
   std::vector<uint8_t> opcode_buffer;
   bool is_extracted;
--- a/include/wabt/ir.h
+++ b/include/wabt/ir.h
@@ -429,7 +429,8 @@ enum class ExprType {
   Unreachable,
 
   First = AtomicLoad,
-  Last = Unreachable
+  Last = Unreachable,
+  OpCodeRaw = Last + 1, // virual type
 };
 
 const char* GetExprTypeName(ExprType type);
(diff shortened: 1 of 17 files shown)


src/binary-reader-logging.cc
```diff
@@ -851,6 +860,8 @@ DEFINE_INDEX_DESC(OnDelegateExpr, "depth");
 DEFINE0(OnDropExpr)
 DEFINE0(OnElseExpr)
 DEFINE0(OnEndExpr)
+DEFINE_SKIP(OnSkipFunctionBodyExpr)
```
sbc100, line 863: I don't think there is any need to define a macro if there is only one caller.

src/binary-reader.cc
```diff
@@ -1939,6 +1940,17 @@ Result BinaryReader::ReadInstructions(Offset end_offset, const char* context) {
   return Result::Error;
 }
 
+Result BinaryReader::ReadSkipedFunctionBody(Offset end_offset) {
+  std::vector<uint8_t> opcode_buffer;
+  opcode_buffer.resize(end_offset - state_.offset);
```
sbc100, line 1945: Can you just combine there two lines?

`std::vector<uint8_t> opcode_buffer(end_offset - state_.offset)`

include/wabt/stream.h
```diff
@@ -195,6 +195,36 @@ class MemoryStream : public Stream {
   std::unique_ptr<OutputBuffer> buf_;
 };
 
+class MemoryAllocStream : public wabt::Stream {
```
sbc100, line 198: I don't see any uses of this new class.   Can you perhaps split this out into its own PR along with a description of why its needed?

sbc100 (commented):
I'm not sure I understand the motivation here.  The wabt interpreter (IIUC) is really just designed a reference interpreter.   I'm not sure the goal should be to run large program efficiently on resource constrained devices.    I think we can leave that the production runtimes out there.  

Perhaps you could link to the motivating issue and we could discuss more there?

include/wabt/stream.h
primstar-cool: sometimes we can't use std mem alloc (such as run in a pre-allocated mempool)
if we use MemoryStream and copy to managed buffer, it will cost double size.
in low-mem device, we need prevent it.
it can be removed from this feature, and impl by proj.

primstar-cool: i'v remove it in new commit

src/binary-reader-logging.cc
primstar-cool: i will expand the macro in next commit

src/binary-reader.cc
primstar-cool: fixed in next commit

include/wabt/ir.h
```diff
@@ -429,7 +429,8 @@ enum class ExprType {
   Unreachable,
 
   First = AtomicLoad,
-  Last = Unreachable
+  Last = Unreachable,
+  OpCodeRaw = Last + 1,  // virtual type
```
SoniEx2, line 433: correct us if we're wrong, but you should be able to put it into the enum as normal, between `Nop` and `RefIsNull`. you'll need to update the matching array in uh `ir.cc` (and maybe other places) but the alternative would be to make the code reject `OpCodeRaw` everywhere else.

also, how about calling this `UndecodedOpcodes` (between `Unary` and `Unreachable`)?

src/binary-reader-ir.cc
```diff
@@ -28,9 +28,24 @@
 #include "wabt/cast.h"
 #include "wabt/common.h"
 #include "wabt/ir.h"
-
+#ifdef WABT_OVERRIDE_ALLOC_EXPR
+extern void* operator new(size_t size, wabt::Module* m);
+#endif
 namespace wabt {
 
+#ifdef WABT_OVERRIDE_ALLOC_EXPR
```
SoniEx2, line 36: would you mind limiting this PR to `skip_function_bodies` for `ReadBinaryIr`?

src/binary-reader.cc
```diff
@@ -3142,4 +3164,16 @@ Result ReadBinary(const void* data,
       BinaryReader::ReadModuleOptions{options.stop_on_first_error});
 }
 
+Result ExtractFunctionBody(const void* data,
+                           size_t size,
+                           Location& loc,
+                           BinaryReaderDelegate* delegate,
+                           const ReadBinaryOptions& options) {
+  BinaryReader reader(static_cast<const uint8_t*>(data) - loc.offset,
```
SoniEx2, line 3172: this is undefined behavior. would it be possible to add a "virtual offset" to the `BinaryReader` instead?

SoniEx2 (changes requested):
we would also like to see some tests, see `src/test-*.cc`.

include/wabt/ir.h
primstar-cool: I was involved in a rush project so I got delayed for a few months, I'll fix them in next commit.
OpCodeRaw -> will renamed as UndecodedOpcodes
I'll put it between Unary and Unreachable. Although I think it is a special opcode, it should have its own location to avoid confusion.

src/binary-reader-ir.cc
primstar-cool: ok

src/binary-reader.cc
primstar-cool: Yes, this will access an illegal address. Although loc.offset will be added when reading, this may not pass some memory checks tools. I will use a new func to avoid this problem.
It will be fixed in next commit

Issue #1015, opened 2019-02-12
Create `wasm-objcopy` tool for manipulating wasm binary files


sbc100:
The real objcopy is the ultimate swiss army knife tool and we don't want to emulate all of its feature but we we probably want to be able to such as 

- Extracting/dumping selected section(s)
- Replacing sections
- Stripping sections

See `objcopy` options: `--remove-section`, `--add-section`, `--update-section`, `--rename-section`
https://linux.die.net/man/1/objcopyhttps://linux.die.net/man/1/objcopy

hugovincent:
It would also be useful to support `--prefix-symbols` and/or `--redefine-sym`.

sbc100:
The plan is for this type of work to happen in llvm-objcopy if possible.

PR #2538, opened 2025-02-11
implement objcopy-like tool

hovertank3d:
I needed a CLI tool to modify binary WASM sections and found that someone had already opened issue #1015. So, I attempted to implement the changes myself. I understand that this tool is still in an early stage and may have several issues that need attention. I will address these as soon as they are pointed out.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -570,6 +570,13 @@ if (BUILD_TOOLS)
     INSTALL
   )
 
+  # wasm-objcopy
+  wabt_executable(
+    NAME wasm-objcopy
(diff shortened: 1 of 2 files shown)


sbc100:
Hi @hovertank3d, thanks for the contribution to wabt.  I don't think we should rule out add such a new tool, but I wonder if llvm-objcopy could suite your needs?  If not I wonder if our efforts might be better spend improving that tool.   Also, you have looked into the existing tools in binaryen and wasm-tools repos?

hovertank3d:
Uh, I completely forgot about llvm-objcopy. Looked into repos you mentioned, they suit my needs, thanks for help! (:

PR #2540, opened 2025-02-13
CI: Use safer GITHUB_OUTPUT

SoniEx2:
apparently this is a thing https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

tho they did go back on it https://github.blog/changelog/2023-07-24-github-actions-update-on-save-state-and-set-output-commands/

but it seems like a good idea to switch anyway

### DIFF
--- a/.github/actions/release-archive/action.yml
+++ b/.github/actions/release-archive/action.yml
@@ -21,8 +21,8 @@ runs:
         mv install wabt-$VERSION
         tar -czf $TARBALL wabt-$VERSION
         scripts/sha256sum.py $TARBALL > $SHASUM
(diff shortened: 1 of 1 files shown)


sbc100 (approved):
Is there some way we can test this or confirm its correct?  Or do we just need to wait for a release?

SoniEx2:
the wasi job uploads artifacts, which is how we came across the deprecation warnings. and it seems to be still working. you can see them here: https://github.com/WebAssembly/wabt/actions/runs/13314673812/attempts/1?pr=2540

PR #2541, opened 2025-02-14
Make ReadBinaryOptions more strict across readers

### LINKED ISSUE #2537: Binary reader options and intended use cases
Currently, we have the following binary reader options:

```
  Features features;
  Stream* log_stream = nullptr;
  bool read_debug_names = false;
  bool stop_on_first_error = true;
  bool fail_on_custom_section_error = true;
  bool skip_function_bodies = false;
```

Now, `features` and `log_stream` are... fairly universal, we wanna be able to select which post-MVP features to enable and disable and oftentimes also enable verbose logging.

However, it's another story with these other options. Most of these aren't available when running the provided tools, but API users can bring their own `ReadBinaryOptions`. In theory, this leads to use-cases like:

- Emitting IR without function bodies. (#2534 / #2535)
- Interpreting a module without function bodies.
- **Interpreting a module with errors.**
- Emitting IR with errors (this may well be useful to someone, but we don't officially support it).

among other, potentially questionable ideas. However, we would like to highlight one issue in particular, and it's that our interpreter is designed to be... pretty fast. It's not trying to be the fastest thing in the world, but it does little run-time checking, instead relying entirely on the `SharedValidator` to do its job. As such, it's a dangerous idea to run (maliciously crafted) broken modules in the interpreter.

Would it make sense to more strictly enforce that many of these options are purely for use with the objdump reader? Ofc, we can still support use-cases like those in #2534 / #2535, but we (Soni) don't believe the API should look like this.

SoniEx2:
this is a major breaking change and we don't expect CIFuzz to survive, but coupling the assumptions of the binary reader delegate with the accepted binary reader options makes some things a lot easier. it does mean CIFuzz won't be able to fuzz the interpreter reader with `stop_on_first_error = false;` but we think that's for the best. (that should never have been a supported configuration, honestly.)

closes #2537


SoniEx2:
(sorry for the force push, forgot to rebase)

SoniEx2:
somehow we have managed to break the `lint` task. that's a new one.

SoniEx2:
(CIFuzz failing is expected because CIFuzz is stored out-of-tree, see https://github.com/google/oss-fuzz/tree/master/projects/wabt - this should actually be changed if we're understanding the oss-fuzz documentation correctly, but either way it'll require coordination between the repos)

PR #2406, opened 2024-03-26
Apple Silicon support 

SoniEx2:
What it says on the tin.

Closes #2366 and closes #2373

PR #2545, opened 2025-02-18
Switch to portable SIMD reads

### LINKED ISSUE #2240: wasm2c/spec/simd_address.txt fails on aarch64
When running the test suite of wabt-1.0.33 on aarch64 on Fedora 38, I get this (new) test failure:
```
- test/wasm2c/spec/simd_address.txt
  expected error code 0, got 1.
  STDERR MISMATCH:
  --- expected
  +++ actual
  @@ -0,0 +1,11 @@
  +Traceback (most recent call last):
  +  File "/builddir/build/BUILD/wabt-1.0.33/test/run-spec-wasm2c.py", line 648, in <module>
  +    sys.exit(main(sys.argv[1:]))
  +             ^^^^^^^^^^^^^^^^^^
  +  File "/builddir/build/BUILD/wabt-1.0.33/test/run-spec-wasm2c.py", line 642, in main
  +    utils.Executable(main_exe, forward_stdout=True).RunWithArgs()
  +  File "/builddir/build/BUILD/wabt-1.0.33/test/utils.py", line 95, in RunWithArgs
  +    raise error
  +utils.Error: Error running "out/test/wasm2c/spec/simd_address/simd_address" (1):
  +None
  +out/test/wasm2c/spec/simd_address/simd_address-main.c:499: assertion failed: expected w2c_simd__address__1__wasm_v1280x2Eload_offset_65521(&simd__address__1__wasm_instance) to trap.
  STDOUT MISMATCH:
  --- expected
  +++ actual
  @@ -1 +1 @@
  -42/42 tests passed.
  +41/42 tests passed.
```

SoniEx2:
Closes #2240

### DIFF
--- a/src/template/wasm2c_simd.declarations.c
+++ b/src/template/wasm2c_simd.declarations.c
@@ -1,16 +1,11 @@
-#if defined(__GNUC__) && defined(__x86_64__)
-#define SIMD_FORCE_READ(var) __asm__("" ::"x"(var));
-#elif defined(__GNUC__) && defined(__aarch64__)
-#define SIMD_FORCE_READ(var) __asm__("" ::"w"(var));
-#else
-#define SIMD_FORCE_READ(var)
-#endif
-// TODO: equivalent constraint for ARM and other architectures
-
 #define DEFINE_SIMD_LOAD_FUNC(name, func, t)                             \
   static inline v128 name##_unchecked(wasm_rt_memory_t* mem, u64 addr) { \
-    v128 result = func(MEM_ADDR(mem, addr, sizeof(t)));                  \
-    SIMD_FORCE_READ(result);                                             \
+    char tmp[sizeof(t)];                                                 \
+    const volatile char* v_addr;                                         \
+    v_addr = (const volatile char*)MEM_ADDR(mem, addr, sizeof(t));       \
+    for (int i = 0; i < sizeof(t); i++)                                  \
+      tmp[i] = v_addr[i];                                                \
+    v128 result = func(&tmp);                                            \
     return result;                                                       \
   }                                                                      \
   DEF_MEM_CHECKS0(name, _, t, return, v128);
@@ -18,8 +13,12 @@
 #define DEFINE_SIMD_LOAD_LANE(name, func, t, lane)                     \
   static inline v128 name##_unchecked(wasm_rt_memory_t* mem, u64 addr, \
                                       v128 vec) {                      \
-    v128 result = func(MEM_ADDR(mem, addr, sizeof(t)), vec, lane);     \
-    SIMD_FORCE_READ(result);                                           \
+    char tmp[sizeof(t)];                                               \
+    const volatile char* v_addr;                                       \
+    v_addr = (const volatile char*)MEM_ADDR(mem, addr, sizeof(t));     \
+    for (int i = 0; i < sizeof(t); i++)                                \
+      tmp[i] = v_addr[i];                                              \
+    v128 result = func(&tmp, vec, lane);                               \
     return result;                                                     \
   }                                                                    \
   DEF_MEM_CHECKS1(name, _, t, return, v128, v128);
--- a/src/prebuilt/wasm2c_simd_source_declarations.cc
+++ b/src/prebuilt/wasm2c_simd_source_declarations.cc
@@ -1,27 +1,18 @@
-const char* s_simd_source_declarations = R"w2c_template(#if defined(__GNUC__) && defined(__x86_64__)
-)w2c_template"
-R"w2c_template(#define SIMD_FORCE_READ(var) __asm__("" ::"x"(var));
-)w2c_template"
-R"w2c_template(#elif defined(__GNUC__) && defined(__aarch64__)
-)w2c_template"
-R"w2c_template(#define SIMD_FORCE_READ(var) __asm__("" ::"w"(var));
-)w2c_template"
-R"w2c_template(#else
-)w2c_template"
-R"w2c_template(#define SIMD_FORCE_READ(var)
-)w2c_template"
-R"w2c_template(#endif
-)w2c_template"
-R"w2c_template(// TODO: equivalent constraint for ARM and other architectures
-)w2c_template"
-R"w2c_template(
-#define DEFINE_SIMD_LOAD_FUNC(name, func, t)                             \
+const char* s_simd_source_declarations = R"w2c_template(#define DEFINE_SIMD_LOAD_FUNC(name, func, t)                             \
 )w2c_template"
 R"w2c_template(  static inline v128 name##_unchecked(wasm_rt_memory_t* mem, u64 addr) { \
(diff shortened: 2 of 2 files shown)


SoniEx2:
(wait, is this slower than the old way? hmm... think we might still be able to do something about it tho...)

SoniEx2:
(regardless of speed, this is certainly way cleaner now.)

shravanrn (changes requested):
I am worried about introducing volatile here. Volatile never quite does what anyone expects in most cases and results in quite inconsistent behavior across compilers. I think the current code although messier is much safer. My suggestion is to add this only in the cases/environments where we don't have existing options working

SoniEx2:
we can understand that.

in theory, guard pages are in-bounds of the memory object as per C rules, so we are arguably not triggering UB by accessing guard pages but merely side-effects. (this is unlike, say, trying to use NULL pointers to access the zero page.) but even then, volatile access is implementation-defined. we would personally consider any compilers that simultaneously:

1. target platforms where we can use guard pages
2. allow us to use guard pages
3. ignore the volatile access

as buggy. but there's a good chance the compiler would emit an *additional* read (and possibly a register spill) for the volatile read (especially with optimizations disabled), which isn't ideal as it degrades performance.

(there's also the question of whether the memory object itself should be considered a volatile object, such that regular accesses to it are UB - at least when guard pages are in use. we don't have a good answer to that one.)

so, considering all of that... yeah, fair enough.

shravanrn (approved):

src/template/wasm2c_simd.declarations.c
```diff
@@ -2,6 +2,9 @@
 #define SIMD_FORCE_READ(var) __asm__("" ::"x"(var));
 #elif defined(__GNUC__) && defined(__aarch64__)
 #define SIMD_FORCE_READ(var) __asm__("" ::"w"(var));
+#elif WASM_RT_MEMCHECK_GUARD_PAGES
+// best-effort using volatile
+#define SIMD_FORCE_READ(var) (void)*(volatile v128*)&var;
```
sbc100, line 7: Why only with `WASM_RT_MEMCHECK_GUARD_PAGES`, why not make this the default fallback?

sbc100: Can we use `_asm__("" ::"x"(var));` on x86 too maybe?

shravanrn: I think we don't want it as the default fallback always because:
1. We don't need this force_read operation when using bounds checks
2. It may have some performance implications

sbc100: Yes, I would expect SIMD_FORCE_READ to not be used at all if its not needed. 

Defining it to something that doesn't actually force anything seems wrong.

SoniEx2: SIMD_FORCE_READ_IF_NECESSARY?

sbc100: Lets just leave the name as is.

I do this think we should not define it to anything unless `WASM_RT_MEMCHECK_GUARD_PAGES` is set though.

SoniEx2: that would make the code way messier tho... hmm...

SIMD_CHECK_PAGE_FAULT? we already have bounds-check macros that do nothing when they're not needed...

keithw:
My understanding is that #2240 (specifically the failure on aarch64) was already fixed by #2406.

It seems like a minimalist option could be to enforce that every platform **either** uses explicit bounds-checks, or (if using guard pages) that it has a working "force read" macro. We could just bomb out if the user tries to compile with guard pages but we don't have a working force-read on that platform.

SoniEx2:
hmm, we would still like to try the volatile-based best-effort support. the tests should catch platforms where it doesn't work, then we can revisit what to do about it (probably include the bomb out for the specific platform, with an error message that instructs how to make it work). how does that sound?

shravanrn:
> hmm, we would still like to try the volatile-based best-effort support. the tests should catch platforms where it doesn't work, then we can revisit what to do about it (probably include the bomb out for the specific platform, with an error message that instructs how to make it work). how does that sound?

I'm not opposed to either trying it out and seeing how it fairs on different platforms, or falling back on bounds checks. I'll let @keithw and @sbc100 weigh in separately.

If we do end up relying on bounds checks due to lack of force_read on some platforms, I think one nice to have would be to use bounds checks for reads but rely on guard pages for writes. This is still a non trivial performance benefit. An easy way to do this would be to adjust the force read macro to do a bounds checks (you'll probably want to rename the macro and adjust its parameters if you go this route)

sbc100 (commented):
lgtm, but defering to @shravanrn

src/template/wasm2c_simd.declarations.c
```diff
@@ -1,8 +1,13 @@
...
 #if defined(__GNUC__) && defined(__x86_64__)
 #define SIMD_FORCE_READ(var) __asm__("" ::"x"(var));
 #elif defined(__GNUC__) && defined(__aarch64__)
 #define SIMD_FORCE_READ(var) __asm__("" ::"w"(var));
 #else
+// best-effort using volatile
+#define SIMD_FORCE_READ(var) (void)*(volatile v128*)&var;
+#endif
+#else
 #define SIMD_FORCE_READ(var)
```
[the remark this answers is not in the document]
sbc100: Agreed but:

(a) this a pre-existing condition that doesn't necessarily need to be fixed as part of this PR
(b) can you think of a better name that isn't overly long like `SIMD_MAYBE_FORCE_READ` or `SIMD_FORCE_READ_IF_NECESSARY`?  Maybe we leave the bikeshedding for a separate PR/issue?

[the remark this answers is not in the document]
keithw: What's new in this PR (in the current version) is that if WASM_RT_MEMCHECK_GUARD_PAGES isn't set, now SIMD_FORCE_READ will become a nop -- that's the change that I didn't super-love.

I don't think we need a `SIMD_MAYBE_FORCE_READ` because we're already careful only to call this in the "unchecked" versions of all the memory ops.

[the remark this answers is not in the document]
sbc100: I see, so maybe WASM_RT_MEMCHECK_GUARD_PAGES isn't set we simply don't need to define `SIMD_FORCE_READ` at all?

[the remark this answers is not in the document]
SoniEx2: keep in mind the "unchecked" versions are the *base* versions. they're used regardless of page size, so even tho custom-page-sizes does manual bounds checking, it's still additionally triggering this force-read when guard pages are enabled. (even tho custom-page-sizes doesn't/can't use guard pages, yes.)

how about we rename it to `SIMD_CHECK_READ`? it'd be nice to only apply this to the default32 version, but that's... difficult, currently.

PR #2546, opened 2025-02-18
Fix lint task

SoniEx2:
git clang-format appears to now return an exit status instead of always succeeding, and the change broke the script. this fixes it.

### DIFF
--- a/scripts/clang-format-diff.sh
+++ b/scripts/clang-format-diff.sh
@@ -25,8 +25,7 @@ else
 fi
 
 MERGE_BASE=$(git merge-base $BRANCH HEAD)
-if ! git clang-format $MERGE_BASE -q --diff -- src/ 2>&1 >/dev/null
-then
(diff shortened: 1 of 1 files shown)


scripts/clang-format-diff.sh
```diff
@@ -25,13 +25,12 @@ else
 fi
 
 MERGE_BASE=$(git merge-base $BRANCH HEAD)
-FORMAT_MSG=$(git clang-format $MERGE_BASE -q --diff -- src/ || true)
-if [ -n "$FORMAT_MSG" -a "$FORMAT_MSG" != "no modified files to format" ]
+if ! git clang-format $MERGE_BASE -q --diff -- src/ 2>&1 >/dev/null
 then
```
sbc100, line 29: Can you move this `; then` to the line above?

sbc100 (approved):

sbc100 (approved):

PR #2548, opened 2025-03-01
build.yml: fix s390x build

keithw:


SoniEx2 (approved):

keithw:
Thanks for the speedy response!

sbc100 (commented):
Why is this needed?

SoniEx2:
it's actually documented here: https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/customizing-github-hosted-runners#installing-software-on-ubuntu-runners

PR #2549, opened 2025-03-01
Update testsuite submodule

keithw:
This PR vendors the old "memory64" tests into "test/old-spec" and runs them from there, then updates the testsuite submodule. We skip the new "wasm-3.0" tests (which rely on function-references and gc).


sbc100 (approved):
I wonder what we will have to resort to once wasm-3.0 merges into the main branch?

keithw:
I guess it doesn't seem like anybody is jumping up to implement conforming GC in WABT (which would probably be a lot of work). My sense is that the most likely graceful evolution would probably be to port wasm2c over to the Bytecode Alliance wasm-tools codebase (in Rust) and improve wasm-tools's "wat2wasm" and "wasm2wat" subtools so they can substitute for WABT's. :-/

PR #2552, opened 2025-03-03
[CI] Remove macos-12 build

sbc100:
See https://github.com/actions/runner-images/issues/10721


keithw (approved):

sbc100:
I believe using mulitple OS images here in the wrong approach anyway.  If we actaully wanted to provide more than one set of macOS binaries would just use the latest OS and set `CMAKE_OSX_DEPLOYMENT_TARGET` to the min supported version.

PR #2555, opened 2025-03-05
wasm2c: Alter how segments are loaded

squk:
This reduces the number of usages of the `LOAD_DATA` macro from the number of segments to one. For large binaries, this greatly reduces the amount of memory required to compile the library

### DIFF
--- a/test/wasm2c/hello.txt
+++ b/test/wasm2c/hello.txt
@@ -848,12 +848,12 @@ static const u8* active_data_segment_data_ptrs[] = {
   data_segment_data_w2c_test_d0,
 };
 
-static const u32 active_data_segment_offsets[] = {
-  8u,  // data_segment_data_w2c_test_d0
-};
-
 static const size_t active_data_segment_sizes[] = {
   14,  // data_segment_data_w2c_test_d0
+};
+
+static const u32 active_data_segment_offsets[] = {
+  8u,  // data_segment_data_w2c_test_d0
 };
 
 static void init_memories(w2c_test* instance) {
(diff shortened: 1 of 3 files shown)


test/wasm2c/hello.txt
```diff
@@ -844,9 +844,24 @@ static const u8 data_segment_data_w2c_test_d0[] = {
...
+  8u,  // data_segment_data_w2c_test_d0
+};
+
 static void init_memories(w2c_test* instance) {
   wasm_rt_allocate_memory(&instance->w2c_memory, 1, 65536, 0, 65536);
-  LOAD_DATA(instance->w2c_memory, 8u, data_segment_data_w2c_test_d0, 14);
+  for (int i = 0; i < sizeof(active_data_segment_data_ptrs) / sizeof(active_data_segment_data_ptrs[0]); i++) {
+    
+    LOAD_DATA(instance->w2c_memory, active_data_segment_offsets[i], active_data_segment_data_ptrs[i], active_data_segment_sizes[i]);
+  }
```
sbc100, line 864: Do you know why the compiler was using more memory in the old form?  Is it because of inlining somehow? 

That amount of static data here is the same right?   Does the program you are having trouble with contains a huge about of segments?

sbc100:
> If possible, if you have a simple benchmark that you are using measure compile time for a wasm module with lots of segment data, that would be great to include in the benchmarks folder as well.

We tracked this down offline to the use of `-Wthread-safety-analysis` in clang when compiling the generated code.  Its not clear yet if this is bug.. but we should probably hold off landing this until we figure it out.

Issue #2551, opened 2025-03-03
Function references support


zherczeg:
I would like to parse some test files using wabt from the function reference (Phase 5) proposal such as `call_ref.wast`.

https://github.com/WebAssembly/function-references/blob/main/test/core/call_ref.wast

I have enabled the feature with `--enable-function-references`.

However, I get errors such as:
```
call_ref.wast:5:15: error: unexpected token $ii, expected ).
    (call_ref $ii (local.get $x) (local.get $f))
              ^^^
```

The format of `call_ref` was changed in:
https://github.com/WebAssembly/function-references/commit/2b194cef87ba7f71c187b063a2ff8fea878e900a

May I ask the status of function reference support in wabt? Do you plan to support these proposal changes?

Thank you very much for the help!

sbc100:
This proposal is not yet fully supported.  If you would like to take a look at adding support that would be great.  Sadly wabt doesn't have many folks contributing these days to its hard to say when the proposal might get implemented.

keithw:
I think my own long-term plan is to get wasm-tools's wat2wasm and wasm2wat (https://github.com/bytecodealliance/wasm-tools/) to the point where they can substitute fully for WABT's wat2wasm/wasm2wat, and then sunset WABT. wasm-tools already supports function-references and GC.

It would be helpful to hear: what are some of the features that you use from WABT, that wasm-tools is currently missing?

zherczeg:
Thank you for the information. Contributing to wabt might be possible, but I cannot promise anything at the moment.

Wabt is C++, and can be added to C++ projects.

zherczeg:
I will try to make a patch for this, but I have little experience with the project, so I need your guidance.

Could you help me with these questions:

1) How can I represent an s33 value?
For example `type` can be a value type or func index:
https://github.com/WebAssembly/wabt/blob/main/include/wabt/binary-reader.h#L298

2) How can I store a type in `Val` structure?
Is this correct? The index is a large value, but still part of `uint32_t` range.
https://github.com/WebAssembly/wabt/blob/main/src/wast-parser.cc#L959

3) Type with argument looks like something new. `-0x1c` type (ref ht) has a  `$t : heaptype`. How can I represent this?

zherczeg:
To start from somewhere, I have created a branch:  #2562

zherczeg:
This issue can be closed. There are 6 patches implementing the whole GC support.

sbc100:
[...]

PR #2562, opened 2025-03-11
Add support for (ref ...) / (ref null ...) constructs

zherczeg:
This large patch adds function references support for those parts of the wabt library, which already present in the code. The current code is designed for an older (outdated) variant of the function references proposal.

Key changes:
- Var variables can optionally store type data. This way there is no need for a Var/Type field pair in many cases.
- CallRef now needs to explicitly provide its reference type (no auto detection)
- Tracking local reference initialization (non nullable refs must be set before using them)
- Supporting type equality comparisons in the validator and in the interpreter
- Adds a code for reading/writing 33 bit integers (leb format)
- Remove EndTypeSection in shared validator, types can be validated earlier
- Improve named reference resolving in the text parser


zherczeg:
@sbc100 @keithw since I don't know enough about the project, these are just some random changes in the parser. 

The aim is parsing https://github.com/WebAssembly/function-references/blob/main/test/core/call_ref.wast

It is doing something (parsing correctly is surely an exaggeration) with `call_ref $ii` and `ref null $ii`. The next item is `ref.null $ii`. It looks like it will need API changes, since a token type will not be enough. Would adding an index to it be a good idea?

Could you check that these changes makes sense? There is a `Type::Reference` enum in the code, but function reference introduces two other enums, and does not use Reference. This is strange to me.

Thank you for the help.

zherczeg:
This patch goes to a new direction. The 32 bit type filed of `Var` is split into two 16 bit fields. The second can be used to store WebAssembly types (negative 7 bit numbers). This way `Var` can be used in a more generic way.

zherczeg:
There is something I don't understand. There is a `Reference = -0x15` in type.h. However, in
https://github.com/WebAssembly/function-references/blob/main/proposals/function-references/Overview.md
`(ref ht)` is defined as -0x1c and `(ref null ht)` is -0x1d. Is the old code obsolete?

The key change of the patch is that `Var` can also behave like `Type`, and transformation between the two is possible after names are resolved.

Btw, may I ask how actively maintained the wabt project?

sbc100:
I believe `ht` refers to Heap Type which would not be needed for function references.   I believe heap types refer to things like structs and arrays in the GC proposal.

zherczeg:
That is possible. Unfortunately not everything is described clearly in the text. For example, `ref.null ht: [] -> [(ref null ht)]` Which encoding is used for `ht` here? 

https://github.com/WebAssembly/function-references/blob/main/proposals/function-references/Overview.md

zherczeg:
I have tried to track down the v8 implementation for `ref.null`.

https://github.com/v8/v8/blob/main/src/wasm/function-body-decoder-impl.h#L386

It calls `read_heap_type`. It is in the same file.

https://github.com/v8/v8/blob/main/src/wasm/function-body-decoder-impl.h#L217

Starts with:
decoder->read_i33v<ValidationTag>(pc, "heap type");

So it looks like it uses the s33 decoding which the specification mentioned.

zherczeg:
I have some good news! This test is fully working now (including the interpreter):
https://github.com/WebAssembly/function-references/blob/main/test/core/call_ref.wast

Only 9 fails remained, they all look like CallRef syntax change related, so some tests need to be updated. 

The patch is quite large. There are some parts which are questionable. I will comment about these parts later to help review.

include/wabt/type.h
```diff
@@ -46,6 +46,8 @@ class Type {
     FuncRef = -0x10,    // 0x70
     ExternRef = -0x11,  // 0x6f
     Reference = -0x15,  // 0x6b
```
zherczeg, line 48: This is completely missing from the proposal. Maybe it was removed at some point? Currently Ref and Reference are considered equal, and randomly used one or the other. I wanted to wait a feedback before doing anything with it.

src/binary-reader-logging.cc
```diff
@@ -719,10 +719,15 @@ Result BinaryReaderLogging::OnGenericCustomSection(std::string_view name,
     return reader_->name(value);                  \
   }
 
-#define DEFINE_TYPE(name)                         \
-  Result BinaryReaderLogging::name(Type type) {   \
-    LOGF(#name "(%s)\n", type.GetName().c_str()); \
-    return reader_->name(type);                   \
+#define DEFINE_TYPE(name)                                        \
```
zherczeg, line 722: Hm, this might be reverted. I will check.

src/binary-reader-objdump.cc
```diff
@@ -897,7 +897,7 @@ Result BinaryReaderObjdumpDisassemble::OnOpcodeType(Type type) {
   if (!in_function_body) {
     return Result::Ok;
   }
-  if (current_opcode == Opcode::SelectT) {
+  if (current_opcode == Opcode::SelectT || current_opcode == Opcode::CallRef) {
```
zherczeg, line 900: Maybe more opcodes could go here (ref.null?). This is confusing for me, the two toString variants works differently, although they could be merged.

src/leb128.cc
```diff
@@ -288,6 +286,40 @@ size_t ReadS32Leb128(const uint8_t* p,
   }
 }
 
+size_t ReadS33Leb128(const uint8_t* p,
```
zherczeg, line 289: These readers don't check that the minimum amount of bytes is used for encoding. This is important for utf, not sure for leb.

src/wast-parser.cc
```diff
@@ -545,6 +546,17 @@ Result ResolveFuncTypes(Module* module, Errors* errors) {
     }
 
     if (func) {
+      if (!func->local_type_list.empty()) {
```
zherczeg, line 549: This is kind of a bugfix. Named references has been never resolved for local types. Maybe we could do it on the compressed list, but that can be done in another patch.

test/spec/function-references/call_ref.txt
```diff
@@ -0,0 +1,19 @@
+;;; TOOL: run-interp-spec
```
zherczeg, line 1: It turned out that `call_ref.wast` is already in the repository, and it is passing now. There are other tests in the directory (e.g. `binary.wast`) which is also passing now, they could be updated later as well.

zherczeg:
I think the patch is ready. There are lots of follow-up work ahead.

src/validator.cc
```diff
@@ -189,13 +189,32 @@ void ScriptValidator::PrintError(const Location* loc, const char* format, ...) {
   errors_->emplace_back(ErrorLevel::Error, *loc, buffer);
 }
 
+static Result CheckType(Type actual, Type expected) {
```
zherczeg, line 192: If I understand correctly, this is only used by comparing expected results. This means "module" context is not present, and expected values are immediate values. So maybe an even restricted checker could work.

include/wabt/ir.h
```diff
@@ -259,8 +284,8 @@ struct FuncSignature {
   // So to use this type we need to translate its name into
   // a proper index from the module type section.
   // This is the mapping from parameter/result index to its name.
-  std::unordered_map<uint32_t, std::string> param_type_names;
-  std::unordered_map<uint32_t, std::string> result_type_names;
+  std::unordered_map<uint32_t, Var> param_type_names;
```
zherczeg, line 287: If the name is not found, we need to throw an error. A string has no location. I think a Var is nicer than creating a name / location pair.

include/wabt/type.h
```diff
@@ -46,6 +46,8 @@ class Type {
     FuncRef = -0x10,    // 0x70
```
zherczeg, line 46: FuncRef is the same as `(ref null func)`, but `(ref func)` is not a possible value at the moment. I would use the index argument to store a bool like argument. I don't plan to do it now, but future tests needs it, so we can discuss this early.

zherczeg:
This patch and #2567 covers the majority of function reference proposal. There are still missing pieces, such as compile time checking that all non-null references are set before use, and typed tables.

Is there anybody here who can review these patches or how reviewing works here?

zherczeg:
I got no reply for my previous comment. This patch perfectly aligns with the aims of the project, it improves the current implementation, and still there is no feedback. I suspect there is something I don't know. Another project was mentioned in #2551 . Is this project in some kind of a maintenance mode, and improvements are not accepted anymore? There is no problem with negative news as long as it is clearly communicated.

SoniEx2:
expect slow responses, we don't know about other maintainers but we've been a bit busy... will review when we have time

include/wabt/type.h
zherczeg: In v8 it is kStructRefCode:

Probably used by GC. Probably kArrayRefCode will be needed as well.

src/shared-validator.cc
```diff
@@ -282,7 +284,7 @@ Result SharedValidator::OnElemSegmentElemType(const Location& loc,
   if (elem.is_active) {
     // Check that the type of the elem segment matches the table in which
     // it is active.
-    result |= CheckType(loc, elem.table_type, elem_type, "elem segment");
+    result |= CheckType(loc, elem_type, elem.table_type, "elem segment");
```
zherczeg, line 287: I just found this bug. The actual/expected arguments are swapped. The tests are updated accordingly.

sbc100: Is this is pre-existing bug?  If so maybe it can be split out into a separate PR?

zherczeg:
Patch is rebased on the top of #2571 Thanks to the fixes, it is now < 900 lines of new code including test changes.

zherczeg:
This patch now fails with a fuzzer error. Reason:

https://github.com/WebAssembly/wabt/blob/main/fuzzers/read_binary_interp_fuzzer.cc#L37
The comment explicitly says that `stop_on_first_error` must be false.

However:
https://github.com/WebAssembly/wabt/blob/main/src/interp/binary-reader-interp.cc#L634
This check can fail now. The failed fuzzer test has a `(ref 5)` global type, and there is no type 5, so this is correct. Then the global is not added to `module_.globals`. Furthermore, the binary parsing does not stop as well.

This time the exports are parsed and this line crashes:
https://github.com/WebAssembly/wabt/blob/main/src/interp/binary-reader-interp.cc#L634

The `validator_.OnExport` does not fail, since it thinks 21 globals are available, although the parsing is failed after the first global, and all globals are skipped.

zherczeg:
I have a question. In this variant of the patch, Globals / Tables / etc. uses `Type` structures, and the parser may update the named references, and validate the index references. However, `call_ref` and `ref.null` uses `Var` structures. I don't know how types should be handled in general, since many WebAssembly structures uses `Var`, but types are optionally use references.

zherczeg:
This is the next patch for supporting the function references proposal. It focuses on improving the wabt supported elements of WebAssembly. It is a feature patch now, the bugfixes were extracted and landed as separate patches.

The key concept changes of this patch:

- Type comparison (type-checker) requires the types hashmap (shared-validator). This circular dependency can be broken by a new type comparison class, which is used by both type-checker and shared-validator.

- Vars can optionally store types now without increasing its structure size. This simplify passing information, but does not necessary fit for the design concepts of the wabt project.

- RefNull and CallRef use (the new) Vars, instead of types. Many structures uses Vars in the internal representation, so this is not necessary new. However, using Vars for types is something new and debatable.

Please let me know your opinion about these design concepts. Since the parser can store pointers to types, and check/resolve them later, it is possible to use types only. But using Vars directly is not necessary a problem.

src/leb128.cc
sbc100: Can you explain a little more what you mean here?

For sure we allow encoding that are over-long, i.e. zero-padded.  This is used a lot in the llvm wasm writer, for example.

sbc100 (commented):
Thanks for all your work on this!

I'm excited to see this land but I'm also having some trouble groking the whole change since its so large.   

I think the parts that could use some more explanation, are the changes to ir.h an in particular the hybrid use of `struct Var`.   Perhaps could you could update the PR description which an explanaation of the various parts of the this change?

include/wabt/interp/interp.h
```diff
@@ -184,6 +184,7 @@ struct FuncType : ExternType {
 
   ValueTypes params;
   ValueTypes results;
+  std::vector<FuncType>* func_types;
```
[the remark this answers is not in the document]
zherczeg: This is needed for supporting a complex requirement:

https://github.com/WebAssembly/testsuite/blob/d76759e746f3564a03f6106ae19679742f2a1831/proposals/function-references/type-equivalence.wast#L112

We need to match the import/export references across modules. To do that, the interpreter needs to see the whole type tree (where the references are extracted to other types).

include/wabt/ir.h
```diff
@@ -63,17 +80,25 @@ struct Var {
...
   }
+  Type::Enum opt_type() const {
+    assert(has_opt_type());
+    return static_cast<Type::Enum>(opt_type_);
+  }
 
   void set_index(Index);
   void set_name(std::string&&);
   void set_name(std::string_view);
+  void set_opt_type(Type::Enum);
```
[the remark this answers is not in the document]
zherczeg: Yes. For most Vars (e.g. names of WebAssembly constructs) there is no need for any type.

include/wabt/ir.h
```diff
@@ -727,9 +753,7 @@ class CallRefExpr : public ExprMixin<ExprType::CallRef> {
   explicit CallRefExpr(const Location& loc = Location())
       : ExprMixin<ExprType::CallRef>(loc) {}
 
-  // This field is setup only during Validate phase,
-  // so keep that in mind when you use it.
```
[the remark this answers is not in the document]
zherczeg: Yes. Call ref needs to explicitly specify the type, auto detection is not allowed anymore:

https://github.com/WebAssembly/testsuite/blob/d76759e746f3564a03f6106ae19679742f2a1831/proposals/function-references/call_ref.wast#L5

src/binary-reader.cc
```diff
@@ -563,18 +602,13 @@ bool BinaryReader::IsConcreteType(Type type) {
...
-      return options_.features.exceptions_enabled();
-
     case Type::Reference:
+    case Type::Ref:
+    case Type::RefNull:
       return options_.features.function_references_enabled();
 
     default:
-      return false;
+      return IsGenericReferenceType(type);
```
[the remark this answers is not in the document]
zherczeg: Huh, that is just for removing code duplication. I have introduced `IsGenericReferenceType`, and this part of the code was exactly the same, so I used it here. Shall I copy back the original code?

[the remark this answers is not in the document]
sbc100: No this looks good, I just wanted to confirm that generic reference types are to be considered concrete types?

src/leb128.cc
```diff
@@ -329,10 +361,8 @@ size_t ReadS64Leb128(const uint8_t* p,
     return 9;
   } else if (p + 9 < end && (p[9] & 0x80) == 0) {
     // The top bits should be a sign-extension of the sign bit.
-    bool sign_bit_set = (p[9] & 0x1);
-    int top_bits = p[9] & 0xfe;
-    if ((sign_bit_set && top_bits != 0x7e) ||
-        (!sign_bit_set && top_bits != 0)) {
+    int top_bits = p[9];
+    if (top_bits != 0x7f && top_bits != 0) {
```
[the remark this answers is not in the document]
zherczeg: No, it is an optimization.

src/leb128.cc
zherczeg: I did not know that. Some encodings, such as utf8 expects a minimum number of byte encodings, so a 7 bit character cannot be encoded more than a single byte, even if the type bits are correct for longer encodings. If padding for leb is allowed then it is perfectly ok.

src/validator.cc
```diff
@@ -321,14 +355,8 @@ Result Validator::OnCallIndirectExpr(CallIndirectExpr* expr) {
...
-  Index function_type_index;
-  result_ |= validator_.OnCallRef(expr->loc, &function_type_index);
-  if (Succeeded(result_)) {
-    expr->function_type_index = Var{function_type_index, expr->loc};
-    return Result::Ok;
-  }
-
-  return Result::Error;
+  result_ |= validator_.OnCallRef(expr->loc, expr->sig_type);
+  return Result::Ok;
```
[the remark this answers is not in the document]
zherczeg: That is strange for me as well. All the other functions accumulate errors in `result_` and returns with `Result::Ok` so I just followed their style:

https://github.com/WebAssembly/wabt/blob/main/src/validator.cc#L320

src/type-checker.cc
```diff
@@ -786,17 +862,7 @@ Result TypeChecker::OnTableFill(Type elem_type, const Limits& limits) {
...
-   * This avoids having to implement funcref subtyping (for now) and matches
-   * the previous behavior where SharedValidator::OnElemSegmentElemExpr_RefFunc
-   * examined only the validity of the function index.
-   */
-  if (features_.function_references_enabled() && !force_generic_funcref) {
-    PushType(Type(Type::Reference, func_type));
-  } else {
-    PushType(Type::FuncRef);
-  }
+  PushType(Type(Type::Ref, func_type));
```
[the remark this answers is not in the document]
zherczeg: Yes, you are right!

zherczeg:
I have updated the description

src/binary-reader.cc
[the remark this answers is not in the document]
zherczeg: I renamed it to `IsConcreteReferenceType`.

zherczeg:
I hope I did all the requests. Thank you for the review!

src/leb128.cc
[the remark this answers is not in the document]
sbc100: This pattern is also repeated in other function in this file such as `ReadS32Leb128`.   Should we make this change for all function?  Maybe this can be its own PR?

src/leb128.cc
sbc100: I wonder if we can avoid duplicating code here and instead just use ReadS64Leb128 here (with a bounds check after reading it)?

include/wabt/ir.h
```diff
@@ -63,17 +80,25 @@ struct Var {
...
+  Type to_type() const;
 
   Location loc;
 
  private:
   void Destroy();
 
   VarType type_;
+  // Can be set to Type::Enum types, 0 represent no optional type.
+  int16_t opt_type_;
```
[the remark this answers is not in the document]
sbc100: Alternatively if you do need a third type for Var why not add another element to VarType and add another element to the union below?  

Sorry, I think I must be misunderstanding something here.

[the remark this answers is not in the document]
zherczeg: For indexed references, you need both a reference type (nullable / non-nullable) and an index or name (only for the text parser). The index or name is stored in the union. For non indexed types the type could go to the union.

Another alternative could be:
```
VarType {
  Name, Index, RefName, RefIndex, RefNullName, RefNullIndex, Type
}
```

But handling this seems far more complex than the current one, and we have more than enough bits (VarType is aligned to 32 bit, because the union is aligned to 32 bit)

src/leb128.cc
zherczeg: That is possible. We could add a default false boolean flag to the ReadS64Leb128 as well.

Anyway, what is the reason for the lot of code duplications (and manual loop unrollings) in this file? Performance? For consistency, I also added another duplication, but this part could be improved (reducing code size) in many ways. It can be done in another patch as well, before or after this patch.

What is your suggestion?

sbc100: For now, to keep this patch focused how about just removing the LEB changes from this PR?

I think its reasonable for now to just read S33 values using ReadS64Leb128.   Apparently this is what binaryen does.   We can come back and try to cleanup this stuff and remove the duplication later perhaps?

include/wabt/ir.h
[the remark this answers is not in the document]
sbc100: I think this is the main part of this PR that I'm still finding confusing.

@tlively WDYT about this part?

src/leb128.cc
zherczeg: Ok, I removed the S33 parsing. We should do something with it later. Please, let me know your opinion, and I will do it.

include/wabt/ir.h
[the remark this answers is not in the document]
zherczeg: @sbc100 do you think @tlively will reply?

zherczeg:
I have found a bug in RecursiveMatch and fixed it. It seems there will be no other comments to this patch, @sbc100 you are my only hope.

sbc100:
> I have found a bug in RecursiveMatch and fixed it. It seems there will be no other comments to this patch, @sbc100 you are my only hope.

I think this change is mostly looking good.   I'm still trying to find some time to really understand what is going on with the Var / opt_type thing as it seems a little odd to me still.

sbc100:
I'm on vacation for the next couple of week, but I will make a concerted effort to review this change when I get back in July.

zherczeg:
Thank you. The GC support will also likely completed by that time. Only 5 instructions remained, all type cast related.

zherczeg:
It looks like WebAssembly 3.0 is released, and GC is part of it: https://webassembly.org/news/2025-09-17-wasm-3.0/
May I ask what is the plans for these patches?

include/wabt/shared-validator.h
```diff
@@ -289,6 +277,11 @@ class SharedValidator {
     Index end;
   };
 
+  struct LocalReferenceMap {
+    Type type;
+    Index bit_index;
```
sbc100: IIRC, this is an index into local_ref_is_set_?   Perhaps it should use be size_t rather than `Index`?

Perhaps it should be called `is_set_index`?  Or something like that?

src/shared-validator.cc
```diff
@@ -470,6 +494,33 @@ Index SharedValidator::GetFunctionTypeIndex(Index func_index) const {
...
+    } else {
+      IgnoreLocalRefs();
+    }
+  }
+}
+
+void SharedValidator::IgnoreLocalRefs() {
+  if (!local_ref_is_set_.empty()) {
+    std::fill(local_ref_is_set_.begin(), local_ref_is_set_.end(), true);
+  }
```
sbc100: Maybe just remove the `if` check here?

src/binary-reader-ir.cc
```diff
@@ -920,8 +920,10 @@ Result BinaryReaderIR::OnCallIndirectExpr(Index sig_index, Index table_index) {
   return AppendExpr(std::move(expr));
 }
 
-Result BinaryReaderIR::OnCallRefExpr() {
-  return AppendExpr(std::make_unique<CallRefExpr>());
+Result BinaryReaderIR::OnCallRefExpr(Type sig_type) {
+  auto expr = std::make_unique<CallRefExpr>();
+  expr->sig_type = Var(sig_type, GetLocation());
```
sbc100, line 925: Should the contructor for CallRefExpr tage this sig_type as an argument?  (Like the RefNullExpr below?  Or is there some reason they should be different?

test/parse/bad-references.txt
```diff
@@ -11,19 +11,19 @@
   )
 )
 (;; STDERR ;;;
-out/test/parse/bad-references.txt:6:31: error: reference type out of range: 2 (max: 2)
+out/test/parse/bad-references.txt:6:4: error: reference 2 is out of range in params
```
sbc100: Is it ok that we will the `mac: xxx` in these error messages?   They seem maybe useful?

test/spec/function-references/br_table.txt
```diff
@@ -0,0 +1,78 @@
+;;; TOOL: run-interp-spec
+;;; STDIN_FILE: third_party/testsuite/proposals/function-references/br_table.wast
```
sbc100, line 2: Should the `update-spec-tests.py` script be updated to remove the exclusion of `function-references`?

include/wabt/shared-validator.h
zherczeg: Ok. I have added a comment which should clear its use.

include/wabt/type-checker.h
```diff
@@ -31,6 +32,18 @@ class TypeChecker {
  public:
   using ErrorCallback = std::function<void(const char* msg)>;
 
+  struct FuncType {
```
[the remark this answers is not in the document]
zherczeg: This things were reworked later in the GC support. The code is good enough.

src/binary-reader-ir.cc
zherczeg: I realized (I made the code months ago) I just followed the other Call structs.

https://github.com/WebAssembly/wabt/blob/main/include/wabt/ir.h#L725
https://github.com/WebAssembly/wabt/blob/main/include/wabt/ir.h#L694

The wast parser first creates the object, then parses its arguments. Shall I change this case?

zherczeg:
The title was misleading, I am sorry. I thought I can do this in one patch, but the full support is divided into three patches. I have updated the title.

test/parse/bad-references.txt
zherczeg: I have added it back. However, I am not sure it is that helpful in this case.  The original messages were produced by the parser, now the shared validator produces them. The difference is, that implicit function declarations are now part of the max, since the shared validator has no idea that they are implicit declarations. Furthermore, just because you pass a number less than max, it is not necessarily valid, which can also be confusing for some people. An interesting side effect is that max also grows with every new type declaration, so it is a bit more precise.

zherczeg:
I have rebased and updated the patch. Please let me know if you need more changes.

src/binary-reader-ir.cc
sbc100: I'm not sure what the best pattern is but we should try to reach consistency eventually.   Happy to land this now and cleanup later.

sbc100:
lgtm! 

(What about update-spec-tests.py?)

sbc100 (approved):

zherczeg:
> (What about update-spec-tests.py?)

I think we should update it in the third patch when the feature is complete.

sbc100:
Thank you Zoltan, and again sorry for the absurd review time.

zherczeg:
No problem. These changes are rather complex which means higher risk from project perspective, and needs more time to review. There are two more patches for function references: #2567 implements new opcodes (I have rebased it) and  #2593 implements table initializers. I will add README and update-spec-tests.py changes to the last one.
Could you review them as well? Thank you!

PR #2564, opened 2025-03-18
wasm2c: Add Segue support for FreeBSD

shravanrn:
FreeBSD requires slightly different system calls than Linux to enable Segue, but otherwise the code is mostly the same. The biggest difference is that Linux provides [getauxval](https://www.kernel.org/doc/html/next/x86/x86_64/fsgs.html) function to check for support for the `wrgsbase` instruction. In FreeBSD, this is always supported if the hardware supports it, so we have to query cpuid rather than an OS functions.


shravanrn:
@sbc100 @keithw Any chance one of you could have a look at this?

sbc100 (approved):

PR #2567, opened 2025-03-20
Implement ref.as_non_null, br_non_null, br_on_non_null instructions

zherczeg:
Depends on #2562

Another 1000 lines addition. Depends on the previous work. Currently implements 3 opcodes, but I might add the remaining opcode (return_call_ref).


sbc100:
Which proposal are these instruction part of?

zherczeg:
Thank you for checking the patch.

These are the instructions:
https://github.com/WebAssembly/function-references/blob/main/proposals/function-references/Overview.md#instructions-1

I am working on function references phase 5 proposal at the moment, which is requirement for the (latest) GC proposal. The work is split into two parts:

#2562 adds the concepts of `(ref ...)` and `(ref null ...)` to many parts of the engine. The big change is that references has an extra argument, and the code needs to handle these non primitive types.

This patch adds the instructions from the proposal.

This is the generic testsuite, also used by wabt:
https://github.com/WebAssembly/testsuite/tree/main/proposals/function-references

My aim is to make all test pass in this directory.

zherczeg:
What can I do with this test:
```
- test/spec/function-references/return_call_ref.txt
    TIMEOUT
```
The test is also slow on x86. It does a lot of recursion. I did not write the test, it is a spec test.

SoniEx2:
you might be able to add `;;; SLOW:` to it. hopefully that gives it enough time to run.

zherczeg:
Great idea! Thanks it worked!

zherczeg:
This patch is also updated. It is top of #2571 / #2562 It is a large patch, but easy to review.
Still missing: tables with initializers. I plan to do it like globals, since a table in theory is a global with multiple elements. The initializers returns with the default element.

zherczeg:
My suggestion: I suspect objdump does not validate the code intentionally, just dumps it. Then validation related sanity checks does not make sense, so maybe setting `in_function_body` to true even for init expressions is a valid solution.

src/shared-validator.cc
```diff
@@ -1192,6 +1210,16 @@ Result SharedValidator::OnReturnCallIndirect(const Location& loc,
   return result;
 }
 
+Result SharedValidator::OnReturnCallRef(const Location& loc,
+                                        Var function_type_var) {
+  Result result = CheckInstr(Opcode::ReturnCallRef, loc);
+  FuncType func_type;
+  result |= typechecker_.OnReturnCallRef(function_type_var.to_type());
+  result |= CheckFuncTypeIndex(function_type_var, &func_type);
+  result |= typechecker_.OnReturnCall(func_type.params, func_type.results);
```
zherczeg, line 1219: This code is based on the original implementation in 9aac0a28896a714b698ab489a473b9bc05deb7c3, although OnReturnCallIndirect uses its own function which report errors as "call_indirect" rather than "call". Maybe adding a string "descriptor" argument would be better than duplicating the whole function. Call indirect has the code duplication as well, perhaps we could simplify the whole thing with the string argument.

src/apply-names.cc
```diff
@@ -350,6 +352,18 @@ Result NameApplier::OnBrIfExpr(BrIfExpr* expr) {
...
+Result NameApplier::OnBrOnNonNullExpr(BrOnNonNullExpr* expr) {
+  std::string_view label = FindLabelByVar(&expr->var);
+  UseNameForVar(label, &expr->var);
+  return Result::Ok;
+}
+
+Result NameApplier::OnBrOnNullExpr(BrOnNullExpr* expr) {
+  std::string_view label = FindLabelByVar(&expr->var);
+  UseNameForVar(label, &expr->var);
+  return Result::Ok;
```
sbc100, line 364: Unrelated to this PR but this seems to be a repeated pattern here.. i wonder if we could refactor/tempetize/macro somehow?

src/shared-validator.cc
sbc100: Seems like a good idea, could certainly be in a followup too

sbc100 (approved):

src/apply-names.cc
zherczeg: True. I will do it.

PR #2568, opened 2025-03-21
feat(demo): Run js in worker in wat2wasm demo

spotandjake:
This pr moves execution of the user js and wasm module from the main page to a web worker. The benefit of this is if you write an infinite loop in either the js or wasm you won't crash the main page and can recompile. I also added a `Run` button as I noticed there wasn't one and it was useful for testing (when you edit the wasm the output doesn't update until you click run or edit the js).

### NOTE
As the js is now running in a worker it is isolated so while before you could have worked with the dom from the js editor you can't really do this anymore. I feel like the old behaviour didn't make much sense anyways though.


spotandjake:
Hope I'm not bugging anyone, but any chance I could get a review on this?

docs/demo/wat2wasm/worker.js
```diff
@@ -0,0 +1,29 @@
...
+      type: "log",
+      data: String(e)
+    });
+  }
+  const fn = new Function('wasmModule', 'console', js + '//# sourceURL=demo.js');
+  fn(wasm, wrappedConsole);
+  postMessage({
+    type: "done"
+  });
+}
```
sbc100: Missing newline at EOF

docs/demo/wat2wasm/worker.js
```diff
@@ -0,0 +1,29 @@
+let wrappedConsole = Object.create(console);
```
sbc100, line 4: Maybe add a comment at the top of this file explaining what it does and why it exists?

spotandjake:
> Fair enough!
> 
> Seems like it might be overkill for a simple demo but if you find it useful to be able to avoid blocking the main page then lgtm.
> 
> Its been while since we changed this I think.. do you know how we publish the new version? Is it automatic?

I know that this is just a demo however I know myself and others in the wasm community sometimes use this tooling to explore the spec and play around hands on with wasm in a convient and quick manner.

I wasn't around the last time this was updated and can't find any specific workflows for it however I did notice that github pages seems to run on every commit to main [logs here](https://github.com/WebAssembly/wabt/deployments/github-pages) so it seems like this should just get updated. 

Seperatly though this is probably best discussed in another issue it might make sense to run `make demo` on release to regenerate the `libwabt.js` file. 


Thankyou for the review.

kevmoo:
Noted this was approved but never landed?

spotandjake:
I think this is good to be merged, I think it was never landed because of @sbc100's queston about the release process.

I'll make an issue about updating `libwabt.js` with `make demo` on release once this is merged.

PR #2571, opened 2025-03-26
Fix reference type processing

zherczeg:
Parse (ref index) form, not just (ref $name) form
Resolve references in types and locals
Display location when a named reference is not found

### DIFF
--- a/include/wabt/ir.h
+++ b/include/wabt/ir.h
@@ -255,12 +255,25 @@ struct FuncSignature {
   TypeVector param_types;
   TypeVector result_types;
 
-  // Some types can have names, for example (ref $foo) has type $foo.
-  // So to use this type we need to translate its name into
-  // a proper index from the module type section.
-  // This is the mapping from parameter/result index to its name.
-  std::unordered_map<uint32_t, std::string> param_type_names;
-  std::unordered_map<uint32_t, std::string> result_type_names;
+  // Reference types can have names, for example (ref $foo) represents
+  // a type which name is $foo. These types are translated to their
+  // corresponding index after the parsing is completed. The position
+  // and names of these reference types are stored in a ReferenceVars
+  // vector. The names are stored as variables, because the error
+  // message construction requires a location when a name is not found.
+  // TODO: move out this parser specific structure from IR.
+  struct ReferenceVar {
+    ReferenceVar(uint32_t index, Var var)
+      : index(index), var(var) {}
+
+    uint32_t index;
+    Var var;
+  };
+
+  typedef std::vector<ReferenceVar> ReferenceVars;
+
+  ReferenceVars param_type_vars;
+  ReferenceVars result_type_vars;
 
   Index GetNumParams() const { return param_types.size(); }
   Index GetNumResults() const { return result_types.size(); }
@@ -925,6 +938,12 @@ struct Func {
   std::string name;
   FuncDeclaration decl;
   LocalTypes local_types;
+  // When some locals are named references, the parser keeps the
+  // non-compressed local vector until the module parsing is
+  // completed. After the locals are resolved, local_types are
+  // constructed from this vector, and then this vector is freed.
+  // TODO: move out this parser specific structure from IR.
+  TypeVector local_type_list;
   BindingHash bindings;
   ExprList exprs;
   Location loc;
--- a/include/wabt/wast-parser.h
+++ b/include/wabt/wast-parser.h
@@ -137,7 +137,8 @@ class WastParser {
   Result ParseValueType(Var* out_type);
   Result ParseValueTypeList(
       TypeVector* out_type_list,
-      std::unordered_map<uint32_t, std::string>* type_names);
+      FuncSignature::ReferenceVars* type_vars,
+      Index binding_index_offset);
   Result ParseRefKind(Type* out_type);
   Result ParseRefType(Type* out_type);
   bool ParseRefTypeOpt(Type* out_type);
@@ -176,13 +177,13 @@ class WastParser {
   Result ParseBoundValueTypeList(TokenType,
                                  TypeVector*,
                                  BindingHash*,
-                                 std::unordered_map<uint32_t, std::string>*,
+                                 FuncSignature::ReferenceVars*,
                                  Index binding_index_offset = 0);
   Result ParseUnboundValueTypeList(TokenType,
                                    TypeVector*,
-                                   std::unordered_map<uint32_t, std::string>*);
+                                   FuncSignature::ReferenceVars*);
   Result ParseResultList(TypeVector*,
-                         std::unordered_map<uint32_t, std::string>*);
+                         FuncSignature::ReferenceVars*);
   Result ParseInstrList(ExprList*);
   Result ParseTerminatingInstrList(ExprList*);
   Result ParseInstr(ExprList*);
--- a/src/wast-parser.cc
+++ b/src/wast-parser.cc
@@ -319,37 +319,59 @@ bool ResolveFuncTypeWithEmptySignature(const Module& module,
   return false;
 }
 
-void ResolveTypeName(
-    const Module& module,
-    Type& type,
-    Index index,
-    const std::unordered_map<uint32_t, std::string>& bindings) {
-  if (type != Type::Reference || type.GetReferenceIndex() != kInvalidIndex) {
-    return;
+Result ResolveRefTypes(const Module& module,
+                       TypeVector& types,
+                       uint32_t base_index,
+                       FuncSignature::ReferenceVars& ref_vars,
+                       Errors* errors) {
+  Result result = Result::Ok;
+  size_t types_size = types.size();
+
+  for (auto& ref_var : ref_vars) {
+    uint32_t index = ref_var.index;
+
+    // The index of resolved variables is converted to kInvalidIndex.
+    if (index == kInvalidIndex || index < base_index ||
+        (index - base_index) >= types_size) {
+      continue;
+    }
+
+    Type type = types[index - base_index];
+    ref_var.index = kInvalidIndex;
+
+    assert(type.IsReferenceWithIndex());
+
+    if (type.GetReferenceIndex() != kInvalidIndex) {
+      continue;
+    }
+
+    const auto type_index = module.type_bindings.FindIndex(ref_var.var.name());
+
+    if (type_index == kInvalidIndex) {
+      errors->emplace_back(ErrorLevel::Error, ref_var.var.loc,
+                           StringPrintf("undefined reference type name %s",
+                                        ref_var.var.name().c_str()));
+      result = Result::Error;
+      continue;
+    }
+
+    types[index - base_index] = Type(static_cast<Type::Enum>(type), type_index);
   }
 
-  const auto name_iterator = bindings.find(index);
-  assert(name_iterator != bindings.cend());
-  const auto type_index = module.type_bindings.FindIndex(name_iterator->second);
-  assert(type_index != kInvalidIndex);
-  type = Type(Type::Reference, type_index);
+  return Result::Ok;
 }
 
-void ResolveTypeNames(const Module& module, FuncDeclaration* decl) {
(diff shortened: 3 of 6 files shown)


zherczeg:
This is another bugfix patch extracted from #2562. The list of changes are above, these constructs should work even in the main branch.

src/wast-parser.cc
```diff
@@ -973,11 +1008,16 @@ Result WastParser::ParseValueTypeList(
     CHECK_RESULT(ParseValueType(&type));
 
     if (type.is_index()) {
-      out_type_list->push_back(Type(type.index()));
+      // TODO: Temporary fix.
```
zherczeg, line 1011: The full fix is in #2562. It is unlikely that close to UINT32_MAX number of types are defined, so this is good as a temporary change.

src/wast-parser.cc
```diff
@@ -545,6 +570,15 @@ Result ResolveFuncTypes(Module* module, Errors* errors) {
     }
 
     if (func) {
+      if (!func->local_type_list.empty()) {
+        result |= ResolveRefTypes(*module, func->local_type_list,
+                                  func->GetNumParams(),
+                                  decl->sig.param_type_names, errors);
+
+        func->local_types.Set(func->local_type_list);
```
zherczeg, line 578: This is a delayed processing of locals. The reasion is that otherwise we cannot tell that `(ref $t)` and `(ref 0)` refers to the same or different locals, so they can be merged or not.

src/ir.cc
```diff
@@ -181,7 +181,7 @@ void LocalTypes::Set(const TypeVector& types) {
   Type type = types[0];
   Index count = 1;
   for (Index i = 1; i < types.size(); ++i) {
-    if (types[i] != type) {
+    if (!types[i].IsSame(type)) {
```
zherczeg, line 184: The name might not be the best. Anyway, the purpose is that `==` only compares the enum, while the index is also important for references.

test/typecheck/missing-ref.txt
```diff
@@ -0,0 +1,26 @@
...
+out/test/typecheck/missing-ref.txt:6:25: error: Unknown reference: $second_type
+           (result (ref $second_type))
+                        ^^^^^^^^^^^^
+out/test/typecheck/missing-ref.txt:7:17: error: Unknown reference: $third_type
+    (local (ref $third_type))
+                ^^^^^^^^^^^
+out/test/typecheck/missing-ref.txt:5:24: error: Unknown reference: $first_type
+  (func $f (param (ref $first_type))
+                       ^^^^^^^^^^^
+out/test/typecheck/missing-ref.txt:6:25: error: Unknown reference: $second_type
```
sbc100, line 23: Looking at the errors that come from `resolve-names.cc` It looks like this should be `undefined reference type variable \"%s\""`.. or something like that?

src/wast-parser.cc
```diff
@@ -457,7 +474,7 @@ class ResolveFuncTypesExprVisitorDelegate : public ExprVisitor::DelegateNop {
       : module_(module), errors_(errors) {}
 
   void ResolveBlockDeclaration(const Location& loc, BlockDeclaration* decl) {
-    ResolveTypeNames(*module_, decl);
+    ResolveTypeNames(*module_, decl->sig, errors_);
```
sbc100, line 477: How is this different to the code in `src/wast-parser.cc` that does name resolving?

sbc100 (commented):
Thanks for working on this.

Mostly LGTM, just trying to page in how this stuff all works.

include/wabt/type.h
```diff
@@ -79,6 +79,11 @@ class Type {
     return IsRef();
   }
 
+  // The == comparison only compares the enum_ member.
+  bool IsSame(const Type& rhs) const {
```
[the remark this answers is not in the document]
zherczeg: Unfortunately I get errors such as `error: ambiguous overload for ‘operator==’ (operand types are ‘const wabt::Type’ and ‘wabt::Type::Enum’)`

The reason is the Type has a `constexpr operator Enum() const { return enum_; }`, and this confuses the compiler.

I am open to any ideas.

src/wast-parser.cc
zherczeg: I am not sure I understand this question. ResolveTypeNames is reworked to support error throwing.

test/typecheck/missing-ref.txt
zherczeg: I cahnged it to `undefined reference type name %s`

include/wabt/type.h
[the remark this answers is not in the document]
sbc100: This seems to work for me locally:

```
 
+  // TODO(sbc) Use =default here once we have C++20
+  bool operator==(const Type& other) const {
+    return enum_ == other.enum_ && type_index_ == other.type_index_;
+  }
+
+  bool operator==(const Enum& other) const { return enum_ == other; }
```

I tried the `explicit` route too and that seems to break a bunch of places.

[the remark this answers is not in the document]
zherczeg: Ok, I change it. Thank you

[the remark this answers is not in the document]
zherczeg: Btw it might fix another bug. Currently if we have two types which are the same except the ref part:

```
(type $t0 (func (param (ref $t0))))
(type $t1 (func (param (ref $t1))))

(func $f (param (ref $t1)) )
```
$f gets type 0, instead of type 1. The index is simply ignored.

include/wabt/wast-parser.h
```diff
@@ -137,7 +137,8 @@ class WastParser {
   Result ParseValueType(Var* out_type);
   Result ParseValueTypeList(
       TypeVector* out_type_list,
-      std::unordered_map<uint32_t, std::string>* type_names);
+      std::unordered_map<uint32_t, Var>* type_names,
```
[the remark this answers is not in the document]
zherczeg: All ...type_names are renamed to ...type_vars.

test/typecheck/missing-ref.txt
```diff
@@ -0,0 +1,26 @@
...
+out/test/typecheck/missing-ref.txt:6:25: error: undefined reference type name $second_type
+           (result (ref $second_type))
+                        ^^^^^^^^^^^^
+out/test/typecheck/missing-ref.txt:7:17: error: undefined reference type name $third_type
+    (local (ref $third_type))
+                ^^^^^^^^^^^
+out/test/typecheck/missing-ref.txt:5:24: error: undefined reference type name $first_type
+  (func $f (param (ref $first_type))
+                       ^^^^^^^^^^^
+out/test/typecheck/missing-ref.txt:6:25: error: undefined reference type name $second_type
```
zherczeg, line 23: Just a note: the error is thrown twice. The reason is that the references are resolved as func type and as a function signature. I don't think it is a problem, but we could fix it (could be later) by removing the items of param_type_vars / result_type_vars, if somebody does not like it.

include/wabt/type.h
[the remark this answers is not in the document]
zherczeg: It seems it does not work :(

`if (types[i] != type) {` still uses the enum auto conversion.
`if (!types[i].operator==(type)) {` works but does not look nice.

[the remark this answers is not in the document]
zherczeg: What shall I do?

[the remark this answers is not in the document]
sbc100: I guess removing the implicit cast is the way to go.. although as you say it might require a bunch of downstream changes.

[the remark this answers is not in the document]
zherczeg: I tried to move the operator== before the Enum casting, and the auto casting still takes precedence. I will never understand c++.

Removing the casting might have many side effects which is too big risk for this patch, so I move back to the original code for now, and we can try it later.

zherczeg:
Did some code improvements. The unordered_map is replaced by a vector, because it is a simpler structure. Furthermore, when a reference is resolved, its index is set to `kInvalidIndex` to avoid duplicated resolves.

include/wabt/ir.h
```diff
@@ -255,12 +255,24 @@ struct FuncSignature {
...
+  // a type which name is $foo. These types are translated to their
+  // corresponding index after the parsing is completed. The position
+  // and names of these reference types are stored in a ReferenceVars
+  // vector. The names are stored as variables, because the error
+  // message construction requires a location when a name is not found.
+  struct ReferenceVar {
+    ReferenceVar(uint32_t index, Var var)
+      : index(index), var(var) {}
+
+    uint32_t index;
```
SoniEx2, line 268: what is the meaning of this index?

zherczeg: Location of the reference.

Example: `(param i32 (ref $a) i64 (ref 0) f32 (ref $b))`, then two `ReferenceVar` is created: `(1, $a)` and `(5, $b)`. The index is 1 and 5.

Shall I add a comment?

zherczeg:
Thank you for the review. Is there any other changes this patch needs?

include/wabt/ir.h
```diff
@@ -255,12 +255,24 @@ struct FuncSignature {
...
+    Var var;
+  };
+
+  typedef std::vector<ReferenceVar> ReferenceVars;
+
+  ReferenceVars param_type_vars;
+  ReferenceVars result_type_vars;
```
sbc100, line 275: I guess at some point it might make sense to move this parser-specific metadata out of the IR representation?  But it seems like this change just adds to the existing pattern?

sbc100 (commented):
LGTM.

I assume that once other PRs in this chain land we can start to run some of the upstream tests instead of constructing our own. 

Would it make sense to then remove the local tests that (I assume) duplicate upstream tests?

include/wabt/ir.h
sbc100: So this whole class is only needed/used during parsing?  Should we add a TODO to move this into the parser itself in the future?

include/wabt/ir.h
zherczeg: I think yes. Yes, the param_type_names/result_type_names was there, I just changed them to another structure.

There are many text parser related stuff in the code, since declaration order is not fixed in text parsing, sometimes things are declared after use. Maybe they could be hidden behind smart pointers, so they have less memory overhead when not used anymore.

Or use class inheritance, and reallocate the signature without parser data after parsing is completed.

Ok I will think about it.

include/wabt/ir.h
```diff
@@ -925,6 +937,11 @@ struct Func {
   std::string name;
   FuncDeclaration decl;
   LocalTypes local_types;
+  // When some locals are named references, the parser keeps the
+  // non-compressed local vector until the parsing is completed.
+  // After the locals are resolved, local_types are constructed
+  // from this vector, and then this vector is freed.
+  TypeVector local_type_list;
```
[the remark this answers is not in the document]
zherczeg: Whole module unfortunately. I will update the comment.

```
(module
  (func $f (local (ref 0)) (local (ref $t)))
  (type $t (func))
)
```

The two locals are the same, and should be merged:
```
0000016: 01                                        ; local decl count
0000017: 02                                        ; local type count
0000018: 64                                        ; (ref 0)
0000019: 00                                        ; (ref 0)
```

But we don't know this until the type is parsed, which can be at the end of the module.

The new `test/roundtrip/typed-func-refs.txt` checks this merging.

zherczeg:
Yes, the follow up patches use spec tests for testing the features.

However, we cannot do it now, because "Type::Reference" is not a function reference anymore.

In v8 engine, the const is assigned to kStructRefCode, and ref/refnull is defined as 0x64/0x63

https://github.com/WebAssembly/function-references/blob/main/proposals/function-references/Overview.md#types-1

zherczeg:
Thank you for the review. I have updated the comments. I will submit a follow-up patch for extracting parser structures out from IR, but it will be a non-trivial change.

sbc100 (approved):
LGTM.  @SoniEx2 WDYT?

zherczeg:
Wait a bit please.  #2581 may replace this patch.

PR #2572, opened 2025-03-27
Remove implicit Enum conversion of types

zherczeg:
Based on #2571

Mostly switches needed the implicit conversion. Not that big patch.

### DIFF
--- a/include/wabt/decompiler-ls.h
+++ b/include/wabt/decompiler-ls.h
@@ -26,7 +26,7 @@ namespace wabt {
 
 // Names starting with "u" are unsigned, the rest are "signed or doesn't matter"
 inline const char* GetDecompTypeName(Type t) {
-  switch (t) {
+  switch (t.code()) {
     case Type::I8: return "byte";
     case Type::I8U: return "ubyte";
     case Type::I16: return "short";
@@ -56,9 +56,9 @@ inline Type GetMemoryType(Type operand_type, Opcode opc) {
     // FIXME: change into a new column in opcode.def instead?
     auto is_unsigned = name.substr(name.size() - 2) == "_u";
     switch (opc.GetMemorySize()) {
-      case 1: return is_unsigned ? Type::I8U : Type::I8;
-      case 2: return is_unsigned ? Type::I16U : Type::I16;
-      case 4: return is_unsigned ? Type::I32U : Type::I32;
+      case 1: return Type(is_unsigned ? Type::I8U : Type::I8);
+      case 2: return Type(is_unsigned ? Type::I16U : Type::I16);
+      case 4: return Type(is_unsigned ? Type::I32U : Type::I32);
     }
   }
   return operand_type;
@@ -70,7 +70,7 @@ inline Type GetMemoryType(Type operand_type, Opcode opc) {
 struct LoadStoreTracking {
   struct LSAccess {
     Address byte_size = 0;
-    Type type = Type::Any;
+    Type type = Type(Type::Any);
     Address align = 0;
     uint32_t idx = 0;
     bool is_uniform = true;
@@ -79,7 +79,7 @@ struct LoadStoreTracking {
   struct LSVar {
     std::map<uint64_t, LSAccess> accesses;
     bool struct_layout = true;
-    Type same_type = Type::Any;
+    Type same_type = Type(Type::Any);
     Address same_align = kInvalidAddress;
     Opcode last_opc;
   };
@@ -155,7 +155,7 @@ struct LoadStoreTracking {
       var.same_align = align;
       var.last_opc = opc;
     } else {
-      var.same_type = Type::Void;
+      var.same_type = Type(Type::Void);
       var.same_align = kInvalidAddress;
     }
   }
--- a/include/wabt/type.h
+++ b/include/wabt/type.h
@@ -59,13 +59,38 @@ class Type {
   };
 
   Type() = default;  // Provided so Type can be member of a union.
-  Type(int32_t code)
+  explicit Type(int32_t code)
       : enum_(static_cast<Enum>(code)), type_index_(kInvalidIndex) {}
-  Type(Enum e) : enum_(e), type_index_(kInvalidIndex) {}
-  Type(Enum e, Index type_index) : enum_(e), type_index_(type_index) {
+  explicit Type(Enum e) : enum_(e), type_index_(kInvalidIndex) {}
+  explicit Type(Enum e, Index type_index) : enum_(e), type_index_(type_index) {
     assert(e == Enum::Reference);
   }
-  constexpr operator Enum() const { return enum_; }
+  constexpr Enum code() const { return enum_; }
+
+  // TODO: These comparisons could be removed in = default in C++20.
+  bool operator==(const Type& other) const {
+    return enum_ == other.enum_ && type_index_ == other.type_index_;
+  }
+
+  bool operator!=(const Type& other) const {
+    return !(*this == other);
+  }
+
+  bool operator<(const Type& other) const {
+    return (enum_ != other.enum_) ? enum_ < other.enum_
+                                  : type_index_ < other.type_index_;
+  }
+
+  bool operator==(const Type::Enum& other) const {
+    // Disallow non-intended comparisons, where the
+    // other argument should have a valid type_index_.
+    assert(!Type(other).IsReferenceWithIndex());
+    return enum_ == other;
+  }
+
+  bool operator!=(const Type::Enum& other) const {
+    return !(*this == other);
+  }
 
   bool IsRef() const {
     return enum_ == Type::ExternRef || enum_ == Type::FuncRef ||
--- a/src/type-checker.cc
+++ b/src/type-checker.cc
@@ -33,7 +33,8 @@ std::string TypesToString(const TypeVector& types,
     Type ty = types[i];
     // NOTE: Reference (and GetName) is also used by (e.g.) objdump, which does
     // not apply validation. do this here so as to not break that.
-    if (ty == Type::Reference && ty.GetReferenceIndex() == kInvalidIndex) {
+    if (ty.code() == Type::Reference &&
+        ty.GetReferenceIndex() == kInvalidIndex) {
       result += "reference";
     } else {
       result += types[i].GetName();
@@ -182,7 +183,7 @@ Result TypeChecker::PeekType(Index depth, Type* out_type) {
   CHECK_RESULT(TopLabel(&label));
 
   if (label->type_stack_limit + depth >= type_stack_.size()) {
-    *out_type = Type::Any;
+    *out_type = Type(Type::Any);
     return label->unreachable ? Result::Ok : Result::Error;
   }
   *out_type = type_stack_[type_stack_.size() - depth - 1];
@@ -190,7 +191,7 @@ Result TypeChecker::PeekType(Index depth, Type* out_type) {
 }
 
 Result TypeChecker::PeekAndCheckType(Index depth, Type expected) {
-  Type actual = Type::Any;
+  Type actual(Type::Any);
   Result result = PeekType(depth, &actual);
   return result | CheckType(actual, expected);
 }
@@ -233,11 +234,6 @@ Result TypeChecker::CheckType(Type actual, Type expected) {
     return Result::Ok;
   }
 
-  if (expected == Type::Reference && actual == Type::Reference) {
-    return expected.GetReferenceIndex() == actual.GetReferenceIndex()
-               ? Result::Ok
-               : Result::Error;
-  }
   if (actual != expected) {
     return Result::Error;
   }
@@ -466,7 +462,7 @@ Result TypeChecker::OnBr(Index depth) {
 }
 
 Result TypeChecker::OnBrIf(Index depth) {
-  Result result = PopAndCheck1Type(Type::I32, "br_if");
+  Result result = PopAndCheck1Type(Type(Type::I32), "br_if");
   Label* label;
   CHECK_RESULT(GetLabel(depth, &label));
   result |= PopAndCheckSignature(label->br_types(), "br_if");
@@ -476,7 +472,7 @@ Result TypeChecker::OnBrIf(Index depth) {
 
 Result TypeChecker::BeginBrTable() {
   br_table_sig_ = nullptr;
-  return PopAndCheck1Type(Type::I32, "br_table");
+  return PopAndCheck1Type(Type(Type::I32), "br_table");
 }
 
 Result TypeChecker::OnBrTableTarget(Index depth) {
@@ -514,8 +510,8 @@ Result TypeChecker::OnCall(const TypeVector& param_types,
 Result TypeChecker::OnCallIndirect(const TypeVector& param_types,
                                    const TypeVector& result_types,
                                    const Limits& table_limits) {
-  Result result = PopAndCheck1Type(table_limits.is_64 ? Type::I64 : Type::I32,
-                                   "call_indirect");
+  Result result = PopAndCheck1Type(
+      Type(table_limits.is_64 ? Type::I64 : Type::I32), "call_indirect");
   result |= PopAndCheckCall(param_types, result_types, "call_indirect");
   return result;
 }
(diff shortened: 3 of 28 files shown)


SoniEx2:
any reason not to use `explicit operator Enum`?

SoniEx2:
ah, fair enough!

src/wat-writer.cc
```diff
@@ -470,7 +470,7 @@ void WatWriter::WriteEndBlock() {
 }
 
 void WatWriter::WriteConst(const Const& const_) {
-  switch (const_.type()) {
```
sbc100, line 473: Can we land this comment as a separate PR?

sbc100:
This PR looks like it contains two commits rather than just the one described in the title

zherczeg:
This patch is the extension of #2571. Please review that first, then I will update this patch, which removes the first commit.

sbc100:
> This patch is the extension of #2571. Please review that first, then I will update this patch, which removes the first commit.

Can we flip it so that this one lands first.  Avoiding the need for isEqual?

zherczeg:
I have removed the other commit.

include/wabt/type.h
```diff
@@ -65,7 +65,28 @@ class Type {
   Type(Enum e, Index type_index) : enum_(e), type_index_(type_index) {
     assert(e == Enum::Reference);
   }
-  constexpr operator Enum() const { return enum_; }
+  constexpr Enum code() const { return enum_; }
+
+  bool operator==(const Type& other) const {
```
sbc100, line 70: Can you add a comment about how this could be removed in `= default` in C++20?

include/wabt/type.h
```diff
@@ -65,7 +65,28 @@ class Type {
...
+    return enum_ == other.enum_ && type_index_ == other.type_index_;
+  }
+
+  bool operator!=(const Type& other) const {
+    return enum_ != other.enum_ || type_index_ != other.type_index_;
```
sbc100, line 75: I think the canonical way to write this is `!(this == other)`

include/wabt/type.h
```diff
@@ -65,7 +65,28 @@ class Type {
...
+                                  : type_index_ < other.type_index_;
+  }
+
+  bool operator==(const Enum& other) const {
+    return enum_ == other;
+  }
+
+  bool operator!=(const Enum& other) const {
+    return enum_ != other;
+  }
```
sbc100, line 89: Do we actually need to these operators for comparing to Enum?

zherczeg: It looks like this is not needed.

zherczeg:
Thank you for the review, patch was updated.

include/wabt/type.h
sbc100: I still see these comparisons with Enum.

zherczeg: I only removed the !=. I thought the compiler can generate a better code when comparing to an enum. Anyway, it looks like the source code compiles without them.

zherczeg:
Interesting. Without the enum specific comparison, one test has an extra fail:
```
  --- expected
  +++ actual
  @@ -7,6 +7,9 @@
   out/test/typecheck/bad-bulk-memory-invalid-segment.txt:13:52: error: elem_segment variable out of range: 0 (max 0)
       i32.const 0 i32.const 0 i32.const 0 table.init 0
                                                      ^
  +out/test/typecheck/bad-bulk-memory-invalid-segment.txt:13:41: error: type mismatch at table.init. got any, expected funcref
  +    i32.const 0 i32.const 0 i32.const 0 table.init 0
  +                                        ^^^^^^^^^^
   out/test/typecheck/bad-bulk-memory-invalid-segment.txt:14:15: error: elem_segment variable out of range: 0 (max 0)
       elem.drop 0
                 ^
```
Shall I update the test or restore the enum comparison.

sbc100:
> Interesting. Without the enum specific comparison, one test has an extra fail:
> 
> ```
>   --- expected
>   +++ actual
>   @@ -7,6 +7,9 @@
>    out/test/typecheck/bad-bulk-memory-invalid-segment.txt:13:52: error: elem_segment variable out of range: 0 (max 0)
>        i32.const 0 i32.const 0 i32.const 0 table.init 0
>                                                       ^
>   +out/test/typecheck/bad-bulk-memory-invalid-segment.txt:13:41: error: type mismatch at table.init. got any, expected funcref
>   +    i32.const 0 i32.const 0 i32.const 0 table.init 0
>   +                                        ^^^^^^^^^^
>    out/test/typecheck/bad-bulk-memory-invalid-segment.txt:14:15: error: elem_segment variable out of range: 0 (max 0)
>        elem.drop 0
>                  ^
> ```
> 
> Shall I update the test or restore the enum comparison.

Strange.  Might be worth investigating.  I think I'd be ok with keeping these in, but maybe worth figuring what is going on there.

Does that new error look correct?  What type is the table in question?  What type is the elem segment at 0?

zherczeg:
The new error is not valid.
https://github.com/WebAssembly/wabt/blob/main/src/type-checker.cc#L232

```
1: expected = {enum_ = wabt::Type::FuncRef, type_index_ = 4294967295}
2: actual = {enum_ = wabt::Type::Any, type_index_ = 0}
```

For whatever reason, the Any is constructed with type_index_ = 0, and the `if (expected == Type::Any || actual == Type::Any) {` fails. In the past, it was harmless, since only the type was compared.

zherczeg:
Memory write breakpoints revealed that it was set here:
https://github.com/WebAssembly/wabt/blob/main/src/shared-validator.cc#L342
`Succeeded(result)` is false, so  the `T{}` construct created it.

I don't know the purpose of that code, and templates make it more complex.

zherczeg:
It seems to me that `T{}` simple clears the memory area, and the 0 is happen to be Type::Any. SharedValidator::ElemType default constructor is `ElemType() = default`. Maybe creating a real constructor could help. What shall we do?

sbc100:
> if (expected == Type::Any || actual == Type::Any) {

Wouldn't this line need to changed `if (expected.code() == Type::Any || actual.code() == Type::Any) {` as part of this PR?   If you remove the Enum comparisons I assume it would force you to change this line?

zherczeg:
That is also an option, but I suspect there are many other similar cases in the code, with similar side effects.

sbc100:
> That is also an option, but I suspect there are many other similar cases in the code, with similar side effects.

But shouldn't comparison `if (expected == Type::Any || actual == Type::Any)` simply fail to compile once you remove the comparison operators with Enum values?

zherczeg:
There is no compile fails because `Type` has an implicit constructor with `Enum`.
https://github.com/WebAssembly/wabt/blob/main/include/wabt/type.h#L64

So `expected == Type::Any` is converted to `expected == Type(Type::Any)`. The problem is that `type_index_` is set to `kInvalidIndex` by the constructor, and then compared to `expected`. I don't know in which places this `kInvalidIndex` comparison matters. For types without index, the index should be ignored during comparison. This executes faster and no side effects.

This `if` for example will always fails in the new code:
https://github.com/WebAssembly/wabt/blob/main/src/type-checker.cc#L236

But the `if (actual != expected) {` handles this part, so this problem is hidden.
I don't know how many such places in the code, it is a minefield.

zherczeg:
I can do that in another patch. Probably a lot of changes.

zherczeg:
Does this patch needs any more changes?

src/type-checker.cc
```diff
@@ -33,7 +33,8 @@ std::string TypesToString(const TypeVector& types,
     Type ty = types[i];
     // NOTE: Reference (and GetName) is also used by (e.g.) objdump, which does
     // not apply validation. do this here so as to not break that.
-    if (ty == Type::Reference && ty.GetReferenceIndex() == kInvalidIndex) {
+    if (ty.code() == Type::Reference &&
```
zherczeg, line 36: I would prefer to keep these this this way, because `ty == Type::Reference` is misleading equality for me, since the index is not checked.

zherczeg:
@SoniEx2 I have merged the two patches. I have added the comparison operators (to Enum::Type) as well. Please check it.

include/wabt/decompiler-ls.h
```diff
@@ -70,7 +70,7 @@ inline Type GetMemoryType(Type operand_type, Opcode opc) {
 struct LoadStoreTracking {
   struct LSAccess {
     Address byte_size = 0;
-    Type type = Type::Any;
+    Type type = Type(Type::Any);
```
zherczeg, line 73: It looks like field initializers cannot be written as `Type type(Type::Any);`

kripken:
> I'd like to see who other engines, for example, binaryen do it.

Here is Binaryen's code:

https://github.com/WebAssembly/binaryen/blob/f77a69d7f64dffdcdda65c478ee2a93c2821349c/src/wasm-type.h#L297

The basic idea is that the fixed types (i32, f64, etc.) are fixed enum values, while the user-declared ones (structs, arrays, etc.) are pointers (to objects that describe the type). Through bit fiddling we store them as a single pointer-sized value (same idea as i31 in wasm, in other words).

zherczeg:
2x32 bit can be efficiently passed as a 64 bit value, which fits into a machine word in 64 bit systems. Most devices are 64 bit systems now.

In Walrus (basically a wabt parser + interpreter + jit, https://github.com/Samsung/walrus/) we also do many bitmagic, but parsing /validating is rarely a bottleneck, and not worth it for a project like wabt.

sbc100:
> 2x32 bit can be efficiently passed as a 64 bit value, which fits into a machine word in 64 bit systems. Most devices are 64 bit systems now.
> 
> In Walrus (basically a wabt parser + interpreter + jit, https://github.com/Samsung/walrus/) we also do many bitmagic, but parsing /validating is rarely a bottleneck, and not worth it for a project like wabt.

I'm don't disagree here, however I think there maybe be something to learn from the `Type` / `BasicType` / `HeapType` distinction in binaryen.   They are able to have an implicit convertiion from basic types like `i32` without needing to write `Type(Type::i32)` .. they can just write `Type::i32`.   Doing it that way could perhaps avoid a lot of the changes in this PR?

zherczeg:
For simple types, setting the key part is enough. The constructor can be changed to not set the invalid part. However, this might cause issues with comparison.

@sbc100 you mentioned in #1860 that micro optimizations are not needed for wabt, and I agree with you. It is a parser, not an execution engine, so the trade-off of a small performance increase compared to the maintenance burden does not worth it.

An option is to use 64 bit signed types. A small negative number can be set by a single instruction on 64 bit systems. For reference types, the upper 32 bit contains the unsigned index. There will be no "undefined part" for the types. This form has a small overhead on the 32 bit (mostly embedded) systems.

I think the `Type(Enum)` form is needed unless the implicit constructor is restored (we go back where we started).

zherczeg:
The in64 variant proposal can be seen #2576 . Only the second patch contains the changes.

sbc100:
> For simple types, setting the key part is enough. The constructor can be changed to not set the invalid part. However, this might cause issues with comparison.
> 
> @sbc100 you mentioned in #1860 that micro optimizations are not needed for wabt, and I agree with you. It is a parser, not an execution engine, so the trade-off of a small performance increase compared to the maintenance burden does not worth it.
> 
> An option is to use 64 bit signed types. A small negative number can be set by a single instruction on 64 bit systems. For reference types, the upper 32 bit contains the unsigned index. There will be no "undefined part" for the types. This form has a small overhead on the 32 bit (mostly embedded) systems.

I'm not so much worried about optimizations, as I am about keep the code clear/simple/concise and compliant with the spec.

> 
> I think the `Type(Enum)` form is needed unless the implicit constructor is restored (we go back where we started).

IIUC the way binaryen does is to have separate enum of BasicType from which and Type can be implicitly constructed, but this doesn't work for reference types which are not part of BasicType.

zherczeg:
I think the current Type model (two 32 bit values) is quite good. Its only weakness is that the index part is not defined for all types (this was the problem with Type::Any comparison). A single 32 bit value is not enough to store a type and a 32 bit index, and next data type is 64 bit. 32 bit systems will handle the 64 bit value as a pair of 32 bit values, so it is basically the same as now. 64 bit systems (most systems) will process the whole value in one step.

zherczeg:
Btw with the int64_t implementation we could keep the implicit Enum<->Type conversions.

SoniEx2:
hmm...

what about... keeping the implicit enum->type conversion, but asserting it isn't a reference? how much would break from that?

zherczeg:
Actually, it was not the references that caused issues, but the non-references, because their "index" part was undefined, and some `T{}` template constructors zeroed them instead of setting them to `kInvalidIndex`. C++ magic. Then comparisons failed. 

Using `int64_t` we can fix this, because all bits are defined, and C++ cannot break it.

sbc100:
> Actually, it was not the references that caused issues, but the non-references, because their "index" part was undefined, and some `T{}` template constructors zeroed them instead of setting them to `kInvalidIndex`. C++ magic. Then comparisons failed.
> 
> Using `int64_t` we can fix this, because all bits are defined, and C++ cannot break it.

Could we use `= kInvalidIndex` in the declaration to avoid zeroing?

zherczeg:
This method does not need it.
Do you like this method or keep the original one?
If the latter, shall I add it to the field declaration?

zherczeg:
I am sorry I don't follow. I am ok with any method, just please decide.

Anyway I think this patch is a quite nice solution. It keeps the Type size (8 byte), has no initializer issues, and keeps the advantages of the current implicit initializers / casts. It is faster on 64 bit, same speed on 32 bit.

zherczeg:
Summary:
Original problem: Implicit type to enum cast, so if you compare two references, they are always equal regardless of the index. Example: when a signature with references is searched for a function (param/result vector comparison) it may choose the wrong signature, because any two references are equal.
https://github.com/WebAssembly/wabt/blob/main/src/ir.cc#L110

We can track down all places, and update the code (I did it in my other patches), but it is hard to maintain (easy to forget).

Adding an operator== does not work, because of ambiguity. Hence we need explicit Type::Enum casting. That is happening in #2572. Easier to maintain, but the source code grows, more multi-line expressions, harder to read. Also there wan an issue with Type::Any initialization in some templates.

This patch is a best of both words, keeps the advantages of the original code, but has no issues with comparisons, initializing.

SoniEx2:
hmm, let me try something... will report back if it works...

SoniEx2:
can you check if this works? (note: copy, not ref)

```diff
diff --git a/include/wabt/type.h b/include/wabt/type.h
index d0e85d97..c8a17983 100644
--- a/include/wabt/type.h
+++ b/include/wabt/type.h
@@ -67,6 +67,20 @@ class Type {
   }
   constexpr operator Enum() const { return enum_; }
 
+  friend constexpr bool operator==(const Type a, const Type b) {
+    return a.enum_ == b.enum_ &&
+           (!a.IsReferenceWithIndex() || a.type_index_ == b.type_index_);
+  }
+  friend constexpr bool operator!=(const Type a, const Type b) {
+    return !(a == b);
+  }
+  friend constexpr bool operator==(const Type ty, const Enum code) {
+    return ty.enum_ == code;
+  }
+  friend constexpr bool operator!=(const Type ty, const Enum code) {
+    return !(ty == code);
+  }
+
   bool IsRef() const {
     return enum_ == Type::ExternRef || enum_ == Type::FuncRef ||
            enum_ == Type::Reference || enum_ == Type::ExnRef;
```

SoniEx2:
looking back on the previous attempts, it seems the only thing we're doing different is explicitly ignoring the `type_index_` for types without an index.

zherczeg:
It looks like the reference indices are fixed for this test:
```
(module
  (type $t1 (func))
  (type $t2 (func))
  (type $t3 (func))

  (func $f1 (param (ref $t3)) )
  (func $f2 (param (ref $t2)) )
  (func $f3 (param (ref $t1)) )
  (func $f4 (param (ref $t3)) )
  (func $f5 (param (ref $t2)) )
  (func $f6 (param (ref $t1)) )
)
```
I don't understand why `friend` works when the normal member `operator==` does not compile.

The `IsReferenceWithIndex()` will be lot more complex later, so that is a runtime disadvantage.

SoniEx2:
it shouldn't be too bad, don't worry about it.

zherczeg:
What should be the next steps? Do you submit a patch with your code? Is this patch rejected then?
Please don't forget to add an `operator<` since a map somewhere uses Type, and it will use the implicit enum conversion.

SoniEx2:
well, do you wanna update the patch to use this `operator==` and also make a suitable `operator<` and update the title/message?

zherczeg:
Code is in #2580

PR #2575, opened 2025-03-31
Remove implicit Enum to Type conversion

zherczeg:
This patch adds another commit to #2572. It removes implicit conversions of Type from Type::Enum.


zherczeg:
Around 300 lines are changed in the new patch

SoniEx2:
hmm, a lot of these are comparisons. this is why we think this change should be done together with #2572 , as well as adding the relevant comparison operators between Type::Enum and Type.

also, `Type type(Type::Any);` instead of `Type type = Type(Type::Any);`.

zherczeg:
Shall I merge the patch then? It is kind of a one way thing. I will close this PR then.

PR #2576, opened 2025-04-01
Type as int64

zherczeg:
This patch is on the top of #2572 and changes the Type internal representation to an i64. This way the reference types and their 32 bit indices can fit into a single value. In the future, the nullable references (e.g. `(ref func)` / `funcref`) can be handled by the upper 32 bit.


zherczeg:
A cmake error appeared on s390x.
```
 CMake Error at out/_deps/libuv-src/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.
```
Patch is not involved.

zherczeg:
@sbc100 @SoniEx2 the patch is reworked to be simple. Please check it.
There are more cmake errors, independent from the patch.

zherczeg:
It looks like uvwasi is very old (Sun May 17 17:49:24 2020), but moving to a newer version does not help as well.

zherczeg:
I keep the branch for a while, but close this PR.

PR #2580, opened 2025-04-03
Add Type comparison methods

zherczeg:
Code is based on the work of SoniEx2.

### DIFF
--- a/include/wabt/type.h
+++ b/include/wabt/type.h
@@ -60,19 +60,40 @@ class Type {
 
   Type() = default;  // Provided so Type can be member of a union.
   Type(int32_t code)
-      : enum_(static_cast<Enum>(code)), type_index_(kInvalidIndex) {}
-  Type(Enum e) : enum_(e), type_index_(kInvalidIndex) {}
+      : enum_(static_cast<Enum>(code)), type_index_(0) {
+    assert(!EnumIsReferenceWithIndex(enum_));
+  }
+  Type(Enum e) : enum_(e), type_index_(0) {
+    assert(!EnumIsReferenceWithIndex(enum_));
+  }
   Type(Enum e, Index type_index) : enum_(e), type_index_(type_index) {
-    assert(e == Enum::Reference);
+    assert(EnumIsReferenceWithIndex(e));
   }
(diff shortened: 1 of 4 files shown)


src/type-checker.cc
```diff
@@ -524,7 +519,7 @@ Result TypeChecker::OnIndexedFuncRef(Index* out_index) {
   Type type;
   Result result = PeekType(0, &type);
   if (!type.IsReferenceWithIndex()) {
-    type = Type::Reference;
+    type = Type(Type::Reference, kInvalidIndex);
```
zherczeg, line 522: This was inconsistent in the code, sometimes `Type(Type::Reference, kInvalidIndex)` is used, sometimes just a `Type::Reference`. Now we must use the `Type(Type::Reference, kInvalidIndex)` form.

include/wabt/type.h
```diff
@@ -60,19 +60,40 @@ class Type {
...
+  Type(Enum e) : enum_(e), type_index_(0) {
+    assert(!EnumIsReferenceWithIndex(enum_));
+  }
   Type(Enum e, Index type_index) : enum_(e), type_index_(type_index) {
-    assert(e == Enum::Reference);
+    assert(EnumIsReferenceWithIndex(e));
   }
   constexpr operator Enum() const { return enum_; }
 
+  friend constexpr bool operator==(const Type a, const Type b) {
```
zherczeg, line 74: Is Type or Type& is better?

SoniEx2: Type is probably better, it's a pair of 32-bit values after all. Any modern compiler should be able to optimize it.

src/type-checker.cc
SoniEx2: sounds good!

SoniEx2 (approved):
Good thought to make `Type(Type::Any)` match zero-initialization. We wonder where that's being relied on (an union not being explicitly set?), but otherwise, well done!

zherczeg:
The `T{}` template constructor does it.
https://github.com/WebAssembly/wabt/blob/main/src/shared-validator.cc#L342

zherczeg:
Thank you!

PR #2581, opened 2025-04-08
Move parser related data from IR to the parser and improve reference type parsing

zherczeg:
This patch moves (and reworks) param_type_names / result_type_names out of IR to the parser, and fix the following issues:

Parse (ref index) form, not just (ref $name) form
Resolve references in types and locals
Display location when a named reference is not found

### DIFF
--- a/src/wast-parser.cc
+++ b/src/wast-parser.cc
@@ -319,39 +319,6 @@ bool ResolveFuncTypeWithEmptySignature(const Module& module,
   return false;
 }
 
-void ResolveTypeName(
-    const Module& module,
-    Type& type,
-    Index index,
-    const std::unordered_map<uint32_t, std::string>& bindings) {
-  if (type != Type::Reference || type.GetReferenceIndex() != kInvalidIndex) {
-    return;
-  }
-
-  const auto name_iterator = bindings.find(index);
-  assert(name_iterator != bindings.cend());
-  const auto type_index = module.type_bindings.FindIndex(name_iterator->second);
-  assert(type_index != kInvalidIndex);
-  type = Type(Type::Reference, type_index);
-}
-
-void ResolveTypeNames(const Module& module, FuncDeclaration* decl) {
-  assert(decl);
-  auto& signature = decl->sig;
-
-  for (uint32_t param_index = 0; param_index < signature.GetNumParams();
-       ++param_index) {
-    ResolveTypeName(module, signature.param_types[param_index], param_index,
-                    signature.param_type_names);
-  }
-
-  for (uint32_t result_index = 0; result_index < signature.GetNumResults();
-       ++result_index) {
-    ResolveTypeName(module, signature.result_types[result_index], result_index,
-                    signature.result_type_names);
-  }
-}
-
 void ResolveImplicitlyDefinedFunctionType(const Location& loc,
                                           Module* module,
                                           const FuncDeclaration& decl) {
@@ -457,7 +424,6 @@ class ResolveFuncTypesExprVisitorDelegate : public ExprVisitor::DelegateNop {
       : module_(module), errors_(errors) {}
 
   void ResolveBlockDeclaration(const Location& loc, BlockDeclaration* decl) {
-    ResolveTypeNames(*module_, decl);
     ResolveFuncTypeWithEmptySignature(*module_, decl);
     if (!IsInlinableFuncSignature(decl->sig)) {
       ResolveImplicitlyDefinedFunctionType(loc, module_, *decl);
@@ -536,7 +502,6 @@ Result ResolveFuncTypes(Module* module, Errors* errors) {
     bool has_func_type_and_empty_signature = false;
 
     if (decl) {
-      ResolveTypeNames(*module, decl);
       has_func_type_and_empty_signature =
           ResolveFuncTypeWithEmptySignature(*module, decl);
       ResolveImplicitlyDefinedFunctionType(field.loc, module, *decl);
@@ -960,9 +925,8 @@ Result WastParser::ParseValueType(Var* out_type) {
   return Result::Ok;
 }
 
-Result WastParser::ParseValueTypeList(
-    TypeVector* out_type_list,
-    std::unordered_map<uint32_t, std::string>* type_names) {
+Result WastParser::ParseValueTypeList(TypeVector* out_type_list,
+                                      ReferenceVars* type_vars) {
   WABT_TRACE(ParseValueTypeList);
   while (true) {
     if (!PeekMatchRefType() && !PeekMatch(TokenType::ValueType)) {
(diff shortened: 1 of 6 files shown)


zherczeg:
This patch can be a superior patch for the issue, thank you @sbc100 . It removes many things added in #2571 so it might better to merge the two patches, and drop the other PR.

The patch keeps all parser related data structures in the parser, and use pointers to the original data. It works as long as the data structures are allocated in memory, and not moved (copied) during parsing.

src/wast-parser.cc
```diff
@@ -3100,10 +3165,9 @@ Result WastParser::ParseBlockDeclaration(BlockDeclaration* decl) {
   WABT_TRACE(ParseBlockDeclaration);
   FuncDeclaration func_decl;
   CHECK_RESULT(ParseTypeUseOpt(&func_decl));
-  CHECK_RESULT(ParseUnboundFuncSignature(&func_decl.sig));
+  CHECK_RESULT(ParseUnboundFuncSignature(&decl->sig));
```
zherczeg, line 3168: Using locals for storing type vector is not possible anymore. I hope this is the only part where it caused an issue. I cannot check it with asserts.

This is the weakness of the patch.

src/wast-parser.cc
```diff
@@ -1335,6 +1349,21 @@ Result WastParser::ParseModuleFieldList(Module* module) {
       CHECK_RESULT(Synchronize(IsModuleField));
     }
   }
+
+  Result result = Result::Ok;
+
+  for (auto it : context_.resolve_types) {
```
zherczeg, line 1355: This is the core part. We resolve all names after the module is parsed in a single loop.

zherczeg:
I have simplified the code, this method does not even need a separate class, everything is inside WabtParser as private constructs. I only needed to move one static function to static private.

I think the two patch could be merged, and #2571 can be closed.

zherczeg:
I have found even more simplifications.

zherczeg:
What do you think about this patch?

src/wast-parser.cc
```diff
@@ -1324,8 +1293,53 @@ bool WastParser::PeekIsCustom() {
...
+
+  return Result::Ok;
+}
+
 Result WastParser::ParseModuleFieldList(Module* module) {
   WABT_TRACE(ParseModuleFieldList);
+
+  // Parser may parse multiple modules. Reset internal status.
+  resolve_types_.clear();
+  resolve_funcs_.clear();
```
sbc100, line 1341: There seems to be whole bunch of other state stored in `WastParser` (e.g. errors_, last_module_index_, etc)  .. are you sure re-use is supported?  Or is the other state reset somewhere else?

sbc100 (commented):
I'm a little confused about the title now.  Does this patch just refactor the internal state or does it also change / fix some behaviour too?

src/wast-parser.cc
zherczeg: I don't know. A spec test may contain several modules,

This is an example call stack:
```
#0  wabt::WastParser::ParseModuleFieldList
#1  wabt::WastParser::ParseScriptModule
#2  wabt::WastParser::ParseModuleCommand
#3  wabt::WastParser::ParseCommand
#4  wabt::WastParser::ParseCommandList
#5  wabt::WastParser::ParseScript
#6  wabt::ParseWastScript
```

The `ParseCommandList` has a loop. But nothing looks like a reset. I suspect the new module construction is the "reset". The point of this patch is moving out parser related stuff from the module, so I suspect this is the reason of the missing reset. Any suggestion?

sbc100: I see, so the rest of the state is not per-module state but per-wast-file state.

In that case I agree this is probably needed.   Maybe  update the commend to say something like "Reset module-specific state"?

zherczeg:
I have updated the patch description.

src/wast-parser.cc
zherczeg: Patch updated. Thank you!

sbc100:
Can you update the PR description?  It seems out of date (unless I'm misunderstanding).

zherczeg:
I have found another issue, which should be solved in some way, although not in this patch.

Currently this is happily accepted `(func $f (param (ref 20)))` even if there is no type 20. The parser only checks the invalid names. The parser could check the index as well, but invalid references can be read by the binary reader, so it seems to me that this is a shared-validator business. However, the shared validator has no location info, which is ok for binary data, but not really for text.

The easiest solution is doing checks both in the parser and the shared validator, but double checking is not that nice. Do you have any suggestions what to do with this issue? I can do the coding.

src/wast-parser.cc
```diff
@@ -1335,6 +1349,19 @@ Result WastParser::ParseModuleFieldList(Module* module) {
       CHECK_RESULT(Synchronize(IsModuleField));
     }
   }
+
+  Result result = Result::Ok;
+
+  for (auto it : resolve_types_) {
+    result |= ResolveRefTypes(module, *it.target_types, it.vars, errors_);
```
sbc100, line 1356: Why not `CHECK_RESULT` here?   I suppose so that we report more than just the first error?   Is that necessary though?  Do we do it elsewhere?

src/wast-parser.cc
```diff
@@ -1335,6 +1349,19 @@ Result WastParser::ParseModuleFieldList(Module* module) {
       CHECK_RESULT(Synchronize(IsModuleField));
     }
   }
+
+  Result result = Result::Ok;
```
sbc100, line 1353: Maybe add a comment to the effect of "Module parsing is now completely.  Type names can now be resolved"

src/wast-parser.cc
```diff
@@ -3100,10 +3159,9 @@ Result WastParser::ParseBlockDeclaration(BlockDeclaration* decl) {
   WABT_TRACE(ParseBlockDeclaration);
   FuncDeclaration func_decl;
   CHECK_RESULT(ParseTypeUseOpt(&func_decl));
```
sbc100, line 3161: Can we simply remove `func_decl` and just use `decl` for the `ParseTypeUseOpt` call too?

src/wast-parser.cc
```diff
@@ -1324,8 +1293,53 @@ bool WastParser::PeekIsCustom() {
          tokens_.front().text() == "custom";
 }
 
+Result WastParser::ResolveRefTypes(const Module* module,
+                                   TypeVector& types,
+                                   ReferenceVars& ref_vars,
```
sbc100, line 1298: I think the style we use it either a const ref (for const arguments) or pointer to mutable argument.    So these two args should either be const ref of pointer I think.

src/wast-parser.cc
zherczeg: I usually get multiple errors. But if you prefer a single error, I can change the code.

src/wast-parser.cc
zherczeg: It looks like the two structures are the same.

sbc100 (approved):

zherczeg:
Patch is rebased to master

Issue #2561, opened 2025-03-11
Parse invalid binary modules


sjamesr:
Is it possible to get `wabt::WastParser::ParseScript` to return the contents of `(module binary ...)` modules even if those modules are invalid?

My use case is that I'm helping to write a WASM JIT compiler and I want to exercise it with invalid inputs.

E.g. if the module looks like `(module binary "\00")`, I want to be able to ask the `wabt::Script` to return me exactly that one null byte.

Sorry if this is obvious and I'm just not seeing it.

sjamesr:
This is similar to #2374, but I'm using the wabt API directly and not the provided executables.

sjamesr:
I think something like https://github.com/WebAssembly/wabt/pull/2583 might work, what do you think?

PR #2583, opened 2025-04-12
[wast-parser] Add parse_binary_modules option

sjamesr:
Enabled by default, this option implements the current behavior of WABT when it encounters "(module binary ...)" script modules. That is, it parses the binary module, reporting any errors that it sees.

When disabled, the binary module will be passed back to the WAST parser as unvalidated raw bytes. This allows for the use of WABT in places which want to provide invalid modules to module loaders (e.g. spec test runners).

Fixes https://github.com/WebAssembly/wabt/issues/2561.

### DIFF
--- a/src/test-wast-parser.cc
+++ b/src/test-wast-parser.cc
@@ -18,6 +18,7 @@
 
 #include <memory>
 
+#include "wabt/cast.h"
 #include "wabt/wast-lexer.h"
 #include "wabt/wast-parser.h"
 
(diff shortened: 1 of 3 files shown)


include/wabt/feature.def
```diff
@@ -19,25 +19,26 @@
...
+WABT_FEATURE(reference_types,         "reference-types",         true,    "Reference types (externref)")
+WABT_FEATURE(annotations,             "annotations",             false,   "Custom annotation syntax")
+WABT_FEATURE(code_metadata,           "code-metadata",           false,   "Code metadata")
+WABT_FEATURE(gc,                      "gc",                      false,   "Garbage collection")
+WABT_FEATURE(memory64,                "memory64",                false,   "64-bit memory")
+WABT_FEATURE(multi_memory,            "multi-memory",            false,   "Multi-memory")
+WABT_FEATURE(extended_const,          "extended-const",          false,   "Extended constant expressions")
+WABT_FEATURE(relaxed_simd,            "relaxed-simd",            false,   "Relaxed SIMD")
+WABT_FEATURE(custom_page_sizes,       "custom-page-sizes",       false,   "Custom page sizes")
+WABT_FEATURE(validate_binary_modules, "validate-binary-modules", true,    "Validate binary modules")
```
sjamesr, line 44: Now that I think about it, perhaps this is best placed in a field of WastParserOptions instead? It would not make sense for a user to specify this flag on the command line, I expect it would break assumptions in the various tools (e.g. if the WastParser returns OK, the Module in a BinaryScriptModule should be populated).

src/wast-parser.cc
```diff
@@ -3527,8 +3527,10 @@ Result WastParser::ParseModuleCommand(Script* script, CommandPtr* out_command) {
       options.features = options_->features;
       Errors errors;
       const char* filename = "<text>";
-      ReadBinaryIr(filename, bsm->data.data(), bsm->data.size(), options,
-                   &errors, module);
+      if (options_->validate_binary_modules) {
+        ReadBinaryIr(filename, bsm->data.data(), bsm->data.size(), options,
+                     &errors, module);
```
zherczeg, line 3532: Based on the code change, this is more like ignoring the module. Not validating usually means the module is processed, but the errors are ignored.
(Note: I am not a reviewer)

sjamesr: The script_module and its binary data are still made available (see the unit test for example usage), it's just the structured wabt::Module that is not provided.

sbc100: In that case perhaps a better name for this option might be "parse_binary_modules" or "no_parse_binary_modules"?  

BTW, how do invalid binaries modules in the test suite currently work without this option?

sjamesr: Renamed to "parse_binary_modules".

Currently invalid modules are parsed and fail validation in WABT, which causes the test to pass. This change does not alter that behavior.

The use case is when you're using WABT to parse WAST files, but you don't want WABT to reject invalid binary modules. Maybe you want to send that invalid module data to a WASM engine that you're building. With this change, you can set `parse_binary_modules` to false and extract that invalid data out without WABT rejecting it.

sbc100 (commented):
LGTM with title updated.

sjamesr:
Thank you. Love WABT btw, it's really nice to work with!

src/test-wast-parser.cc
```diff
@@ -86,3 +87,33 @@ TEST(WastParser, LongTokenSpace) {
   ASSERT_STREQ(R"(unexpected token notmodule, expected EOF.)",
                errors[1].message.c_str());
 }
+
+TEST(WastParser, InvalidBinaryModule) {
+  std::string text = R"((module binary "\00asm\bc\0a\00\00"))";
+  std::vector<uint8_t> expected_data = {
+      0, 'a', 's', 'm', static_cast<uint8_t>('\xbc'), '\x0a', 0, 0};
```
sbc100, line 94: This static_cast seems a little odd.  Is there a better way to write this?   Is this needed because C chars are signed by default?    Does 0xbc work here instead?

sbc100 (approved):

PR #2584, opened 2025-04-16
Validate type references by index

zherczeg:
This patch is another bugfix.

### DIFF
--- a/src/shared-validator.cc
+++ b/src/shared-validator.cc
@@ -76,6 +76,21 @@ Result SharedValidator::OnArrayType(const Location&, TypeMut field) {
   return Result::Ok;
 }
 
+Result SharedValidator::EndTypeSection() {
+  Result result = Result::Ok;
+
+  for (auto func_type : func_types_) {
+    for (auto type : func_type.second.params) {
+      result |= CheckReferenceType(Location(), type, "params");
+    }
+
+    for (auto type : func_type.second.results) {
+      result |= CheckReferenceType(Location(), type, "results");
+    }
+  }
+  return result;
+}
+
 Result SharedValidator::OnFunction(const Location& loc, Var sig_var) {
   Result result = Result::Ok;
   FuncType type;
@@ -194,6 +209,22 @@ Result SharedValidator::CheckType(const Location& loc,
                actual.GetName().c_str(), expected.GetName().c_str());
     return Result::Error;
   }
+  return Result::Ok;
+}
+
+Result SharedValidator::CheckReferenceType(const Location& loc,
+                                           Type type,
+                                           const char* desc) {
+  if (type.IsReferenceWithIndex()) {
+    Index index = type.GetReferenceIndex();
+    auto iter = func_types_.find(index);
+
+    if (iter == func_types_.end()) {
+      return PrintError(loc, "reference %d is out of range in %s",
+                        static_cast<int>(index), desc);
+    }
+  }
+
   return Result::Ok;
 }
 
@@ -479,6 +510,9 @@ Result SharedValidator::OnLocalDecl(const Location& loc,
     PrintError(loc, "local count must be < 0x10000000");
     return Result::Error;
   }
+
+  CHECK_RESULT(CheckReferenceType(loc, type, "locals"));
+
   locals_.push_back(LocalDecl{type, GetLocalCount() + count});
   return Result::Ok;
 }
--- a/src/wast-parser.cc
+++ b/src/wast-parser.cc
@@ -542,6 +542,11 @@ void AppendInlineExportFields(Module* module,
   module->AppendFields(fields);
 }
 
+bool IsIndexLikelyType(Index index) {
+  // TODO: Incorrect values can be misinterpreted by the parser.
+  return index >= static_cast<Index>(Type::Void);
+}
+
 }  // End of anonymous namespace
 
 WastParser::WastParser(WastLexer* lexer,
@@ -937,10 +942,10 @@ Result WastParser::ParseValueTypeList(TypeVector* out_type_list,
     CHECK_RESULT(ParseValueType(&type));
 
     if (type.is_index()) {
-      // TODO: Incorrect values can be misinterpreted by the parser.
-      if (type.index() >= static_cast<Index>(Type::Void)) {
+      if (IsIndexLikelyType(type.index())) {
         out_type_list->push_back(Type(type.index()));
       } else {
+        type_vars->push_back(ReferenceVar(out_type_list->size(), type));
         out_type_list->push_back(Type(Type::Reference, type.index()));
       }
     } else {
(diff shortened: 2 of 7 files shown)


src/shared-validator.cc
```diff
@@ -76,6 +76,21 @@ Result SharedValidator::OnArrayType(const Location&, TypeMut field) {
   return Result::Ok;
 }
 
+Result SharedValidator::EndTypeSection() {
+  Result result = Result::Ok;
```
zherczeg, line 80: This is a new event in SharedValidator. This check could be done in `OnFuncType`, but then we needed the `OnTypeCount` event, which is also not used by the validator.

Since the Google v8 engine has separate reference types for array/struct, we might need to check the reference type in the future as well, and we cannot do that before reading all types.

src/shared-validator.cc
```diff
@@ -76,6 +76,21 @@ Result SharedValidator::OnArrayType(const Location&, TypeMut field) {
...
+
+  for (auto func_type : func_types_) {
+    for (auto type : func_type.second.params) {
+      result |= CheckReferenceType(Location(), type, "params");
```
zherczeg, line 84: We don't have the locations at this point.

src/wast-parser.cc
```diff
@@ -2057,7 +2070,13 @@ Result WastParser::ParseBoundValueTypeList(TokenType token,
       bindings->emplace(name,
                         Binding(loc, binding_index_offset + types->size()));
       if (type.is_index()) {
-        types->push_back(Type(type.index()));
+        // TODO: Incorrect values can be misinterpreted by the parser.
```
zherczeg, line 2073: This is a bugfix. There is a similar code above in the patch.

zherczeg:
The parser also validates the references to provide nice error locations, which is a double check when the validator is used. However the check in the parser has an advantage: implicit function type indicies are not accepted, which I think is a user friendly thing.

test/gen-wasm.py
```diff
@@ -43,6 +43,7 @@
     'f64': 0x7c,    # -4
     'v128': 0x7b,    # -5
     'funcref': 0x70,    # -0x10
+    'reference': 0x6b,    # -0x15
```
sbc100, line 46: Should we just use "ref" here like the spec does?

test/parse/bad-references.txt
```diff
@@ -0,0 +1,29 @@
...
+  ;; Implicit type declarations are not allowed
+  (type $t0 (func (param (ref 2))))
+  (type $t1 (func (param (ref 33)) (result (ref 444))))
+
+  (func $f (param $p1 (ref 5555))
+    (local $p2 (ref 66666))
+  )
+)
+(;; STDERR ;;;
+out/test/parse/bad-references.txt:6:31: error: reference type out of range: 2 (max: 2)
```
sbc100, line 14: The error message seems to suggest that 2 should be within range?

src/wast-parser.cc
```diff
@@ -2057,7 +2071,13 @@ Result WastParser::ParseBoundValueTypeList(TokenType token,
...
+        if (type.index() >= static_cast<Index>(Type::Void)) {
```
sbc100, line 2075: What does it mean to compare to `Type::Void` here?

src/wast-parser.cc
```diff
@@ -1313,11 +1314,24 @@ Result WastParser::ResolveRefTypes(const Module& module,
 
     assert(type.IsReferenceWithIndex());
 
+    if (ref_var.var.is_index()) {
+      if (type.GetReferenceIndex() >= module.types.size()) {
```
sbc100, line 1318: Is `module.types` complete at this point? i.e. might it not be possible that there are more types to be be parsed?

src/shared-validator.cc
```diff
@@ -479,8 +510,11 @@ Result SharedValidator::OnLocalDecl(const Location& loc,
     PrintError(loc, "local count must be < 0x10000000");
     return Result::Error;
   }
+
+  Result result = CheckReferenceType(loc, type, "locals");
```
sbc100, line 514: Use `CHECK_RESULT` here and revert line 483?

test/gen-wasm.py
zherczeg: The current wabt implementation uses this name and constant.
The spec has different names (ref/refnull) and constants for function references: 0x64 / 0x63

https://github.com/WebAssembly/function-references/blob/main/proposals/function-references/Overview.md#reference-types-2

The 0x6b also has a new name (`kStructRefCode`) according to v8 engine:

This whole naming needs to be changed, and #2562 does this, but this patch (should) precedes that.

src/wast-parser.cc
zherczeg: This is a short term workaround until #2562 lands. The index is used as reference index (0..`void`) or type (`void`..0xffffffff). The current code does not allow passing a reference at all, so making a test is impossible. Also converting any index to type is dangerous.

This code is used twice at the moment, and will be removed.

sbc100: Could we add a helper here to convey what we are trying to do ? Something like `type.isXXX()`?

src/shared-validator.cc
zherczeg: Good question. What happens if the variable is not added to `locals_`?

src/wast-parser.cc
zherczeg: Yes, all names are resolved after the module is fully parsed. The implicit function types has not been added yet.

src/wast-parser.cc
zherczeg: I can. However, this code should be dropped as soon as possible, since it is incorrect. Now a `(ref 0xffffffff)` is converted to `i32`.

test/parse/bad-references.txt
zherczeg: I just copied an existing error template:
```
(module
  (func $f (type 0))
)
```

I get:
```
test.wast:8:18: error: function type variable out of range: 0 (max 0)
  (func $f (type 0))
```

What should we use instead? `(< 0)` or `(less than 0)`

src/shared-validator.cc
sbc100: I think if we return error then all best are off, no?  Should be fine I think

test/parse/bad-references.txt
sbc100: Fair enough.  LGTM as is then.

test/gen-wasm.py
sbc100: Fair enough lgtm as is then.

zherczeg:
Btw, I agree that the current `max:` error message is not the best, but we should improve it everywhere for consistency. Perhaps the `value - 1` for `max:` could be better with a special string when the value is zero. Please decide it, and I will update the code.

src/shared-validator.cc
zherczeg: Ok, I changed it to CHECK_RESULT.

zherczeg:
Thank you for the review, I have updated the patch.

sbc100:
> Btw, I agree that the current `max:` error message is not the best, but we should improve it everywhere for consistency. Perhaps the `value - 1` for `max:` could be better with a special string when the value is zero. Please decide it, and I will update the code.

Leaving as is, for consistency, seems fine to me.

src/wast-parser.cc
```diff
@@ -937,10 +942,10 @@ Result WastParser::ParseValueTypeList(TypeVector* out_type_list,
     CHECK_RESULT(ParseValueType(&type));
 
     if (type.is_index()) {
-      // TODO: Incorrect values can be misinterpreted by the parser.
-      if (type.index() >= static_cast<Index>(Type::Void)) {
+      if (IsIndexLikelyType(type.index())) {
```
sbc100, line 945: From the code that follows it looks like we are trying to distinguish reference types from function types?  Is that right?  How about something like `IsReferenceType`?   I'm not really sure what `IsIndexLikelyType` means here... perhaps I'm misunderstanding?

sbc100 (approved):

src/wast-parser.cc
zherczeg: It is not just reference type, it is currently any type. The `ParseValueType` encodes the value type as index.

For example: the binary code of `(func $f (param i64 (ref 0xfffffffe)))` is

```
; func type 0
000000b: 60                                        ; func
000000c: 02                                        ; num params
000000d: 7e                                        ; i64
000000e: 7e                                        ; i64
000000f: 00                                        ; num results
```

The i64 is encoded as `0xfffffffe`, which is stored in the index. Also, the `(ref 0xfffffffe)` is converted to i64, because currently it is impossible to tell the difference (at the moment). The `function_ref` patch will fix this issue.

sbc100: SGTM then.

zherczeg:
Rebased this patch (no change). Can you land it? Thank you!

zherczeg:
@sbc100 could you land this approved patch? Thank you!

PR #2590, opened 2025-04-24
wasm2c: Remove implicit void* conversions in tail calls

shravanrn:
Tail call wrappers rely on implicit conversion of `void*` to the target module's instance types. Modifying this to explicitly cast to the target type in order to play nicely with the `-Wc++-compat` flag in gcc.


include/wabt/ir.h
```diff
@@ -916,6 +917,7 @@ struct Func {
   Index GetLocalIndex(const Var&) const;
 
   std::string name;
+  std::string ext_module_name;
```
sbc100, line 920: This feels like it should be part of the `ImportMixin` rather then the function itself.   Is that not possible?

sbc100 (commented):
lgtm % comment

include/wabt/ir.h
shravanrn: I think the problem is that the `FuncImport` information (and thus any info brought by the `ImportMixin` that is inherited) seems to be lost by the time wasm2c is outputting code in `c-writer.cc`. So, i've made this change to capture whether functions are from external modules directly on the `Func` object. Happy to explore any alternatives you have in mind though. Let me know your thoughts/if the above approach is fine.

shravanrn: Unfortunately, it does seem like `c-writer` does not receive this information at all today. From what I can tell, the `binary-reader` and `binary-reader-ir` parse wasm files and create instances of objects such as `FuncImport`. It then creates unique names for all function calls (of the form <module>_<func>) and only these unique names are passed into c-writer as objects of type `Func`. Since`c-writer' only receives `Func`, it does not actually receive information about the external module. This was the smallest change I could think of to get things work. Let me know if you'd prefer if I do things differently.

sbc100: Since c-writer has access the whole `Module` IR object I would hope that it can use the `imports` member to figure out the name.   If we are going to do it we should explain what why we are duplicating information that already exists elsewhere in the IR.

sbc100: The c-writer already is using the module_name of imported functions.

See 1779:

```
    if (import->kind() == ExternalKind::Func) {                                       
      import_func_module_set_.insert(import->module_name);                            
    }     
```

If this data is not available in the right time and place perhaps we could build an additional data structure here in the C-writer to cache the module names of functions?

src/c-writer.cc
```diff
@@ -2351,7 +2351,11 @@ void CWriter::WriteFuncRefWrapper(const Func* func) {
   Write(") ", OpenBrace());
   Write("return ");
   Write(ExternalRef(ModuleFieldType::Func, func->name));
-  Write("(instance");
+  std::string target_module_name =
+      func->ext_module_name == ""
+          ? ModuleInstanceTypeName()
+          : ModuleInstanceTypeName(func->ext_module_name);
+  Write("((struct ", target_module_name, "*) instance");
```
sbc100, line 2358: Can we not somehow do dynamic check here and downcast to a `FuncImport`?   I guess we don't have the RTTI info in the IR classes to do a downcast?   

It looks like we have `IsImport` which can check if the function is imported.. and then do a cast mabye?

include/wabt/ir.h
sbc100: I'm not sure what you mean by "only these unique names are passed into c-writer".  The c-writer operates on the complete IR Module object, doesn't it?

shravanrn: > Can we not somehow do dynamic check here and downcast to a FuncImport? I guess we don't have the RTTI info in the IR classes to do a downcast? It looks like we have IsImport which can check if the function is imported.. and then do a cast mabye?

Oh gotcha. You're saying that the func pointer we are using here is a pointer to the member of a FuncImport due to the mixin system used? So i have to figure out the correct way to do `FuncImport* import = (char*)func - offsetof(FuncImport, func);`. I am trying to figure out if this is safe to assume in which case I can try this. **Any suggestions on this welcome.**

> I'm not sure what you mean by "only these unique names are passed into c-writer". The c-writer operates on the complete IR Module object, doesn't it?

Oh I was just saying that all we have are datastructres like `import_func_module_set_` or `funcs_` which have only the function names, not the name of the IR Module. If the above casting approach doesn't work, I can try your suggestion of `an additional data structure here in the C-writer to cache the module names of function`.

shravanrn: Oh ignore my prior comments. You were right that the information is around and all I needed to do was preserve this in a separate data structure. I've made the change. Let me know what you think

src/c-writer.cc
shravanrn: Yup, I am using `IsImport` to figure this out now.

include/wabt/ir.h
shravanrn: @sbc100 Let me know if this looks ok. If yes, I'll go ahead and land it

keithw:
lgtm but would be good to have a look from @SoniEx2 since I think that's who contributed the original fix here.

keithw:
BTW should we add this warning flag to the testing to catch this kind of thing going forward?

shravanrn:
> BTW should we add this warning flag to the testing to catch this kind of thing going forward?

I'd say we can hold off on enabling the flag for now. At the moment, the use case is to run some custom tools that identify better optimization opportunities and potential UB/vulnerabilities (the tools need C code to be compile-able with that flag). Since I am still only in the testing phase, this spot fix should be fine for now.

@SoniEx2 Please have a look when you have a chance.

shravanrn:
Gonna go ahead and merge in the interest of time. @SoniEx2 Feel free to open an issue if you have concerns

sbc100 (approved):

PR #2591, opened 2025-04-25
Add missing in_function_body check to objdump

zherczeg:
Fixes the fuzzer bug in #2567

I noticed all objdump logs are guarded by `if (!in_function_body)`. The code crashes without this check.


sbc100 (approved):
Thanks!

Issue #2348, opened 2023-12-05
Plan for GC and beyond


keithw:
The function-references and gc proposals are going to be merged into the spec soon (as well as the "v4" revision of the exception-handling proposal, which we don't support yet). It could be good to talk about our plan for supporting these features in WABT. I think there's a few options to put on the table:

- We work on the current best-effort basis to update our exception-handling support and finish the implementation of function-references and gc (in the binary/text readers/writers, the interpreter, and wasm2c). I'm not sure who has a ton of spare cycles for this, but if there is continued interest in WABT, I and other contributors will probably eventually make it happen. Could take a long time though, and I'm a little nervous about losing community trust/interest if we allow WABT to fall behind the living spec for too long.
- We try to (re-)increase investment of engineering resources in WABT by some of the companies who use it. Based on conversations at the CG meeting, this seems like a bit of a long shot, but maybe I misread the situation or there is something these companies would want from us that we could provide. (E.g. how to make a persuasive case for Google and Mozilla to devote more engineering resources or to find a way to pay Igalia or others to start contributing more again.)
- We deprecate most of the WABT tools except for wasm2c, which we'd port to run in binaryen. An upside is that binaryen attracts a lot of investment/velocity and seems to be Google's focus right now in terms of non-browser Wasm tools, and wasm2c would get to ride its coattails; downside is that binaryen isn't really intended (as I understand it) as a spec-conforming implementation of a Wasm binary/text reader/writer + validator/interpreter, given its IR that is slightly different from Wasm. (And, smaller point, binaryen currently seems to be much slower than WABT at some operations like a wasm-merge.) So the community would lose that part of WABT's benefits.
- We replace the internals of WABT with the Bytecode Alliance's Rust wasm-tools project, and port wasm2c to run in wasm-tools. Upside is that this is also a heavily depended-on building block from the BA folks and a newer Rust codebase that also passes the spec tests, and it might be worthwhile to combine our efforts (aiui there are minor differences, e.g. I think wasm-tools can't currently write the folded text format which seems minor and probably addressable); downside is that this would be a pretty avulsive change (moving to Rust among other things) and would really need to have buy-in, might lose a lot of the current contributors, and would take significant investment to pull off.

I don't think anything is forcing us to make a decision here, but tentatively I'd rather make one than sort of avoid it.

sbc100:
[...]

PR #2593, opened 2025-04-29
Support table initializers

zherczeg:
Another large patch on the top of the other two. Now there are 3-4000 lines of new code is waiting for review, and it looks like the end is still far.

The painful part so far:
https://webassembly.github.io/function-references/core/binary/modules.html#element-section

Since type 0-3 is reserved for `(ref func)`, all `(elem ... funcref ...)` must be encoded with 4-7, which increase binary size. This affect several tests, and probably real word programs :( 

There are spec tests to check this behavior.


sbc100:
> This amount of work this feature needs was unexpected for me. Unfortunately, this is unavoidable. I made many patches, and these cover most specification tests, although some of them are still just partly supported. However, I don't get any reviews for them. I would like to understand why.
> 
> * Is this a feature which does not align with the aims of wabt?
> * Are these changes too big / too risky?
> * Any other reasons?
> 
> It is very hard to make progress when I don't get any feedback, and I need to maintain several patches. Is there a chance these patches will land?

Sorry for the lack of feedback on some of your patches.  Unfortunately there are very few folks to who have any time to work on wabt these days (or review patches).

Please expect fairly long feedback cycles.  If you are waiting for a review from me, feel free to ping me a every couple of days, but please understand that I have very few cycles to spend on wabt myself.

For some background on where some folks see wabt heading see this issue: https://github.com/WebAssembly/wabt/issues/2348.  If you would like to help maintain wabt going forward, and potentially work on major additions such as GC please join that discussion.

zherczeg:
Thank you for the suggestion!

zherczeg:
The patch set (#2562, #2567, and this patch) covers 25 tests from the 26 spec tests in testsuite/proposals/function-references. That is 96% pass rate. The missing one is ref.wast, which needs removing the invalid reference test during parsing.

sbc100:
> The patch set (#2562, #2567, and this patch) covers 25 tests from the 26 spec tests in testsuite/proposals/function-references. That is 96% pass rate. The missing one is ref.wast, which needs removing the invalid reference test during parsing.

Nice!   Thanks for all the hard work on this.

I assume #2562 is the first in the series and the one we should look at first?

BTW can you have this PR a better title?   `Support table initializers` sounds odd because that is obviously something that wabt already supports.  Perhaps you mean a certain type of new table intializer?

zherczeg:
> I assume #2562 is the first in the series and the one we should look at first?

Yes. It focuses on extending the existing features, e.g. tables / globals now supports refs. Recursive refs, and not initalized non nullable refs are detected as well.
The next patch focuses on adding new byte codes.
The last focuses on a new initializer value for tables. It breaks the current API.

> BTW can you have this PR a better title? `Support table initializers` sounds odd because that is obviously something that wabt already supports. Perhaps you mean a certain type of new table intializer?

I am open to any ideas.

https://github.com/WebAssembly/function-references/blob/main/proposals/function-references/Overview.md#tables
Tables can have an optional initializer value, and all elements of the table is set to that value when the table is constructed (before active elems are processed). If `ref.null` is not possible for a table (e.g. it has `(ref func)` type), the initializer is mandatory.

Btw I found a case, which has no spec test, and the specification is unclear:

`(table (ref func) (elem 1))` -> since all elements of the table is initialized, it seems the initializer is unnecessary.

However, if I do the inlining here, I should get a missing initializer error:
https://webassembly.github.io/function-references/core/text/modules.html#tables
And I have no idea how the binary should be constructed. Maybe the first function could be the initializer.

keithw:
FWIW, when I hit a case where the prose spec is unclear, I usually just go try it in the reference interpreter and in wasm-tools. If those two disagree, it's an even better opportunity to submit a new test case. :-) Andreas is usually very quick about accepting new test cases into the spec repo.

zherczeg:
The last test is done, now all spec test are pass.

include/wabt/binary-reader-logging.h
```diff
@@ -96,9 +96,13 @@ class BinaryReaderLogging : public BinaryReaderDelegate {
...
-  Result OnTable(Index index,
-                 Type elem_type,
-                 const Limits* elem_limits) override;
+  Result BeginTable(Index index,
+                    Type elem_type,
+                    const Limits* elem_limits,
+                    bool has_init_expr) override;
+  Result BeginTableInitExpr(Index index) override;
+  Result EndTableInitExpr(Index index) override;
+  Result EndTable(Index index) override;
```
zherczeg, line 105: Since tables can have an initializer expressions now, this is a breaking change. I don't know how these are handled in this project.

README.md
```diff
@@ -63,6 +63,7 @@ Wabt has been compiled to JavaScript via emscripten. Some of the functionality i
 | [extended-const][]    | `--enable-extended-const`   |   | ✓ | ✓ | ✓ | ✓ | ✓ |
 | [relaxed-simd][]      | `--enable-relaxed-simd`     |   | ✓ | ✓ | ✓ | ✓ |   |
 | [custom-page-sizes][] | `--enable-custom-page-sizes`|   | ✓ | ✓ | ✓ | ✓ | ✓ |
+| [function-references][] | `--enable-function-references` |   | ✓ | ✓ | ✓ | ✓ |   |
```
zherczeg, line 66: I haven't used `update-spec-tests.py` and updated README before, so these are just speculative changes.

test/dump/typed_func_refs_params.txt
```diff
@@ -25,7 +25,7 @@
     )
   )
 
-  (elem declare funcref (ref.func $square))
+  (elem declare func $square)
```
zherczeg, line 28: There are similar changes below. The reason is a backward incompatible change. As for elem types 0-3, they have changed the result type from `(ref null func)` (same as `funcref`) to `(ref func)`. This change affects table type matches.

zherczeg (commented):
@sbc100 this is the last patch of the function references proposal.

include/wabt/ir.h
```diff
@@ -987,13 +987,14 @@ struct Table {
...
   Type elem_type;
+  ExprList init_expr;
 };
 
 using ExprListVector = std::vector<ExprList>;
 
 struct ElemSegment {
   explicit ElemSegment(std::string_view name) : name(name) {}
-  uint8_t GetFlags(const Module*) const;
+  uint8_t GetFlags(const Module*, bool) const;
```
sbc100: Can you give this new argument a name?

include/wabt/shared-validator.h
```diff
@@ -75,7 +75,7 @@ class SharedValidator {
   Result OnArrayType(const Location&, TypeMut field);
 
   Result OnFunction(const Location&, Var sig_var);
-  Result OnTable(const Location&, Type elem_type, const Limits&);
+  Result OnTable(const Location&, Type elem_type, const Limits&, bool, bool);
```
sbc100: Can you name these new arguments?   (seems especially important for bool)

src/binary-reader-objdump.cc
```diff
@@ -1127,9 +1127,10 @@ class BinaryReaderObjdump : public BinaryReaderObjdumpBase {
   Result OnFunction(Index index, Index sig_index) override;
 
   Result OnTableCount(Index count) override;
-  Result OnTable(Index index,
-                 Type elem_type,
-                 const Limits* elem_limits) override;
+  Result BeginTable(Index index,
+                    Type elem_type,
+                    const Limits* elem_limits,
+                    bool) override;
```
sbc100: Name here too please

src/binary-reader.cc
```diff
@@ -2787,8 +2787,32 @@ Result BinaryReader::ReadTableSection(Offset section_size) {
     Index table_index = num_table_imports_ + i;
     Type elem_type;
     Limits elem_limits;
+    bool has_init_expr = false;
+
+    if (options_.features.function_references_enabled() &&
+        state_.offset < read_end_ && state_.data[state_.offset] == 0x40) {
```
sbc100: What is this 0x40 value?   Maybe at least deserves a comment, if not a name.

It looks like you are peeking at the next byte here.. is there perhaps a helper function for that?

src/binary-reader.cc
```diff
@@ -2787,8 +2787,32 @@ Result BinaryReader::ReadTableSection(Offset section_size) {
...
+      state_.offset++;
+      has_init_expr = true;
+
+      uint8_t value;
+      CHECK_RESULT(ReadU8(&value, "table init"));
+      if (value != 0) {
+        PrintError("unsupported table intializer: 0x%x\n",
+                   static_cast<int>(value));
```
sbc100, line 2806: This looks like it should fit on one line?

src/interp/interp-wasm-c-api.cc
```diff
@@ -960,7 +960,8 @@ void wasm_global_set(wasm_global_t* global, const wasm_val_t* val) {
 own wasm_table_t* wasm_table_new(wasm_store_t* store,
                                  const wasm_tabletype_t* type,
                                  wasm_ref_t* init) {
-  return new wasm_table_t{Table::New(store->I, *type->As<TableType>())};
+  return new wasm_table_t{
+      Table::New(store->I, *type->As<TableType>(), Ref::Null)};
```
sbc100: It looks like `init` is ignored here.. is this something we should implement later?   Should we have a TODO here maybe?

src/ir.cc
```diff
@@ -703,7 +703,8 @@ void Var::Destroy() {
   }
 }
 
-uint8_t ElemSegment::GetFlags(const Module* module) const {
+uint8_t ElemSegment::GetFlags(const Module* module,
+                              bool function_references_enabled) const {
```
sbc100, line 707: It seems a little odd that that we don't have any other features in `ir.cc`, I guess we have gotten away to a feature-agnostic IR up until this point.

src/shared-validator.cc
```diff
@@ -134,19 +136,18 @@ Result SharedValidator::OnTable(const Location& loc,
   if (limits.is_shared) {
     result |= PrintError(loc, "tables may not be shared");
   }
-  if (elem_type != Type::FuncRef &&
-      !options_.features.reference_types_enabled()) {
+  if (options_.features.reference_types_enabled()) {
+    if (!elem_type.IsRef()) {
+      result |= PrintError(loc, "tables must have reference types");
```
sbc100: How about "tables may only contain reference types"?   i'm not sure its any better actually :)

src/validator.cc
```diff
@@ -224,6 +224,12 @@ static Result CheckType(Type actual, Type expected) {
     }
   }
 
+  if (actual_type == Type::FuncRef && expected_type == Type::RefNull) {
+    // Specification tests pass expected functions
+    // directly, their type is not relevant.
```
sbc100: I'm not sure I understand this comment here, does it need updating?

src/validator.cc
```diff
@@ -801,8 +807,8 @@ Result Validator::CheckModule() {
 
         case ExternalKind::Table: {
           auto&& table = cast<TableImport>(f->import.get())->table;
-          result_ |=
-              validator_.OnTable(field.loc, table.elem_type, table.elem_limits);
+          result_ |= validator_.OnTable(field.loc, table.elem_type,
+                                        table.elem_limits, true, false);
```
sbc100: Might be nice to have Enums to avoid all these `true` / `false` arguments, but not a huge deal, and I think we have a lot of them already probably.

src/wast-parser.cc
```diff
@@ -1344,7 +1344,7 @@ Result WastParser::ResolveTargetRefType(const Module& module,
   assert(type->IsReferenceWithIndex() && !var.is_index());
 
   if (type->GetReferenceIndex() != kInvalidIndex) {
-    // Resolved earlier.
+    // Index or resolved earlier.
```
sbc100: Comment doesn't seem to maker sense.  Is there a typo?

sbc100 (approved):
Great!  Thanks for all the work on this!

Can you update the PR description to match what you would like in the final commit?

sbc100:
What do you think about re-titling: "Support table initializer expressions; completes function references support"

include/wabt/ir.h
zherczeg: Yes. Since a bool (`options_.features.function_references_enabled()`) is passed, I would not introduce an enum for this.

src/ir.cc
zherczeg: Normally, new features are backward compatible, so you don't need flags for them (the parser will check the feature). This is probably the first exception.

src/binary-reader.cc
zherczeg: I have added the comment. I could not find a helper for peaking a byte, probably this is something new as well. I will add something then.

src/binary-reader.cc
zherczeg: It is a bit over 80 column. The style checker checks these things.

src/interp/interp-wasm-c-api.cc
zherczeg: Nice catch! The API always provided an init value, but it was ignored before. Now we don't need to ignore it. This is kind of a bugfix then. I copied the code from `wasm_table_set`.

src/validator.cc
zherczeg: I have updated the description. There is a test: `(assert_return (invoke "ref") (ref.null))`, where the value is null without any actual type. Unfortunately we only get the types, not the values, so this check is not perfect, but I have no idea how to do this differently.

src/binary-reader.cc
zherczeg: I chose a different direction at the end. It does not need peak, but some code duplication is needed. However, it is clearer, that the 0x40 is the leb32 representation of Type::Void.

zherczeg:
Thank you for the review. I changed the title, and I hope I fixed every request. Please check the patch again.

PR #2602, opened 2025-05-17
wasm2c: Mmap+guard on big-endian won't move memory (fix #2599)

### LINKED ISSUE #2599: wasm2c: big endian support moves memory on growth which will cause overheads

shravanrn:


shravanrn:
Sorry, it took a few iterations to get all the tests passing. Looks like things are working now.

shravanrn:
@sbc100 @keithw any chance one of you could review this pr?

wasm2c/wasm-rt-mem-impl-helper.inc
```diff
@@ -76,26 +76,65 @@ void MEMORY_API_NAME(wasm_rt_allocate_memory)(MEMORY_TYPE* memory,
...
 
-  if (WASM_RT_USE_MMAP &&
-      MEMORY_API_NAME(wasm_rt_memory_is_default32)(memory)) {
-#if WASM_RT_USE_MMAP  // mmap-related functions don't exist unless this is set
+#if WASM_RT_USE_MMAP
+  if (MEMORY_API_NAME(wasm_rt_memory_is_default32)(memory)) {
     const uint64_t mmap_size =
         get_alloc_size_for_mmap_default32(memory->max_pages);
-    void* addr = os_mmap(mmap_size);
+    unsigned char* addr = os_mmap(mmap_size);
```
sbc100: Why not use MEMORY_CELL_TYPE here and avoid the cast below?

sbc100: Or just `uint8_t*`?

wasm2c/wasm-rt.h
```diff
@@ -492,6 +496,8 @@ typedef struct {
    * volatile.
    */
   _Atomic volatile uint8_t* data;
+  /** The location after the the reserved space for the linear memory data. */
```
sbc100: Might be worth clarifying if is the address after the last byte of memory, or if it can also include padding/page alignment. From the usage it looks like former?

wasm2c/wasm-rt-mem-impl-helper.inc
```diff
@@ -76,26 +76,65 @@ void MEMORY_API_NAME(wasm_rt_allocate_memory)(MEMORY_TYPE* memory,
...
       abort();
     }
-    memory->data = addr;
+    memory->data = (MEMORY_CELL_TYPE)addr;
+    memory->data_end = (MEMORY_CELL_TYPE)data_end;
+    return;
+  }
+#endif
+
+  memory->data = calloc(byte_length, 1);
```
sbc100: Why is an explicit cast to `MEMORY_CELL_TYPE` need above but not here?

wasm2c/examples/fac/fac.c
```diff
@@ -771,7 +771,7 @@ wasm_rt_func_type_t wasm2c_fac_get_func_type(uint32_t param_count, uint32_t resu
   
   if (param_count == 1 && result_count == 1) {
     va_start(args, result_count);
-    if (true && va_arg(args, wasm_rt_type_t) == WASM_RT_I32 && va_arg(args, wasm_rt_type_t) == WASM_RT_I32) {
+    if (true && va_arg(args, int) == WASM_RT_I32 && va_arg(args, int) == WASM_RT_I32) {
```
sbc100, line 768: Are these change from a different PR?   If so, can they be landed separately to avoid confusion.

wasm2c/wasm-rt-mem-impl-helper.inc
shravanrn: Fixed!

wasm2c/wasm-rt.h
shravanrn: Fixed!

wasm2c/wasm-rt-mem-impl-helper.inc
shravanrn: Fixed!

wasm2c/examples/fac/fac.c
shravanrn: This PR is now rebased on #2619 which does this cleanup

wasm2c/wasm-rt-impl.c
```diff
@@ -238,11 +238,14 @@ static void os_disable_and_deallocate_altstack(void) {
...
-    : "=a" (*rax), "=b" (*rbx), "=c" (*rcx), "=d" (*rdx) // output operands
-    : "a" (*rax), "c" (*rcx) // input operands
+static void call_cpuid(uint64_t* rax,
+                       uint64_t* rbx,
+                       uint64_t* rcx,
+                       uint64_t* rdx) {
+  __asm__ volatile(
+      "cpuid"
+      : "=a"(*rax), "=b"(*rbx), "=c"(*rcx), "=d"(*rdx)  // output operands
+      : "a"(*rax), "c"(*rcx)                            // input operands
```
sbc100, line 248: This change (and the above one) look like maybe just reformatting changes?  Can they be split out or reverted?  (I'm sure you know, but you can use `git clang-format` to format just the touched parts of a file)

shravanrn: Yup, this is already split out into #2619. It just shows up here since this PR is rebased on that one.

PR #2603, opened 2025-05-17
wasm2c: allow test script to specify a runner app

shravanrn:
This is useful to specify debuggers, launch the tests in qemu emulators for different platforms etc.


test/run-spec-wasm2c.py
```diff
@@ -665,7 +665,9 @@ def main(args):
 
             # Run the resulting binary
             if options.run:
-                return subprocess.run([main_exe]).returncode
+                test_runner = os.getenv('WASM2C_TEST_RUNNER')
```
sbc100, line 668: I think can just write this in one line with  `command = os.getenv('WASM2C_TEST_RUNNER', '').split()`

shravanrn: @sbc100 Fixed!

sbc100 (approved):

Issue #2530, opened 2025-01-14
wat2wasm doesn't support wasm GC?


chennbnbnb:
code example:
```wasm
;; wat2wasm --enable-gc ./test.wat
(module
    (type $i32_arr (array (mut i32)) )

    (func $createArray (param $len i32) (result i32) 
        (array.new
            $i32_arr
            (i32.const 42)      ;; initial value 
            (local.get $len)   ;; array len
        )
        (array.get
            $i32_arr
            (i32.const 0)   ;; element idx
        )
    )

    (export "createArray" (func $createArray) )
)
```
error msg:
```
./test.wat:6:10: error: unexpected token "array.new", expected an instr.
        (array.new
```

wat2wasm doesn't support `array` instr?

sbc100:
Indeed wasm-gc is not supported by wabt.   See https://github.com/WebAssembly/wabt/pull/2363 and https://github.com/WebAssembly/wabt/issues/2348

nuskey8:
It's very inconvenient that wat2wasm lacks this feature. Are there any new developments or alternative tools regarding this?

sbc100:
I suggest binaryen, or wasm-tools.  Both support wasm GC and both have very good wat-to-wasm conversion.

PR #2604, opened 2025-05-17
Remove --enable-gc from help text

soegaard:
According to issue #2530 the GC proposal is not supported by wat2asm. Having --enable-gc in the help text gives a false impression.

### DIFF
--- a/test/help/wat2wasm.txt
+++ b/test/help/wat2wasm.txt
@@ -36,7 +36,6 @@ options:
       --disable-reference-types                Disable Reference types (externref)
       --enable-annotations                     Enable Custom annotation syntax
       --enable-code-metadata                   Enable Code metadata
-      --enable-gc                              Enable Garbage collection
       --enable-memory64                        Enable 64-bit memory
       --enable-multi-memory                    Enable Multi-memory
       --enable-extended-const                  Enable Extended constant expressions


sbc100 (commented):
I looks like you are just updating the test expectation.  

I think you would also need to update the wat2wasm source code itself.

soegaard:
Oh! I didn't look closely enough. I thought, that was the help text printed by the tool `wat2wasm`.
Looking at the code, the option `--enable-gc` is added here:

https://github.com/WebAssembly/wabt/blob/main/include/wabt/feature.def#L38

But - and it is a big but - the options are shared between the tools, 
so if other tools support GC then we can't remove it from the `feature.def`.

What would be the least intrusive way of handling this?
A simple idea: make `wat2wasm` print a note after the examples in:

https://github.com/WebAssembly/wabt/blob/main/src/tools/wat2wasm.cc#L51

The note could simply say that the `--enable-gc` option has no effect.

Not ideal, but at least low effort.

sbc100:
How about just renaming it to `--enable-gc-experimental`?

soegaard:
> How about just renaming it to --enable-gc-experimental?

I would interpret a flag `--enable-gc-experimental` as "GC is implemented, but it is still a WIP".
Or perhaps that the specification might change.
My understanding is that `--enable-gc` is a noop for `wat2wasm`.

From the perspective of a user new to Wasm that just wants to try out Wasm:
he sees `i31` in the GC spec. Checks the help of `wat2wasm` and sees he must
use `--enable-gc` to use the GC proposal. To his surprise none of the example programs works.

Adding a line in the README would help.
Removing the flag from the help text of `wat2wasm` even better.

I'll close this PR (since the fix was wrong) and create an issue instead.

PR #2607, opened 2025-05-23
Implement Garbage Collector type system

zherczeg:
This patch supports parsing the new GC types
  - Abstract types
  - Recursive types
  - Composite types

The patch also improves type comparison.


include/wabt/binary-reader.h
```diff
@@ -52,13 +52,20 @@ struct ReadBinaryOptions {
...
 
-// TODO: Move somewhere else?
+// TODO: Move both TypeMut and GCTypeInformation somewhere else?
 struct TypeMut {
   Type type;
   bool mutable_;
 };
 using TypeMutVector = std::vector<TypeMut>;
 
+// Garbage Collector specific type information
```
zherczeg, line 62: This is a core part of the patch. It contains the `(sub ...)` part of the type. It is declared as a structure, because it is not mandatory.

include/wabt/binary-reader.h
```diff
@@ -99,15 +106,20 @@ class BinaryReaderDelegate {
   /* Type section */
   virtual Result BeginTypeSection(Offset size) = 0;
   virtual Result OnTypeCount(Index count) = 0;
+  virtual Result OnRecursiveRange(Index start_index, Index type_count) = 0;
   virtual Result OnFuncType(Index index,
+                            GCTypeExtension* gc_ext,
```
zherczeg, line 111: This structure is passed as a second argument, because it is a header, but it could go to the last argument since it is an extra (and optional) information. Which one you prefer?

include/wabt/type-checker.h
```diff
@@ -20,17 +20,94 @@
...
+    ArrayType() = default;
+    ArrayType(TypeMut field) : field(field) {}
+
+    TypeMut field;
+  };
+
+  struct RecursiveRange {
+    Index start_index;
+    Index type_count;
+  };
```
zherczeg, line 80: This is another core structure to encode `(rec ...)` constructs. It represents the range.

zherczeg:
I have reworked the type validation system of the patch. Now it is capable of detecting the first type index for all equal types. This first type index is called canonical index. If I have two types (`t1`/`t2`), and their canonical index is computed,  then type comparison is `t1.canonical_index == t2.canonical_index`. Sub type indices can also be turned to canonical sub indices. This is not only useful for validation, but also very important for high speed execution, since it simplifies type comparison a lot. To compute these canonical indices, a hash code is computed for each type. When two types have different hash codes, they are never equal. My hash computation algorithm might not be good, I don't have much experience with these algorithms.

zherczeg:
The `type-*` gc tests are running except the runtime part of type-subtyping.wast
I think the typing system in the validator and interpreter are ok now. This is another huge change with 1500 lines of new code.

rossberg:
It sounds like you are canonicalising wrt type indices. But type indices are meaningless in any program that consists of more than one module. Type canonicalisation must happen globally, across module boundaries, based on the types' structure. I suspect that is the reason for the link/run-time tests failing.

zherczeg:
The link tests are not failing, although the interpreter do a slow type comparison for import/exports. As far as I understand the interpreter here is just a demonstration, so this is probably ok. The runtime tests fail because the operations (such as `ref.cast`) is not implemented. I will do that in a follow-up patch.

The global type canonicalisation sounds like a very good idea! A high performance engine should do that!

zherczeg:
@sbc100 there is a fuzzer issue in the code. The code is correct though.

https://github.com/WebAssembly/wabt/blob/main/src/interp/binary-reader-interp.cc#L772
As for the fuzzer generated test, it wants to allocate 16190847 entries, which is pretty large for a 38 byte input, but not an invalid value in general.

https://github.com/llvm/llvm-project/blob/main/compiler-rt/lib/fuzzer/FuzzerLoop.cpp
LLVM considers this as a large value, and reports it as an error. There is an `-rss_limit_mb` to modify this limit.

What shall I do?

sbc100:
> @sbc100 there is a fuzzer issue in the code. The code is correct though.
> 
> https://github.com/WebAssembly/wabt/blob/main/src/interp/binary-reader-interp.cc#L772 As for the fuzzer generated test, it wants to allocate 16190847 entries, which is pretty large for a 38 byte input, but not an invalid value in general.
> 
> https://github.com/llvm/llvm-project/blob/main/compiler-rt/lib/fuzzer/FuzzerLoop.cpp LLVM considers this as a large value, and reports it as an error. There is an `-rss_limit_mb` to modify this limit.
> 
> What shall I do?

We don't tend to have time to worry about fixing all the fuzz tests issues, unless they could conceivable show up in real world programs.  i.e. we tend to assume trusted and save inputs, since we don't have the resources the harden wabt against other things.

Having said that we obviously would be happy to accept fixes for such issues if folks come up with them.

zherczeg:
There is nothing to fix here, the code is correct (and not related to this patch). It is simply a limitation of the fuzzer, it assumes too much memory allocation is likely a bug.

zherczeg:
Is there a compile time define (or can be defined during build in some way) to detect that the fuzzer is active? Then I could guard out the failing `.reserve` call. Generating huge numbers is perfectly valid for a fuzzer.

zherczeg:
I have checked the type hashing system in the whole test system. It performs 45157 comparisons, and the hash/type_count is equal 513 times. From that, 467 times the types are really equal, and 46 times they are not. Overall the efficiency is 99.9%.

I could not improve the hash by using more complex hashes. However, `(hash * 33) + value` has the same efficiency.
I will check the fails, maybe it gives some ideas.

Edit: Now the efficiency is 100%. It does not mean it is perfect or anything. The missing cases were valid, worth improving the code.

zherczeg:
@sbc100 This is the first patch of the GC support. It triggers a fuzzer fail I described above. Reserving memory is correct for valid wasm files. However, a random value generated by a fuzzer causes a large memory allocation, which is considered as an error by the fuzzer. The fuzzer has an option to raise this allocation limit. What do you suggest?

zherczeg:
Note: the issue is visible on the fuzzer backtrace:
```
     #11 0x56421033ba2f in __allocate_at_least<std::__1::allocator<wabt::interp::DataDesc> > /usr/local/bin/../include/c++/v1/__memory/allocate_at_least.h:41:19
    #12 0x56421033ba2f in __split_buffer /usr/local/bin/../include/c++/v1/__split_buffer:330:25
    #13 0x56421033ba2f in std::__1::vector<wabt::interp::DataDesc, std::__1::allocator<wabt::interp::DataDesc>>::reserve(unsigned long) /usr/local/bin/../include/c++/v1/__vector/vector.h:1109:49
    #14 0x564210325038 in wabt::interp::(anonymous namespace)::BinaryReaderInterp::OnDataCount(unsigned int) /src/wabt/src/interp/binary-reader-interp.cc:925:17
    #15 0x5642103d8474 in wabt::(anonymous namespace)::BinaryReader::ReadDataCountSection(unsigned long) /src/wabt/src/binary-reader.cc:3113:3
```
The OnDataCount gets a huge number, runs the `.reserve()` and aborts.

src/interp/binary-reader-interp.cc
```diff
@@ -615,21 +717,24 @@ Result BinaryReaderInterp::OnImportTag(Index import_index,
 }
 
 Result BinaryReaderInterp::OnFunctionCount(Index count) {
-  module_.funcs.reserve(count);
+  module_.funcs.reserve(std::min(count, 1024u));
```
zherczeg: Fuzzer fix: avoid allocating too much memory. If count is really > 1024, the array will still grow, just it happens later.  However, some elements at the end of the buffer might not be used.  I don't know what should we do here. There are similar cases below.

src/interp/binary-reader-interp.cc
```diff
@@ -615,21 +717,24 @@ Result BinaryReaderInterp::OnImportTag(Index import_index,
...
   return Result::Ok;
 }
 
 Result BinaryReaderInterp::OnFunction(Index index, Index sig_index) {
-  CHECK_RESULT(
-      validator_.OnFunction(GetLocation(), Var(sig_index, GetLocation())));
-  FuncType& func_type = module_.func_types[sig_index];
+  Result result =
```
zherczeg, line 731: Fuzzer bug: we need to add something to the list, even if the validation fails. The `module_.funcs` will be used later.

zherczeg (commented):
The patch is finally green. I have fixed some fuzzer bugs, which are unrelated to the patch. These are interpreter related, which main purpose is testing.

include/wabt/binary-reader-nop.h
```diff
@@ -45,20 +45,29 @@ class BinaryReaderNop : public BinaryReaderDelegate {
...
                     Index result_count,
-                    Type* result_types) override {
+                    Type* result_types,
+                    GCTypeExtension* gc_ext) override {
     return Result::Ok;
   }
   Result OnStructType(Index index,
                       Index field_count,
-                      TypeMut* fields) override {
+                      TypeMut* fields,
```
sbc100, line 62: I wonder if we should be using C++ std::array rather then size + ptr in these API ?   But I see its a pre-existing thing so no worries for this PR.

include/wabt/binary-reader.h
```diff
@@ -52,13 +52,20 @@ struct ReadBinaryOptions {
...
-// TODO: Move somewhere else?
+// TODO: Move both TypeMut and GCTypeInformation somewhere else?
 struct TypeMut {
   Type type;
   bool mutable_;
 };
 using TypeMutVector = std::vector<TypeMut>;
 
+// Garbage Collector specific type information
+struct GCTypeExtension {
```
sbc100: Why "Extension" here?   Maybe `GCTypeInfo`?

sbc100: Oh I see its like info about the types that extend a given type maybe?

include/wabt/interp/interp-inl.h
```diff
@@ -42,7 +42,30 @@ inline bool FuncType::classof(const ExternType* type) {
...
 
 inline FuncType::FuncType(ValueTypes params, ValueTypes results)
-    : ExternType(ExternKind::Func), params(params), results(results), func_types(nullptr) {}
+    : ExternType(ExternKind::Func),
+      kind(FuncType::TypeKind::Func),
+      canonical_index(kInvalidIndex),
+      canonical_sub_index(kInvalidIndex),
+      is_final_sub_type(true),
+      recursive_start(0),
+      recursive_count(0),
```
sbc100: Is there some reason not just just put the defaults like `= 0` into the class definition itself?

include/wabt/interp/interp.h
```diff
@@ -178,14 +178,37 @@ struct FuncType : ExternType {
...
   static bool classof(const ExternType* type);
 
+  enum class TypeKind {
+    Func,
+    Struct,
+    Array,
+  };
+
+  // To simplify the implementation, FuncType may also represent
+  // Struct and Array types. To do this, the mutability is stored
```
sbc100: This seems odd.. is the idea to fix this later?

include/wabt/ir.h
```diff
@@ -1107,7 +1132,8 @@ enum class ModuleFieldType {
   Memory,
   DataSegment,
   Start,
-  Tag
+  Tag,
+  EmptyRec
```
sbc100: What is `EmptyRec`?

include/wabt/type-checker.h
```diff
@@ -44,6 +65,75 @@ class TypeChecker {
...
+
+  struct ArrayType {
+    ArrayType() = default;
+    ArrayType(TypeMut field) : field(field) {}
+
+    TypeMut field;
+  };
+
+  struct RecursiveRange {
```
sbc100: Maybe we should use `RecGroup` and `rec_group` rather then `Recursive` or `rec` ?   Or maybe I'm just not used to seeing "rec" / "recurisive" and thinking of recursion group?

sbc100 (commented):
Wow!  This is awesome.    

There is a lot going on here though.   I didn't get time to look through all of it get.

@tlively could you take a quick look over this and see if anything jumps out?   Does the general approach look reasonable?

include/wabt/binary-reader.h
zherczeg: That was the idea. Normally, you don't use them (except passing them to the validator). And they might be changed in the future.

I am not a native speaker, so naming is never my strength. Shall I change it?

include/wabt/interp/interp.h
zherczeg: It can be changed. My idea was that the interpreter is just a test program, so this choice should not affect library users, and the patch is already quite big. It can be improved later.

include/wabt/ir.h
zherczeg: I don't know the best choice here. Rec = recursive type. Normally, it contains several types, so it can be reconstructed without storing any extra data. This is not true for empty recursive types. Their position must be marked by something, and that is the purpose of that type.

I don't really understand the purpose of empty recursive types. They cannot be used for anything.

include/wabt/type-checker.h
zherczeg: Done

include/wabt/ir.h
rossberg: Drive-by: Empty recursion groups don't declare anything, they are like an empty section or an empty group of locals. I don't think they require any representation in the IR, you can just ignore them. Or, if you really want to represent them, the natural form would be as empty RecursionRanges.

include/wabt/binary-reader.h
rossberg: Why "GC", though? Subtyping isn't limited to GC types like structs and arrays, it's used for all types.

include/wabt/ir.h
zherczeg: The use case is a bit more complicated. Wabt supports load/store module operations, and when an empty group is present during loading, we need to serialize it during storing to produce the same module.

There is another problem. Wabt has a list of types, and I don't want to change it to a list of recursive types, because that would require a large code refactor (and I suspect the new code would be even slower). Instead the types know when a new recursive type starts, and its size. But empty recursive blocks has no types, so this trick don't work with them.

include/wabt/binary-reader.h
zherczeg: I am open to better names! Naming is not my best skill. I just wanted a name, which represents the extension where it comes, and "gc" is the name of the extension in wabt.

zherczeg: Probably the "Garbage Collector specific type information" comment is misleading. The "Extra type information added by the Garbage Collector proposal" would be better.

rossberg: Well, the originating proposal is just historical accident, it is not semantically relevant. Func types are unrelated to GC, as will be other proposed types like cont. (How about just `TypeExtension` or `Supertypes`?)

include/wabt/ir.h
rossberg: Does Wabt preserve empty sections? If not, there is no reason to expect it to preserve empty recursion groups.

As for the list of types, your RecursionRange is a start index and a count, so isn't an empty range representable just fine?

zherczeg: Well, spec test and wasm-tools preserves empty recs (they even have text form representation: `(rec)`), so I think wabt should do that as well. I don't know about empty sections. Can you represent empty sections in text form?

As for the second question, wabt has a unique feature: when a module is parsed and dumped in text form, it wants to preserve the declaration order in the module:
https://github.com/WebAssembly/wabt/blob/main/src/wat-writer.cc#L1776

Example:
```
(module
  (type (func (param i32)))
  (func (param i32))
  (type (func (param i32)))
)
```
wat-desugar tool output:
```
(module
  (type (;0;) (func (param i32)))
  (func (;0;) (param i32))
  (type (;1;) (func (param i32))))
```

I don't think this feature matters in practice too much, but it is working this way. If we want to keep this feature it is not enough to know that an empty rec group is present, its exact field location needs to be known as well.

include/wabt/interp/interp-inl.h
zherczeg: Done

include/wabt/binary-reader.h
zherczeg: I picked `SupertypesInfo`, which is the combination of the two suggestions. I think this name can avoid symbol clashes.

include/wabt/ir.h
rossberg: Right, sections do not have a direct text representation. But there are other forms in the text format that have. For example, I can declare an empty list of locals:`(local)`, similarly with a few other lists. I would be very surprised if Wabt preserved those artefacts, so why bother with recursion groups in particular?

zherczeg:
```
(module (func $f (local i32) (local) (local i32)))
```
output:
```
(module
  (func $f
    (local i32 i32))
  (type (;0;) (func)))
```
You are right, the empty local is dropped by wabt.

Using the spec test interpreter (the ocaml one)
```
(module
  (rec)
  (type (func))
  (rec)
  (rec)
  (type (func))
  (rec)

  (func (local i32) (local) (local i64))
)
```
The `wasm-tools dump` output:
```
  0x0 | 00 61 73 6d | version 1 (Module)
      | 01 00 00 00
  0x8 | 01 8f 80 80 | type section
      | 80 00
  0xe | 06          | 6 count
--- rec group 0 (explicit) ---
  0xf | 4e 00       |
--- rec group 1 (implicit) ---
 0x11 | 60 00 00    | [type 0] SubType { is_final: true, supertype_idx: None, composite_type: CompositeType { inner: Func(FuncType { params: [], results: [] }), shared: false } }
--- rec group 2 (explicit) ---
 0x14 | 4e 00       |
--- rec group 3 (explicit) ---
 0x16 | 4e 00       |
--- rec group 4 (implicit) ---
 0x18 | 60 00 00    | [type 1] SubType { is_final: true, supertype_idx: None, composite_type: CompositeType { inner: Func(FuncType { params: [], results: [] }), shared: false } }
--- rec group 5 (explicit) ---
 0x1b | 4e 00       |
 0x1d | 03 82 80 80 | func section
      | 80 00
 0x23 | 01          | 1 count
 0x24 | 00          | [func 0] type 0
 0x25 | 0a 8c 80 80 | code section
      | 80 00
 0x2b | 01          | 1 count
============== func 0 ====================
 0x2c | 86 80 80 80 | size of function
      | 00
 0x31 | 02          | 2 local blocks
 0x32 | 01 7f       | 1 locals of type I32
 0x34 | 01 7e       | 1 locals of type I64
 0x36 | 0b          | end
```
The `(rec)` groups are kept, the `(local)` is dropped.

Wow, it looks like it produces a lot of unnecessary 80 bytes! Is this a bug?

The `wasm-tools parse` is not affected:

```
  0x0 | 00 61 73 6d | version 1 (Module)
      | 01 00 00 00
  0x8 | 01 0f       | type section
  0xa | 06          | 6 count
--- rec group 0 (explicit) ---
  0xb | 4e 00       |
--- rec group 1 (implicit) ---
  0xd | 60 00 00    | [type 0] SubType { is_final: true, supertype_idx: None, composite_type: CompositeType { inner: Func(FuncType { params: [], results: [] }), shared: false } }
--- rec group 2 (explicit) ---
 0x10 | 4e 00       |
--- rec group 3 (explicit) ---
 0x12 | 4e 00       |
--- rec group 4 (implicit) ---
 0x14 | 60 00 00    | [type 1] SubType { is_final: true, supertype_idx: None, composite_type: CompositeType { inner: Func(FuncType { params: [], results: [] }), shared: false } }
--- rec group 5 (explicit) ---
 0x17 | 4e 00       |
 0x19 | 03 02       | func section
 0x1b | 01          | 1 count
 0x1c | 00          | [func 0] type 0
 0x1d | 0a 08       | code section
 0x1f | 01          | 1 count
============== func 0 ====================
 0x20 | 06          | size of function
 0x21 | 02          | 2 local blocks
 0x22 | 01 7f       | 1 locals of type I32
 0x24 | 01 7e       | 1 locals of type I64
 0x26 | 0b          | end
```
My wabt code generates exactly the same bytes as `wasm-tools parse` command.

I agree the tools are not consistent from this perspective. Do you suggest we should drop the `(rec)` blocks? This would simplify the code, so I would be happy. Hopefully all tools will follow this method as well.

zherczeg: I have rebased the spec test interpreter to the latest `wasm-3.0` branch, and the 80 bytes are still there. I suspect this is a bug in the spec test interpreter!

rossberg: This is working as intended. The interpreter generates the binary in a single pass, backpatching section/function sizes, so no attempt is made to optimise their length. They will all be 5-byte LEBs.

rossberg: > Do you suggest we should drop the (rec) blocks? This would simplify the code, so I would be happy. Hopefully all tools will follow this method as well.

I'd suggest they do whatever is most convenient with their respective implementation. It just doesn't matter either way.

zherczeg: @sbc100 please decide, the code is there, but I can remove it.

sbc100: The `EmptyRec` thing does seem out of place here.   If its only there to support round-tripping of empty `(rec)` groups, and this is not necessary for spec compliance (am I understanding that correctly) then I suggest we drop it.

zherczeg: The `EmptyRec` is removed from the patch. Please let me know if you need more changes.

zherczeg:
@sbc100 may I ask whether you have time to review this patch?

include/wabt/interp/interp-inl.h
```diff
@@ -42,7 +42,24 @@ inline bool FuncType::classof(const ExternType* type) {
 }
 
 inline FuncType::FuncType(ValueTypes params, ValueTypes results)
-    : ExternType(ExternKind::Func), params(params), results(results), func_types(nullptr) {}
+    : ExternType(ExternKind::Func),
+      params(params),
+      results(results),
+      func_types(nullptr) {}
```
sbc100: This line looks like it did not change?     Unless I'm missing something?   Maybe revert this line?

include/wabt/interp/interp.h
```diff
@@ -178,14 +178,37 @@ struct FuncType : ExternType {
...
   static bool classof(const ExternType* type);
 
+  enum class TypeKind {
+    Func,
+    Struct,
+    Array,
+  };
+
+  // To simplify the implementation, FuncType may also represent
+  // Struct and Array types. In the latter case, the mutability
```
sbc100: Can you explain a little more?   Why would FuncType represent struct or array?

src/interp/binary-reader-interp.cc
sbc100: Why is this relevant to this change? 

Surely this list will need to grow to count elements anyway?  Why not allocate it all here?

src/shared-validator.cc
```diff
@@ -46,44 +46,80 @@ void SharedValidator::OnTypecheckerError(const char* msg) {
   PrintError(expr_loc_, "%s", msg);
 }
 
+Result SharedValidator::OnRecursiveType(Index first_type_index,
```
sbc100: Should this be called `OnRecGroup` maybe?

test/roundtrip/rec-groups.txt
```diff
@@ -0,0 +1,36 @@
+;;; TOOL: run-roundtrip
```
sbc100, line 1: Should this test be called `rec-groups.txt`?

sbc100 (commented):
There is a lot of code here.  I've not yet had a chance to look at all of it but LGTM so far.

include/wabt/interp/interp.h
zherczeg: FuncType is a type, which is used in a lot of code in the interpreter. When I worked on this patch, I wanted to avoid adding a lot of new code to the interpreter (the patch is large enough), which primary purpose is just demonstration as far as I understood. This could be reworked later if we want the extra code. An option is just renaming FuncType to something generic.

include/wabt/interp/interp-inl.h
zherczeg: The line is longer than 80 columns, it is surprising for me that it is not a style error.

src/interp/binary-reader-interp.cc
zherczeg: Without this change, the fuzzer CI fails. The random byte sequence contains a huge "count" value, and when the `reserve` is executed, the fuzzer reports it as a "too big allocation". I tried to find a configuration parameter for the fuzzer, but I could not. I would also prefer a fuzzer change rather than this code change. At least it is in the interpreter, and not in the main code.

src/shared-validator.cc
zherczeg: I have changed all similar cases to OnRecGroup

zherczeg:
Thank you for the review. I have updated the patch.

src/interp/binary-reader-interp.cc
sbc100: But that issues seems unrelated to adding the GC type system no?  Perhaps this could be split into a separate PR?

I'm not so sure failing with `too big allocation` is such a bad outcome from such crazy inputs anyway.  

If we do use some hardcode value here it might make sense for it be the implementation limit on the number of function.

include/wabt/interp/interp.h
sbc100: OK, maybe add a TODO to refactor or rename it?

include/wabt/interp/interp-inl.h
sbc100: Hmm.. i wonder if we should clang-format the codebase as a separate PR?

zherczeg: I have no objection.

src/interp/binary-reader-interp.cc
zherczeg: I really don't know what is the best action here. The current code is perfect for correct WebAssembly files. The best thing would be to tell to the fuzzer that large allocations are valid. Ok I will move this to another patch.

include/wabt/interp/interp.h
zherczeg: Added

src/interp/binary-reader-interp.cc
zherczeg: #2713

zherczeg:
I have updated this patch. If there are follow-up works I should do (e.g changing the param/result types for functions from pointer to vector reference), please open issues for them and assign me. It is easy to forget these things.

zherczeg:
I made some small changes. I hope this patch is in good shape now. I will fix issues related to this patch.

zherczeg:
The patch is updated with `exn` support.

zherczeg:
Based on recent patches, it looks like lot of people needs GC support. However, these patches are big, and reviewing them in detail takes a lot of time.

Would it be possible to find a middle ground for these patches? Only do a coarse level review, and do the necessary changes later as other, smaller scope patches? These three patches should work in general.

src/wast-parser.cc
```diff
@@ -945,20 +969,43 @@ Result WastParser::ParseRefDeclaration(Var* out_type) {
...
+    case TokenType::Exn:
+    case TokenType::I31:
+    case TokenType::NoExn:
+    case TokenType::NoExtern:
+    case TokenType::NoFunc:
+    case TokenType::None:
+    case TokenType::Struct:
+      if (options_->features.gc_enabled()) {
+        is_generic_ref = true;
+      }
```
chicoxyzzy: `TokenType::Exn` is in the `gc_enabled()` group. `(ref func)` / `(ref extern)` work with function-references only. `(ref exn)` is exception-handling + typed refs and must not require `--enable-gc`.

Without GC this falls through to `ParseVar` and fails with `unexpected token "exn"` (same as #2818).

`Exn` belongs with `Func` / `Extern`. Keep `NoExn` behind exceptions + GC.

src/binary-reader.cc
```diff
@@ -377,9 +378,12 @@ Result BinaryReader::ReadType(Type* out_value, const char* desc) {
...
       Type::Enum heap_type_code = static_cast<Type::Enum>(heap_type);
-      ERROR_UNLESS(
-          heap_type_code == Type::FuncRef || heap_type_code == Type::ExternRef,
-          "Reference type is limited to func and extern: %s", desc);
+      ERROR_UNLESS(heap_type_code == Type::ExnRef ||
+                       heap_type_code == Type::FuncRef ||
+                       heap_type_code == Type::ExternRef ||
+                       (options_.features.gc_enabled() &&
+                        Type::EnumIsNonTypedGCRef(heap_type_code)),
+                   "not allowed reference type: %s", desc);
```
chicoxyzzy: Heap type `ExnRef` (`0x69`) is accepted whenever `reference_types` is on. It never checks `exceptions_enabled()`.

`EnumIsNonTypedGCRef` also includes `NullExnRef`, so `(ref noexn)` only requires GC, not exceptions.

`ExnRef` should require `exceptions_enabled()`. `NullExnRef` should require exceptions + GC.

src/wast-parser.cc
```diff
@@ -339,14 +340,37 @@ void ResolveImplicitlyDefinedFunctionType(const Location& loc,
...
+    case Type::NullExternRef:
+    case Type::NullRef:
+    case Type::AnyRef:
+    case Type::EqRef:
+    case Type::I31Ref:
+    case Type::StructRef:
+    case Type::ArrayRef:
+      return options_->features.gc_enabled();
+    default:
+      return true;
```
chicoxyzzy, line 373: `NullExnRef` is missing. It hits `default` and returns true, so `nullexnref` / `(ref noexn)` are treated as always enabled.

`ExnRef` correctly needs `exceptions_enabled()`. `NullExnRef` should need exceptions + GC.

include/wabt/type.h
```diff
@@ -155,12 +196,20 @@ class Type {
...
+      case Type::NullRef:
+        return (type_index_ == kBottomRef) ? "any" : "none";
+      case Type::FuncRef:       return "func";
+      case Type::ExternRef:     return "extern";
+      case Type::ExnRef:        return "exn";
+      case Type::AnyRef:        return "any";
+      case Type::EqRef:         return "eq";
+      case Type::I31Ref:        return "i31";
+      case Type::Struct:        return "struct";
+      case Type::Array:         return "array";
```
chicoxyzzy: These are composite opcodes (`0x5f` / `0x5e`), not ref kinds. Use `Type::StructRef` and `Type::ArrayRef`. As written, `structref` / `arrayref` print as `<invalid>`.

include/wabt/type.h
```diff
@@ -135,15 +153,38 @@ class Type {
...
       case Type::Any:       return "any";
+      case Type::NullExnRef:
+        return type_index_ == ReferenceOrNull ? "nullexnref" : "(ref noexn)";
+      case Type::NullFuncRef:
+        return type_index_ == ReferenceOrNull ? "nullfuncref" : "(ref nofunc)";
+      case Type::NullExternRef:
+        return type_index_ == ReferenceOrNull ? "nullexternref" : "(ref noextern)";
+      case Type::NullRef:
+        if (type_index_ == kBottomRef) {
+          return "(ref any)";
```
chicoxyzzy: `BottomRef` is an unknown/bottom reference used by the validator. Printing it as `(ref any)` is a non-nullable `any`, not a bottom type. That will confuse validator dumps.

chicoxyzzy:
Thanks for pushing GC types through wabt. The rec-group / one-super / per-module canon approach looks right for this slice.

Inline notes cover the type-system holes.

zherczeg:
Thank you for the review. I have updated the patch.

PR #2611, opened 2025-05-31
wasm2c: update Windows build to always use internal sha256 library to resolve libcrypto DLL not found issue

### LINKED ISSUE #2609: The wasm2c command has no effect
windows 11, Download 1.0.37 from release, there is no response when running commands, while commands from other tools are normal.

Image

Image

silbinarywolf:
Fixes #2609


keithw:
I don't use Windows myself, but I think this libcrypto stuff has been causing us enough trouble (on platforms that are maybe not as strong at versioning/managing dependencies) that we should probably just switch to an always-vendored BLAKE3 hash for this and ditch the external library. I'll send a PR on this.

PR #2618, opened 2025-06-11
Implement GC opcodes

zherczeg:
Implementing garbage collection byte codes. Starting with struct.


zherczeg:
Status update: all struct/array opcodes are implemented. That is 20 opcodes. The patch is 4K lines now, and the end is still very far.

zherczeg:
All 32 GC related opcodes are done. Tests might reveal some issues.

zherczeg:
I have added a `OnGCUnaryExpr` method here, and `RefEq`, which is a binary operation, also handled by it. Shall I add a separate method for this, or rename OnGCUnaryExpr to something else.

zherczeg:
Who is responsible for the fuzzer? It really hit something unfixable this time.
https://github.com/WebAssembly/wabt/blob/main/src/interp/interp.cc#L581
This is exactly working as expected. Just because the initial value is big, it is not an error. This memory limit thing needs to be disabled. But I simply don't understand how can I configure the fuzzer, since its code comes from somewhere else.

PR #2621, opened 2025-06-16
wasm2c: fix big endian guard page support

shravanrn:


shravanrn:
> lgtm, but are you trying to fix this just to make CI happy are your using s390 yourself in production. (If that latter I wonder how much effort it is worth investing here).

@sbc100 There is an effort to make sure Firefox works well in the s390x. A part of that effort was to see if the wasm2c-based RLBox library sandboxing works there as well. It's not the highest priority item, but I'm cleaning up the support in the chance that this is something that can get adopted.

PR #2622, opened 2025-06-27
Support all garbage collector tests

zherczeg:
This patch adds support for all tests in `testsuite/proposals/gc`. Two tests are still missing.


zherczeg:
It looks like garbage collection tests are finally working. It took me 4 months, but the initial version of garbage collection support is completed. I think I understand the concept of references in WebAssembly, and know why it works that way. The code probably needs a lot of changes.

kevmoo:
So exciting to see this moving along!

zherczeg:
Thank you! I try to update this patchset reguraly. The longer term aim is full wasm 3.0 support, but I also do some work on WebAssembly Components. Do you need this code for something?

kevmoo:
I work with the `dart2wasm` team. We have our own `wasm2wat` impl for our output. I'd be awesome if we could use the default tools.

kevmoo:
If you want a sample file: https://bouncing.web.app/main.dart.wasm

kevmoo:
Sweet!

zherczeg:
I am updating these patches but I don't know when they land. The first one in the set has a birthday soon!

PR #2626, opened 2025-08-06
Divide data segments across all output files when using wasm2c

white-axe:
wasm2c has a `--num-outputs` flag for splitting the output C file into smaller files to reduce compilation time and memory usage by the compiler, but it only splits the functions across the output files. The data segments are all still put in the first output file instead of being split across the output files. That can cause problems on some PowerPC platforms, where the PowerPC TOC places an upper limit on the size of a translation unit. I've modified the code generator to distribute the data segments across all the output files when the `--num-outputs` flag is set to something greater than one.


keithw:
Cool! Can you add a test to https://github.com/WebAssembly/wabt/tree/main/test/wasm2c/spec-multi-output that exercises this please? I'm not sure we have anything that uses data segments, much less multiple ones.

white-axe:
The memory_init tests already test reading from memory with multiple data segments. Each of the `data` S-expressions in memory_init.wast is a data segment.

white-axe:
Do I still have to write the test or anything else for this pull request?

PR #2642, opened 2025-09-21
Add doc of --enable-annotations of wasm2wat

qouteall:


sbc100:
I believe these html pages are auto-generated.    See scripts/generate-html-docs.sh

qouteall:
I will fix it later.

PR #2644, opened 2025-09-23
Update HTML doc

qouteall:
Previously I found that the web doc of wasm2wat doesn't mention `--enable-annotations`. Previous PR https://github.com/WebAssembly/wabt/pull/2642

After knowing that the HTML docs are auto-generated, I run `generate-html-docs.sh`. Now it includes `--enable-annotations` option because `wasm2wat.1` already includes it. 

That also caused many other changes, probably caused by HTML doc not updated for long time. I opened these html and they looks fine.

### DIFF
--- a/docs/doc/spectest-interp.1.html
+++ b/docs/doc/spectest-interp.1.html
@@ -2,17 +2,17 @@
 <html>
 <head>
   <meta charset="utf-8"/>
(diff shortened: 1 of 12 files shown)


sbc100:
lgtm!

For the record can you mention exactly which version of mandoc you used to generated these?

qouteall:
mandoc version 1.14.6-1

PR #2650, opened 2025-10-30
Add arm64 linux release binaries. NFC

KKiiim:
Support ARM64 builds and improve package naming in the GitHub Actions release workflow.

### 1. Added ARM64 Support
- Added `ubuntu-24.04-arm64` runner for native Linux ARM64 compilation
- Utilized existing `macos-14` runner which provides native macOS ARM64 (Apple Silicon) compilation
- New Linux ARM64 release package will be named `aarch64-ubuntu`

### 2. Updated Package Naming
- Renamed existing Ubuntu x86_64 package from `ubuntu-20.04` to `x86-ubuntu` for clarity
- Package names now clearly indicate the architecture

### 3. Matrix Configuration
The build matrix now includes:
- `amd64-ubuntu` (ubuntu-20.04)
- `arm64-ubuntu` (ubuntu-24.04-arm64)
- `arm64-macos-14` (macOS Apple Silicon)
- `amd64-windows-latest` (Windows-lastest)

## Release Artifacts

After these changes, each release will include:
- `wabt-{version}-amd64-ubuntu.tar.gz` - Linux x86_64 build
- `wabt-{version}-arm64-ubuntu.tar.gz` - Linux ARM64 build
- `wabt-{version}-arm64-macos-14.tar.gz` - macOS ARM64 build (Apple Silicon)
- `wabt-{version}-amd64-windows-latest.tar.gz` - Windows x86_64 build
- `wabt-{version}-wasi.tar.gz` - WASI build

### DIFF
--- a/.github/workflows/build_release.yml
+++ b/.github/workflows/build_release.yml
@@ -17,10 +17,10 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os: ubuntu-20.04
+          - os: ubuntu-latest
             arch: x86_64
             name: linux-x64
-          - os: ubuntu-24.04-arm64
+          - os: ubuntu-24.04-arm
             arch: aarch64
             name: linux-arm64
           - os: macos-14


.github/workflows/build_release.yml
```diff
@@ -15,7 +15,19 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-20.04, macos-14, windows-latest]
+        include:
+          - os: ubuntu-20.04
+            arch: x86_64
+            name: x86-ubuntu-20.04
```
sbc100: `x86-` seems wrong here.   How about `x64`, `x86_64` or `amd64`?

.github/workflows/build_release.yml
```diff
@@ -15,7 +15,19 @@ jobs:
...
+          - os: ubuntu-24.04-arm64
+            arch: aarch64
+            name: aarch64-ubuntu-24.04
```
sbc100: If we are going to go with friendly arch names names here in the `name` field I think `arm64` is more readable then `aarch64`.

KKiiim:
Standard GitHub-hosted runners:
https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

.github/workflows/build_release.yml
```diff
@@ -16,7 +16,19 @@ jobs:
...
+            name: amd64-ubuntu-20.04
+          - os: ubuntu-24.04-arm64
+            arch: aarch64
+            name: arm64-ubuntu-24.04
+          - os: macos-14
+            arch: aarch64
+            name: arm64-macos-14
+          - os: windows-latest
+            arch: x86_64
+            name: amd64-windows-latest
```
sbc100: Can we just call this one `windows` for now?  (At least until we support more than one windows arch).

.github/workflows/build_release.yml
```diff
@@ -16,7 +16,19 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-20.04, macos-14, windows-latest]
+        include:
+          - os: ubuntu-20.04
+            arch: x86_64
+            name: amd64-ubuntu-20.04
```
sbc100: can we these `linux-x64` and `linux-arm64`?

.github/workflows/build_release.yml
```diff
@@ -16,7 +16,19 @@ jobs:
...
+          - os: ubuntu-24.04-arm64
+            arch: aarch64
+            name: arm64-ubuntu-24.04
+          - os: macos-14
+            arch: aarch64
+            name: arm64-macos-14
```
sbc100: Can we call this `macos-arm64`?

sbc100: (i.e. can we drop the `-14` here)

sbc100 (commented):
Is there some way we can test this change?   Or do we need to wait for the next release to know if it actually works?

.github/workflows/build_release.yml
KKiiim: name "windows-x64" looks more friendly for developer to integration.
I think it is acceptable ?

KKiiim:
> Is there some way we can test this change? Or do we need to wait for the next release to know if it actually works?

# I have tested it on my fork repo and works well.

image
image

.github/workflows/build_release.yml
```diff
@@ -30,7 +42,7 @@ jobs:
 
     - name: install ninja (linux)
       run: sudo apt-get install ninja-build
-      if: startsWith(matrix.os, 'ubuntu-')
+      if: contains(matrix.os, 'ubuntu')
```
sbc100, line 45: This line can be reverted now I think

KKiiim:
Hi, @sbc100 
Should we support different ubuntu-version release?
Since different ubuntu have different glibc. Such as ubuntu-2204 don't have GLIBC_2.38 that may failed to executed.
`wasm2wat: /lib/x86_64-linux-gnu/libc.so.6: version GLIBC_2.38 not found (required by wasm2wat)`

sbc100:
> Hi, @sbc100 Should we support different ubuntu-version release? Since different ubuntu have different glibc. Such as ubuntu-2204 don't have GLIBC_2.38 that may failed to executed. `wasm2wat: /lib/x86_64-linux-gnu/libc.so.6: version GLIBC_2.38 not found (required by wasm2wat)`

No I think its fine to just support a single version, but we should build with the oldest version we can for max compatabiliy.

Folks to obscure/old linux distros and just rely on their distro package manager or build it themselves.

I did just notices that this PR changed ubuntu-20.04 to ubuntu-latest.   I failed to notice that in review.  Can you revert that part please?   We should use `ubuntu-20.04` to increase compat.

KKiiim:
> > Hi, @sbc100 Should we support different ubuntu-version release? Since different ubuntu have different glibc. Such as ubuntu-2204 don't have GLIBC_2.38 that may failed to executed. `wasm2wat: /lib/x86_64-linux-gnu/libc.so.6: version GLIBC_2.38 not found (required by wasm2wat)`
> 
> No I think its fine to just support a single version, but we should build with the oldest version we can for max compatabiliy.
> 
> Folks to obscure/old linux distros and just rely on their distro package manager or build it themselves.
> 
> I did just notices that this PR changed ubuntu-20.04 to ubuntu-latest. I failed to notice that in review. Can you revert that part please? We should use `ubuntu-20.04` to increase compat.

Sorry for the mistake. Have created a fix patch:
- https://github.com/WebAssembly/wabt/pull/2662

Issue #2655, opened 2025-11-02
test/wasm2c/spec/memory64/memory_copy.txt is failing on s390x


sbc100:
I'm not sure when this started failing but its failing on the main branch.

sbc100:
It looks like the first commit where it started failing was https://github.com/WebAssembly/wabt/pull/2634.. which seems completely unrelated.  So my guess is that something changes with the github CI infra of the qemu package we use to run the test.

sbc100:
@SoniEx2 would you have time to look into this?   It looks like that the s390 stuff was addeding #2380 but this test started failing for some external reason.

shravanrn:
@sbc100 @keithw I did a quick test to see if this issue still exists and unfortunately this is the case. https://github.com/WebAssembly/wabt/actions/runs/33046846822

We could lengthen the timeout of the failing tests. However, even without the failing tests, the CI for s390x takes about 25mins. Given this, I am not sure it makes sense to turn on CI even if these tests were fixed as this would more than double the current CI time of about 11mins. Not sure if y'all have a better suggestion.

keithw:
I don't have a strong view, but if we care about BE support, it would be nice to have a test for it. If it's a non-required test and we're willing to merge without waiting for it to complete, is it that bad if it takes 25 minutes?

shravanrn:
> If it's a non-required test and we're willing to merge without waiting for it to complete, is it that bad if it takes 25 minutes?

oh, i guess that's fine. Not sure if there are CI limits that we are eating up or something, but I assume this is not the case?

In that case, the easiest path I think right now is to simply disable the tests in question in CI and at least let the remaining tests run

PR #2656, opened 2025-11-02
Temporarily disable s390 CI build

sbc100:
See https://github.com/WebAssembly/wabt/issues/2655


sbc100:
I'm happy to re-enable once we fix #2655

PR #2662, opened 2025-11-06
Lower the Ubuntu version for the release

KKiiim:


KKiiim:
Thanks for review @sbc100

KKiiim:
> Have you tested these?

Yes. It works well in ubuntu-24.04 and 22.04 and mac-os

KKiiim:
Thanks. BTW, could you please kindly release the latest version. @sbc100

sbc100:
Do you think its worth doing another release right now?  Can't we wait a while for 1.0.40?

KKiiim:
> Do you think its worth doing another release right now? Can't we wait a while for 1.0.40?

My concern is the user who use ubuntu22 or lower version may failed to integrate latest wabt because of the glibc mismatch.
Of course, it's okay for me to wait for 1.0.40. Thanks !

PR #2663, opened 2025-11-09
Code improvements for call ref and br_on_[non_]null

zherczeg:
This is just a small set of changes which should follow #2567 . They are not that high risk, so it could be merged with #2567 , but partly unrelated, so finding bugs in the future could be easier if it is kept separate. @sbc100 let me know your opinion.

### DIFF
--- a/src/apply-names.cc
+++ b/src/apply-names.cc
@@ -98,6 +98,7 @@ class NameApplier : public ExprVisitor::DelegateNop {
   Result UseNameForDataSegmentVar(Var* var);
   Result UseNameForElemSegmentVar(Var* var);
   Result UseNameForParamAndLocalVar(Func* func, Var* var);
+  Result UseNameForLabelVar(Var* var);
   Result VisitFunc(Index func_index, Func* func);
   Result VisitGlobal(Global* global);
   Result VisitTag(Tag* tag);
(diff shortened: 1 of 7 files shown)


src/apply-names.cc
```diff
@@ -342,37 +349,31 @@ Result NameApplier::OnStoreExpr(StoreExpr* expr) {
 }
 
 Result NameApplier::OnBrExpr(BrExpr* expr) {
-  std::string_view label = FindLabelByVar(&expr->var);
-  UseNameForVar(label, &expr->var);
+  UseNameForLabelVar(&expr->var);
   return Result::Ok;
```
sbc100: Should we just do `return UseNameForLabelVar(&expr->var);` in all these places?

Or should `UseNameForLabelVar` return void?

zherczeg: I just followed the pattern. See the functions below `NameApplier::UseNameForLabelVar`. I don't understand why they return with a Result, that is not checked in most places. Shall I change them to void as well? Probably this is the right patch to do this.

zherczeg: I am sorry I misunderstood the code. The other cases are virtual functions with fixed declarations. I changed `UseNameForLabelVar` to void.

sbc100: Sorry to go back and forth on this, but what about keeping the Result return type and just doing 

```
Result NameApplier::OnBrIfExpr(BrIfExpr* expr) {
  return UseNameForLabelVar(&expr->var);
}
```

It just saves one line per function but there are quite a few functions?

zherczeg: You are right. This method is good for tail call optimizations.

zherczeg: I have also simplified the other CHECK_RESULT / return pairs. Patch is updated.

sbc100 (approved):

PR #2668, opened 2025-11-17
Fix use of uninitialized value in WastParser::ParseSimdV128Const.

wffurr:
Found this from msan.  See b/461044733.

### DIFF
--- a/src/wast-parser.cc
+++ b/src/wast-parser.cc
@@ -2952,7 +2952,7 @@ Result WastParser::ParseF32(Const* const_, ConstType const_type) {
   }
 
   auto literal = token.literal();
-  uint32_t f32_bits;
+  uint32_t f32_bits = 0;
   Result result = ParseFloat(literal.type, literal.text, &f32_bits);
   const_->set_f32(f32_bits);
   return result;
@@ -2972,7 +2972,7 @@ Result WastParser::ParseF64(Const* const_, ConstType const_type) {
   }
 
   auto literal = token.literal();
-  uint64_t f64_bits;
+  uint64_t f64_bits = 0;
   Result result = ParseDouble(literal.type, literal.text, &f64_bits);
   const_->set_f64(f64_bits);
   return result;


src/wast-parser.cc
```diff
@@ -2873,24 +2873,28 @@ Result WastParser::ParseSimdV128Const(Const* const_,
         case 16: {
           uint8_t value = 0;
           result = ParseInt8(sv, &value, ParseIntType::SignedAndUnsigned);
+          if (Failed(result)) break;
```
sbc100: But value is initialized above, no?   What uninitialized value would be used in `const_->set_v128_u8(lane, value);` below?

I guess if you want to keep these checks then you should remove the `= 0` initialization above?   (i.e. we should either initialize it so its safe to use regardless of ParseInt result, or we should check the ParseInt result)

wffurr: Aha I got carried away with the code pattern.  It's the float/double cases below that have the actual use of uninitialized value.  I'll amend the commit.

sbc100 (approved):

src/wast-parser.cc
wffurr: Thanks Sam; I revised the PR to only fix the use of an uninitialized uint32_t in the ParseF32 and ParseF64 functions, which is sufficient to fix the msan finding.  PTAL

sbc100:
BTW, while changes like this are great, and very much appreciated, you should note that we (wabt contributors) don't currently spend any time trying to fix fuzzer reported bugs in wabt, because (a) we don't the the resources and (b) the inputs tend to come from trusted sources. 

Obviously its fantastic when somebody is able to take the time like this, but just a word of warning that if you start down this road it might be a very long one :)

wffurr:
We use WABT in our test harness for running spec tests, but the sanitizers can't tell the difference between WABT and our own code that we actually do want to sanitize.  I'll keep sending these as the sanitizers find them, unless it becomes untenable at which point we'll find another solution.

Thanks for the prompt review!

Issue #2670, opened 2025-11-22
Assertion `local_ref_is_set_.size() == label->local_ref_is_set_.size()' failed at wasm-interp


Q1IQ:
## Current State

Run the following command:

```bash
wasm-interp --enable-all --run-export=main test.wasm
```

[test.wasm.txt](https://github.com/user-attachments/files/23687320/test.wasm.txt)

Observed output:
```
wasm-interp: /work/harnesses/sources/wabt/src/shared-validator.cc:512: void wabt::SharedValidator::RestoreLocalRefs(wabt::Result): Assertion `local_ref_is_set_.size() == label->local_ref_is_set_.size()' failed.

```

## GDB Backtrace

```
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Program received signal SIGABRT, Aborted.
__pthread_kill_implementation (no_tid=..., signo=..., threadid=...) at ./nptl/pthread_kill.c:44
#0  __pthread_kill_implementation (no_tid=..., signo=..., threadid=...) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=..., threadid=...) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=..., signo=...) at ./nptl/pthread_kill.c:89
#3  0x00007ffff6c26476 in __GI_raise (sig=...) at ../sysdeps/posix/raise.c:26
#4  0x00007ffff6c0c7f3 in __GI_abort () at ./stdlib/abort.c:79
#5  0x00007ffff6c0c71b in __assert_fail_base (fmt=..., assertion=..., file=..., line=..., function=...) at ./assert/assert.c:94
#6  0x00007ffff6c1de96 in __GI___assert_fail (assertion=..., file=..., line=..., function=...) at ./assert/assert.c:103
#7  0x0000555555766433 in wabt::SharedValidator::RestoreLocalRefs (this=..., result=...) at /work/harnesses/sources/wabt/src/shared-validator.cc:512
#8  0x000055555577366e in wabt::SharedValidator::OnEnd (this=..., loc=...) at /work/harnesses/sources/wabt/src/shared-validator.cc:917
#9  0x00005555555df32e in wabt::interp::(anonymous namespace)::BinaryReaderInterp::OnEndExpr (this=...) at /work/harnesses/sources/wabt/src/interp/binary-reader-interp.cc:358
[...]

PR #2672, opened 2025-11-22
Fix reference tracking in SharedValidator

zherczeg:
Clear "reference is set" bitset and hasmap in BeginFunctionBody.
Furtermore support try_tables.

Fixes: #2670

### DIFF
--- a/src/shared-validator.cc
+++ b/src/shared-validator.cc
@@ -540,6 +540,8 @@ Result SharedValidator::BeginFunctionBody(const Location& loc,
                                           Index func_index) {
   expr_loc_ = loc;
   locals_.clear();
+  local_ref_is_set_.clear();
+  local_refs_map_.clear();
   if (func_index < funcs_.size()) {
     for (Type type : funcs_[func_index].params) {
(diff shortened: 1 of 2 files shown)


sbc100:
Thanks for the quick fix!

src/shared-validator.cc
```diff
@@ -1444,6 +1447,7 @@ Result SharedValidator::EndTryTable(const Location& loc, Type sig_type) {
   TypeVector param_types, result_types;
   result |= CheckBlockSignature(loc, Opcode::TryTable, sig_type, &param_types,
                                 &result_types);
+  RestoreLocalRefs(result);
```
zherczeg, line 1450: I realized I misunderstood how BeginTryTable/EndTryTable work. This is a new part of the code for me. `BinaryReaderInterp::OnTryTableExpr` calls `EndTryTable` after catch blocks are processed, not after the `end`.

In this case this is not needed (although harmless), and the normal OnEnd should call RestoreLocalRefs (it does).
I am sorry, I will submit another fix.

PR #2674, opened 2025-12-02
added wasm 3 types

odewdney:
Added types and opcodes to get wasm-objdump to dump a kotlin wasm file


odewdney:
this is not needed if PR 2618 is processed...

PR #2681, opened 2026-01-14
wast-parser: make IsPowerOfTwo support memory64 addresses

sjamesr:


src/wast-parser.cc
```diff
@@ -36,7 +36,7 @@ namespace {
 
 static const size_t kMaxErrorTokenLength = 80;
 
-bool IsPowerOfTwo(uint32_t x) {
+bool IsPowerOfTwo(const Address& x) {
```
sbc100: I think that can just be `Address x`, no?

src/test-wast-parser.cc
```diff
@@ -88,6 +88,30 @@ TEST(WastParser, LongTokenSpace) {
...
 }
 
+TEST(WastParser, Memory64Alignment) {
+  // from align.wast, the large align param should be OK with memory64 enabled
+  std::string text = R"(
+  (module
+    (memory 1)
+    (func
+      i32.const 0
+      i32.load offset=0xFFFF_FFFF_FFFF_FFFF align=0x8000_0000_0000_0000
```
sbc100, line 101: If this exists in align.wast why doesn't the normal test suite run catch this bug?

sbc100 (approved):
lgtm % testing question

src/test-wast-parser.cc
sjamesr: > If this exists in align.wast why doesn't the normal test suite run catch this bug?

The test came into existence here at https://github.com/WebAssembly/testsuite/commit/4b24564c844e3d34bf46dfcb3c774ee5163e31cc, which I suspect is newer than the version in wabt's copy of the spec tests in third_party.

src/wast-parser.cc
sjamesr: Changed

src/test-wast-parser.cc
sbc100: Ah, I see.  In that case can you maybe add a comment with a link to that commit and a TODO to remove this test once we update the testsuite?

sjamesr: Done, thanks

sbc100 (approved):

PR #2684, opened 2026-01-17
[binary-reader] Re-use ReadExternalKind for import as well as exports. NFC

sbc100:
Another refactor to prepare for compact imports.

I don't feel strongly about `malformed import kind` vs `invalid import external kind` but we should probably be consistent between imports and exports.


sbc100:
See #2685 which depends on this change.

dschuff (approved):

PR #2685, opened 2026-01-17
Add initial support for compact import section. NFC

sbc100:
See https://github.com/WebAssembly/compact-import-section

For this initial commit I've just added a single simple test in form of test/binary/compact-imports.txt.  Once we update the `testsuite` repo we can pull in the official tests.

### DIFF
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -70,33 +70,6 @@ namespace wabt {
 
 namespace {
 
-struct ExternalType {
-  // Union into which we read extern type details.
-  ExternalType() : memType{} {}
-  ~ExternalType() {}
-  union {
-    // Global
-    struct {
-      Type type;
-      bool mutable_;
-    } globalType;
-    // Function/Tag
-    struct {
-      Index sig_index;
-    } funcType;
-    // Memory
-    struct {
-      Limits page_limits;
-      uint32_t page_size;
-    } memType;
-    // Table
-    struct {
-      Limits elem_limits;
-      Type elem_type;
-    } tableType;
-  };
-};
-
 class BinaryReader {
  public:
   struct ReadModuleOptions {
@@ -204,12 +177,10 @@ class BinaryReader {
   [[nodiscard]] Result ReadCustomSection(Index section_index,
                                          Offset section_size);
   [[nodiscard]] Result ReadTypeSection(Offset section_size);
-  [[nodiscard]] Result ReadImportDetails(ExternalKind kind, ExternalType* out);
-  [[nodiscard]] Result ReportImport(Index i,
-                                    std::string_view module_name,
-                                    std::string_view field_name,
-                                    ExternalKind kind,
-                                    ExternalType& type);
+  [[nodiscard]] Result ReadImport(Index i,
+                                  std::string_view module_name,
+                                  std::string_view field_name,
+                                  ExternalKind kind);
   [[nodiscard]] Result ReadImportSection(Offset section_size);
   [[nodiscard]] Result ReadFunctionSection(Offset section_size);
   [[nodiscard]] Result ReadTableSection(Offset section_size);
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -82,6 +83,7 @@ Wabt has been compiled to JavaScript via emscripten. Some of the functionality i
 [relaxed-simd]: https://github.com/WebAssembly/relaxed-simd
 [custom-page-sizes]: https://github.com/WebAssembly/custom-page-sizes
 [function-references]: https://github.com/WebAssembly/function-references
+[compat-imports]: https://github.com/WebAssembly/compact-import-section
```
bvisness: Typo

bvisness (approved):
All looks reasonable to me, although I've not looked at WABT's code before.

src/binary-reader.cc
```diff
@@ -2687,72 +2720,132 @@ Result BinaryReader::ReadTypeSection(Offset section_size) {
   return Result::Ok;
 }
 
+Result BinaryReader::ReadImportDetails(ExternalKind kind, ExternalType* out) {
```
aheejin: Given that `ReportImport` only calls callbacks, can this be merged into `ReadImport`? We also don't have `WriteImportDetails` and only have `WriteImport`.

src/binary-reader.cc
```diff
@@ -2687,72 +2720,132 @@ Result BinaryReader::ReadTypeSection(Offset section_size) {
...
+      CHECK_RESULT(ReadCount(&num_compact_imports, "compact import count"));
+      for (Index j = 0; j < num_compact_imports; ++j) {
+        CHECK_RESULT(ReadStr(&field_name, "compact import field name"));
+        if (kind_u8 != 0x7E) {
+          CHECK_RESULT(
+              ReadExternalKind(&kind, "compact import kind", "import"));
+        }
+        ExternalType type;
+        CHECK_RESULT(ReadImportDetails(kind, &type));
+        CHECK_RESULT(ReportImport(i++, module_name, field_name, kind, type));
```
aheejin: I feel separating the two cases is easier to read, like
```cpp
if (kind_u8 == 0x7E) {
  read external kind
  for (...) {
    read name
    read import
  }
} else { // kind_u8 == 0x7F
  for (...) {
    read name
    read external kind
    read import
  }
}
```
but this maybe can be a little longer? I don't feel too strongly though.

src/binary-writer.cc
```diff
@@ -1466,37 +1494,49 @@ Result BinaryWriter::WriteModule() {
...
-          WriteGlobalHeader(&cast<GlobalImport>(import)->global);
-          break;
-
-        case ExternalKind::Tag:
-          WriteTagType(&cast<TagImport>(import)->tag);
-          break;
+      bool compact = false;
+      if (options_.features.compact_imports_enabled()) {
+        // Write compact imports when they are available.
+        // Currentsly we donly support grouping by module name (0x7F mode)
```
aheejin:
```suggestion
        // Currently we only support grouping by module name (0x7F mode)
```

src/binary-writer.cc
```diff
@@ -1466,37 +1494,49 @@ Result BinaryWriter::WriteModule() {
...
+        // and not the module name + kind grouping (0x7E mode).
+        size_t group_size = 1;
+        for (size_t j = i + 1; j < module_->imports.size(); ++j) {
+          if (import->module_name == module_->imports[j]->module_name) {
+            group_size++;
+          } else {
+            break;
+          }
+        }
```
aheejin, line 1514:
```suggestion
        size_t group_size = 1, j = i + 1;
        while (j < module_->imports.size() && 
               module_->imports[j]->module_name == import->module_name) {
          group_size++;
          j++;
        }
```

src/binary-reader.cc
sbc100: Ok, done

src/binary-reader.cc
sbc100: Oh you're right!  No need for that new union type either.  Much better.

dschuff:
This does sort of raise the question of how a tool should handle compact imports. Since it looks like there will be both a text and a binary format, you could argue that a true "pass-through/roundtrip" encoding would only encode compact imports in binary where they are encoded in the text. But of course it also makes sense to encode them everywhere possible, just to reduce binary size. Maybe wabt should default to only round-tripping? (And optionally have a flag to opportunistically use compact imports?)

sbc100:
> This does sort of raise the question of how a tool should handle compact imports. Since it looks like there will be both a text and a binary format, you could argue that a true "pass-through/roundtrip" encoding would only encode compact imports in binary where they are encoded in the text. But of course it also makes sense to encode them everywhere possible, just to reduce binary size. Maybe wabt should default to only round-tripping? (And optionally have a flag to opportunistically use compact imports?)

Right now that flag is `--enable-compact-imports`.    Do you think we would ever want the binaryen writer to not opportunistically use compact imports when the feature is enabled?    

I guess if we might have users who want to enable them in the reader but disabled them in the writer, but so far that hasn't come up for other features.

dschuff:
Right, for Binaryen, I don't know of any case where we wouldn't want to opportunistically use compact imports when they are enabled. Wabt is more interesting, because its job is to be a low-level tool that round-trips text and binary, rather than an optimization/transformation tool, and changing the encoding is a transformation.

sbc100:
Yeah, right no the compact imports are not remembered on read, and we write them purely opportunistically.     We should probably followup by preserving them in the IR instead.

PR #2692, opened 2026-02-12
Fix wrong iterator in InterpDropKeep ref tracking

sumleo:
## Summary

- In the `InterpDropKeep` handler in `interp.cc`, the second loop ("find dropped refs") checked `*iter` instead of `*drop_iter` in its break condition, causing incorrect erasure of ref-tracking entries in the `refs_` vector.
- This could lead to reference-typed values losing their GC root tracking while still alive on the value stack.
- Add a regression test that exercises `drop=2`/`keep=1` with three `externref` values on the stack.

## Details

The `InterpDropKeep` handler has two loops over `refs_` (reverse iterators):

1. **First loop** (`iter`): shifts kept refs down by `drop` positions.
2. **Second loop** (`drop_iter`): finds dropped refs to erase.

The bug is in the second loop:

```cpp
auto drop_iter = iter;
for (; drop_iter != refs_.rend(); ++drop_iter) {
  if (*iter < values_.size() - keep - drop) {  // BUG: should be *drop_iter
    break;
  }
}
```

Since `iter` is not advanced in this loop, the break condition is evaluated against a stale value on every iteration. This causes the loop to either break too early (erasing too few entries) or never break (erasing too many entries), depending on the stale `*iter` value relative to the boundary.

**Fix:** Change `*iter` to `*drop_iter`.

## Test plan

- [x] New test `InterpTest.DropKeepRefTracking` creates a wasm module with 3 externref params, a block that pushes all three to the stack, and a `br` that triggers `drop=2`/`keep=1`
- [x] All existing unit tests pass (128 tests)

### DIFF
--- a/src/interp/interp.cc
+++ b/src/interp/interp.cc
@@ -1684,7 +1684,7 @@ RunResult Thread::StepInternal(Trap::Ptr* out_trap) {
       // Find dropped refs.
       auto drop_iter = iter;
       for (; drop_iter != refs_.rend(); ++drop_iter) {
-        if (*iter < values_.size() - keep - drop) {
+        if (*drop_iter < values_.size() - keep - drop) {
           break;
         }
       }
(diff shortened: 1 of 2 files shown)


sumleo:
Closing as duplicate of #2702 which has been merged.

PR #2693, opened 2026-02-12
Add bounds check for function body size in binary reader

sumleo:
## Summary

- `ReadCodeSection` computed `end_offset = body_start_offset + body_size` without validating that `end_offset` falls within the code section boundary (`read_end_`).
- A crafted module with an oversized `body_size` could cause the reader to parse past the section end.
- Add an `ERROR_UNLESS` check that validates `end_offset` against `read_end_`, matching the validation pattern used elsewhere for section and subsection sizes.

## Details

In `binary-reader.cc`, `ReadCodeSection` reads `body_size` via LEB128 and computes the expected end offset:

```cpp
uint32_t body_size;
CHECK_RESULT(ReadU32Leb128(&body_size, "function body size"));
Offset body_start_offset = state_.offset;
Offset end_offset = body_start_offset + body_size;
// No validation that end_offset <= read_end_
CALLBACK(BeginFunctionBody, func_index, body_size);
```

The `end_offset` is later used to set `state_.offset` (when `skip_function_bodies` is true) and passed to `ReadFunctionBody` as the expected end position. Without bounds validation, a malformed `body_size` allows reading past the code section boundary.

**Fix:** Add `ERROR_UNLESS(end_offset >= body_start_offset && end_offset <= read_end_, ...)` immediately after computing `end_offset`.

## Test plan

- [x] New test `BinaryReader.InvalidFunctionBodySize` exercises the fix with a wasm module whose function body size (255 bytes) exceeds the code section
- [x] All existing unit tests pass (128 tests)

### DIFF
--- a/src/test-binary-reader.cc
+++ b/src/test-binary-reader.cc
@@ -73,3 +73,29 @@ TEST(BinaryReader, DisabledOpcodes) {
         << "Got error message: " << message;
   }
 }
+
+TEST(BinaryReader, InvalidFunctionBodySize) {
+  // A wasm module where the function body size extends past the end of the
+  // code section.  Without the bounds check this would allow the binary reader
+  // to read past the section boundary.
+  // TODO: Move this test upstream into the spec repo.
+
+  uint8_t data[] = {
+      0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00,  // magic + version
+      0x01, 0x04, 0x01, 0x60, 0x00, 0x00,  // type section: 1 type, (func)
+      0x03, 0x02, 0x01, 0x00,              // func section: 1 func, type 0
+      // Code section: 1 func, but body_size claims 0xFF bytes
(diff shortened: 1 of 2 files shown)


src/test-binary-reader.cc
```diff
@@ -73,3 +73,28 @@ TEST(BinaryReader, DisabledOpcodes) {
         << "Got error message: " << message;
   }
 }
+
+TEST(BinaryReader, InvalidFunctionBodySize) {
+  // A wasm module where the function body size extends past the end of the
+  // code section.  Without the bounds check this would allow the binary reader
+  // to read past the section boundary.
```
sbc100, line 80: Presumably this means we are lacking an upstream spec test for this case?

Maybe add a TODO here to move this test upstream into the spec repo?

sumleo: Done, added a TODO to move this test upstream into the spec repo.

PR #2694, opened 2026-02-12
Fix integer overflow in section and subsection size checks

sumleo:
## Summary

- Multiple places in the binary reader computed boundary offsets by adding a size value to the current position before checking bounds (e.g. `read_end_ = state_.offset + section_size`). On platforms where `size_t` is 32-bit, large size values cause the addition to wrap around, producing a small result that incorrectly passes subsequent bounds checks.
- Replace all addition-based checks with subtraction-based equivalents (`size > end - offset`), which are safe because the invariant `state_.offset <= read_end_` is always maintained.
- Add regression tests with oversized section and subsection sizes.

## Details

The vulnerable pattern appears in 6 locations:

| Function | Line | Expression |
|----------|------|------------|
| `ReadSections` | 3109 | `read_end_ = state_.offset + section_size` |
| `ReadNameSection` | 2036 | `subsection_end = state_.offset + subsection_size` |
| `ReadDylinkSection` | 2227 | `subsection_end = state_.offset + subsection_size` |
| `ReadLinkingSection` | 2359 | `subsection_end = state_.offset + subsection_size` |
| `ReadStr` | 420 | `state_.offset + str_len <= read_end_` |
| `ReadBytesWithSize` | 445 | `state_.offset + size <= read_end_` |

On 32-bit platforms where `Offset` is `size_t` (= `uint32_t`), a `section_size` of e.g. `0xFFFFFFF0` added to an offset of `0x20` produces `0x10`, which passes the check `read_end_ <= state_.size` even though the section extends far past the buffer.

**Fix:** Use subtraction from the known-good bound instead:
```cpp
// Before (overflow-prone):
ERROR_UNLESS(state_.offset + size <= read_end_, ...);
// After (overflow-safe):
ERROR_UNLESS(size <= read_end_ - state_.offset, ...);
```

For `ReadSections`, add an explicit overflow check before computing `read_end_`.

## Test plan

- [x] New test `BinaryReader.OversizedSectionSize` with a 2 GiB section size in a tiny module
- [x] New test `BinaryReader.OversizedSubsectionSize` with a name section subsection exceeding the section boundary
- [x] All existing unit tests pass (129 tests)

### DIFF
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -417,7 +417,7 @@ Result BinaryReader::ReadStr(std::string_view* out_str, const char* desc) {
   uint32_t str_len = 0;
   CHECK_RESULT(ReadU32Leb128(&str_len, "string length"));
 
-  ERROR_UNLESS(state_.offset + str_len <= read_end_,
+  ERROR_UNLESS(str_len <= read_end_ - state_.offset,
                "unable to read string: %s", desc);
 
   *out_str = std::string_view(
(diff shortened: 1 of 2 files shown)


sumleo:
Done, added TODO comments to both tests to move them upstream into the spec repo.

PR #2696, opened 2026-02-24
[wasm-objdump] Fix missing error on bad opcode

sbc100:
In binary-reader-objdump were we suppressing all error except those in pre-pass, in order to avoid reporting them mulitple time.

However, the pre-pass is run with `skip_function_bodies` so opcode error are not shown in that pass.

### DIFF
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -738,6 +738,22 @@ Result BinaryReader::ReadAddress(Address* out_value,
 }
 
 Result BinaryReader::ReadFunctionBody(Offset end_offset) {
+  uint64_t total_locals = 0;
+  Index num_local_decls;
+  CHECK_RESULT(ReadCount(&num_local_decls, "local declaration count"));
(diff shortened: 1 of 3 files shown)


src/binary-reader.cc
```diff
@@ -738,6 +738,22 @@ Result BinaryReader::ReadAddress(Address* out_value,
 }
 
 Result BinaryReader::ReadFunctionBody(Offset end_offset) {
+  uint64_t total_locals = 0;
```
dschuff, line 741: Is moving this code related to the error checking or just a cleanup?

sbc100: Yes, with this code being moved we end up seeing any error that relate to locals being shown twice, once in the pre-pass and once in the disassembly.

Once we move this code into `ReadFunctionBody` it means that `skip_function_bodies` sill skip this too, which means that pre-pass will skip this, and the only disassembly will show local-related errors.

PR #2699, opened 2026-02-25
Fix GC root leak in RefPtr assignment operators

sumleo:
## Summary

The copy and move assignment operators for `RefPtr` (both same-type and cross-type) overwrite `root_index_` with a new root without first calling `DeleteRoot` on the old one. While the destructor correctly releases the root via `reset()`, assignments bypass this cleanup, leaking one GC root per reassignment.

In a long-running interpreter session where `RefPtr` objects are reassigned frequently, the leaked roots accumulate in the `Store`'s root list. Because roots prevent the GC from collecting the referenced objects, this leads to unbounded memory growth.

## Changes

- Call `reset()` at the top of each assignment operator so the previous root is released before a new one is acquired.
- Add self-assignment guards (`if (this == &other) return *this;`) to the same-type copy and move assignment operators to avoid use-after-reset on self-assignment.

The cross-type assignment operators (`RefPtr<T>::operator=(const RefPtr<U>&)` and `RefPtr<T>::operator=(RefPtr<U>&&)`) do not need a self-assignment guard because `T` and `U` are different types, so `this` and `&other` cannot alias.

## Test plan

- [x] Verified that `wasm-interp` (which exercises `RefPtr` heavily through the interpreter) builds cleanly with no warnings.
- The fix is a mechanical correction of a well-understood RAII pattern; existing interpreter tests cover `RefPtr` usage paths.

### DIFF
--- a/include/wabt/interp/interp-inl.h
+++ b/include/wabt/interp/interp-inl.h
@@ -317,6 +317,7 @@ RefPtr<T>::RefPtr(const RefPtr<U>& other)
 template <typename T>
 template <typename U>
 RefPtr<T>& RefPtr<T>::operator=(const RefPtr<U>& other) {
+  assert(static_cast<const void*>(this) != static_cast<const void*>(&other));
   reset();
   obj_ = other.obj_;
   store_ = other.store_;
@@ -336,6 +337,7 @@ RefPtr<T>::RefPtr(RefPtr&& other)
 template <typename T>
 template <typename U>
 RefPtr<T>& RefPtr<T>::operator=(RefPtr&& other) {
+  assert(static_cast<const void*>(this) != static_cast<const void*>(&other));
   reset();
   obj_ = other.obj_;
   store_ = other.store_;


include/wabt/interp/interp-inl.h
```diff
@@ -313,6 +317,7 @@ RefPtr<T>::RefPtr(const RefPtr<U>& other)
 template <typename T>
 template <typename U>
 RefPtr<T>& RefPtr<T>::operator=(const RefPtr<U>& other) {
+  reset();
```
zherczeg, line 321: Why `if (this == &other) return *this;` is not added here?
I suspect the point of this check, is that if you run a reset on yourself, then you copy null values later (set by the reset), and just loose the object forever.

zherczeg (commented):
Yes, it looks like a serious bug.

include/wabt/interp/interp-inl.h
zherczeg: Sorry, I see in your description that this case cannot happen. I think an assert would be still useful here that it really never happens.

sumleo: Good point. Added an assert to both cross-type assignment operators to document the invariant. Since T and U are different types self-assignment can't happen through normal use, but the assert catches any unexpected aliasing.

zherczeg (approved):

PR #2704, opened 2026-02-26
Fix off-by-one in Select instruction ref tracking

sumleo:
## Summary
- The `Select` instruction's ref tracking compares against wrong stack index after popping the condition value, causing reference operands to lose GC tracking.
- Fix adjusts both ref checks to use `values_.size() - 1` instead of `values_.size()`.

## Details
In the `Select` case (interp.cc), `Pop<u32>()` removes the condition from the value stack before the ref checks run. At that point, `values_.size()` is the index one past `false_val`, not the index of `false_val` itself. Similarly, after popping `false_`, `values_.size()` is one past `true_val`.

The `refs_` vector stores indices of reference-typed values on the stack. Comparing `refs_.back()` against `values_.size()` (instead of `values_.size() - 1`) means neither operand is detected as a ref. The result is that when both select operands are references, the pushed result loses its ref tracking, which can cause premature GC collection.

The fix changes both comparisons from `values_.size()` to `values_.size() - 1` to correctly identify the top-of-stack value.

### DIFF
--- a/test/interp/select-ref.txt
+++ b/test/interp/select-ref.txt
@@ -1,0 +1,71 @@
+;;; TOOL: run-interp
+;;; ARGS: --enable-function-references
+
+(module
+  (type $i32_func (func (result i32)))
+
+  ;; Two simple functions to use as reference values.
+  (func $f1 (type $i32_func) i32.const 42)
+  (func $f2 (type $i32_func) i32.const 99)
+
+  ;; Declare functions in elem so ref.func is valid.
+  (elem declare funcref (ref.func $f1) (ref.func $f2))
+
+  ;; Typed select choosing between two funcref values.
+  ;; When cond is nonzero, should return first operand (ref to $f1).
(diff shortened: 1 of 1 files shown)


zherczeg:
The patch looks good. Thank you for finding these and fixing them.
I think adding a test would be a good idea since an early collection should have visible side effects.

sumleo:
Added a test in `test/interp/select-ref.txt` that exercises typed `select` with `funcref` operands. It covers:

- Selecting between two non-null funcrefs (true and false branches)
- Selecting between a non-null ref and a null ref
- Storing the select result into a global

Note that a GC-level test (verifying early collection has visible side effects) would ideally need low-level Thread control to trigger `Store::Collect()` mid-execution, which the existing test infrastructure doesn't support from WAT — there's a TODO about this at `test-interp.cc:734`. The WAT-level test verifies correct behavior with reference operands, which is the practical regression test for this fix.

PR #2707, opened 2026-02-26
Fix typed select pushing Type::Any instead of expected type

sumleo:
## Summary
- The typed select instruction's type checker validates operands against the expected type but forgets to set `result_type`, leaving it as `Type::Any`.
- Fix adds `result_type = expected[0]` in the typed select branch.

## Details
In `TypeChecker::OnSelect` (type-checker.cc), `result_type` is initialized to `Type::Any`. The untyped select path (when `expected` is empty) correctly sets `result_type = type1` on line 1012. However, the typed select path (when `expected` is non-empty) validates both operands against `expected[0]` but never assigns `result_type`.

The typed select instruction (`select t`) exists specifically to provide precise type information for cases where the operand types alone are insufficient (e.g., reference types). Pushing `Type::Any` as the result type defeats this purpose and can cause downstream type checking to be less precise than intended.

The fix adds `result_type = expected[0]` after the CheckType calls in the typed select branch, mirroring the pattern used in the untyped path.

## Tests
- `typed-select-result-type.txt`: Verifies that typed select with funcref/externref annotations properly propagates the result type, allowing the result to be used in contexts expecting that specific reference type (e.g., `global.set` to a `funcref` global, return from an `externref` function).
- `bad-typed-select-type-mismatch.txt`: Verifies that using a typed select `(result funcref)` where `externref` is expected correctly produces a type mismatch error.

### DIFF
--- a/test/typecheck/bad-typed-select-type-mismatch.txt
+++ b/test/typecheck/bad-typed-select-type-mismatch.txt
@@ -1,0 +1,18 @@
+;;; TOOL: wat2wasm
+;;; ERROR: 1
+(module
+  ;; The result of typed select (result funcref) should be funcref,
+  ;; so using it where externref is expected should fail.
+  (global (mut externref) (ref.null extern))
+  (func
+    ref.null func
+    ref.null func
+    i32.const 1
+    select (result funcref)
+    global.set 0
+  ))
+(;; STDERR ;;;
+out/test/typecheck/bad-typed-select-type-mismatch.txt:12:5: error: type mismatch in global.set, expected [externref] but got [funcref]
+    global.set 0
+    ^^^^^^^^^^
+;;; STDERR ;;)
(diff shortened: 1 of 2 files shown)


zherczeg:
This looks valid. Even without GC there are some reference types, so it is possible to add a test for this. With GC, unknown reference types in dead code could be an interesting test as well.

sbc100:
@zherczeg should we land this now or ask for a test for be added?

sumleo:
Added two tests:

- `test/typecheck/typed-select-result-type.txt` — positive test that typed select with `funcref` and `externref` annotations properly propagates the result type (e.g., storing the result in a `(mut funcref)` global, returning it from an `externref` function).
- `test/typecheck/bad-typed-select-type-mismatch.txt` — negative test that using a typed select `(result funcref)` where `externref` is expected correctly produces a type mismatch error.

PR #2715, opened 2026-03-12
Implement quoted identifier parsing

zherczeg:
It turned out that webassembly 3.0 support `$"quoted identifiers"`. It seems it is part of the annotations proposal:

https://github.com/WebAssembly/annotations/blob/main/proposals/annotations/Overview.md
Extend the grammar of identifiers as follows: `id ::= "$"idchar+ | "$"name`

This patch should add support for it.

### DIFF
--- a/include/wabt/wast-lexer.h
+++ b/include/wabt/wast-lexer.h
@@ -89,7 +89,7 @@ class WastLexer {
     return ReadReservedChars() == ReservedChars::None;
   }
   void ReadSign();
-  Token GetStringToken();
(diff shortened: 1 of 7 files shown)


zherczeg:
This patch is bigger than I hoped for, but at least not complex.

zherczeg:
More tests are always useful I think. And they test the error message as well, not just the operation. Can I land the patch?

sbc100:
Yup, feel free to land if you have LGTM and tests pass.

PR #2717, opened 2026-03-14
ci: add -y to apt-get commands and unify cross-build install

kevmoo:
- Add `-y` to all `apt-get install` and `apt-get update` commands to prevent process hangs.
- Unify `apt-get install` commands in `build-cross`.
- Fix minor typo (`Temporatily` -> `Temporarily`).


.github/workflows/build.yml
```diff
@@ -13,14 +13,16 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/setup-python@v1
+    - uses: actions/setup-python@v6
       with:
         python-version: '3.x'
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v6
+      with:
+        fetch-depth: 0
```
sbc100: Do we need `fetch-depth: 0` for some reason?   IIRC this is slower since it fetches the whole history

.github/workflows/build.yml
```diff
@@ -34,35 +36,29 @@ jobs:
...
     - name: install ninja (osx)
       run: brew install ninja
       if: startsWith(matrix.os, 'macos-')
     - name: install ninja (win)
       run: choco install ninja
       if: matrix.os == 'windows-latest'
-    - name: mkdir
-      run: mkdir -p out
+    - uses: ilammy/msvc-dev-cmd@v1
+      if: matrix.os == 'windows-latest'
```
sbc100: What is this new external dependency doing?

sbc100 (commented):
Thanks for all the fixes!

Would you mind splitting out the build.yaml changes that are just updates and typo fixes from the win32 build stuff which seems like more of an actual change?

.github/workflows/build.yml
kevmoo: The fetch-depth: 0 setting was necessary because the lint job relies on the 

`./scripts/clang-format-diff.sh` script (called on line 27 of build.yml).

By default, the actions/checkout step only performs a shallow fetch of the repository (fetching just a single commit, i.e., depth=1).

However, the `clang-format-diff.sh` script uses Git commands that need to inspect your repository's history to figure out what changed so it only formats modified code:


```bash
# Finds the common ancestor commit
# Runs the formatter only on the diff against the merge base
git clang-format $MERGE_BASE -q --diff -- src/
```

Without the full Git history fetched (`fetch-depth: 0`), `git merge-base` and `git clang-format` wouldn't be able to find the common ancestor commit locally, causing the CI lint step to fail.

kevmoo: I totally wrote that 😛

.github/workflows/build.yml
kevmoo: That `step` (the `ilammy/msvc-dev-cmd@v1` action) activates the Microsoft Visual C++ (MSVC) Developer Command Prompt environment on the `windows-latest` GitHub Actions runner. 

It was introduced in the very same commit (`ea995cd92`) you asked about previously, which fixed Windows CI failures. 

Here's why it was needed:
The WABT build uses the **Ninja** build system across all platforms (`cmake -G Ninja`). On Windows runners, GitHub Actions has multiple compilers installed (both MSVC and MinGW/GCC). If you run CMake with Ninja without explicitly setting up the MSVC environment, CMake might default to using the MinGW `gcc` compiler instead of Microsoft's `cl.exe`, or it might fail entirely to find the C/C++ compiler.

By calling `- uses: ilammy/msvc-dev-cmd@v1`, it injects all the environment variables (like `PATH`, `LIB`, and `INCLUDE`) required for CMake and Ninja to locate and use `cl.exe` + `link.exe` correctly for the native Windows build.

kevmoo: We could also revert the bits here that move the Windows build to use Ninja, but it seemed nice to have everything using ninja 🤷

kevmoo: I'm happy to add more notes so there are bread crumbs if that's helpful – in the actual source

.github/workflows/build.yml
sbc100: Ah yes, of course, this is only the lint step, that makes sense.

Why did the `v1` version not need this?  Did it fetch all the history by default maybe?

kevmoo:
Yep! They changed the default to be more efficient.

.github/workflows/build.yml
kevmoo: v1 did the full checkout. v6 is being efficient. But too efficient in this case.

.github/workflows/build.yml
sbc100: Can you split out the windows->ninja parts?   We can discuss those more in a separate PR.  The other changes here seem trivially good.

kevmoo: Done! Split the Windows Ninja and MSVC updates out into #2846.

kevmoo: Done! https://github.com/WebAssembly/wabt/pull/2846

kevmoo:
PTAL!

sbc100 (approved):
lgtm, but it looks like there are conflicts so maybe we need another merge or rebase?

kevmoo:
Merged `main` to resolve conflicts—thanks!

sbc100 (approved):

PR #2719, opened 2026-03-15
Optimize text with escapes parsing

zherczeg:
This patch is a small optimization for quoted text parsing, which reduces the number of allocations, and the final string is allocated once (capacity == size, so no bytes wasted at the buffer end).

The final string is allocated once without wasting memory at the end. An internal buffer is allocated at most once, when the inline buffer is too small. The final size of the unescaped buffer is always less or equal than the original size. This observation can be used for reducing memory allocations.

### DIFF
--- a/src/wast-parser.cc
+++ b/src/wast-parser.cc
@@ -40,14 +40,15 @@ bool IsPowerOfTwo(Address x) {
   return x && ((x & (x - 1)) == 0);
 }
 
-template <typename OutputIter>
-void RemoveEscapes(std::string_view text, OutputIter dest) {
+void RemoveEscapes(std::string_view text, char* dest, size_t* out_length) {
   // Remove surrounding quotes; if any. This may be empty if the string was
   // invalid (e.g. if it contained a bad escape sequence).
   if (text.size() <= 2) {
+    *out_length = 0;
     return;
   }
 
+  const char* start = dest;
   const char* src = text.data();
   const char* end = text.data() + text.size() - 1;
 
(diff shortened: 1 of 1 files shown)


src/wast-parser.cc
```diff
@@ -752,11 +746,14 @@ Result WastParser::ErrorIfLpar(const std::vector<std::string>& expected,
   return Result::Ok;
 }
 
+static const size_t kInlineBufferSize = 96;
```
sbc100, line 749: How did you pick this value?

src/wast-parser.cc
```diff
@@ -768,14 +765,27 @@ Result WastParser::ParseVarText(Token& token, std::string* out_text) {
       return Result::Error;
     }
 
-    RemoveEscapes(token.text(), std::back_inserter(*out_text));
-    size_t length = out_text->length();
+    // The length of the final string is
+    // always less or equal than original size.
```
sbc100: Can this comment fit on a single line?

src/wast-parser.cc
```diff
@@ -867,11 +877,39 @@ Result WastParser::ParseTextList(std::vector<uint8_t>* out_data) {
...
+  while (PeekMatch(TokenType::Text)) {
     texts.push_back(Consume().text());
+    length += texts.back().length();
+  }
+
+  // The length of the final data is
+  // always less or equal than original size.
+  char inline_buffer[kInlineBufferSize];
+  std::vector<char> buffer;
+  char* data;
```
sbc100, line 888: Maybe just initialize this as `char* data = inline_buffer;` to make the below condition only have one branch?

zherczeg:
I made a measurement with cpu cycle counters. Parsing time of "[Method] 012345778\61::func" is reduced from 90960 cycles to 77543 cycles, which is 17% improvement (in release mode). The overall runtime is 0.0000379 sec, so this is negligible.

This patch is intended to be a small improvement. Of course it is not that important, I don't mind if it is rejected. The GC patches are more important for me.

src/wast-parser.cc
zherczeg: No reason. a 100 byte stack is not big, and people prefer shorter names.

sbc100: Maybe worth a comment, even if to mention that its somewhat arbitrary.

sbc100:
I'm somewhat conflicted because I think we should aim to keep to keep wabt simple where possible.   But performance is nice.    Perhaps we could leave this open and we can re-consider if folks are noticing slow parse times for text files.

Another measurement I'd be curious about:  Does this change have any measurable effect on the time it takes to run the test suite?

PR #2577, opened 2025-04-01
Remove -sSINGLE_FILE and other emscripten link flags

sbc100:

PR #2720, opened 2026-03-15
demo cleanup and modernization

kevmoo:
- **emscripten: Export compact_imports feature**
- **docs/demo: Modernize JS syntax and remove empty HTML tags**
- **Split libwabt into separate JS and WASM files**
- **Features: get features from the binary.**
- **Migrate WABT web demos to use CDN dependencies**
- **Add WebAssembly favicon to WABT web demos**

### DIFF
--- a/scripts/clang-format-diff.sh
+++ b/scripts/clang-format-diff.sh
@@ -3,6 +3,12 @@
 set -o errexit
 set -o pipefail
 
+FIX=0
+if [ "$1" = "--fix" ]; then
+  FIX=1
+  shift
+fi
+
 if [ -n "$1" ]; then
   BRANCH="$1"
 elif [ "$CI" != "true" ]; then
@@ -25,8 +31,15 @@ else
 fi
 
 MERGE_BASE=$(git merge-base $BRANCH HEAD)
+
+if [ $FIX -eq 1 ]; then
+  echo "Applying clang-format..."
+  git clang-format $MERGE_BASE -q -- src/ docs/demo/ ':(exclude)docs/demo/third_party.bundle.js'
+  exit 0
+fi
+
 if ! git clang-format $MERGE_BASE -q --diff -- src/ docs/demo/ ':(exclude)docs/demo/third_party.bundle.js' 2>&1 >/dev/null; then
-  echo "Please run git clang-format before committing, or apply this diff:"
+  echo "Please run '$0 --fix $BRANCH' before committing, or apply this diff:"
   echo
   # Run git clang-format again, this time with output.  This lets us add
   # the above message.
--- a/src/wabt.post.js
+++ b/src/wabt.post.js
@@ -52,7 +52,7 @@ function allocateBuffer(buf) {
     size = buf.byteLength;
     addr = malloc(size);
     (new Uint8Array(HEAPU8.buffer, addr, size))
-      .set(new Uint8Array(buf.buffer, buf.byteOffset, buf.byteLength));
+        .set(new Uint8Array(buf.buffer, buf.byteOffset, buf.byteLength));
   } else if (typeof buf == 'string') {
     const encoded = new TextEncoder().encode(buf);
     size = encoded.length;
@@ -61,7 +61,7 @@ function allocateBuffer(buf) {
   } else {
     throw new Error('unknown buffer type: ' + buf);
   }
-  return { addr: addr, size: size };
+  return {addr: addr, size: size};
 }
 
 function allocateCString(s) {
@@ -70,7 +70,7 @@ function allocateCString(s) {
   const addr = malloc(size + 1);
   (new Uint8Array(HEAPU8.buffer, addr, size)).set(encoded);
   HEAPU8[addr + size] = 0;  // null-terminate
-  return { addr: addr, size: size };
+  return {addr: addr, size: size};
 }
 
 /// Features
@@ -87,13 +87,13 @@ class Features {
   }
 }
 
-Object.keys(FEATURES).forEach(function (feature) {
+Object.keys(FEATURES).forEach(function(feature) {
   Object.defineProperty(Features.prototype, feature, {
     enumerable: true,
-    get: function () {
+    get: function() {
       return Module['_wabt_' + feature + '_enabled'](this.addr);
     },
-    set: function (newValue) {
+    set: function(newValue) {
       Module['_wabt_set_' + feature + '_enabled'](this.addr, newValue | 0);
     }
   });
@@ -105,8 +105,8 @@ class Lexer {
     this.filenameObj = allocateCString(filename);
     this.bufferObj = allocateBuffer(buffer);
     this.addr = Module._wabt_new_wast_buffer_lexer(
-      this.filenameObj.addr, this.bufferObj.addr, this.bufferObj.size,
-      errors.addr);
+        this.filenameObj.addr, this.bufferObj.addr, this.bufferObj.size,
+        errors.addr);
   }
 
   destroy() {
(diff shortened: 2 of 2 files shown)


kevmoo:
CC @sbc100 – unlike the C++ changes, I actually understand most of this. 😄

kevmoo:
I'm happy to submit these as stacked PRs, but I figured it might be easier to just land them all together if that's okay w/ you

CMakeLists.txt
```diff
@@ -829,8 +829,11 @@ if (EMSCRIPTEN)
     -sALLOW_TABLE_GROWTH
     -sALLOW_MEMORY_GROWTH
     -sMODULARIZE
+    -sEXPORT_ES6=1
     -sEXPORT_NAME=WabtModule
     -sEXPORTED_RUNTIME_METHODS=writeAsciiToMemory
+    -sWASM=1
+    -sSINGLE_FILE=0
```
sbc100: You don't need these two options since they are both the default.

CMakeLists.txt
```diff
@@ -829,8 +829,11 @@ if (EMSCRIPTEN)
     -sALLOW_TABLE_GROWTH
     -sALLOW_MEMORY_GROWTH
     -sMODULARIZE
+    -sEXPORT_ES6=1
```
sbc100: We can drop `-sMODULARIZE` if we add `-sEXPORT_ES6` since the later implies the format.

You can also drop `=1` here.

Makefile
```diff
@@ -151,6 +151,7 @@ update-wasm2c-fac:
 .PHONY: demo
 demo: emscripten-release
 	cp out/emscripten/Release/libwabt.js docs/demo
+	cp out/emscripten/Release/libwabt.wasm docs/demo
```
sbc100, line 154: This is s a little confusing.. I guess we dropped `-sSINGLE_FILE` a while back but never updated this Makefile? 

Ah yes, indeed: #2577

docs/demo/third_party/.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
sbc100, line 1: Why move to `docs/third_party` from `docs/demo/third_party`?     Keeping this stuff under the `demo` directory makes sense to me (since its only used for the demo, right?)

kevmoo: I was trying to keep "turds" out of the deployed directory.
Happy to put back if you like.

sbc100: I see yes, I agree we wouldn't want to deploy node_modules or anything like that.

I just want to make sure its clear to an observer that these deps are just for the demo, and nothing to do with wabt itself (or even wabt's docs)

kevmoo: fixed!

scripts/clang-format-diff.sh
```diff
@@ -25,11 +25,11 @@ else
 fi
 
 MERGE_BASE=$(git merge-base $BRANCH HEAD)
-if ! git clang-format $MERGE_BASE -q --diff -- src/ 2>&1 >/dev/null; then
+if ! git clang-format $MERGE_BASE -q --diff -- src/ docs/demo/ ':(exclude)docs/demo/third_party.bundle.js' 2>&1 >/dev/null; then
```
sbc100, line 28: I think a better way to do this is to create a `.clang-format-ignore` file.  See https://github.com/emscripten-core/emscripten/blob/main/.clang-format-ignore

Maybe revert this file from this PR and we can add `.clang-format-ignore` separately.

Makefile
```diff
@@ -20,7 +20,7 @@ MAKEFILE_NAME := $(lastword $(MAKEFILE_LIST))
 ROOT_DIR := $(dir $(abspath $(MAKEFILE_NAME)))
 
 USE_NINJA ?= 1
-EMSCRIPTEN_DIR ?= $(dir $(shell which emcc))
+EMSCRIPTEN_DIR ?= $(shell em-config EMSCRIPTEN_ROOT 2>/dev/null || dirname $(shell which emcc))
```
sbc100, line 23: Is this change needed?  it seems like if `em-config` is the PATH then `emcc` will always be in the PATH too.

Maybe revert this line?

CMakeLists.txt
```diff
@@ -813,32 +813,46 @@ if (EMSCRIPTEN)
...
+    COMMAND ${Python3_EXECUTABLE} ${WABT_SOURCE_DIR}/scripts/gen-emscripten-exports.py
+      --feature-def ${WABT_SOURCE_DIR}/include/wabt/feature.def
+      --features-out ${WABT_FEATURES_JS}
+      --exports-in ${WABT_SOURCE_DIR}/src/emscripten-exports.txt.in
+      --exports-out ${EMSCRIPTEN_EXPORTS}
+    DEPENDS
+      ${WABT_SOURCE_DIR}/scripts/gen-emscripten-exports.py
+      ${WABT_SOURCE_DIR}/include/wabt/feature.def
+      ${WABT_SOURCE_DIR}/src/emscripten-exports.txt.in
+  )
```
sbc100, line 831: In that past we've had trouble with adding python dependencies to the default build.  See #1970 for example and https://github.com/webassembly/wabt/issues/1385.

It might be worth leaving this exact python script stuff out of this PR and landing a version that just has the static feature list.

We can then consider this separately once this base PR has landed.

Makefile
kevmoo: Ah! this was so that folks who symlink binaries work (like on homebrew).

The directory with `emcc` is not necessarily the EMSCRIPTEN_ROOT

sbc100:
In general I will always prefer to land each commit as its own PR (we always squash on landing anyway to be sure that bisect keeps working).  

I'm not saying all of these commits need to be their own PR, but maybe there are at least 2 or 3 here.

Makefile
sbc100: I've not heard of this... but I can see that it could be an issue.   I wonder why no homebrew folks have complained about this issue in the part.

I any case, this seems like a separate bugfix that we don't need to make part of the demo upgrade.

(Seems like if do need to deal with symlinks we could also do `$(realpath $(dir ($shell which emcc)))` instead maybe?).

kevmoo: Ran command: `echo "1. \$(dirname \$(which emcc))" && echo "2. \$(realpath \$(dirname \$(which emcc)))" && echo "3. \$(dirname \$(realpath \$(which emcc)))"`
Ran command: `emcc_path=$(which emcc); echo "dirname:" $(dirname $emcc_path); echo "realpath dirname:" $(realpath $(dirname $emcc_path)); echo "dirname realpath:" $(dirname $(realpath $emcc_path))`
Ran command: `ls -la /opt/homebrew/Cellar/emscripten/5.0.3/bin/cmake`

If we use `$(realpath $(dir $(shell which emcc)))`, it actually wouldn't quite work for two reasons:

1. **Resolution order:** Because `dir` runs *before* `realpath`, it checks `dir` of `/opt/homebrew/bin/emcc`, which returns `/opt/homebrew/bin/`. Calling `realpath` on that directory just returns `/opt/homebrew/bin/` because the directory itself isn't a symlink (only the file inside it is). 
2. **Homebrew's directory structure:** Even if we reversed it to `$(dir $(realpath $(shell which emcc)))`, it would resolve the symlink and return `/opt/homebrew/Cellar/emscripten/5.0.3/bin`. However, Homebrew places Emscripten's CMake toolchain files in `libexec/cmake`, not `bin/cmake`. 

`em-config EMSCRIPTEN_ROOT` is a utility provided directly by Emscripten precisely for this reason! It knows its own internal paths and confidently returns the correct toolchain root (`.../libexec`) regardless of how the package manager scrambled the symlinks.

PR #2721, opened 2026-03-17
Refactor browser demos and Emscripten build output

kevmoo:
This comprehensive modernization of the WABT web demos and Emscripten
build output includes the following major improvements:

* Separate WASM payload: Updated CMake and Makefile to output a separate
  `libwabt.wasm` file instead of inlining it in JS. This drastically
  reduces the `libwabt.js` size from ~1.24MB to ~26KB.
* ES6 and async/await Modernization: Refactored `wabt.post.js` and demo
  scripts to use modern ES6 syntax (classes, const/let, arrow functions)
  and properly load the WebAssembly binary asynchronously.
* Dynamic feature flags: Demos now dynamically generate feature checkboxes
  by inspecting `wabt.FEATURES`, rather than hardcoding them in HTML.
* NPM Dependency Bundling: Migrated third-party demo dependencies
  (CodeMirror, Split.js) into a `package.json`-based build, generating
  a clean `third_party.bundle.js`. Consolidated shared demo logic into
  `share.js`.
* Emscripten exports: Added missing features (`compact_imports`,
  `custom_page_sizes`, and `wide_arithmetic`) to `emscripten-exports.txt`
  to prevent instantiation errors.

### DIFF
--- a/docs/demo/third_party/README.md
+++ b/docs/demo/third_party/README.md
@@ -1,27 +1,37 @@
 # WABT Demo Third-Party Dependencies
 
-This directory contains the build configuration for third-party dependencies used by the WABT web demos (located in `docs/demo`).
+This directory contains the build configuration for third-party dependencies
+used by the WABT web demos (located in `docs/demo`).
 
-Currently, this is used to generate a single ES Module bundle for **CodeMirror 6**, which is required because CodeMirror 6 consists of many highly-scoped NPM packages that are difficult to load individually via CDNs in a browser without instance-duplication issues.
+Currently, this is used to generate a single ES Module bundle for **CodeMirror
+6**, which is required because CodeMirror 6 consists of many highly-scoped NPM
+packages that are difficult to load individually via CDNs in a browser without
+instance-duplication issues.
 
 ## Building the Third-Party Bundle
 
-The demo applications (`wasm2wat/index.html` and `wat2wasm/index.html`) expect to find a compiled bundle at `docs/demo/third_party.bundle.js`.
+The demo applications (`wasm2wat/index.html` and `wat2wasm/index.html`) expect
+to find a compiled bundle at `docs/demo/third_party.bundle.js`.
 
 To regenerate this bundle:
 
 1. Ensure you have [Node.js and npm](https://nodejs.org/) installed.
-2. From this directory (`docs/third_party`), execute:
+2. From this directory (`docs/demo/third_party`), execute:
 
 ```bash
 npm install
 npm run build
 ```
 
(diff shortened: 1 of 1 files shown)


docs/demo/third_party/index.js
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export { basicSetup, EditorView } from "codemirror";
+export { EditorState } from "@codemirror/state";
+export { StreamLanguage } from "@codemirror/language";
+export { wast } from "@codemirror/legacy-modes/mode/wast";
```
sbc100: Hmm.. I wonder why "wast" is under legacy-modes here?

kevmoo: Ah! It's because it hasn't been updated in a while – although gemini told me that it's still more up-to-date than the wast coloring that was in the repo!

kevmoo:
Let me dig through your questions before you land this!

kevmoo:
> Thanks for splitting this out! Much appreciated.
> 
> LGTM.
> 
> With a couple of questions:
> 
> 1. I assume you tested this locally somehow to make sure it works?

Yep! Both apps. Make sure all of the boxes work and that I can edit code and it runs!

> 2. What cmake command was used to build the demo binary (i.e. debug vs release,etc)

Just `make demo`. Double checked that it is a release build. I also added that `Makefile` fix for emscripten. See the convo in the other PR.

I think this is ready to land!

Makefile
```diff
@@ -20,7 +20,7 @@ MAKEFILE_NAME := $(lastword $(MAKEFILE_LIST))
 ROOT_DIR := $(dir $(abspath $(MAKEFILE_NAME)))
 
 USE_NINJA ?= 1
-EMSCRIPTEN_DIR ?= $(dir $(shell which emcc))
+EMSCRIPTEN_DIR ?= $(shell em-config EMSCRIPTEN_ROOT 2>/dev/null || dirname $(shell which emcc))
```
sbc100: I agree this might be needed, but maybe in a different PR? 

BTW, did you run into this issue yourself building against homebrew version of emscripten?

I wonder if we should instead switch to `emcmake` which is a command we takes care of the  `cmake/Modules/Platform/Emscripten.cmake` thing for us.

kevmoo: Okay...

sbc100:
Let me know when you are ready and I will land this, and we can test the live version.

kevmoo:
Done! Land it!

kevmoo:
gah! no! Let me fix!

kevmoo:
oh. nevermind. all good. 😮‍💨

PR #2722, opened 2026-03-17
Fix wasm2wat demo

kevmoo:
This was broken by #2721


docs/demo/wasm2wat/demo.js
sbc100, line 51: (unrelated) Can this be an arrow function?

kevmoo: likely?

PR #2724, opened 2026-03-17
Fix Emscripten warnings by removing deprecated writeAsciiToMemory

kevmoo:
The Emscripten function writeAsciiToMemory has been deprecated and logs warnings
during the build process.

- Replaced usages of writeAsciiToMemory in wabt.post.js with TextEncoder
- Fixed a 1-byte buffer overflow in allocateCString
- Removed writeAsciiToMemory from CMake EXPORTED_RUNTIME_METHODS


src/wabt.post.js
```diff
@@ -78,19 +78,22 @@ function allocateBuffer(buf) {
     addr = malloc(size);
     (new Uint8Array(HEAP8.buffer, addr, size)).set(buf);
   } else if (typeof buf == 'string') {
-    size = buf.length;
+    const encoded = new TextEncoder().encode(buf);
+    size = encoded.length;
     addr = malloc(size);
-    writeAsciiToMemory(buf, addr, true);  // don't null-terminate
+    (new Uint8Array(HEAP8.buffer, addr, size)).set(encoded);
```
sbc100: I think you can do `addr = stringToNewUTF8(buf)` instead of these 4 lines.

src/wabt.post.js
```diff
@@ -78,19 +78,22 @@ function allocateBuffer(buf) {
...
 
 function allocateCString(s) {
-  const size = s.length;
-  const addr = malloc(size);
-  writeAsciiToMemory(s, addr);
+  const encoded = new TextEncoder().encode(s);
+  const size = encoded.length;
+  const addr = malloc(size + 1);
+  (new Uint8Array(HEAP8.buffer, addr, size)).set(encoded);
+  HEAP8[addr + size] = 0;
```
sbc100: It looks like you can just return `addr` from this function since it only has one caller `this.filenameObj = allocateCString(filename);` and you just just make `this.filename` be the address of the string.

kevmoo:
PTAL

src/wabt.post.js
```diff
@@ -201,18 +185,14 @@ class Errors {
...
-    }
   }
 }
 
 
 /// parseWat
 function parseWat(filename, buffer, options) {
   let errors = new Errors('text');
-  const lexer = new Lexer(filename, buffer, errors);
-  errors.lexer = lexer;
```
kevmoo, line 215: This was just silly!

docs/demo/wat2wasm/worker.js
```diff
@@ -5,12 +5,9 @@ let wrappedConsole = Object.create(console);
 
 wrappedConsole.log = (...args) => {
   const line = args.map(String).join('') + '\n';
-  postMessage({
-    type: "log",
-    data: line
-  });
+  postMessage({ type: 'log', data: line });
   console.log(...args);
-}
```
kevmoo, line 13: This is just a reformat – adding the missing `;`

src/wabt.post.js
```diff
@@ -42,17 +42,9 @@ const FEATURES = Object.freeze({
...
-  if (value === undefined) {
-    return default_;
-  }
-  return value;
-}
-
 /// Coerce value to boolean if not undefined. Otherwise return default_.
 function booleanOrDefault(value, default_) {
-  return !!maybeDefault(value, default_);
+  return !!(value !== undefined ? value : default_);
```
sbc100: Can this just be `!!(value ?? default_)`?

sbc100: In fact I'd be tempted to just inline it everywhere e.g.:

```
const foldExprs = options.foldExprs ?? false;
```

kevmoo: PTAL

PR #2738, opened 2026-04-06
Use unsigned line info in Locations

zherczeg:


sbc100 (approved):

sbc100:
The emscripten CI failures are from the fact that emscripten uses the latest version of clang/llvm which is a little more strict here.   Maybe create a separate PR to fix those first?

zherczeg:
Good idea!

zherczeg:
Is it ok to land this patch?

sbc100:
lgtm!

PR #2739, opened 2026-04-06
Remove debug-parser option and other unused variables

zherczeg:
The support for --debug-parser is removed from the project after the parser has been reworked. This patch removes the    command line option as well.

Would not be bad to change the `"Use multiple times for more info"` to something else.

### DIFF
--- a/src/tools/wast2json.cc
+++ b/src/tools/wast2json.cc
@@ -40,10 +40,8 @@ using namespace wabt;
 
 static const char* s_infile;
 static std::string s_outfile;
-static int s_verbose;
 static WriteBinaryOptions s_write_binary_options;
 static bool s_validate = true;
-static bool s_debug_parsing;
 static Features s_features;
 
 static std::unique_ptr<FileStream> s_log_stream;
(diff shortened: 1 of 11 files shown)


src/tools/wast2json.cc
```diff
@@ -62,11 +60,8 @@ static void ParseOptions(int argc, char* argv[]) {
   OptionParser parser("wast2json", s_description);
 
   parser.AddOption('v', "verbose", "Use multiple times for more info", []() {
-    s_verbose++;
     s_log_stream = FileStream::CreateStderr();
   });
-  parser.AddOption("debug-parser", "Turn on debugging the parser of wast files",
-                   []() { s_debug_parsing = true; });
```
sbc100, line 69: Did this flag used to work in the past?  Maybe we should make it work rather than removing it?

src/tools/wat2wasm.cc
```diff
@@ -71,8 +70,6 @@ static void ParseOptions(int argc, char* argv[]) {
     s_verbose++;
     s_log_stream = FileStream::CreateStderr();
   });
-  parser.AddOption("debug-parser", "Turn on debugging the parser of wat files",
-                   []() { s_debug_parsing = true; });
```
sbc100, line 75: Ditto.

sbc100 (approved):
lgtm % question

src/tools/wast2json.cc
zherczeg: What does it do? It is added here, replacing another flag:
https://github.com/WebAssembly/wabt/commit/bc3deb3288252866d9b571cd28c4104b9ffffd56

zherczeg:
```
   parser.AddOption("debug-parser", "Turn on debugging the parser of wast files",
-                   []() { s_parse_options.debug_parsing = true; });
+                   []() { s_debug_parsing = true; });
```
This was the actual change

zherczeg: A flag still present here, but it seems nothing is using it.
https://github.com/WebAssembly/wabt/blob/main/include/wabt/wast-parser.h#L37

Perhaps some plans changed?

sbc100:
It looks like the debug parser setting was removed back in #591 when the parser was re-written.  So I think its safe to remove these command line flags, but maybe mention that in the PR description and maybe even add to the PR title?

sbc100 (approved):

PR #2744, opened 2026-05-06
Validator: thread table limits through `OnReturnCallIndirect`

ggreif:
`SharedValidator::OnCallIndirect` passes `table_type.limits` to `TypeChecker::OnCallIndirect`, which expects i64 vs i32 for the table-index based on `is_64`. `OnReturnCallIndirect` doesn't — so under `table64`, plain `call_indirect` with an i64 index validates, but the matching `return_call_indirect` rejects it with `expected [i32] but got [i64]`.

Mirrors the signature. Adds a `test/typecheck/` regression for the negative case.

### DIFF
--- a/include/wabt/type-checker.h
+++ b/include/wabt/type-checker.h
@@ -106,7 +106,8 @@ class TypeChecker {
   Result OnReturnCall(const TypeVector& param_types,
                       const TypeVector& result_types);
   Result OnReturnCallIndirect(const TypeVector& param_types,
-                              const TypeVector& result_types);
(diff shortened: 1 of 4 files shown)


sbc100 (approved):

sbc100:
Thanks!

ggreif:
@sbc100 thanks for reviewing and merging! What is the schedule for 1.0.41? Any plans? It would be cool to have a release before the `nixpkgs` (26.05, stable) happens...

sbc100:
I can do a release today

PR #2748, opened 2026-05-28
Fix Windows/MSVC handling of large WAT input files and increase stack reserve

aydgelee:
# Fix Windows/MSVC handling of large WAT input files and increase stack reserve
## Summary

This PR fixes Windows/MSVC handling of very large WAT input files in WABT tools, mainly `wat2wasm.exe` and `wasm2wat.exe`.

On Windows, `wat2wasm` can fail when reading WAT files larger than 2GB with errors such as:

```text
value too large
unable to read file
```

The failure is caused by the MSVC file-reading path using APIs and types that are not safe for large files.

After fixing the large-file read path, processing very large WAT files can also trigger a Windows stack overflow during parsing:

```text
0xC00000FD
STATUS_STACK_OVERFLOW
```

This PR addresses both issues for Windows/MSVC builds while keeping non-Windows behavior unchanged.

### Windows/MSVC large-file reading

In `src/common.cc`, the Windows/MSVC file-reading path is updated to use 64-bit file APIs and file-size types:

```text
_stat      -> _stat64
fseek      -> _fseeki64
ftell      -> _ftelli64
long size  -> __int64 / 64-bit file size
```

The MSVC path also performs size checks before resizing buffers and reads large files in chunks.

### Windows/MSVC stack reserve

In `CMakeLists.txt`, Windows/MSVC builds now set a larger stack reserve for both tools:

```cmake
if(MSVC)
  if(TARGET wasm2wat)
    target_link_options(wasm2wat PRIVATE "/STACK:536870912")
  endif()
  if(TARGET wat2wasm)
    target_link_options(wat2wasm PRIVATE "/STACK:536870912")
  endif()
endif()
```

`536870912` bytes equals 512MB.

This is only applied for MSVC builds.

### Non-Windows behavior preserved

Linux and macOS code paths remain unchanged. The large-file API changes are limited to `COMPILER_IS_MSVC` / `MSVC` branches.

No WebAssembly parser semantics are changed.

## Validation

Validation was performed on Windows x64 using a WAT file larger than 2GB.

### Baseline behavior

Using the baseline code, even with a larger stack reserve, `wat2wasm` still failed during file reading:

```text
value too large
unable to read file
```

This confirms that increasing stack size alone does not fix the issue. The file-reading path must be updated.

### Large-file IO fix without larger stack

After updating the Windows/MSVC file-reading path, `wat2wasm` no longer failed with `value too large`, but processing the large WAT file exited with:

```text
0xC00000FD
STATUS_STACK_OVERFLOW
```

This confirms that the large-file reading fix works, and the next failure point is parser stack usage.

### Full fix

With both changes applied:

1. Windows/MSVC large-file reading fix
2. 512MB stack reserve for `wasm2wat` and `wat2wasm`

The round-trip conversion succeeds:

```powershell
wasm2wat.exe input.wasm -o output.wat
wat2wasm.exe output.wat -o roundtrip.wasm
```

Result:

```text
ExitCode=0
roundtrip.wasm generated successfully
```

### Windows/MSVC

Windows builds can now handle WAT files larger than 2GB in the tested scenario.

Affected tools:

```text
wasm2wat.exe
wat2wasm.exe
```

### Linux/macOS

Linux and macOS behavior is not changed.

The non-MSVC file-reading path remains the original WABT behavior, and `/STACK` is only applied under `MSVC`.

## Build instructions

```powershell
cmake -S . -B build -A x64 -DBUILD_TESTS=OFF
cmake --build build --config Release --target wasm2wat wat2wasm
```

Expected output files:

```text
build/Release/wasm2wat.exe
build/Release/wat2wasm.exe
```

### Generate WAT from wasm

```powershell
$wasm2wat = "build\\Release\\wasm2wat.exe"
$wasm = "input.wasm"
$wat = "output.wat"

& $wasm2wat $wasm -o $wat

"LASTEXITCODE=$LASTEXITCODE"
```

Expected result:

```text
LASTEXITCODE=0
output.wat generated successfully
```

### Convert WAT back to wasm

```powershell
$wat2wasm = "build\\Release\\wat2wasm.exe"
$out = "roundtrip.wasm"

& $wat2wasm $wat -o $out

"LASTEXITCODE=$LASTEXITCODE"
```

Expected result:

```text
LASTEXITCODE=0
roundtrip.wasm generated successfully
```

## Notes for reviewers

- The large-file API changes are limited to the Windows/MSVC path.
- The `/STACK:536870912` option is only applied under `MSVC`.
- Non-Windows code paths are intended to remain unchanged.
- This PR does not change WebAssembly parsing semantics.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -511,6 +511,10 @@ function(wabt_executable)
   add_dependencies(everything ${EXE_NAME})
   target_link_libraries(${EXE_NAME} ${EXE_LIBS})
 
+  if (MSVC)
+    target_link_options(${EXE_NAME} PRIVATE "/STACK:8388608")
+  endif ()
+
   if (EMSCRIPTEN)
     set(EXTRA_LINK_FLAGS
       "${EXTRA_LINK_FLAGS} -sNODERAWFS -Oz -sALLOW_MEMORY_GROWTH"
--- a/src/stream.cc
+++ b/src/stream.cc
@@ -20,6 +20,11 @@
 #include <cctype>
 #include <cerrno>
 
+#if COMPILER_IS_MSVC
+#include <cstdint>
+#include <limits>
+#endif
+
 #define DUMP_OCTETS_PER_LINE 16
 #define DUMP_OCTETS_PER_GROUP 2
 
@@ -27,6 +32,39 @@
 #define ERROR(fmt, ...) fprintf(stderr, "wabt: " fmt, __VA_ARGS__)
 
 namespace wabt {
+
+#if COMPILER_IS_MSVC
+static Result WriteDataChunked(FILE* file,
+                               const void* data,
+                               size_t size,
+                               const char* name) {
+  static constexpr size_t kWriteChunkSize = 64 * 1024 * 1024;
+  const uint8_t* src = static_cast<const uint8_t*>(data);
+  size_t offset = 0;
+
+  while (offset < size) {
+    size_t remaining = size - offset;
+    size_t chunk_size =
+        remaining < kWriteChunkSize ? remaining : kWriteChunkSize;
+    size_t bytes_written = fwrite(src + offset, 1, chunk_size, file);
+    if (bytes_written != chunk_size) {
+      ERROR("failed to write %" PRIzd " bytes to %s\n", size, name);
+      return Result::Error;
+    }
+    offset += bytes_written;
+  }
+
+  return Result::Ok;
+}
+
+static int SeekFile(FILE* file, size_t offset) {
+  if (offset > static_cast<size_t>(std::numeric_limits<int64_t>::max())) {
+    errno = EINVAL;
+    return -1;
+  }
+  return _fseeki64(file, static_cast<int64_t>(offset), SEEK_SET);
+}
+#endif
 
 Stream::Stream(Stream* log_stream)
     : offset_(0), result_(Result::Ok), log_stream_(log_stream) {}
@@ -145,6 +183,13 @@ Result OutputBuffer::WriteToFile(std::string_view filename) const {
     return Result::Ok;
   }
 
+#if COMPILER_IS_MSVC
+  if (Failed(WriteDataChunked(file, data.data(), data.size(),
+                              filename_str.c_str()))) {
+    fclose(file);
+    return Result::Error;
+  }
+#else
   ssize_t bytes = fwrite(data.data(), 1, data.size(), file);
   if (bytes < 0 || static_cast<size_t>(bytes) != data.size()) {
     ERROR("failed to write %" PRIzd " bytes to %s\n", data.size(),
@@ -152,6 +197,7 @@ Result OutputBuffer::WriteToFile(std::string_view filename) const {
     fclose(file);
     return Result::Error;
   }
+#endif
 
   fclose(file);
   return Result::Ok;
@@ -161,11 +207,17 @@ Result OutputBuffer::WriteToStdout() const {
   if (data.empty()) {
     return Result::Ok;
   }
+#if COMPILER_IS_MSVC
+  if (Failed(WriteDataChunked(stdout, data.data(), data.size(), "stdout"))) {
+    return Result::Error;
+  }
+#else
   ssize_t bytes = fwrite(data.data(), 1, data.size(), stdout);
   if (bytes < 0 || static_cast<size_t>(bytes) != data.size()) {
     ERROR("failed to write %" PRIzd " bytes to stdout\n", data.size());
     return Result::Error;
   }
+#endif
   return Result::Ok;
 }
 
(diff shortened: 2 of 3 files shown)


sbc100:
For the 64-bit file ops can we do something simpler like?

```
#ifdef _MSC_VER
#define ftell _ftelli64
#define fseek _fseeki64
#endif
```

For the stack size do you know why we would need such a high value?  IIUC the 8mb limit on linux seems to be enough there.

aydgelee:
For the file APIs, I agree the goal is simply to use the 64-bit MSVC CRT variants. I avoided globally redefining `ftell` / `fseek` because that changes all uses in the translation unit through macros. Keeping the `_fseeki64` / `_ftelli64` usage local to the large-file read path makes the change more explicit and avoids affecting unrelated code.

For the stack size, one clarification: `/STACK` on Windows controls the PE stack reserve size, not memory committed at process startup. It is an upper bound for the thread stack address range, and pages are committed on demand as the stack grows.

The reason this is included here is that, after the 64-bit file I/O fix, the large WAT test no longer failed with `value too large`; instead, it hit:

```text
0xC00000FD
STATUS_STACK_OVERFLOW
```
on Windows.

MSVC-linked executables typically default to a 1MB stack reserve, while Linux/macOS environments commonly have larger or configurable stack limits, for example an 8MB soft stack limit on many Linux systems. That likely explains why the same workload can pass on Linux/macOS but fail on Windows.

So the larger /STACK value is intended to raise the Windows stack reserve ceiling for very large WAT parsing workloads. It does not mean the process commits that amount of memory at startup.

aydgelee:
CI only reported a clang-format issue. No behavior change. Applied the suggested formatting diff.

CMakeLists.txt
```diff
@@ -556,6 +556,15 @@ if (BUILD_TOOLS)
     INSTALL
   )
 
+  if (MSVC)
+    if (TARGET wasm2wat)
+      target_link_options(wasm2wat PRIVATE "/STACK:536870912")
+    endif ()
+    if (TARGET wat2wasm)
+      target_link_options(wat2wasm PRIVATE "/STACK:536870912")
```
sbc100: If we need a bigger stack on windows can we just go with 8mb (to match the linux default).

Also, can we just include this in `wabt_executable` so that it applies to all binaries?   We might as well be consistent.

sbc100:
> For the file APIs, I agree the goal is simply to use the 64-bit MSVC CRT variants. I avoided globally redefining `ftell` / `fseek` because that changes all uses in the translation unit through macros. Keeping the `_fseeki64` / `_ftelli64` usage local to the large-file read path makes the change more explicit and avoids affecting unrelated code.

I'd much rather do it by via redefining these macros, keeping the the code otherwise the same on all platforms. 

There really aren't very many file I/O codepaths in wabt, and I imagine we want to be able to handle large files in all of them.

aydgelee:
Updated.

- Switched MSVC large-file file I/O to macro-based redirection:
  - stat -> _stat64
  - fseek -> _fseeki64
  - ftell -> _ftelli64
- Kept the main ReadFile path shared across platforms.
- Kept the MSVC-only size_t bounds check and chunked fread for 2GB+ inputs.
- Moved the MSVC stack reserve into wabt_executable().
- Changed the stack reserve to /STACK:8388608.
- Non-MSVC behavior is unchanged.

Windows x64 verification:

- Commit: 180b6bc7
- wasm2wat.exe --version: 1.0.41
- wat2wasm.exe --version: 1.0.41
- unity_temp.wat: 2,978,116,509 bytes
- roundtrip.wasm: 92,039,608 bytes
- wat2wasm exit code: 0

No "value too large", no "unable to read file", and no STATUS_STACK_OVERFLOW with the 8MB stack reserve.

CMakeLists.txt
aydgelee: Done. I moved the MSVC stack reserve into `wabt_executable` so it applies consistently to all WABT executables, and changed it to `/STACK:8388608`.

Verified on Windows x64 with a 2,978,116,509-byte WAT file. `wat2wasm` completed with exit code 0 and generated a 92,039,608-byte wasm output.

src/stream.cc
```diff
@@ -28,6 +33,39 @@
...
+      return Result::Error;
+    }
+    offset += bytes_written;
+  }
+
+  return Result::Ok;
+}
+
+static int SeekFile(FILE* file, size_t offset) {
+  if (offset > static_cast<size_t>(std::numeric_limits<int64_t>::max())) {
```
sbc100, line 61: If this `SeekFile` really needed?   

We are never going be seeing outside of `int64_t` range right?  If we did that would fail on all platforms, no just windows.

In fact, fseek on non-windows is limited to `long` which can be 32-bit on all systems right?

Maybe just add this check for `data.size()` <= LONG_MAX`  before the call to `fseek` rather than adding a separate function.

sbc100:
Why do we need to read and write the files in chunks on windows?  But not on other platforms?   Could you update the PR description to include that maybe?

sbc100:
Are you seeing these error on 64-bit windows too?  Or is this mostly a fix for 32-bit windows?

PR #2752, opened 2026-06-01
[wasi-interp] fix integer overflow in wasi getMemPtr size calculation

aizu-m:
`getMemPtr` computes the access size as `num_elems * sizeof(T)` in `size_t` (interp-wasi.cc:608), which truncates on 32-bit targets so an oversized guest iovec count passes `IsValidAccess` and the caller then reads out of bounds. Widen the multiplication to 64 bits to match the `IsValidAccess` signature.

### DIFF
--- a/src/interp/interp-wasi.cc
+++ b/src/interp/interp-wasi.cc
@@ -605,12 +605,12 @@ class WasiInstance {
                    uint32_t num_elems,
                    T** result,
                    Trap::Ptr* trap) {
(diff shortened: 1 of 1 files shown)


sbc100:
lgtm, although I'm not sure `interp-wasi.cc` has any actual users. Perhaps we should delete it?

aizu-m:
No strong opinion on deleting it. If it does stick around the widening is basically free, so probably worth landing either way. Happy to close this if you decide to drop the file.

sbc100:
Out of interest, are you using this wasi-interp yourself?

PR #2754, opened 2026-06-02
Avoid unbounded alloca in FloatParser::ParseFloat

aizu-m:
ParseFloat sizes alloca() by the literal length, so a multi-megabyte float constant in untrusted .wat text overflows the stack; use a heap buffer instead.

### DIFF
--- a/src/literal.cc
+++ b/src/literal.cc
@@ -31,6 +31,7 @@
 #include <cstring>
 #include <limits>
 #include <type_traits>
+#include <vector>
 
 #if defined(__APPLE__)
 #include <xlocale.h>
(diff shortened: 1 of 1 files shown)


sbc100 (approved):

aizu-m:
Rebased onto main to clear the BEHIND state. The branch had drifted six commits behind since the approval, so the queued merge was sat on the up-to-date check. CI is now running on the new head (95e3712), and the change itself is unchanged from your approval (clean rebase, no conflicts in literal.cc).

One thing worth flagging: updating the branch looks to have dropped the auto-merge that was enabled, so it will likely need re-enabling once the checks go green.

src/literal.cc
```diff
@@ -182,14 +183,14 @@ Result FloatParser<T>::ParseFloat(const char* s,
   // so remove them first.
   assert(s <= end);
   const size_t kBufferSize = end - s + 1;  // +1 for \0.
-  char* buffer = static_cast<char*>(alloca(kBufferSize));
-  auto buffer_end =
-      std::copy_if(s, end, buffer, [](char c) -> bool { return c != '_'; });
-  assert(buffer_end < buffer + kBufferSize);
+  std::vector<char> buffer(kBufferSize);
```
sbc100, line 186: Since IIUC that size doesn't change should we just use `std::array` here?

aizu-m: The size isn't fixed here. kBufferSize is `end - s + 1`, the length of the literal token, so it's a runtime value that varies per call. The `k` prefix reads like a compile-time constant but it isn't one. std::array would need a constant size, and pinning it to a fixed maximum would just move the buffer back onto the stack, which is the overflow this change removes. The token length comes straight from untrusted .wat text with no upper bound, so it has to be heap-backed. std::vector is the smallest way to do that, and it's the same std::vector<char> idiom StringPrintf already uses in string-format.h.

sbc100: Ah I see

PR #2756, opened 2026-06-06
[wasm2c] Avoid size_t overflow in wasm_rt_grow table reallocation

aizu-m:
On a 32-bit (ILP32) build, growing a funcref table in `wasm_rt_grow`:

    new_elems            = 230000000          (<= table->max_size)
    new_elems * sizeof   = 4600000000 bytes   (computed as uint64_t)
    realloc() size_t arg = 305032704 bytes    (truncated to 32-bit)

`new_elems` is a `uint64_t` but only bounded by `table->max_size`, which can be `UINT32_MAX`. The element is a multi-pointer struct, so the byte count crosses 2^32 long before `new_elems` does. The product is then narrowed to `realloc`'s `size_t` parameter, so the reallocation is far too small. The loop right after writes `new_elems` elements into it. A guest `table.grow` with a large delta on a table with a large max reaches this.

Found while reading the table path looking for the realloc-side equivalent of the recent wasi size fix. `calloc` in `wasm_rt_allocate` is safe because it overflows-checks its two operands; the grow path does the multiply itself and hands the result straight to `realloc`, which does not.

Reject the grow when the byte size would not fit in `size_t`, before the realloc. No effect on 64-bit. The shared include covers the funcref, externref and exnref variants.

### DIFF
--- a/wasm2c/wasm-rt-impl-tableops.inc
+++ b/wasm2c/wasm-rt-impl-tableops.inc
@@ -69,6 +69,9 @@ uint32_t WASM_RT_TABLE_APINAME(wasm_rt_grow)(WASM_RT_TABLE_TYPE* table,
   if ((new_elems < old_elems) || (new_elems > table->max_size)) {
     return (uint32_t)-1;
   }
+  if (new_elems > SIZE_MAX / sizeof(WASM_RT_TABLE_ELEMENT_TYPE)) {
+    return (uint32_t)-1;
+  }
   void* new_data =
       realloc(table->data, new_elems * sizeof(WASM_RT_TABLE_ELEMENT_TYPE));
   if (!new_data) {


sbc100 (approved):

sbc100:
Can we write a test for this somehow?

aizu-m:
The awkward part is that the truncation only happens on an ILP32 target. On LP64 the guard can't fire: new_elems is already clamped to max_size (<= UINT32_MAX) by the check right above, and SIZE_MAX / sizeof(elem) is ~2^60 there, so new_elems * sizeof never wraps. With no 32-bit job in the matrix, a 64-bit run-wasm2c test of a huge grow would just hit a giant realloc that fails and returns -1, which never touches this guard. Same reason #2752 landed without one.

It does reproduce directly against the real table API when built -m32:

```c
#include "wasm-rt.h"
#include <assert.h>
#include <stdint.h>
#include <string.h>

int main(void) {
  wasm_rt_funcref_table_t t;
  wasm_rt_funcref_t init;
  memset(&init, 0, sizeof init);

  /* table with no declared max -> max_size = UINT32_MAX */
  wasm_rt_allocate_funcref_table(&t, 0, UINT32_MAX);

  /* new_elems = 230000000, * sizeof(funcref) = 4.6e9 bytes,
     truncated to 305032704 when narrowed to realloc's size_t on ILP32 */
  uint32_t r = wasm_rt_grow_funcref_table(&t, 230000000u, init);

  /* before: short buffer, the init loop writes 230000000 elems past it
             -> heap-buffer-overflow (ASan aborts here).
     after:  grow is rejected. */
  assert(r == (uint32_t)-1);

  wasm_rt_free_funcref_table(&t);
  return 0;
}
```

Under ASan on a -m32 build that aborts before the patch and passes after. On LP64 it just allocates ~4.6GB and succeeds, so it's a no-op there. Happy to wire it up as a 32-bit-guarded check if you want it in the tree, though it won't run under the current CI.

sbc100:
I think we can land as is.

PR #2757, opened 2026-06-09
parse decimal float literals independent of the C locale

aizu-m:
Noticed while embedding libwabt in a tool that had called setlocale(LC_ALL, "") on a German machine: a valid float literal would not parse.

    LC_NUMERIC=de_DE.UTF-8
    ParseFloat(LiteralType::Float, "1.5")  ->  Result::Error

`FloatParser::Strto` hands the token to `strtof`/`strtod`, which read the radix character from the current locale's `LC_NUMERIC`. The wat grammar always writes the radix as `.`, so under a locale whose decimal point is `,` strtod stops at the `.` and the literal is rejected. The caller is free to change the locale; `c-writer.cc` already hand-rolls C-locale ctype helpers for the same reason.

Normalise `.` to the locale's decimal point before the `strto*` call. The `.` case keeps the original path, so overflow and underflow behaviour is unchanged, and hex floats (`ParseHex`) were never affected.

The test parses `1.5` under a comma-decimal locale, and skips when none is installed.

### DIFF
--- a/src/literal.cc
+++ b/src/literal.cc
@@ -14,6 +14,11 @@
  * limitations under the License.
  */
 
+// strtof_l/strtod_l are only declared by glibc and musl when _GNU_SOURCE is set.
+#if !defined(_WIN32) && !defined(_GNU_SOURCE)
+#define _GNU_SOURCE
+#endif
+
 #include "wabt/literal.h"
 
 #include <cassert>
@@ -23,11 +28,53 @@
 #include <cstdlib>
 #include <cstring>
 #include <limits>
+#include <locale.h>
 #include <type_traits>
+
+#if defined(__APPLE__)
+#include <xlocale.h>
+#endif
 
 namespace wabt {
 
 namespace {
+
+// strtof/strtod read the radix character from the current C locale's LC_NUMERIC
+// category, but the wat grammar always writes the radix as '.'.  Under a locale
+// whose decimal point is not '.' (for example ',' in the German locale) strtod
+// stops at the '.', so a valid literal like "1.5" fails to parse.  The embedder
+// is free to set the locale, so parse against a fixed "C" locale instead.
+
+#if defined(_WIN32)
+using CLocale = _locale_t;
+CLocale MakeCLocale() {
+  return _create_locale(LC_ALL, "C");
+}
+float StrtofC(const char* s, char** endptr, CLocale loc) {
+  return _strtof_l(s, endptr, loc);
+}
+double StrtodC(const char* s, char** endptr, CLocale loc) {
+  return _strtod_l(s, endptr, loc);
+}
+#else
+using CLocale = locale_t;
+CLocale MakeCLocale() {
+  return newlocale(LC_ALL_MASK, "C", nullptr);
+}
+float StrtofC(const char* s, char** endptr, CLocale loc) {
+  return strtof_l(s, endptr, loc);
+}
+double StrtodC(const char* s, char** endptr, CLocale loc) {
+  return strtod_l(s, endptr, loc);
+}
+#endif
+
+// The "C" locale never changes, so one handle is shared by every parse.  It is
+// created on first use and intentionally lives for the rest of the process.
+CLocale CLocaleHandle() {
+  static CLocale loc = MakeCLocale();
+  return loc;
+}
 
 template <typename T>
 struct FloatTraitsBase {};
@@ -43,7 +90,9 @@ struct FloatTraitsBase<float> {
   static constexpr float kHugeVal = HUGE_VALF;
   static constexpr int kMaxHexBufferSize = WABT_MAX_FLOAT_HEX;
 
-  static float Strto(const char* s, char** endptr) { return strtof(s, endptr); }
+  static float Strto(const char* s, char** endptr) {
+    return StrtofC(s, endptr, CLocaleHandle());
+  }
 };
 
 template <>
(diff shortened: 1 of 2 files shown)


sbc100:
Can we not find a way to configure the existing code to ignore the locale?

aizu-m:
Makes sense, that's cleaner. Reworked it that way: Strto now parses against a fixed "C" locale instead of the global one, via strtof_l/strtod_l (and _strtof_l/_strtod_l on MSVC) with a locale_t created once on first use. The embedder's LC_NUMERIC never feeds the parse anymore, so the '.' radix is honoured whatever they've set. The radix-rewriting helper is gone.

Verified the same way I hit the bug: under de_DE the "1.5" regression test now passes, and the full unittest suite stays at 131/131. The hex and integer paths were never affected.

sbc100:
What about just doing `setlocale(LC_NUMERIC, "C");` on startup?   I think that might be enough to make the existing code work.

aizu-m:
For wabt's own binaries that would be enough, but they are already unaffected: a process starts in the "C" locale and none of the tools call setlocale, so the existing code only breaks when something else changes the locale. That something is the embedder. Where I hit this, libwabt was linked into a host app that had called setlocale(LC_ALL, "") for its own UI formatting. There is no wabt startup to hook in that case, and the library calling setlocale(LC_NUMERIC, "C") itself would flip the host's printf/scanf formatting behind its back; it is process-global and not thread-safe either. The _l variants confine the "C" parse to literal.cc and leave the embedder's locale alone, which is why I went that way.

src/literal.cc
```diff
@@ -23,12 +28,54 @@
...
+}
+double StrtodC(const char* s, char** endptr, CLocale loc) {
+  return strtod_l(s, endptr, loc);
+}
+#endif
+
+// The "C" locale never changes, so one handle is shared by every parse.  It is
+// created on first use and intentionally lives for the rest of the process.
+CLocale CLocaleHandle() {
+  static CLocale loc = MakeCLocale();
```
sbc100: Can you just do `static CLocale c_locale = MakeCLocale();` at the top level, avoiding the need for this function?

src/literal.cc
```diff
@@ -23,12 +28,54 @@
...
+CLocale MakeCLocale() {
+  return newlocale(LC_ALL_MASK, "C", nullptr);
+}
+float StrtofC(const char* s, char** endptr, CLocale loc) {
+  return strtof_l(s, endptr, loc);
+}
+double StrtodC(const char* s, char** endptr, CLocale loc) {
+  return strtod_l(s, endptr, loc);
+}
+#endif
```
sbc100, line 51: Instead of adding these function can you do:

```
#ifdef _WIN32
#define strtof_l _strtof_l
#define strtod_l _strtod_l
#endif
```

src/literal.cc
aizu-m: Done, it's a top-level static now.

src/literal.cc
aizu-m: Done. The two defines map the MSVC names and the wrapper functions are gone; Strto calls strtof_l/strtod_l directly with the shared handle. Suite still 131/131, locale test included.

src/literal.cc
```diff
@@ -23,12 +28,41 @@
...
+
 namespace wabt {
 
 namespace {
 
+// strtof/strtod read the radix character from the current C locale's LC_NUMERIC
+// category, but the wat grammar always writes the radix as '.'.  Under a locale
+// whose decimal point is not '.' (for example ',' in the German locale) strtod
+// stops at the '.', so a valid literal like "1.5" fails to parse.  The embedder
+// is free to set the locale, so parse against a fixed "C" locale instead.
```
sbc100: How about a shorter comment like:

```
Always use the C locale for parsing floats since the Wat grammar requires `.` for the radix.
```

src/literal.cc
```diff
@@ -23,12 +28,41 @@
...
+
+#if defined(_WIN32)
+#define strtof_l _strtof_l
+#define strtod_l _strtod_l
+using CLocale = _locale_t;
```
sbc100: `#define locale_t _locale_t`

This avoid the new CLocale type.

src/literal.cc
```diff
@@ -23,12 +28,41 @@
...
+#else
+using CLocale = locale_t;
+CLocale MakeCLocale() {
+  return newlocale(LC_ALL_MASK, "C", nullptr);
+}
+#endif
+
+// The "C" locale never changes, so one handle is shared by every parse and
+// intentionally lives for the rest of the process.
+static CLocale c_locale = MakeCLocale();
```
sbc100: How about just:

```
#if defined(_WIN32)
static _locale_t c_locale = _create_locale(LC_ALL, "C");
#else
static locale_t c_locale = newlocale(LC_ALL_MASK, "C", nullptr);
#endif
```

This avoids needing a MakeCLocale function

sbc100 (approved):
lgtm, with couple of final cleanups

src/literal.cc
aizu-m: Done, swapped in your wording.

src/literal.cc
aizu-m: The CLocale alias is gone. With your initialiser snippet below the two branches just use _locale_t/locale_t directly, so the extra define wasn't needed either.

src/literal.cc
aizu-m: Done, took your snippet as-is. MakeCLocale and the alias are both gone; the block is now four lines plus the comment. Rebuilt and the suite is still 131/131.

sbc100 (approved):

sbc100:
Looks like you just need to fix the lint errors (`git clang-format origin/main`) now.

Thanks!

aizu-m:
Done. git clang-format wrapped the long _GNU_SOURCE comment and moved the locale.h include up; suite still 131/131 locally.

sbc100 (approved):

PR #2760, opened 2026-06-17
[interp] Use u64 table operands so out-of-range table64 indices trap

aizu-m:
On a table declared with the i64 index type, an index past the end of the table does not trap as it should:

    $ wasm-interp --enable-memory64 --run-all-exports t.wasm
    get_oob() => funcref:0

The table has 10 slots and the index was `(i64.const 0x1_0000_0003)`. That is out of bounds, but it read slot 3.

table.get/set/fill/copy/init/grow pop their operands as u64 for an i64 table (PopPtr), then pass them to Table::Get/Set/Fill/Copy/Init/Grow, which take u32. The high half is dropped, so any index in [2^32, 2^32 + size) wraps to an in-range slot and the bounds check passes. table.set overwrites a live entry that way, and table.grow truncates a delta above 2^32 and reports success instead of failing.

Memory already takes u64 throughout for memory64, so this widens the Table side to match and rejects a grow delta that does not fit in u32. Behaviour for i32 tables is byte-identical. Added an interp test covering the wrap window.

### DIFF
--- a/include/wabt/interp/interp.h
+++ b/include/wabt/interp/interp.h
@@ -827,26 +827,26 @@ class Table : public Extern {
 
   Result Match(Store&, const ImportType&, Trap::Ptr* out_trap) override;
 
-  bool IsValidRange(u32 offset, u32 size) const;
+  bool IsValidRange(u64 offset, u64 size) const;
 
-  Result Get(u32 offset, Ref* out) const;
-  Result Set(Store&, u32 offset, Ref);
-  Result Grow(Store&, u32 count, Ref);
-  Result Fill(Store&, u32 offset, Ref, u32 size);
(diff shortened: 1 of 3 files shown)


sbc100 (approved):

sbc100:
I wonder if there are new tests unstream in the testsuite repo that cover this case.

Sadly I think upgrade the testsuite at this point is tricky because don't yet support Wasm GC.

aizu-m:
Agreed. That GC dependency is exactly why I didn't try to sync the suite here. Bumping the in-tree testsuite would pull in the GC tests wabt can't parse yet, so I kept the regression in test/interp instead, scoped to just the wrap window. Easy to swap for the upstream cases whenever the suite can be upgraded.

PR #2762, opened 2026-06-19
Avoid Var::name() on function-level branches in decompiler

aizu-m:
Found while fuzzing wasm-decompile.

    Assertion failed: (is_name()), function name, file ir.h, line 82.

A branch whose target is the function's implicit outermost label is left as an index by the name resolution passes, since there is no block label to attach to it. DecompileExpr then called Var::name() on that index var. ir.h keeps index_ and name_ in a union, so this reads the inactive std::string member; with NDEBUG the assert is gone and it dereferences a garbage pointer instead.

Smallest repro, decompiles to a return:

    (func (block (loop (br 2))))

br_if and br_table with a function-level target reach the same path. Routed the three branch sites through a helper that emits return when the target is unnamed.

### DIFF
--- a/src/decompiler.cc
+++ b/src/decompiler.cc
@@ -220,6 +220,17 @@ struct Decompiler {
     return name[0] == '$' ? name.substr(1) : name;
   }
 
+  // A branch that targets the function's implicit outermost label is left as
+  // an index by the name-resolution passes (there is no block label to use),
+  // so it must not be routed through Var::name(). Such a branch leaves the
+  // function, i.e. it is a return.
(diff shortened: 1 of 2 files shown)


sbc100:
I think maybe we should should delete the decompiler since its not maintained at all, and I'm not sure it has any users either.

sbc100 (approved):

aizu-m:
No objection from me. It does look unmaintained, and there's a fair bit of this class of bug still lurking in there. This at least stops the crash for anyone still using it in the meantime.

PR #2764, opened 2026-06-20
Fix operand arity of v128.storeN_lane in GetExprArity

aizu-m:
Found this fuzzing wasm-decompile under a debug build: a valid module whose only instruction is v128.store8_lane aborts on the value-stack assert in decompiler-ast.h Construct.

1. GetExprArity groups SimdLoadLane and SimdStoreLane together and reports both as {2, 1}.
2. store_lane pops the address and the vector and pushes nothing, so its real effect is {2, 0}; the spare result throws value_stack_depth out by one.

The decompiler has since been removed, but GetExprArity still drives the folded wat writer, where the phantom result is equally observable: wasm2wat --fold-exprs folds the store into the next value consumer and strands the value beneath it. Split the case so SimdStoreLane returns {2, 0} and SimdLoadLane keeps {2, 1}. A reduced repro is added under test/roundtrip.


test/decompile/simd-store-lane.txt
```diff
@@ -0,0 +1,32 @@
+;;; TOOL: run-wasm-decompile
+(module
+  (memory 1)
+
+  ;; store_lane consumes the address and the vector and returns nothing, so the
+  ;; first local.get must survive as a separate return value rather than being
+  ;; folded into the store.
+  (func (param i32 v128) (result i32 i32)
+    local.get 0
```
zherczeg: I would prefer an f32 arg for stack value testing purposes (i32 is too common as a result), but the patch is ok this way.

zherczeg (approved):
LGTM

test/decompile/simd-store-lane.txt
aizu-m: Good point. Switched the surviving value to an f32 arg so it stands out from the i32 address. The store still consumes the i32/v128 pair and the float drops out as its own return value.

aizu-m:
any update?

zherczeg:
The patch is ok for me. @sbc100 what do you think?

test/decompile/simd-store-lane.txt
```diff
@@ -0,0 +1,32 @@
+;;; TOOL: run-wasm-decompile
```
sbc100: wasm-decompile was removed from wabt so I guess this test needs to be changed in some way?

sbc100 (commented):
The change itself lgtm.

test/decompile/simd-store-lane.txt
aizu-m: Right, rebased past the decompiler removal and reworked the test. GetExprArity still drives wasm2wat --fold-exprs, and the phantom result is observable there too: the writer folds the store into the next consumer and strands the value beneath it. Replaced the decompile test with test/roundtrip/fold-simd-store-lane.txt (kept the f32 stack value). It fails with the shared {2, 1} case and passes with the split one; the rest of test/roundtrip still passes locally.

sbc100 (approved):

sbc100 (approved):

PR #2765, opened 2026-06-20
Avoid front() on an empty element expression in the validator

aizu-m:
the validator picks each element expression's diagnostic location with elem_expr.front().loc, which dereferences a null list head when the element expression is empty (an init expr that is only the trailing end byte, which the binary reader accepts), so fall back to the segment location and let the existing constant-expression check reject it with the same type-mismatch error already produced for an empty global or data init expr; reachable from untrusted .wasm via wasm-validate, wasm2wat and wasm-decompile, which all validate by default.


zherczeg (approved):
LGTM

PR #2766, opened 2026-06-21
Avoid front() on empty element expression in ElemSegment::GetFlags

aizu-m:
Turned up running `wasm2wat --no-check` over hand-built binaries in a debug build.

The input is a module with one element segment holding a single empty element expression (the init expr is just the `end` byte):

    Assertion failed: (!empty()), function front, file intrusive-list.h, line 394.
    frame #4: wabt::intrusive_list<wabt::Expr>::front() const at intrusive-list.h:394
    frame #5: wabt::ElemSegment::GetFlags(...) const at ir.cc
    frame #6: wabt::WatWriter::WriteElemSegment(...) at wat-writer.cc:1605

The binary reader accepts an empty element expression, so `elem_exprs` can hold an empty `ExprList`. `GetFlags` calls `front()` on each expression to spot the `ref.func` shortcut before picking the encoding, and on an empty list that reads a null head pointer. A release build reads out of bounds instead of asserting. The flag helper is shared by the wat and binary writers, and `--no-check` skips validation, so nothing rejects the segment earlier.

`NameResolver::VisitElemSegment` already guards the same read with `elem_expr.size() == 1`. Used the same check here. A `ref.func` element expression is always one instruction, so output for valid modules does not change.

Regression test added under test/binary.

### DIFF
--- a/test/binary/elem-expr-empty-nocheck.txt
+++ b/test/binary/elem-expr-empty-nocheck.txt
@@ -1,0 +1,14 @@
+;;; TOOL: run-gen-wasm-nocheck
+magic
+version
+section(ELEM) {
+  count[1]
+  flags[5]
+  funcref
+  count[1]
+  end
+}
+(;; STDOUT ;;;
+(module
+  (elem (;0;) funcref))
+;;; STDOUT ;;)
--- a/test/run-tests.py
+++ b/test/run-tests.py
@@ -112,6 +112,11 @@ TOOLS = {
         ('ERROR', '1'),
         ('VERBOSE-ARGS', ['--print-cmd', '-v']),
     ],
+    'run-gen-wasm-nocheck': [
+        ('RUN', '%(gen_wasm_py)s %(in_file)s -o %(temp_file)s.wasm'),
+        ('RUN', '%(wasm2wat)s --no-check %(temp_file)s.wasm'),
+        ('VERBOSE-ARGS', ['--print-cmd', '-v']),
+    ],
     'run-gen-wasm-interp': [
         ('RUN', '%(gen_wasm_py)s %(in_file)s -o %(temp_file)s.wasm'),
         ('RUN', '%(wasm-interp)s --run-all-exports %(temp_file)s.wasm'),
--- a/src/ir.cc
+++ b/src/ir.cc
@@ -734,7 +734,8 @@ uint8_t ElemSegment::GetFlags(const Module* module,
         elem_type == Type::FuncRef &&
         std::all_of(elem_exprs.begin(), elem_exprs.end(),
                     [](const ExprList& elem_expr) {
-                      return elem_expr.front().type() == ExprType::RefFunc;
+                      return elem_expr.size() == 1 &&
+                             elem_expr.front().type() == ExprType::RefFunc;
                     });
 
     if (!all_ref_func) {


test/binary/elem-expr-empty-nocheck.txt
```diff
@@ -0,0 +1,14 @@
+;;; TOOL: run-gen-wasm-nocheck
+magic
+version
+section(ELEM) {
+  count[1]
+  flags[5]
+  funcref
+  count[1]
+  0x0b
```
zherczeg: Why `end` cannot be used here?

test/run-tests.py
```diff
@@ -112,6 +112,11 @@
         ('ERROR', '1'),
         ('VERBOSE-ARGS', ['--print-cmd', '-v']),
     ],
+    'run-gen-wasm-nocheck': [
```
zherczeg, line 115: What is the problem with `run-gen-wasm-bad`? This file should not be valid.

test/binary/elem-expr-empty-nocheck.txt
aizu-m: No reason, just an oversight. Switched it to `end`. gen-wasm maps `end` to 0x0b so the bytes come out identical.

test/run-tests.py
aizu-m: The file is invalid, agreed. The snag is that run-gen-wasm-bad runs wasm2wat with checks on, and the crash this fixes lives in the writer (GetFlags), past validation.

Traced it on a debug build. wasm-validate on this module aborts inside the validator's own front() on the empty expr, which is the separate site in #2765. wasm2wat without --no-check hits that same validator abort before it ever reaches the writer. Only wasm2wat --no-check skips the checks and gets as far as GetFlags, where this front() sits.

So the bad driver can't reach the code this touches: it stops (today, crashes) at validation. That's why the test needs --no-check, and there's no existing gen-wasm plus wasm2wat --no-check driver to reuse. On the unpatched tree the new test aborts mid-write at `(elem (;0;)`; with the guard it prints the whole module.

zherczeg (approved):
LGTM

test/run-tests.py
sbc100: I think I might be misunderstanding here, but I think its reasonable to the IR code to assume valid inputs.  I'm not crashes that require `--no-check` to reproduce I really worth going after.  Maybe better to deprecate / remove the `--no-check` flag rather than replicating a subset of its checks further into the codebase.

sbc100: Can I ask what your use case is for requiring `--no-check` here?

zherczeg: I suspect the test is not malformed, but not valid as well.
https://webassembly.github.io/spec/core/binary/modules.html#element-section
https://webassembly.github.io/spec/core/binary/instructions.html#binary-expr
`expr = in*`, so no instruction is a non-malformed expression, but also not a valid initializer.

aizu-m: No production use case behind it. I was running wasm2wat --no-check over hand-built and untrusted .wasm in a debug build, watching for crashes in the writer, and this was one of them. So the use case is really just feeding the tools untrusted input, nothing specific.

Fair point on the layer, though. If --no-check means the caller vouches for the input, then guarding inside the IR is the wrong place, and dropping --no-check would cover this plus the sibling sites in one go rather than scattering size checks deeper in. Happy to close this and leave it to that if that's the direction you'd rather take.

sbc100: Sounds good.  How about we try instead to remove `--no-check` completely.   Probably there are some users of it out there, maybe we can start but putting a big warning on it?

aizu-m: Makes sense. And zherczeg's right that the segment is well-formed but not a valid initialiser, so it only gets through with checks off. A warning on `--no-check` is a good first step, and removing it later fences this plus the sibling writer paths in one place instead of scattering size checks through the IR. I'll close this.

aizu-m:
Closing per the thread. The --no-check warning/removal direction is the cleaner fix; guarding the individual read in the IR only chases one symptom.

PR #2767, opened 2026-06-22
Support 64 bit limits for tables

zherczeg:


src/shared-validator.cc
```diff
@@ -134,7 +134,8 @@ Result SharedValidator::OnTable(const Location& loc,
   if (tables_.size() > 0 && !options_.features.reference_types_enabled()) {
     result |= PrintError(loc, "only one table allowed");
   }
-  result |= CheckLimits(loc, limits, UINT32_MAX, "elems");
+  uint64_t absolute_max = limits.is_64 ? UINT64_MAX : UINT32_MAX;
+  result |= CheckLimits(loc, limits, absolute_max, "elems");
```
zherczeg, line 138: Alternative: `if (!limits.is_64) { .... }` since limits is 64 bit.

src/binary-reader.cc
```diff
@@ -644,15 +642,29 @@ Result BinaryReader::ReadTable(Limits* out_elem_limits) {
...
+    out_elem_limits->max = max;
+  } else {
+    uint64_t initial;
+    uint64_t max = 0;
+    CHECK_RESULT(ReadU64Leb128(&initial, "table initial elem count"));
+    if (has_max) {
+      CHECK_RESULT(ReadU64Leb128(&max, "table max elem count"));
+    }
+    out_elem_limits->initial = initial;
+    out_elem_limits->max = max;
```
sbc100: Can we just always read these as 64-bit values and leave it to validation to check them?

If not then at least I think we can limit this if/else to just the 4 lines that do that `ReadUXXLeb128`.

zherczeg: How the code would it look like? The `ReadU32Leb128` uses a pointer to return its value, and we cannot pass the address of an uint64_t value to it.

I would not delegate this check to validation (too late for me), but we could read a 64 bit value and throw an error for `!is_64 && >UINT32_MAX` values. I am not sure we could make the code much nicer though.

zherczeg:
Created another code variant. The new `ReadU32OrU64Leb128` function could be used elsewhere, it could even replace ReadU32Leb128/ReadU64Leb128.

src/binary-reader.cc
sbc100: Can we just always read U64 in this case?

zherczeg: Always? For 32 bit tables, I think a `>UINT32_MAX` value is malformed.

sbc100: But won't the validator then take care of that?

zherczeg: It is too late for that. See these tests:
https://github.com/WebAssembly/testsuite/blob/d76759e746f3564a03f6106ae19679742f2a1831/table.wast#L27

These must fail even if validation is disabled, since they are malformed. If I read the limits as U64, they will pass.
E.g. `(table 5 1 funcref)` with --no-check should work (and it does), but `(table 0 0x1_0000_0000 funcref)` should not.

src/binary-reader.cc
```diff
@@ -350,6 +353,18 @@ Result BinaryReader::ReadU64Leb128(uint64_t* out_value, const char* desc) {
...
+Result BinaryReader::ReadU32OrU64Leb128(uint64_t* out_value,
+                                        bool is_64,
+                                        const char* desc) {
+  const uint8_t* p = state_.data + state_.offset;
+  const uint8_t* end = state_.data + read_end_;
+  size_t bytes_read = wabt::ReadU64Leb128(p, end, out_value);
+  ERROR_UNLESS(bytes_read > 0 && (is_64 || *out_value <= UINT32_MAX),
+               "unable to read u%s leb128: %s", is_64 ? "64" : "32", desc);
+  state_.offset += bytes_read;
+  return Result::Ok;
```
sbc100: Can you just call `ReadU32Leb128` / `ReadU64Leb128` here?

e.g. `return is_64 ? ReadU64Leb128(out_value, desc) : ReadU32Leb128(out_value, desc)`

zherczeg: Changed the code. The `ReadU32Leb128` requires `uint32_t` pointer, so the number of lines does not dropped.

src/binary-reader.cc
```diff
@@ -350,6 +353,18 @@ Result BinaryReader::ReadU64Leb128(uint64_t* out_value, const char* desc) {
...
 
+Result BinaryReader::ReadU32OrU64Leb128(uint64_t* out_value,
+                                        bool is_64,
+                                        const char* desc) {
+  if (is_64) {
+    return ReadU64Leb128(out_value, desc);
+  }
+  uint32_t value = 0;
+  Result result = ReadU32Leb128(&value, desc);
+  *out_value = value;
```
zherczeg: I could add a `if (result == Result::Ok)` here

sbc100 (approved):

PR #2768, opened 2026-06-22
[wasm2c] Write float constants independent of the host locale

aizu-m:
Noticed while running wasm2c with the embedder's locale set to de_DE, where LC_NUMERIC makes the radix ','. The constant writer produced:

    instance->w2c_g0 = 1,5;

snprintf("%.9g"/"%.17g") in CWriter::Write(const Const&) takes the radix from the active locale, but the emitted C must use '.'. The comma is either a compile error or, in argument/initialiser position, a comma expression that quietly drops the fractional part. The f64 path is worse: the following strchr(buf, '.') no longer finds the radix, so it also appends ".0" and writes 1,5.0.

The .wat writer sidesteps this by emitting hex floats, and the parsing direction was already made locale-independent in literal.cc. This is the output side. The fix formats as before, then rewrites the locale's radix back to '.', so the generated C is byte-identical regardless of the locale the caller set. Output under the C locale is unchanged.

Added a unit test that writes a module with f32/f64 constants under a comma locale and checks the result uses '.'. It skips when no comma locale is installed, matching the existing parser test.

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -18,6 +18,8 @@
 
 #include <cctype>
 #include <cinttypes>
+#include <clocale>
+#include <cstring>
 #include <iterator>
 #include <limits>
 #include <map>
@@ -765,6 +767,28 @@ static char internal_toupper(uint8_t ch) {
   return (ch >= 'a' && ch <= 'z') ? (ch - 'a' + 'A') : ch;
 }
 
+// printf-family conversions render the radix character according to the
+// current LC_NUMERIC locale, but the generated C source must always use '.'.
+// Rewrite the locale's radix back to '.' so the emitted float constants don't
+// depend on a locale the caller might have changed (e.g. de_DE, where it is
+// ',').
+static void NormalizeFloatRadix(char* buffer) {
+  const char* point = localeconv()->decimal_point;
+  if (point == nullptr || point[0] == '\0' ||
+      (point[0] == '.' && point[1] == '\0')) {
+    return;
+  }
+  char* found = strstr(buffer, point);
+  if (found == nullptr) {
+    return;
+  }
+  *found = '.';
+  size_t point_len = strlen(point);
+  if (point_len > 1) {
+    memmove(found + 1, found + point_len, strlen(found + point_len) + 1);
+  }
+}
+
 // static
 std::string CWriter::Mangle(std::string_view name, bool double_underscores) {
   /*
@@ -1318,7 +1342,10 @@ void CWriter::Write(const Const& const_) {
         // Negative zero. Special-cased so it isn't written as -0 below.
         Writef("-0.f");
       } else {
-        Writef("%.9g", Bitcast<float>(f32_bits));
+        char buf[128];
+        snprintf(buf, sizeof(buf), "%.9g", Bitcast<float>(f32_bits));
+        NormalizeFloatRadix(buf);
+        Writef("%s", buf);
       }
       break;
     }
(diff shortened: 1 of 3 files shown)


src/c-writer.cc
```diff
@@ -1318,7 +1343,10 @@ void CWriter::Write(const Const& const_) {
         // Negative zero. Special-cased so it isn't written as -0 below.
         Writef("-0.f");
       } else {
-        Writef("%.9g", Bitcast<float>(f32_bits));
+        char buf[128];
+        snprintf(buf, sizeof(buf), "%.9g", Bitcast<float>(f32_bits));
```
sbc100, line 1346: Can we instead use `sprintf_l` like we already use the `_l` flavors of other libc functions?

aizu-m: Switched to the c_locale approach from literal.cc. One snag: glibc and musl only ship the strto*_l family, not the printf side, so there's no snprintf_l/sprintf_l there. On POSIX I format through the same C locale_t via uselocale; Windows uses _snprintf_l. That drops the radix-rewriting helper and keeps the C-locale output byte-identical.

keithw:
No objection to this, but wondering if users might prefer to do #2389 . The benefit is we really don't have to think about floating-point text representations at all, but the downside is it makes the generated C code a little less comprehensible (which is why we never merged it the first time). @aizu-m what do you think?

src/c-writer.cc
```diff
@@ -765,6 +777,31 @@ static char internal_toupper(uint8_t ch) {
...
+static int snprintf_c_locale(char* buf,
+                             size_t size,
+                             const char* format,
+                             double value) {
+#if defined(_WIN32)
+  return _snprintf_l(buf, size, format, c_locale, value);
+#else
+  locale_t prev = uselocale(c_locale);
+  int result = snprintf(buf, size, format, value);
+  uselocale(prev);
```
sbc100: I think sadly this is not thread safe, right?

So if we can't have `snprintf_l` then maybe we have to do with the original approach.

aizu-m: Makes sense. uselocale is per-thread so it wouldn't race other threads, but it still swaps thread state, and glibc/musl don't ship the _l printf side anyway. Reverted to the NormalizeFloatRadix version: format with snprintf, then rewrite localeconv()->decimal_point back to '.'. No locale state touched, and the C-locale output stays byte-identical. Test still passes under de_DE.

aizu-m:
@keithw my vote is to land this as the targeted fix. It's a few lines and the emitted constants stay readable. #2389 is the more complete answer since it drops decimal text entirely, but it bloats the output and reads less clearly, which is why it stalled last time. The two don't really conflict: this stops the comma-locale miscompile now, and #2389 can still land later on its own merits if the readability tradeoff is acceptable. Happy to follow your call though.

src/c-writer.cc
sbc100: Oh, TIL that uselocale is per-thread.

src/test-c-writer.cc
```diff
@@ -0,0 +1,103 @@
+/*
+ * Copyright 2024 WebAssembly Community Group participants
```
sbc100: 2026

src/c-writer.cc
```diff
@@ -18,6 +18,8 @@
 
 #include <cctype>
 #include <cinttypes>
+#include <clocale>
```
sbc100, line 21: Maybe not needed anymore?

sbc100 (approved):

src/test-c-writer.cc
aizu-m: Done, bumped to 2026.

src/c-writer.cc
aizu-m: Still needed, I'm afraid. NormalizeFloatRadix calls localeconv(), which is declared in <clocale>, so I left the include in.

sbc100 (approved):

sbc100:
Landing this for now.  We can always land  #2389 later if we decide that is the way to go.

PR #2769, opened 2026-06-22
Remove wasm-decompile

sbc100:
This tools was unmainatined and acting mostly active as source of fuzzer bugs.


zherczeg:
I never touched this tool. It seems it uses some custom syntax. If it is unlikely to be used by people, it is a good idea to remove it.

dschuff (approved):

PR #2771, opened 2026-06-23
[binary-reader] Templatize LEB reading functions. NFC

sbc100:


src/binary-reader.cc
```diff
@@ -332,40 +336,32 @@ Result BinaryReader::ReadV128(v128* out_value, const char* desc) {
   return ReadT(out_value, "v128", desc);
 }
 
-Result BinaryReader::ReadU32Leb128(uint32_t* out_value, const char* desc) {
+template <typename T, size_t (*ReadFn)(const uint8_t*, const uint8_t*, T*)>
+Result BinaryReader::ReadLeb128(T* out_value,
+                                const char* desc,
+                                const char* type_name) {
```
zherczeg, line 342: Could we encode this as template arg?

zherczeg:
I like this idea!

src/binary-reader.cc
sbc100: I think that would mean defining separate static consts like:

```
    constexpr char kU32Name[] = "u32";
    constexpr char kU64Name[] = "u64";
    constexpr char kI32Name[] = "i32";
    constexpr char kI64Name[] = "i64";
```

Which I think makes the code a little less readable.  Is there some big advantage to doing that?

zherczeg: What about two args: `bool signed` (or `char signed`) and `int bits`?
The advantage of templates is pre-computing consts. 
If you prefer we can leave the code this way.

sbc100: How about we land this change and you can propose a following you like?

zherczeg: Ok.

zherczeg (approved):
LGTM

PR #2772, opened 2026-06-23
Update from C++17 to C++20

sbc100:
Binaryen did this a while back in
https://github.com/WebAssembly/binaryen/pull/8218.

There we a couple of issue related to building on macOS 13.2.1 but I don't think we need to support building on such old versions of macOS here.  We still support deploying to macOS 10.14 via CMAKE_OSX_DEPLOYMENT_TARGET.

Use std::string::starts_with and std::string_view::starts_with in various places now that C++20 is available.

Also, adopt `.contains()` for std::map, std::set, and their unordered counterparts to replace verbose find/end and count checks.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,7 @@ project(WABT LANGUAGES C CXX VERSION 1.0.41)
 
 include(GNUInstallDirs)
 
-set(CMAKE_CXX_STANDARD 17)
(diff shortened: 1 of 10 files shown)


zherczeg (approved):
LGTM.
We need to support old compilers, so maintenance will be a bit harder for us.

PR #2773, opened 2026-06-23
Use C++20 standard bit utilities and std::bit_cast. NFC

sbc100:
- Replace custom Clz, Ctz, and Popcount implementations in config.h.in with standard templates using std::countl_zero, std::countr_zero, and std::popcount from <bit>.
- Redefine wabt::Bitcast to use std::bit_cast.
- This removes a significant amount of compiler-specific boilerplate.


zherczeg:
Question: it does not seem the general library needs these functions. Are these intrerpreter specific? If yes, we should move these to there.

sbc100:
> Question: it does not seem the general library needs these functions. Are these intrerpreter specific? If yes, we should move these to there.

I had the same thought but it seems at least Clz is used in literal.cc.    Maybe we could split them up be it seem like a good idea to keep them together.

sbc100:
OK, I ended up just inlining these calls.. since they were all operating on unsigned types anyway.

zherczeg (approved):
LGTM. I like this code variant.

PR #2774, opened 2026-06-23
Add -Wno-gnu-folding-constant when compiling c api examples

sbc100:
I'm not sure why this starting showing up as a failure on the macos-latest build.  I'm guess it was some kind of system updated that give us a newer version of clang.

I'm also not sure why `-std=gnu11` wasn't enough to enable this extension.


zherczeg (approved):
LGTM

PR #2775, opened 2026-06-24
[cpp20] Modernize binary reader and writer APIs to use std::span

sbc100:
Replace legacy raw pointer and size pairs with std::span<const uint8_t>
(aliased to `ByteSpan`) across WABT's binary reader interfaces, writer
streams, parser callbacks, and public entry points.

This greatly improves type safety, simplifies call sites (allowing vectors
and arrays to be passed directly), and aligns WABT with modern C++20
standards.

Highlights:
- Fixes a performance bug in Stream::WriteData where passing std::vector
  accidentally triggered a full copy-by-value of the vector's contents.
  The template overload now correctly accepts std::span (zero-copy).
- Simplifies various writing call sites (e.g. in binary-writer, c-writer,
  and wasm-strip) by passing vectors directly or leveraging subspan.

### DIFF
--- a/fuzzers/read_binary_interp_fuzzer.cc
+++ b/fuzzers/read_binary_interp_fuzzer.cc
@@ -40,7 +40,6 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
   // validator's counts out of step).
   wabt::ReadBinaryOptions options(features, nullptr, false, true, false);
   std::vector<uint8_t> text = data_provider.ConsumeRemainingBytes<uint8_t>();
-  ReadBinaryInterp("<fuzzer>", text.data(), text.size(), options, &errors,
-                   &module);
+  ReadBinaryInterp("<fuzzer>", text, options, &errors, &module);
   return 0;
 }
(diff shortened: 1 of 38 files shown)


zherczeg (approved):
LGTM.
This will be harder for me to follow, but I will figure out something.

sbc100:
> LGTM. This will be harder for me to follow, but I will figure out something.

Why does it make things harder?   Maybe its not worth it?

zherczeg:
Unfortunately we need to support older systems, even where std::string_view is not available :) With some custom headers and `sed` commands I am able to minimize the manual source code changes after an update. These might work if I can get a custom std::span implementation. Don't worry, I will be able to solve these things in some way.

sbc100:
Oh I didn't realize you have those limitations.  What version of C++ are you stuck on?

zherczeg:
C++ 14

sbc100:
> C++ 14

That is a unfortunate.  Do you know when you might be able to update?

zherczeg:
Don't know, it is not my decision. Anyway, I will figure something out to maintain backwards compatiblity.

zherczeg (approved):
Still LGTM

PR #2776, opened 2026-06-24
Remove ReadLeb128 type argument

zherczeg:


src/binary-reader.cc
```diff
@@ -339,24 +339,25 @@ Result BinaryReader::ReadV128(v128* out_value, const char* desc) {
...
+template <typename T,
+          char prefix,
+          size_t (*ReadFn)(const uint8_t*, const uint8_t*, T*)>
+Result BinaryReader::ReadLeb128(T* out_value, const char* desc) {
   const uint8_t* p = state_.data + state_.offset;
   const uint8_t* end = state_.data + read_end_;
   size_t bytes_read = ReadFn(p, end, out_value);
-  ERROR_UNLESS(bytes_read > 0, "unable to read %s leb128: %s", type_name, desc);
+  ERROR_UNLESS(bytes_read > 0, "unable to read %c%d leb128: %s", prefix,
+               static_cast<int>(sizeof(T) * 8), desc);
```
sbc100: Can you skip the cast here and use %zu in the format string?

zherczeg: Updated

sbc100 (approved):

PR #2778, opened 2026-06-25
[interp] Stop on first error in the binary reader fuzzer

aizu-m:
The interp binary reader is only meant to read valid modules. Every `ReadBinaryInterp` caller runs with `stop_on_first_error=true` except `read_binary_interp_fuzzer.cc`, which passed `false`.

In collect-all-errors mode the `SharedValidator` keeps counting declarations the reader has already rejected. For example a function whose type index is out of range: `OnFunction` skips the append, but the validator still counts it. So the reader's `func_types_`/`module_.funcs` vectors and the validator's counts drift apart, and later callbacks index past the end (`BeginFunctionBody`, `OnExport`, `OnReturnCallExpr`).

Per the review, rather than adding reader-side guards for a mode the interp reader was never meant to support, this stops the fuzzer on the first error like every other caller, so the reader is no longer driven over invalid modules.

### DIFF
--- a/fuzzers/read_binary_interp_fuzzer.cc
+++ b/fuzzers/read_binary_interp_fuzzer.cc
@@ -33,8 +33,12 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
   if (data_provider.ConsumeBool()) { features.enable_##variable(); }
 #include "wabt/feature.def"
 #undef WABT_FEATURE
-  // Add only feature related options, but no logging, stop_on_first_error, etc.
-  wabt::ReadBinaryOptions options(features, nullptr, false, false, false);
+  // Add only feature related options, but no logging etc. The interp reader is
+  // only meant to read valid modules, so stop on the first error instead of
+  // continuing in collect-all-errors mode (the validator keeps counting
+  // declarations the reader has already rejected, leaving its vectors and the
+  // validator's counts out of step).
+  wabt::ReadBinaryOptions options(features, nullptr, false, true, false);
   std::vector<uint8_t> text = data_provider.ConsumeRemainingBytes<uint8_t>();
   ReadBinaryInterp("<fuzzer>", text.data(), text.size(), options, &errors,
                    &module);


src/interp/binary-reader-interp.cc
```diff
@@ -1246,6 +1269,10 @@ Result BinaryReaderInterp::OnReturnCallExpr(Index func_index) {
   CHECK_RESULT(
       validator_.OnReturnCall(GetLocation(), Var(func_index, GetLocation())));
 
+  if (func_index >= func_types_.size()) {
+    PrintError("invalid function index: %" PRIindex, func_index);
+    return Result::Error;
```
sbc100: Why is this needed for `OnReturnCallExpr` but not `OnCallExpr`?

sbc100 (approved):

src/test-interp.cc
```diff
@@ -734,3 +734,48 @@ TEST_F(InterpGCTest, Collect_DeepRecursion) {
 // TODO: Test for Thread keeping references alive as locals/params/stack values.
 // This requires better tracking of references than currently exists in the
 // interpreter. (see TODOs in Select/LocalGet/GlobalGet)
+
+TEST_F(InterpTest, FunctionSectionTypeOutOfRange) {
+  // The function section declares a function whose type index is out of range,
+  // followed by a code section with a matching body. When reading in
+  // collect-all-errors mode (stop_on_first_error=false), OnFunction rejects the
```
zherczeg: Does this officially supported? As far as I remember this part is a testing utility, and never intended to support invalid code. It would require a lot of changes in the code, increasing its complexity heavily. 

See: https://github.com/WebAssembly/wabt/blob/main/src/interp/binary-reader-interp.cc#L526

Supporting invalid wasm is useful for text<->binary conversions, but not really for executing it.

src/interp/binary-reader-interp.cc
aizu-m: OnCallExpr never dereferences func_types_. It just emits the opcode with the raw func_index, so a stale index can't read out of bounds there. OnReturnCallExpr is different: it reads func_types_[func_index] to work out the drop/keep counts, and the same index feeds GetFuncOffset, so the stale index actually gets dereferenced. That's why only this one needed the guard.

src/test-interp.cc
aizu-m: Fair point, the interp reader isn't meant to execute invalid wasm, and I'm not trying to make it do that. None of these three sites are on the execution path. They're hit purely while reading, via read_binary_interp_fuzzer.cc, which sets stop_on_first_error=false. That's how I found them: fuzzing the reader in that mode under ASan.

The sequence is: OnFunction rejects the out-of-range type and skips the append, but the reader carries on, so the validator's func count and the reader's func_types_/module_.funcs are now out of step. The next callback then indexes past the end. BeginElemSegment and BeginDataSegment already bounds-check that exact mismatch, so this just adds the same guard at the three sites that were missing it, no new recovery machinery.

If you'd rather the harness set stop_on_first_error=true so the reader bails on the first error instead, that's fine by me too.

zherczeg: I understand the problem. I think we should only support `stop_on_first_error=true`, since continuing the parsing does not add any user value to the interpreter (other than make it complex). But lets wait @sbc100 's opinion.

aizu-m: Fine by me. The harness side is a one-liner: read_binary_interp_fuzzer.cc passes stop_on_first_error=false today, flip it to true and the parse bails in OnFunction the moment the validator rejects the type index, well before BeginFunctionBody/OnExport/OnReturnCallExpr ever run with the stale index. If that's the direction you'd both prefer, I'll drop the three reader-side guards and just change the harness. Happy to follow whatever you and @sbc100 settle on.

sbc100: Yes, maybe lets remove the `stop_on_first_error=false` feature completely.

aizu-m: Done. Dropped the three reader-side guards and the two tests, and flipped read_binary_interp_fuzzer.cc to stop_on_first_error=true. That was the only caller driving the interp reader in collect-all-errors mode, so the OOB sites are no longer reachable from it.

On removing it completely: objdump still sets stop_on_first_error=false (with fail_on_custom_section_error=false) so it can dump as much of a corrupt module as possible, so the flag and the recovery path in binary-reader.cc need to stay for that. The interp reader's only use of it is now gone. Happy to scope a wider removal as a separate change if you want, but that one would touch objdump's behaviour.

sbc100: I see, yes I suppose its useful of objdump

sbc100 (approved):

src/test-interp.cc
zherczeg: Input reading and validation are two different steps, and the core specification tests has `assert_malformed` and `assert_invalid` tests. So for those tools, which just reads and writes raw webassebly (but does not interpret it in any way), supporting well-formed but invalid webassembly should be supported. For example: IR, wat2wasm wasm2wat or objdump. On the contrary, the c writer should not, since it "interprets" the code. Some tools even try to support malformed webassembly, but I don't know whether it is useful.

PR #2780, opened 2026-06-29
[validator] Allow call_indirect/return_call_indirect table elements to be subtypes of funcref

### LINKED ISSUE #2523: Missing Table Type Validation in call_indirect Instruction
`wasm-interp` fails to validate the type of the table used in the `call_indirect` instruction. According to the [WebAssembly specification](https://webassembly.github.io/spec/core/valid/instructions.html#valid-call-indirect), the table associated with `call_indirect` must have a `funcref` type.

### Test case

```wasm
(module
  (type (func (param i32 i32) (result i32)))

  (func $func1 (type 0) (param i32 i32) (result i32)
    local.get 0
    local.get 1
    i32.add)

  (func $func2 (type 0) (param i32 i32) (result i32)
    local.get 0
    local.get 1
    i32.sub)

  (table 2 externref)
  (elem (i32.const 0) externref (ref.null extern) (ref.null extern))

  (func $main (result i32)
    i32.const 0x12341234
    i32.const 0x1234
    i32.const 1
    call_indirect (type 0)
    )

  (export "main" (func $main))
)
```
[program.wasm.txt](https://github.com/user-attachments/files/18267273/program.wasm.txt)

### Environment

* OS: Ubuntu 20.04
* CPU: amd64
* WABT version: 1.0.36
* Commands:
```bash
./wasm-interp --enable-all --run-export=main ./program.wasm
```

### Actual behavior

The output from `wasm-interp` is:

```
main() => error: uninitialized table element
```

### Expected behavior

The runtime should detect that the table type (`externref`) is invalid for the `call_indirect` instruction and raise a validation error. The expected error message should be similar to:

```
Validation error: call_indirect is only valid when a table has type funcref
```

6d6f6d656e746f:
This PR resolves an issue where the validator rejects valid WebAssembly modules using typed function references (such as `(ref null $sig)`) with `call_indirect` and `return_call_indirect`. 
Currently, `SharedValidator` performs a strict equality check (`table_type.element != Type::FuncRef`), which rejects valid subtypes of `funcref` in indirect calls.
**Changes:**
- Replaced the strict equality check in `src/shared-validator.cc` with `typechecker_.CheckType` to correctly check for subtypes of `funcref`.
- Handled out-of-range table index checks to maintain compatibility with existing spec test outputs.

Fixes: #2523

### DIFF
--- a/src/shared-validator.cc
+++ b/src/shared-validator.cc
@@ -856,7 +856,8 @@ Result SharedValidator::OnCallIndirect(const Location& loc,
   TableType table_type;
   result |= CheckFuncTypeIndex(sig_var, &func_type);
   result |= CheckTableIndex(table_var, &table_type);
-  if (table_type.element != Type::FuncRef) {
+  if (table_type.element == Type::Any ||
+      Failed(typechecker_.CheckType(table_type.element, Type::FuncRef))) {
     result |= PrintError(
         loc,
         "type mismatch: call_indirect must reference table of funcref type");
@@ -1221,7 +1222,8 @@ Result SharedValidator::OnReturnCallIndirect(const Location& loc,
   TableType table_type;
   result |= CheckFuncTypeIndex(sig_var, &func_type);
   result |= CheckTableIndex(table_var, &table_type);
-  if (table_type.element != Type::FuncRef) {
+  if (table_type.element == Type::Any ||
+      Failed(typechecker_.CheckType(table_type.element, Type::FuncRef))) {
     result |= PrintError(loc,
                          "type mismatch: return_call_indirect must reference "
                          "table of funcref type");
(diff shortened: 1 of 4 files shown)


zherczeg:
The referenced issue used `(table 2 externref)`, which was invalid for call_indirect. Isn't this patch is the opposite of it? GC patches probably affect this problem.

6d6f6d656e746f:
Thanks for the reply,
The current check (table_type.element != Type::FuncRef) rejects valid tables containing typed function references that are subtypes of funcref like (ref null $sig).

This patch replaces it with a subtype check, which still rejects externref because its not a subtype of funcref and now correctly allows typed function reference tables.

Please let me know if I'm missing something

zherczeg:
Tests should be added to show which matches / don't match to the new check.

zherczeg (approved):
LGTM

Issue #2779, opened 2026-06-28
Unexpected opcode: 0xfd


Honya2000:
Hello,

Today i tried to decompile my compiled wasm binary and got error:
02d6ac0: error: unexpected opcode: 0xfd 0x85 0x2

sbc100:
What command are you running?   Could you include the full command line and perhaps attach the binary.
 
It looks like `0xfd` + `0x85` is the I16X8NarrowI32X4S simd instruction which it looks that wabt does support.

sbc100:
https://github.com/WebAssembly/wabt/blob/23b62e9665fa96fa6d94e9e08adb3bc5ce1f3a34/include/wabt/opcode.def#L409

Honya2000:
Hmm. not sure why it complains then. I just pulled latest wabt and recompiled it...
 I guess the issue appeared after i added Halide pipeline to my project. Will investigate in wabt under debugger and let you know.

P.S. Didn't use any command line. Just: 
wasm2wat my.wasm > out.wat

Honya2000:
Hmm, just compiled the same binary without debug info, and now it gives another opcode error:
0020b38: error: unexpected opcode: 0xfd 0x80 0x2

Honya2000:
<img width="818" height="242" alt="Image" src="https://github.com/user-attachments/assets/9da344fc-5497-444d-a128-f5429339d6c6" />

This part looks suspicious. Looks like it uses dynamic libs ?

sbc100:
Can you attach the failing file?

Honya2000:
Ok, removed dynamic lib from the binary, still the same error.
Attached binary.
Executed this way:
wasm2wat tmrwNativeAI-nt-simd.wasm > tmrwNativeAI-nt-simd.wat

[tmrwNativeAI-nt-simd.zip](https://github.com/user-attachments/files/29464697/tmrwNativeAI-nt-simd.zip)

sbc100:
OK, I added `--enable-all` and to the `wasm2wat` command it got a lot further.

The new error is `b93575: error: unexpected opcode: 0xfd 0x3f`, which looks like it should be I32X4GeS.

The error message is a actually misleading though.  It seems like the bad op-code is from FP16 SIMD (Half-Precision SIMD), which wabt does not support.

Are you trying use the FP16 proposal?

Honya2000:
I'm using Halide to process float and uint8_t buffers. It shouldn't use fp16 theoretically...
[...]

PR #2781, opened 2026-06-29
Fix DecodeInvalidOpcode for 9-bit opcode codes

sbc100:
When MAX_OPCODE_BITS was increased to 9, DecodeInvalidOpcode still used a hardcoded 0xff mask when unpacking the invalid opcode code. This caused any invalid opcode with a code >= 0x100 (such as unsupported SIMD opcodes) to be truncated to 8 bits when reported in error messages.

Use ((1 << MAX_OPCODE_BITS) - 1) instead of 0xff so that 9-bit opcode codes are correctly preserved when reporting invalid opcodes.

See #2779


zherczeg (approved):
LGTM

PR #2782, opened 2026-06-29
wasm2c: uvwasi adapter layer for wasm2c output

shravanrn:
Implementation of a uvwasi adapter for wasm2c output

Last discussed in a comment with @sbc100 here a while back
https://github.com/WebAssembly/wabt/pull/2395#discussion_r1653120242

wasm2c can now use uvwasi for the wasi implementation by including `wasm-rt-uvwasi-adapter.c` (and of course linking libuv ad libuvwasi)

I have added an example of using this adapter in the examples folder.

**Full disclosure**: While the initial adapter was written by me on an old fork of wasm2c, in a pre-LLM era, I used an LLM to adapt that patch for this PR and the current wasm2c codebase. 

**LLM Mitigations**: I audited every line of the generated code. I fixed several double-fetch bugs, edge cases on assuming null termination and off-by-ones. I also fixed more minor stylistic things to keep the generated code similar to the wasm2c codebase. To the extent possible, I believe my audit has made the code about as secure as it would be if I hand wrote it from scratch.

### DIFF
--- a/wasm2c/examples/hello-wasi/.gitignore
+++ b/wasm2c/examples/hello-wasi/.gitignore
@@ -1,0 +1,5 @@
+*.o
+hello-wasi
+hello-wasi.wasm
+hello-wasi-wasm.*
+test.out
--- a/wasm2c/wasm-rt-uvwasi-adapter.c
+++ b/wasm2c/wasm-rt-uvwasi-adapter.c
@@ -1,0 +1,926 @@
+/*
+ * Copyright 2018 WebAssembly Community Group participants
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "wasm-rt-uvwasi-adapter.h"
+
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+#ifndef WASM_RT_CORE_TYPES_DEFINED
+#define WASM_RT_CORE_TYPES_DEFINED
+typedef uint8_t u8;
+typedef int8_t s8;
+typedef uint16_t u16;
+typedef int16_t s16;
+typedef uint32_t u32;
+typedef int32_t s32;
+typedef uint64_t u64;
+typedef int64_t s64;
+typedef float f32;
+typedef double f64;
+#endif
+
+#define WASM2C_UVWASI_MAX_IOVS 128
+#define TRAP(x) (wasm_rt_trap(WASM_RT_TRAP_##x), 0)
+
+#if WABT_BIG_ENDIAN
+#define MEM_ADDR(mem, addr, n) ((mem)->data_end - (addr) - (n))
+#else
+#define MEM_ADDR(mem, addr, n) &((mem)->data[addr])
+#endif
+
+void wasm2c_uvwasi_link(struct w2c_wasi__snapshot__preview1* wasi,
+                        wasm_rt_memory_t* memory,
+                        uvwasi_t* uvwasi) {
+  wasi->memory = memory;
+  wasi->uvwasi = uvwasi;
+}
+
+static inline uint64_t checked_add_u64(uint64_t a, uint64_t b) {
+  uint64_t ret;
+#if __has_builtin(__builtin_add_overflow)
+  if (__builtin_add_overflow(a, b, &ret))
+    TRAP(OOB);
+#elif defined(_MSC_VER)
+  if (_addcarry_u64(0, a, b, &ret))
+    TRAP(OOB);
+#else
+  ret = a + b;
+  if (ret < a)
+    TRAP(OOB);
+#endif
+  return ret;
+}
+
+#define RANGE_CHECK(mem, offset, len)            \
+  do {                                           \
+    uint64_t res = checked_add_u64(offset, len); \
+    if (UNLIKELY(res > (mem)->size))             \
+      TRAP(OOB);                                 \
+  } while (0);
+
+static inline void* mem_get_checked_pointer(
+    struct w2c_wasi__snapshot__preview1* wasi,
+    u32 addr,
+    u64 len) {
+  RANGE_CHECK(wasi->memory, addr, len);
+  return MEM_ADDR(wasi->memory, addr, len);
+}
+
+static inline void memory_fill(struct w2c_wasi__snapshot__preview1* wasi,
+                               u32 addr,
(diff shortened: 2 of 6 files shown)


shravanrn:
@sbc100 Just checking if there is interest in landing this, before i spend time rebasing or cleanups

sbc100:
I'm not against landing this, I'm just a little stretched on review time right now.  

I can try to get to it soon.   We also have @zherczeg who is been quite active recently on the project and could maybe review ?

shravanrn:
@sbc100 thanks and no worries! I'll push a rebase soon to keep the diff/merge clean

zherczeg:
I can review it, but I don't know the context, so I need to understand the purpose of this code first. It adds several files to gitignore, two files to examples, and an adapter. Is it adds some built ins? How wasm and libuw is connected? Is it something wasm-only, or do you need wasm-host communication as well?

wasm2c/wasm-rt-uvwasi-adapter.c
```diff
@@ -0,0 +1,923 @@
...
+  }
+
+  RANGE_CHECK(wasi->memory, argv, (u64)count * 4);
+  char* buf_base = mem_get_checked_pointer(wasi, buf, buf_size);
+  char** uv_argv = calloc(count, sizeof(char*));
+  if (!uv_argv) {
+    return UVWASI_ENOMEM;
+  }
+
+  ret = uvwasi_args_get(wasi->uvwasi, uv_argv, buf_base);
```
zherczeg, line 266: Where are these calls coming from? Which include provides them?

shravanrn:
> I can review it, but I don't know the context, so I need to understand the purpose of this code first. It adds several files to gitignore, two files to examples, and an adapter. Is it adds some built ins? How wasm and libuw is connected? Is it something wasm-only, or do you need wasm-host communication as well?

Broadly the idea here is that Wasm programs can be compiled as a "pure compute" Wasm program (target of `--target=wasm32-unknown` in Clang) or a Wasm program that uses a system interface to do things like access the file system, and perform network requests (target of `--target=wasm32-wasi` in Clang). The latter is done through a standardized interface called [WASI](https://wasi.dev/). 

For a program compiled to the WASI backend, the Wasm compiler includes a wasi-libc that provides the standard C library functions like `fopen` and implements this internally treating the WASI APIs like its system call layer. The WASI APIs themselves are expected to be supported by the Wasm runtime. There are some standard WASI API implementations like uvwasi and Wasm compilers can add WASI support simply by writing adapters to uvwasi. I think uvwasi is already used by the wasm interpreter in the wabt project to implement WASI APIs. So the sub-modules in wabt already pull in uvwasi, and the CMakeLists.txt in the wabt repo can be configured to build uvwasi by passing in the flag `-DWITH_WASI=ON`

This PR implements the uvwasi adapter for wasm2c.

wasm2c/wasm-rt-uvwasi-adapter.c
shravanrn: This comes from uvwasi.h that is included in wasm-rt-uvwasi-adapter.h

zherczeg:
Thank you for the explanation.

wasm2c/examples/hello-wasi/.gitignore
```diff
@@ -0,0 +1,5 @@
+*.o
+hello-wasi
+hello-wasi.wasm
+hello-wasi-wasm.*
+test.out
```
zherczeg, line 5: Usually tmp files goes to the `bin/` `build/` or `out/` directories when wabt is compiled. Can these files goes there as well (and no need to add them to gitignore)?

wasm2c/wasm-rt-uvwasi-adapter.h
```diff
@@ -0,0 +1,41 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+struct w2c_wasi__snapshot__preview1 {
+  wasm_rt_memory_t* memory;
+  uvwasi_t* uvwasi;
+};
+
+void wasm2c_uvwasi_link(struct w2c_wasi__snapshot__preview1* wasi,
```
zherczeg, line 33: Is this an extension to wasi interface? If not, can it be a static func?

wasm2c/wasm-rt-uvwasi-adapter.c
```diff
@@ -0,0 +1,926 @@
...
+    store_u64(wasi, pos, uvpos);
+  }
+  return ret;
+}
+
+u32 w2c_wasi__snapshot__preview1_fd_filestat_set_size(
+    struct w2c_wasi__snapshot__preview1* wasi,
+    u32 fd,
+    u64 filesize) {
+  return uvwasi_fd_filestat_set_size(wasi->uvwasi, fd, filesize);
```
zherczeg, line 418: How does this adapter works? Does the wasm file contains external references such as `w2c_wasi__snapshot__preview1_fd_filestat_set_size`? Then these references are redirected to other functions such as `uvwasi_fd_filestat_set_size`? Then, when the output `.c` file is compiled, do you need a wasi library?

wasm2c/wasm-rt-uvwasi-adapter.c
```diff
@@ -0,0 +1,926 @@
...
+}
+
+#define RANGE_CHECK(mem, offset, len)            \
+  do {                                           \
+    uint64_t res = checked_add_u64(offset, len); \
+    if (UNLIKELY(res > (mem)->size))             \
+      TRAP(OOB);                                 \
+  } while (0);
+
+static inline void* mem_get_checked_pointer(
```
zherczeg, line 77: Does wasi use a linear memory in a `.c` program? Is it better than raw pointers? Do you have multiple `w2c_wasi__snapshot__preview1` structures or is it a global value in the `.c` implementation of wasi?

zherczeg (commented):
The code looks valid. Does the patch contains all wasi 0.1 functions or it is a partial implementation of them?

shravanrn:
@zherczeg Thanks for the review. I'll address these shortly

I also spoke to @keithw offline and he pointed out that this implementation currently traps on bad parameter rather than returning system call error codes. I will need to address this as well. Moving this to a draft, and will convert back to an active PR once I address this

zherczeg:
As far as I remember trap on memory error is normal for wasi. This is even true for wasi 0.2 and component model. It is practically the same as a segmentation fault. Personally I would not change that, providing a wrong buffer is an error which should be fixed by the developers of the application.

shravanrn:
> As far as I remember trap on memory error is normal for wasi. This is even true for wasi 0.2 and component model. It is practically the same as a segmentation fault. Personally I would not change that, providing a wrong buffer is an error which should be fixed by the developers of the application.

Ah interesting. Ok, I should probably have clarified that I am not 100% sure of keith's interpretation is correct or wrong, but I wanted to go investigate before we went ahead and landed. And while discussing with keith, i remembered that there is a [wasi test suite repo](https://github.com/WebAssembly/wasi-testsuite), so I'll go check if this is something I can integration as well. I'm hoping it'll have checks on whether I am preserving the semantics of errors correctly.

(This might take me some time. I have a separate commitment that will keep me busy till 22nd. But I should have something fixed and uploaded shortly after this)

shravanrn:
Hmm... looks like there are more edge cases from the wasi tests, than I appreciated. Will likely take me into august to have a pr for this

PR #2783, opened 2026-07-03
wasm2c: (Security) fix runtime overflows when running Wasm modules with memory64

shravanrn:
As part of routine audit we found a bugs in the wasm2c runtime for modules supporting memory64. Memory32 is not affected. 

**Security bug** When a memory64 module calls `memory.grow` with a delta value that causes `new_pages * page_size` to overflow `uint64_t` to zero, `realloc(ptr, 0)` is called, which frees the linear memory buffer on glibc. `expand_data_allocation` then returns an error code as `realloc` returns NULL. The caller `grow_memory_impl` returns `-1` (indicating failure) without updating `memory->data`, leaving a dangling pointer. Any subsequent memory access from the Wasm module reads or writes freed heap memory. 

**Non security bug** the delta check for a `0` delta has a typo. So there is a code path where `new_pages` wraps to 0 and returns 0 indicating successful memory grow even though memory wasn't updated. Since this doesn't update the memory field, this is just a functional bug.

### DIFF
--- a/wasm2c/wasm-rt-mem-impl.c
+++ b/wasm2c/wasm-rt-mem-impl.c
@@ -140,6 +140,16 @@ static uint64_t get_alloc_size_for_mmap_default32(uint64_t max_pages) {
 
 #endif /* WASM_RT_USE_MMAP */
 
+static bool u64_mult_overflow(uint64_t a, uint64_t b, uint64_t* result) {
+  uint64_t x = a * b;
+  *result = x;
+  if (a != 0 && x / a != b) {
+    // overflow
+    return true;
+  }
+  return false;
+}
+
 // Include operations for memory
 #define WASM_RT_MEM_OPS
 #include "wasm-rt-mem-impl-helper.inc"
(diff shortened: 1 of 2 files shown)


wasm2c/wasm-rt-mem-impl.c
```diff
@@ -140,6 +140,16 @@ static uint64_t get_alloc_size_for_mmap_default32(uint64_t max_pages) {
 
 #endif /* WASM_RT_USE_MMAP */
 
+static bool u64_mult_overflow(uint64_t a, uint64_t b, uint64_t* result) {
```
sbc100, line 143: Is there nothing in libc or compiler-rt for this?    I guess gcc and clang have `__builtin_mul_overflow`, but maybe nothing more portable than that?

Maybe mention in a comment why can't use `__builtin_mul_overflow`?

sbc100 (approved):
Does it make sense to write a test for this?  Maybe tricky?

wasm2c/wasm-rt-mem-impl.c
shravanrn: For maximum portability, I was generally trying to avoid any compiler specific builtins where (1) they are not strictly needed (2) and are not in the performance-critical path. We could add a path to use `__builtin_mul_overflow` if available, but this feels like overkill here given that memory grow should be relatively rare. I'll plan to leave it as, but let me know if you have a difference preference

shravanrn:
> Does it make sense to write a test for this? Maybe tricky?

It's a bit tricky to write a generic test that tests functionality for such scenarios because a wasm engine will likely fail to allocate enough memory to cause these sort of overflows anwyay. I could write a very-specific test for this failure, but I don't think that that would achieve anything after this patch lands. So I think its best to leave this one as is.

PR #2784, opened 2026-07-03
wasm2c: fix errors in atomic ops implementation

shravanrn:
Fixes two bugs in the wasm2c atomics implementation
- The Wasm spec requires DEFINE_ATOMIC_RMW to return the value prior to modification.
- The Wasm spec requires DEFINE_ATOMIC_CMP_XCHG to return the value read if the compare fails.

All spec tests continue to pass after these changes.


sbc100 (approved):
I guess there is not good test coverage for multi-threading in the spec tests?

shravanrn:
> I guess there is not good test coverage for multi-threading in the spec tests?

Yup, looks like it. I just rechecked the latest test suite upstream and they aren't any changes from what we have (modulo features we don't support).

PR #2785, opened 2026-07-03
wasm2c: fix to check address + offset addition for memory64

shravanrn:
As part of routine audit we found a bugs in the wasm2c code-generation for modules supporting memory64. Memory32 is not affected. 

The wasm2c code generator for `LoadExpr` and `StoreExpr` emits `(u64)(addr) + offset` as `uint64_t` addition. This is fine for memory32 due to the `uint64_t` promotion but hides an overflow for memory64 load/store that will violate Wasm semantics (but not break security).

When the addr and offset addition overflow in Memory64 (e.g., addr=1 + offset=0xFFFFFFFFFFFFFFFF wraps to 0), the wrapped value is passed to `i32_store`/`i32_load`, where `MEMCHECK_GENERAL` -> `RANGE_CHECK` only checks the already-wrapped (small) value against `mem->size`. The overflow is undetected and the access succeeds at the wrong address instead of trapping.

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -5335,10 +5335,18 @@ void CWriter::Write(const LoadExpr& expr) {
 
   Type result_type = expr.opcode.GetResultType();
   Write(StackVar(0, result_type), " = ", func, "(",
-        ExternalInstancePtr(ModuleFieldType::Memory, memory->name), ", (u64)(",
-        StackVar(0), ")");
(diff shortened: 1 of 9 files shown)


sbc100 (approved):

shravanrn:
Just an fyi, looks like I accidentally made the commit title include a (Security) prefix. To clarify, this is not a security issue.

PR #2787, opened 2026-07-05
Bounds-check function index in name section local handlers

aizu-m:
ASAN, calling `ReadBinaryIr` with `read_debug_names=true` and `stop_on_first_error=false` on a crafted module:

    binary-reader-ir.cc:1721: AddressSanitizer: container-overflow
    READ of size 8 in BinaryReaderIR::OnLocalNameLocalCount

Found while going over the name-section handlers. `OnLocalNameLocalCount` and `OnLocalName` index `module_->funcs[func_index]` using the function index taken from the name section's local subsection. The reader only checks `function_index < NumTotalFuncs()`, i.e. the declared count `num_func_imports_ + num_function_signatures_`. With `stop_on_first_error` off, a truncated function section (declares two funcs, but the second signature LEB runs past the section end) leaves `funcs` one short of that count, so a local name for the missing function reads past the vector. `OnLocalNameLocalCount` had a debug-only assert; `OnLocalName` had no check at all.

Every sibling handler here already does the `index >= size()` check, and `OnFunctionBodyCount` already guards this same `stop_on_first_error` case for the code path. Added the matching check to both local-name handlers, plus a regression test.

### DIFF
--- a/src/binary-reader-ir.cc
+++ b/src/binary-reader-ir.cc
@@ -1717,7 +1717,10 @@ Result BinaryReaderIR::OnNameEntry(NameSectionSubsection type,
 }
 
 Result BinaryReaderIR::OnLocalNameLocalCount(Index index, Index count) {
-  assert(index < module_->funcs.size());
+  if (index >= module_->funcs.size()) {
+    PrintError("invalid function index: %" PRIindex, index);
+    return Result::Error;
+  }
   Func* func = module_->funcs[index];
   Index num_params_and_locals = func->GetNumParamsAndLocals();
   if (count > num_params_and_locals) {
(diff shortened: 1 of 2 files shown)


sbc100:
I'm not sure we want to support `stop_on_first_error` in most configurations (for example `ReadBinaryIr`) since it seems like an endless source of such issues, and means we are doing more checking that we need otherwise do on the normal code path.

See the discussion in https://github.com/WebAssembly/wabt/pull/2778

sbc100 (approved):

aizu-m:
Agreed. I went through the callers after reading the #2778 thread: every in-tree ReadBinaryIr user already runs with stop_on_first_error=true (wasm2wat sets it explicitly, the wasm2wat fuzzer and the emscripten helpers get the default). objdump is the only tool that still wants false, and it goes through its own reader, not the IR one. So dropping collect-all-errors mode from the IR path wouldn't cost any tool, just the option surface. Can put that together as a follow-up if that's the direction you'd like. Thanks for merging this one.

PR #2789, opened 2026-07-06
Zero-init an InitInst to avoid accessing unintialized memory

lexi-nadia:


sbc100 (approved):

src/binary-reader-objdump.cc
```diff
@@ -1995,7 +1995,7 @@ Result BinaryReaderObjdump::OnRefNullExpr(Type type) {
 Result BinaryReaderObjdump::OnOpcode(Opcode opcode) {
   BinaryReaderObjdumpBase::OnOpcode(opcode);
   if (ReadingInitExpr() && opcode != Opcode::End) {
-    InitInst i;
+    InitInst i = {};
```
sbc100: Maybe `InitInst i{}` instead?

lexi-nadia: Done.

PR #2790, opened 2026-07-07
Check memory64 addr+offset overflow in wasm2c store/atomic/simd ops

aizu-m:
Comparing wasm2c output either side of 4bedf7c ("check address + offset addition for memory64"):

    // load  (fixed in 4bedf7c)
    i64_load(&M0, (u64)checked_add_u64(var_j0, 16u));
    // store (still unchecked)
    i64_store(&M0, (u64)(var_j0) + 16, var_j1);

That commit only converted `CWriter::Write(const LoadExpr&)`. For an i64 memory the store still adds the static offset with a wrapping `+`, and so do the SIMD lane/splat/zero and atomic emitters.

How I hit it: memory64 module, `i64.store offset=16` at address `0xFFFFFFFFFFFFFFF0`. The sum overflows u64 to 0, the generated `RANGE_CHECK` sees an in-range address, and the write lands at guest address 0. Per the spec that access should trap; the load form already does after 4bedf7c.

Applied the same `checked_add_u64` guard to `StoreExpr`, `SimdLoadLaneExpr`, `SimdStoreLaneExpr`, `LoadSplatExpr`, `LoadZeroExpr`, `AtomicLoadExpr`, `AtomicStoreExpr`, `AtomicRmwExpr` and `AtomicRmwCmpxchgExpr` for `is_64` memories. The 32-bit output is byte for byte unchanged. The wasm2c spec tests (memory64, threads, simd) still pass.

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -5376,9 +5376,16 @@ void CWriter::Write(const StoreExpr& expr) {
   func = GetMemoryAPIString(*memory, func);
 
   Write(func, "(", ExternalInstancePtr(ModuleFieldType::Memory, memory->name),
-        ", (u64)(", StackVar(1), ")");
-  if (expr.offset != 0)
-    Write(" + ", expr.offset);
+        ", (u64)");
+  if (expr.offset == 0) {
+    Write("(", StackVar(1), ")");
+  } else if (memory->page_limits.is_64) {
+    // two u64's being added. Use checked addition
+    Write("checked_add_u64(", StackVar(1), ", ", expr.offset, "u)");
+  } else {
+    // two u32's being added and promoted to u64
+    Write("(", StackVar(1), ") + ", expr.offset);
+  }
   Write(", ", StackVar(0), ");", Newline());
   DropTypes(2);
 }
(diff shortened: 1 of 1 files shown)


src/c-writer.cc
```diff
@@ -6091,10 +6125,14 @@ void CWriter::Write(const AtomicRmwCmpxchgExpr& expr) {
...
-  if (expr.offset != 0)
-    Write(" + ", expr.offset);
+        ExternalInstancePtr(ModuleFieldType::Memory, memory->name), ", (u64)");
+  if (expr.offset == 0) {
+    Write("(", StackVar(2), ")");
+  } else if (memory->page_limits.is_64) {
+    Write("checked_add_u64(", StackVar(2), ", ", expr.offset, "u)");
+  } else {
+    Write("(", StackVar(2), ") + ", expr.offset);
+  }
```
sbc100: This code pattern seems to be repeated in many places.  Can we extract a helper function?

shravanrn (changes requested):
Good catch! I have the same comment as @sbc100 : it would be helpful if you could pull this into a helper function that can be invoked in load/store, simd load/store, atomic load/store to simplify the code.

src/c-writer.cc
aizu-m: Done. Pulled it into a WriteMemoryAddress helper that writes the whole address operand, and switched all ten emitters (load/store, simd lane load/store, load splat/zero, atomics) over to it. Net -45 lines.

aizu-m:
Pushed the helper. One note on output: the old emitters disagreed on the literal suffix for the 32-bit path (loads wrote `+ 16u`, most stores `+ 16`). The helper normalises to `u` everywhere, so 32-bit store output changes cosmetically. No checked-in test expectations contain that pattern, so nothing needed rebasing.

Re-ran the checks after the refactor: wasm2c tests 265/265, unit tests 134/134, clang-format clean. The memory64 repro (store at 0xFFFFFFFFFFFFFFF0 with offset=16) still traps instead of wrapping to guest address 0.

sbc100 (approved):

sbc100:
@shravanrn  can you take another look?

shravanrn (approved):
@aizu-m Thanks for addressing this! Looks good to me.

aizu-m:
Thanks for the reviews and the merge. The helper ended up tidier than the original anyway.

PR #2791, opened 2026-07-07
Mark Result as [[nodiscard]] and fix associated errors

lexi-nadia:
This CL fixes a fuzzer error caused by unintentionally ignoring a failure. Instead of just fixing the one case, I figured I'd fix them all.

In a handful of cases, I wasn't sure how errors should be propagated. Please feel free to suggest improvements!

### DIFF
--- a/include/wabt/type-checker.h
+++ b/include/wabt/type-checker.h
@@ -176,7 +176,7 @@ class TypeChecker {
   void PushLabel(LabelType label_type,
                  const TypeVector& param_types,
                  const TypeVector& result_types);
-  Result PopLabel();
+  void PopLabel();
   Result CheckLabelType(Label* label, LabelType label_type);
   Result Check2LabelTypes(Label* label,
                           LabelType label_type1,
--- a/src/tools/wasm2c.cc
+++ b/src/tools/wasm2c.cc
@@ -138,7 +138,7 @@ Result Wasm2cMain(Errors& errors) {
   CHECK_RESULT(GenerateNames(&module));
   /* TODO(binji): This shouldn't fail; if a name can't be applied
    * (because the index is invalid, say) it should just be skipped. */
-  ApplyNames(&module);
+  (void)ApplyNames(&module);
 
   if (!s_outfile.empty()) {
     std::string header_name_full =
--- a/src/wast-lexer.cc
+++ b/src/wast-lexer.cc
@@ -400,7 +400,8 @@ Token WastLexer::GetStringToken(TokenType token_type) {
             uint32_t scalar_value = 0;
 
             while (IsHexDigit(PeekChar())) {
-              ParseHexdigit(*cursor_++, &digit);
+              Result result = ParseHexdigit(*cursor_++, &digit);
+              assert(Succeeded(result));
 
               scalar_value = (scalar_value << 4) | digit;
               // Maximum value of a unicode code point.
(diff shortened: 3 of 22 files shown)


include/wabt/type-checker.h
```diff
@@ -176,7 +176,7 @@ class TypeChecker {
   void PushLabel(LabelType label_type,
                  const TypeVector& param_types,
                  const TypeVector& result_types);
-  Result PopLabel();
+  void PopLabel();
```
sbc100, line 179: In these places where you have remove the `Result` like this is that because these methods cannot fail?

src/tools/wasm2c.cc
```diff
@@ -138,7 +138,7 @@ Result Wasm2cMain(Errors& errors) {
   CHECK_RESULT(GenerateNames(&module));
   /* TODO(binji): This shouldn't fail; if a name can't be applied
    * (because the index is invalid, say) it should just be skipped. */
-  ApplyNames(&module);
+  (void)ApplyNames(&module);
```
sbc100, line 141: Why not `CHECK_RESULT` here?

src/wast-lexer.cc
```diff
@@ -400,7 +400,8 @@ Token WastLexer::GetStringToken(TokenType token_type) {
             uint32_t scalar_value = 0;
 
             while (IsHexDigit(PeekChar())) {
-              ParseHexdigit(*cursor_++, &digit);
+              Result result = ParseHexdigit(*cursor_++, &digit);
+              assert(Succeeded(result));
```
sbc100, line 404: I guess this logically cannot happen because of `IsHexDigit` above?

src/wat-writer.cc
```diff
@@ -1199,17 +1199,19 @@ Result WatWriter::ExprVisitorDelegate::OnLoadZeroExpr(LoadZeroExpr* expr) {
 }
 
 void WatWriter::WriteExpr(const Expr* expr) {
-  WABT_TRACE(WriteExprList);
+  WABT_TRACE(WriteExpr);
```
sbc100, line 1202: You found a type too.  Thanks for fixing.

sbc100 (commented):
lgtm!

include/wabt/type-checker.h
lexi-nadia: Yes, that's right. It's a very simple method and I can't even imagine what it would mean for it to fail.

src/tools/wasm2c.cc
lexi-nadia: Maybe I misinterpreted the comment? I thought binji was saying that a failure in ApplyNames shouldn't cause the method itself to fail. If you think I should put `CHECK_RESULT` here anyway, let me know and I'll update it!

src/wast-lexer.cc
lexi-nadia: Yes, that's my logic.

src/tools/wasm2c.cc
sbc100: Oh I see.   That makes sense yes.    I guess he means "This shouldn't be treated as a failure" rather than "This shouldn't fail"?

include/wabt/type-checker.h
sbc100: Presumably if the Label stack if empty then it would fail?  But I'm not familiar with this code right now so I don't really know.

lgtm though since there it clearly doesn't fail today.

sbc100 (approved):

PR #2792, opened 2026-07-07
Fix WABT_PRINTF_TYPE_CODE to correctly handle INT_MIN

lexi-nadia:
This fixes a fuzzer-discovered edge case: the result of std::abs(INT_MIN) is undefined behavior.

### DIFF
--- a/include/wabt/string-format.h
+++ b/include/wabt/string-format.h
@@ -29,7 +29,7 @@
 #define WABT_PRINTF_STRING_VIEW_ARG(x) \
   static_cast<int>((x).length()), (x).data()
 
-#define PRItypecode "%s%" PRIx64
+#define PRItypecode "%s%#" PRIx64
 #define WABT_PRINTF_TYPE_CODE(x) \
   (static_cast<int32_t>(x) < 0 ? "-" : ""), std::abs(static_cast<int64_t>(x))
 
--- a/include/wabt/string-format.h
+++ b/include/wabt/string-format.h
@@ -17,6 +17,8 @@
 #ifndef WABT_STRING_FORMAT_H_
 #define WABT_STRING_FORMAT_H_
 
+#include <inttypes.h>
+
 #include <cstdarg>
 #include <string>
 #include <vector>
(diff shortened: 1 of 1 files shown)


sbc100:
Don't our type codes have a very limited range though?   Can this be called with arbitrary values?

lexi-nadia:
> Don't our type codes have a very limited range though? Can this be called with arbitrary values?

https://github.com/WebAssembly/wabt/blob/f4908362bd398b31853c4754664a2ea01ff31966/src/binary-reader.cc#L2005

If this is the real problem, I'm happy to fix this instead!

include/wabt/string-format.h
```diff
@@ -28,8 +28,10 @@
   static_cast<int>((x).length()), (x).data()
 
 #define PRItypecode "%s%#x"
-#define WABT_PRINTF_TYPE_CODE(x) \
-  (static_cast<int32_t>(x) < 0 ? "-" : ""), std::abs(static_cast<int32_t>(x))
```
sbc100, line 32: How about if use `int64_t` there instead?

lexi-nadia: Done. I confirmed that this works. Note that I had to change PRItypecode slightly, which means the output format might change.

sbc100 (approved):

include/wabt/string-format.h
```diff
@@ -27,9 +27,9 @@
 #define WABT_PRINTF_STRING_VIEW_ARG(x) \
   static_cast<int>((x).length()), (x).data()
 
-#define PRItypecode "%s%#x"
+#define PRItypecode "%s%#lx"
```
sbc100: I think there is some kind of PRIi64 thing you can use here.

lexi-nadia: Done.

include/wabt/string-format.h
lexi-nadia: It turns out some tests are sensitive to the output format, which means I can guarantee it's not changing after all. :)

PR #2793, opened 2026-07-08
Add WABT_USE annotations for Result values only used in assertions

lexi-nadia:


sbc100 (approved):

sbc100:
(In future can you capitalize the first letter of the PR title?)

PR #2794, opened 2026-07-09
[interp] Add bounds checks in func/table/memory/global/tag handlers

lexi-nadia:
This fixes a case of undefined behavior caught by fuzzing, and defends against other similar possible errors.

I don't know how to convert a fuzz test case into regression tests, but I'm happy to add it if someone can point me at an example!

### DIFF
--- a/src/interp/binary-reader-interp.cc
+++ b/src/interp/binary-reader-interp.cc
@@ -549,6 +549,9 @@ Result BinaryReaderInterp::OnImportFunc(Index import_index,
                                         Index sig_index) {
   CHECK_RESULT(
       validator_.OnFunction(GetLocation(), Var(sig_index, GetLocation())));
+  if (sig_index >= module_.func_types.size()) {
+    return Result::Error;
+  }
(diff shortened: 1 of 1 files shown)


src/interp/binary-reader-interp.cc
```diff
@@ -549,6 +549,9 @@ Result BinaryReaderInterp::OnImportFunc(Index import_index,
                                         Index sig_index) {
   CHECK_RESULT(
       validator_.OnFunction(GetLocation(), Var(sig_index, GetLocation())));
+  if (sig_index >= module_.func_types.size()) {
+    return Result::Error;
+  }
```
sbc100, line 554: I'm curious.. does hitting this (and the other errors here) require disabling the validator?  If so I think we should instead remove the ability for the fuzzers to skip validation.    I think its reasonable to more of the code in wabt to assume valid inputs only (and use the validator and single point of protection against such).

In I'm right then these could maybe be assertions instead? 

If I'm wrong then this change lgtm :)

PR #2796, opened 2026-07-12
Fix out-of-bounds write in MakeTypeBindingReverseMapping

aizu-m:
ASan, release build, `wasm2wat` on a crafted module:

    ==…==ERROR: AddressSanitizer: heap-buffer-overflow
      #0 wabt::MakeTypeBindingReverseMapping(...) ir.cc
      #1 wabt::ApplyNames(wabt::Module*)
      #2 ProgramMain
    ... located 1535 bytes after 24-byte region

Found while going over the name-section handlers. The local subsection stores a (local index, name) pair per entry. `OnLocalNameLocalCount` checks the entry count against the function's param+local count, but the per-entry local index in `OnLocalName` is never range-checked before it becomes a `Binding`. `MakeTypeBindingReverseMapping` sizes the reverse map to that local count and writes each name at `binding.index`, guarded only by an `assert`. A name whose index sits past the function's locals writes off the end of the vector: a debug build trips the assert, a release build corrupts the heap. Reached from an untrusted `.wasm` through `wasm2wat` and `wasm2c` (both call `ApplyNames` and read debug names by default), on an otherwise valid module.

Skip entries that land outside the mapping. Valid modules only name real locals, so their output does not change. The added test names local index 5 of a one-param function; before the fix `wasm2wat` aborts, after it prints the module with the stray name dropped.


test/binary/bad-name-section-local-index.txt
```diff
@@ -0,0 +1,34 @@
+;;; TOOL: run-gen-wasm
+;; The name section's local subsection can carry a local index that is out of
```
sbc100, line 2: Should we give the test the `bad-` prefix, since the name section in this case contains bad data?

How about `bad-name-section-local-index` or `bad-local-index-in-name-section` ?

aizu-m: Agreed, the entry is bad data so the bad- prefix fits. Renamed it to bad-name-section-local-index.txt and pushed.

sbc100 (approved):

PR #2797, opened 2026-07-13
Avoid out-of-bounds imports read in wasm_instance_new

aizu-m:
ASan, instantiating a 2-import module through the wasm C API with a 1-element imports vector:

    ==ERROR: AddressSanitizer: heap-buffer-overflow
    READ of size 8 ... in wasm_instance_new interp-wasm-c-api.cc:746
    0 bytes after 8-byte region [...] (the caller's imports array)

Spotted while running modules through libwasm with a fixed imports list. `wasm_instance_new` copies the imports into a `RefVec` by looping `i` over the module's declared import count and reading `imports->data[i]`. Nothing checks the caller actually supplied that many, so a module whose import section declares more imports than the embedder passes reads past the end of the imports array and then derefs the wild pointer through `->I->self()`.

`Instance::Instantiate` already rejects a short list with a "not enough imports!" trap, but it runs after this loop, so the over-read happens first. Bounding the loop by `imports->size` lets the short case fall through to that existing trap instead. Valid and over-supplied lists build the same `RefVec` as before.

### DIFF
--- a/src/interp/interp-wasm-c-api.cc
+++ b/src/interp/interp-wasm-c-api.cc
@@ -742,7 +742,11 @@ own wasm_instance_t* wasm_instance_new(wasm_store_t* store,
   assert(store);
 
   RefVec import_refs;
-  for (size_t i = 0; i < module->As<Module>()->import_types().size(); i++) {
+  // Bound the read by imports->size so a module that declares more imports
+  // than the caller supplied cannot walk off the end of the imports array;
+  // the short list is then rejected by Instance::Instantiate below.
+  size_t import_count = module->As<Module>()->import_types().size();
+  for (size_t i = 0; i < import_count && i < imports->size; i++) {
     import_refs.push_back(imports->data[i]->I->self());
   }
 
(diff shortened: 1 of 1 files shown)


src/interp/interp-wasm-c-api.cc
```diff
@@ -742,7 +742,11 @@ own wasm_instance_t* wasm_instance_new(wasm_store_t* store,
   assert(store);
 
   RefVec import_refs;
-  for (size_t i = 0; i < module->As<Module>()->import_types().size(); i++) {
+  // Bound the read by imports->size so a module that declares more imports
+  // than the caller supplied cannot walk off the end of the imports array;
+  // the short list is then rejected by Instance::Instantiate below.
+  size_t import_count = module->As<Module>()->import_types().size();
+  for (size_t i = 0; i < import_count && i < imports->size; i++) {
```
sbc100: IIRC that number of imports supplied should match the number expected by the module.  Maybe we can instead error out here if they are not equal?

aizu-m: Makes sense. Changed it to an explicit equality check that traps with "wrong number of imports provided" and returns null before touching the array. One behavioural note: an over-supplied list used to be silently accepted, and now errors too. Reran the ASan repro (2-import module, 1 supplied): clean trap, no over-read. All the C API examples still pass.

sbc100 (approved):

PR #2798, opened 2026-07-16
Fix out-of-bounds type stack read after rethrow in c-writer

aizu-m:
wasm2c on a valid exceptions module, ASan/debug build:

    Assertion failed: (index < type_stack_.size()), function Write, file c-writer.cc, line 1246.

Worked backwards from the StackVar read. The rethrow case in `Write(const ExprList&)` ends in `break`, so the loop carries on and emits the instructions after it. The validator has already marked those unreachable (`OnRethrow` makes the stack polymorphic), so a valid module can put operations there that pop operands which are not really on the stack. StackVar then underflows `type_stack_` and reads off the end. Release builds drop the assert and take the out-of-bounds read; `DropTypes` does an out-of-bounds erase on the same tail.

br, br_table, return, unreachable, throw, throw_ref and the return_call family all return here to stop walking the list. rethrow was the one that fell through, so it now does the same.

Repro: `(func try nop catch_all rethrow 0 i32.store end)` assembled with `--enable-exceptions`, then run through wasm2c.

### DIFF
--- a/src/test-c-writer.cc
+++ b/src/test-c-writer.cc
@@ -45,13 +45,15 @@ const uint8_t s_float_globals[] = {
     0x03, 0x00, 0x05, 0x67, 0x5f, 0x66, 0x33, 0x32, 0x03, 0x01,
 };
 
-std::string WriteCSource(const uint8_t* data, size_t size) {
+std::string WriteCSource(const uint8_t* data,
+                         size_t size,
+                         const Features& features = Features{}) {
   Errors errors;
   Module module;
   ReadBinaryOptions read_options;
+  read_options.features = features;
   EXPECT_EQ(Result::Ok,
             ReadBinaryIr("test", data, size, read_options, &errors, &module));
-  Features features;
   ValidateOptions validate_options(features);
   EXPECT_EQ(Result::Ok, ValidateModule(&module, &errors, validate_options));
   EXPECT_EQ(Result::Ok, GenerateNames(&module));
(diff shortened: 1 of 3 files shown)


src/test-c-writer.cc
```diff
@@ -101,3 +103,33 @@ TEST(CWriter, FloatConstLocaleIndependent) {
...
+      0x05, 0x03, 0x01, 0x00, 0x01,                    // memory: min 1
+      0x0a, 0x0e, 0x01, 0x0c, 0x00,                    // code: 1 body, size 12
+      0x06, 0x40,                                      // try (void)
+      0x01,                                            // nop
+      0x19,                                            // catch_all
+      0x09, 0x00,                                      // rethrow 0
+      0x36, 0x02, 0x00,  // i32.store (unreachable)
+      0x0b,              // end try
+      0x0b,              // end func
+  };
```
sbc100: Why make this a unit test rather then a normal file-based test in test/wasm2c/?

Also does this mean non of the spec tests have this case?  Perhaps we should upstream something?

aizu-m: Moved it to a file-based test: test/wasm2c/unreachable-after-rethrow.txt, run-spec-wasm2c with --enable-exceptions. Verified it aborts on the unpatched tree at c-writer.cc:1246 and passes (1/1) with the fix in place. Dropped the gtest and the Features plumbing it needed.

On the spec suite: the existing exception-handling tests all translate through wasm2c cleanly, so none of them put an operand-consuming instruction in the unreachable tail after rethrow, which is what drives the StackVar underflow. legacy/rethrow.wast is the closest but its rethrow sits at the end of the block. Agreed it's worth having upstream. I can send a case over to the testsuite repo separately and keep this local regression test in the meantime.

sbc100 (approved):

PR #2799, opened 2026-07-17
Fix memory64 offset truncation in the interp istream

aizu-m:
Running a memory64 module under wasm-interp, `i64.load offset=0x1_0000_0000` from address 0 returned 0 instead of trapping. A large dynamic address (5 GiB) does trap, so the static offset was the suspect.

1. OnLoadExpr hands the memarg offset over as an `Address` (u64), but `istream_.Emit(opcode, memidx, offset)` resolved to `Emit(Opcode, u32, u32)` and narrowed it to 32 bits.
2. For an i64-indexed memory a static offset of 2^32 or more is then lost, so an access whose effective address is out of bounds wraps to a small in-range offset and reads or writes a valid location rather than trapping.

Carried the offset as u64 the whole way: added `imm_index_offset`/`imm_index_offset_lane` immediates and `Emit(Opcode, u32, u64)` overloads, widened the four `Imm_Index_Offset_*` decode groups, the load/store/atomic/simd-lane handlers and the disassembler. 32-bit memories are untouched since the offset always fits u32. Regression test added under test/interp.

### DIFF
--- a/include/wabt/interp/istream.h
+++ b/include/wabt/interp/istream.h
@@ -94,8 +94,8 @@ struct Instr {
     } imm_index_offset;
     struct {
       u32 memidx;
-      u64 offset;
       u8 lane;
+      u64 offset;
     } imm_index_offset_lane;
   };
 };
--- a/include/wabt/interp/istream.h
+++ b/include/wabt/interp/istream.h
@@ -88,6 +88,15 @@ struct Instr {
       u32 fst, snd;
       u8 idx;
     } imm_u32x2_u8;
+    struct {
+      u32 memidx;
+      u64 offset;
+    } imm_index_offset;
+    struct {
+      u32 memidx;
+      u64 offset;
+      u8 lane;
+    } imm_index_offset_lane;
   };
 };
 
@@ -115,6 +124,8 @@ class Istream {
   void Emit(Opcode::Enum, v128);
   void Emit(Opcode::Enum, u32, u32);
   void Emit(Opcode::Enum, u32, u32, u8);
+  void Emit(Opcode::Enum, u32, u64);
+  void Emit(Opcode::Enum, u32, u64, u8);
   void EmitDropKeep(u32 drop, u32 keep);
   void EmitCatchDrop(u32 drop);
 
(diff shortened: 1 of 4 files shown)


include/wabt/interp/istream.h
```diff
@@ -88,6 +88,15 @@ struct Instr {
...
       u8 idx;
     } imm_u32x2_u8;
+    struct {
+      u32 memidx;
+      u64 offset;
+    } imm_index_offset;
+    struct {
+      u32 memidx;
+      u64 offset;
+      u8 lane;
```
zherczeg, line 97: I would move the lane to first member (or after memIndex) to reduce the structure size increase.
u32+u64+u8 = 24 bytes, while u8+u32+u64 = 16 bytes due to padding.

aizu-m: Good point. Moved lane to sit between memidx and offset, so the struct packs to 16 bytes now. Only the in-memory Instr layout changes, the serialised istream order is untouched. interp and spec suites still pass.

test/interp/memory64-offset-oob.txt
```diff
@@ -0,0 +1,25 @@
...
+(module
+  (memory i64 1)
+
+  ;; A static offset >= 2^32 must not be truncated to 32 bits: address 0 with
+  ;; offset 2^32 is out of bounds and must trap, not read a wrapped in-range
+  ;; address.
+  (func (export "load_offset_wrap") (result i64)
+    (i64.load offset=0x1_0000_0000 (i64.const 0)))
+  (func (export "store_offset_wrap")
+    (i64.store offset=0x1_0000_0000 (i64.const 0) (i64.const 1)))
```
zherczeg, line 12: I would use i32.load here just to have some variation, but not strictly necessary.

zherczeg (approved):
LGTM

test/interp/memory64-offset-oob.txt
aizu-m: Done. The load is now i32.load and the store stays i64.store, so the test covers two access widths. Expected trap message updated to 4294967296+4. interp and spec suites still pass.

zherczeg:
Please rebase this patch.

aizu-m:
Rebased on current main. No conflicts. interp and spec suites pass on the rebased branch.

zherczeg:
@sbc100 if you have no objection, I will land this patch

sbc100:
lgtm, please go ahead

PR #2801, opened 2026-07-23
[binary-reader] Fix out-of-bounds read on invalid name sub-section

aizu-m:
ASan, `wasm2wat -v` on a valid module whose name section carries one
unknown subsection (type LEB `0xffffffff`):

    binary.cc:66 GetNameSectionSubsectionName
    READ of size 8 ... 8 bytes before global 'wabt::NameSubsectionName' (size 96)

Worked back from the read to the subsection id. `NameSectionSubsection`
has no explicit underlying type, so it is a signed int. `ReadNameSection`
reads the id into a `uint32_t`, casts it, and only gates the
`OnNameSubsection` callback with `type <= Last`, a signed compare with no
lower bound. An id of `0xffffffff` casts to a negative enum, passes the
gate, and reaches `GetNameSectionSubsectionName`, which indexed
`NameSubsectionName[size_t(subsec)]` with no range check. The logging
delegate then prints the returned wild `const char*` with `%s`.

The module is valid: an unknown name subsection is skipped, so this only
shows with verbose logging (`wasm2wat -v`, `wasm-objdump -v`,
`wasm-interp -v`). Found feeding fuzz-style name sections through
`wasm2wat -v`.

Fixed in the caller: `ReadNameSection` compares the raw `name_type`
against `Last` before the cast, so an out-of-range id is skipped as an
unknown subsection and never becomes a `NameSectionSubsection`.
`GetNameSectionSubsectionName` keeps the plain index and asserts the
in-range precondition. Regression test added under
`src/test-binary-reader.cc`.

### DIFF
--- a/src/binary.cc
+++ b/src/binary.cc
@@ -63,9 +63,8 @@ const char* GetNameSectionSubsectionName(NameSectionSubsection subsec) {
   static_assert(WABT_ENUM_COUNT(NameSectionSubsection) ==
                     WABT_ARRAY_SIZE(NameSubsectionName),
                 "Malformed ExprTypeName array");
-  return size_t(subsec) < WABT_ARRAY_SIZE(NameSubsectionName)
-             ? NameSubsectionName[size_t(subsec)]
-             : "<error_name_subsection>";
+  assert(size_t(subsec) < WABT_ARRAY_SIZE(NameSubsectionName));
+  return NameSubsectionName[size_t(subsec)];
 }
 
 }  // namespace wabt
--- a/src/binary.cc
+++ b/src/binary.cc
@@ -63,7 +63,9 @@ const char* GetNameSectionSubsectionName(NameSectionSubsection subsec) {
   static_assert(WABT_ENUM_COUNT(NameSectionSubsection) ==
                     WABT_ARRAY_SIZE(NameSubsectionName),
                 "Malformed ExprTypeName array");
-  return NameSubsectionName[size_t(subsec)];
+  return size_t(subsec) < WABT_ARRAY_SIZE(NameSubsectionName)
+             ? NameSubsectionName[size_t(subsec)]
+             : "<error_name_subsection>";
 }
 
 }  // namespace wabt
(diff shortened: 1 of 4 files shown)


src/binary.cc
sbc100, line 62: Can we instead have the caller of this function (who creates the NameSectionSubsection from bytes) ensure that the subsec is valid?

Either that or have this function take an integer type rather then a NameSectionSubsection?  

Basically I think its function take a enum as an argument it should a pre-condition that the enum is in-range.

zherczeg: (In this case, an assert should still be added here)

aizu-m: Moved the check to the caller. `ReadNameSection` now compares the raw `name_type` before the cast, so an id past `Last` is skipped as an unknown subsection and never becomes a `NameSectionSubsection` in the first place (the cast itself was unspecified for those values anyway). The accessor is back to a plain index with an assert on the precondition, per @zherczeg.

Checked against the original module (subsection type LEB `0xffffffff`): the assert fires with the old reader, and `wasm2wat -v` runs clean with the new one. Kept the regression test, it still covers that path.

src/test-binary-reader.cc
```diff
@@ -222,6 +222,38 @@ TEST(BinaryReaderLogging, DeepIndentDoesNotOverflowBuffer) {
...
+  MemoryStream log_stream;
+  BinaryReaderError reader;
+  ReadBinaryOptions options(Features{}, &log_stream, /*read_debug_names=*/true,
+                            /*stop_on_first_error=*/true,
+                            /*fail_on_custom_section_error=*/false);
+  // Just ensure the out-of-range subsection type does not read off the name
+  // table when it is logged.
+  Result result = ReadBinary(data, &reader, options);
+  (void)result;
+}
```
sbc100: Can this be a normal file-based test like `name-section-location.txt`?

aizu-m: Done. Dropped the gtest and added test/binary/name-section-unknown-subsection.txt as a run-objdump-gen-wasm test. It runs objdump with --debug so the unknown subsection gets logged, which is the path that indexed off NameSubsectionName. Confirmed it fails under ASan on the old reader and passes on the new one.

sbc100 (approved):

PR #2803, opened 2026-08-01
wasm2c: Extend segue to simd and unshared atomics

shravanrn:
This is part 1 of the wasm2c performance optimization round I am conducting. Second part is [here](https://github.com/WebAssembly/wabt/pull/2804)

This part, part 1, is just minor cleanup of segue --- the major performance optimization will come in part 2.

This PR extends the use of segment registers to simd and unshared atomics. The initial implementation of segue didn't cover simd because simd everywhere didn't support taking pointers to the gs segment. However, it turns out there is an easy way to solve it --- all we need are wrappers that copy to a local variable first and then pass that to simde. The optimizations in clang are able to eliminate all the extra memcpy operations and the result is what we'd want.

### DIFF
--- a/src/prebuilt/wasm2c_atomicops_source_declarations.cc
+++ b/src/prebuilt/wasm2c_atomicops_source_declarations.cc
@@ -113,7 +113,9 @@ R"w2c_template(    ATOMIC_ALIGNMENT_CHECK(addr, t1);                            
 )w2c_template"
 R"w2c_template(    t1 result;                                                              \
 )w2c_template"
-R"w2c_template(    wasm_rt_memcpy(&result, MEM_ADDR(mem, addr, sizeof(t1)), sizeof(t1));   \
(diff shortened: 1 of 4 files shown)


sbc100:
Can you explain what all the MEM_ADDR ->  MEM_ADDR_MEMOP changes are for?

shravanrn:
> Can you explain what all the MEM_ADDR -> MEM_ADDR_MEMOP changes are for?

@sbc100 MEM_ADDR_MEMOP is a macro that uses segments to access memory if this feature is enabled, and falls back to using instance parameter's memory field to access memory otherwise. MEM_ADDR is the macro that unconditionally uses the instance parameter's memory field to access memory. 

Right now using segments was only allowed in the scalar loads/stores, but there is no reason to restrict this as the hardware supports segments on Simd operations as well. So this change ensures simd operations can use segments as well.

sbc100 (approved):
Thanks for the reminder.

PR #2805, opened 2026-08-03
Fix use-after-free resolving reference type names in wast-parser

aizu-m:
drop the deferred reference-type resolutions a module field registered before it failed to parse, otherwise ParseModuleFieldList dereferences pointers into a field that was destroyed instead of appended.

### DIFF
--- a/src/wast-parser.cc
+++ b/src/wast-parser.cc
@@ -1464,20 +1464,7 @@ Result WastParser::ParseModuleFieldList(Module* module) {
       CHECK_RESULT(ParseCustomSectionAnnotation(module));
       continue;
     }
-    // A field that fails to parse is destroyed instead of being appended to
-    // the module, so the deferred resolutions registered while parsing it
-    // point into freed memory. Drop them again.
-    size_t ref_types_size = resolve_ref_types_.size();
-    size_t type_vectors_size = resolve_type_vectors_.size();
-    size_t funcs_size = resolve_funcs_.size();
     if (Failed(ParseModuleField(module))) {
-      resolve_ref_types_.erase(resolve_ref_types_.begin() + ref_types_size,
-                               resolve_ref_types_.end());
-      resolve_type_vectors_.erase(
-          resolve_type_vectors_.begin() + type_vectors_size,
-          resolve_type_vectors_.end());
-      resolve_funcs_.erase(resolve_funcs_.begin() + funcs_size,
-                           resolve_funcs_.end());
       CHECK_RESULT(Synchronize(IsModuleField));
     }
   }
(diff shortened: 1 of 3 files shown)


sbc100:
Should we have `ParseModuleField` clean up after itself on error instead of expecting the caller to track this stuff?

aizu-m:
Agreed, that's better. Moved it into `ParseModuleField`, which now records the three resolve list sizes, dispatches, and truncates them if the field failed. `ParseModuleFieldList` is back to its original shape.

The dispatch switch moved into `ParseModuleFieldImpl` unchanged so `ParseModuleField` has a single exit point to hang the cleanup off. Writing it as `result = ParseXModuleField(module); break;` in place worked too, but clang-format then exploded all eleven cases onto three lines each, which seemed a worse trade than one small helper.

Re-checked after the change: all six registration sites (func param, func result, func local, table elem type, elem segment type, import func signature) exit 1 with only the expected parse diagnostics and no ASan report, and reverting just the truncation brings the heap-use-after-free straight back in `ResolveTargetTypeVector`. Test suite is green in Release, and green again under a Debug ASan build.

src/wast-parser.cc
```diff
@@ -1494,6 +1494,31 @@ Result WastParser::ParseModuleFieldList(Module* module) {
...
+  Result result = ParseModuleFieldImpl(module);
+
+  if (Failed(result)) {
+    resolve_ref_types_.erase(resolve_ref_types_.begin() + ref_types_size,
+                             resolve_ref_types_.end());
+    resolve_type_vectors_.erase(
+        resolve_type_vectors_.begin() + type_vectors_size,
+        resolve_type_vectors_.end());
+    resolve_funcs_.erase(resolve_funcs_.begin() + funcs_size,
+                         resolve_funcs_.end());
```
sbc100, line 1515: I'm still a little confused, if we are returning failure, when why does it matter about the state of these members?   If we fail to parse why is anybody accessing these after this point?

aizu-m: The failure doesn't propagate. `ParseModuleFieldList` swallows it:

```c++
if (Failed(ParseModuleField(module))) {
  CHECK_RESULT(Synchronize(IsModuleField));
}
```

It resynchronises to the next module field and carries on parsing, so wat2wasm reports every parse error in one pass rather than stopping at the first. The failed field's result is dropped right there.

So control still reaches the three resolve loops at the bottom of that same function, which walk `resolve_ref_types_` / `resolve_type_vectors_` / `resolve_funcs_` and dereference the pointers recorded while the field was being parsed. Those point into the field's local `unique_ptr`, which was destroyed on the error path instead of being handed to `AppendField`. That's the freed read, and a freed write via `local_types.Set()` in the func-local case.

The one case where nobody touches them is when `Synchronize` fails too, i.e. the bad field is the last thing in the file, since then we return early and never reach the loops. That's why the test has a `(memory 1)` after the broken fields: it gives Synchronize somewhere to land so the loops are actually reached.

sbc100 (approved):

PR #2806, opened 2026-08-06
Allow initializer expressions to reference defined immutable globals

### LINKED ISSUE #2526: Incorrect Validation for Element Segment Initialization
`wasm-interp` does not correctly handle element segment initialization using `global.get` for non-imported globals.

### Test case
program.wasm
[program.wasm.txt](https://github.com/user-attachments/files/18361702/program.wasm.txt)

### Environment

* OS: Ubuntu 20.04
* CPU: amd64
* WABT version: 1.0.36
* Commands:
```bash
./wasm-interp --enable-all --run-export=main ./program.wasm
```

### Actual behavior
```

/path/to/wabt/bin/wasm-validate --enable-all --run-export=main /path/to/test/program.wasm
error: initializer expression can only reference an imported global

```

### Expected Behavior
Element segment offset expressions should be able to reference both imported and non-imported globals. The module should pass validation without errors.

ANAMASGARD:
Fixes #2526

## Problem

WABT currently rejects `global.get` of any non-imported global in initializer expressions. This prevents element/data segment offsets and global initializers from referencing immutable globals defined earlier in the same module.

The current WebAssembly validation rules allow `global.get` to reference a previously defined immutable global. The referenced global must still be immutable and previously defined; self-references, forward references, invalid indexes, and wrong types remain invalid.

## Fix

Remove the imported-global restriction from `SharedValidator::OnGlobalGet` while preserving the mutable-global check.

Split global handling into `OnGlobalType` (declaration validation via `CheckReferenceType`) and `OnGlobal` (publication into `globals_`). Validate each initializer before publishing the global so it is visible to later globals but not to itself. Failed type checks do not publish the global.

Apply the same ordering semantics to both `Validator::CheckModule` and `BinaryReaderInterp`.

Remove the now-unused `num_imported_globals_` bookkeeping.

## Tests

Add regression coverage in `test/regress/issue-2526.txt` for:
- element segment offsets using a defined immutable global (runtime-observable via `call_indirect` → 42)
- data segment offsets using a defined immutable global (runtime-observable via `i32.load8_u` → 97)
- global initializers referencing an earlier immutable global
- global chaining (`$a` → `$b` → `$c`)
- self-references, forward references, mutable globals, and invalid offset types

Update affected spec-test expectations using minimal patched snapshots under `test/old-spec/` (core `global`/`elem`/`data` plus proposal suites) and extend `test/update-spec-tests.py` to preserve those overrides during regeneration.

## Test plan

- [x] `./test/run-tests.py issue-2526`
- [x] `./test/run-tests.py spec/global spec/elem spec/data`
- [x] `./test/run-tests.py wasm2c/spec/global wasm2c/spec/elem wasm2c/spec/data`
- [x] `cmake --build out --target run-unittests`
- [x] `cmake --build out --target run-c-api-tests`
- [x] `cmake --build out --target run-tests`
- [x] Feature flags: default, `--enable-all`, `--enable-extended-const` on valid repro
- [x] Self-reference rejected by `wasm-validate` / `wasm-interp` paths

### DIFF
--- a/include/wabt/shared-validator.h
+++ b/include/wabt/shared-validator.h
@@ -367,7 +367,6 @@ class SharedValidator {
   std::vector<TagType> tags_;         // Includes imported and defined.
   std::vector<ElemType> elems_;
   Index starts_ = 0;
-  Index num_imported_globals_ = 0;
   Index data_segments_ = 0;
 
   // Includes parameters, since this is only used for validating
--- a/include/wabt/shared-validator.h
+++ b/include/wabt/shared-validator.h
@@ -85,7 +85,8 @@ class SharedValidator {
   Result OnTable(const Location&, Type elem_type, const Limits&, TableImportStatus import_status, TableInitExprStatus init_provided);
   Result OnMemory(const Location&, const Limits&, uint32_t page_size);
   Result OnGlobalImport(const Location&, Type type, bool mutable_);
-  Result OnGlobal(const Location&, Type type, bool mutable_);
+  Result OnGlobalType(const Location&, Type type);
+  Result OnGlobal(Type type, bool mutable_);
   Result OnTag(const Location&, Var sig_var);
 
   Result OnExport(const Location&,
--- a/src/interp/binary-reader-interp.cc
+++ b/src/interp/binary-reader-interp.cc
@@ -359,6 +359,14 @@ class BinaryReaderInterp : public BinaryReaderNop {
   std::vector<GlobalType> global_types_;  // Includes imported and defined.
   std::vector<TagType> tag_types_;        // Includes imported and defined.
 
+  // The defined global whose initializer is currently being validated. It is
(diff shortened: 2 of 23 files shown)


zherczeg:
This is also part of the GC patchset.

ANAMASGARD:
@zherczeg So should I close this PR ??

zherczeg:
@sbc100 what should be the strategy here? Add as many features independently? I think the gc code is less complex, it simply uses a counter for validated globals.

sbc100:
If this duplicating work that is part your your wider GC work then lets just wait for that to land. 

Unless there is some urgent need to fix this issue right now?  But I can't think of one.

zherczeg:
@ANAMASGARD thank you for the contribution. However, this is part of another patch, so we should close this PR.

PR #2807, opened 2026-08-06
wasm2c: Optimize call indirect code footprint for better performance

shravanrn:
Currently call indirects in Wasm2c add a lot of inline checks. This bloats code size impacting icache and has a visible impact on Wasm-sandboxed code. For example, in one of the Firefox code paths that is indirect call-heavy on a Wasm-sandboxed libexpat (for XML parsing), this code change improved code performance by 4.33% (i.e. eliminated 13.5% of wasm2c's overhead over native code in this benchmark)

The changes in this PR are
- Remove the unnecessary null check on the function pointer in `CHECK_CALL_INDIRECT` and null check on the local function type. This will generally be populated by Wasm2c runtime code, and will never populated to be null
- Move the slowpath for call_indirect type checking (the code paths used when invoking a funcref whose type was created in a different module) into a non-inline function to reduce the call_indirect inline footprint

This PR does not change any functionality; it only optimizes performance.

(Note this PR can be landed independently of other perf PRs I submit. I will rebase PRs depending on which land first)

### DIFF
--- a/src/prebuilt/wasm2c_source_declarations.cc
+++ b/src/prebuilt/wasm2c_source_declarations.cc
@@ -137,22 +137,65 @@ R"w2c_template(
 #define UNREACHABLE TRAP(UNREACHABLE)
 )w2c_template"
 R"w2c_template(
-static inline bool func_types_eq(const wasm_rt_func_type_t a,
+#if defined(__clang__) || defined(__GNUC__)
 )w2c_template"
-R"w2c_template(                                 const wasm_rt_func_type_t b) {
+R"w2c_template(#define W2C_COLD_FUNC __attribute__((noinline, cold))
 )w2c_template"
-R"w2c_template(  return (a == b) || LIKELY(a && b && !memcmp(a, b, 32));
+R"w2c_template(#elif defined(_MSC_VER)
+)w2c_template"
+R"w2c_template(#define W2C_COLD_FUNC __declspec(noinline)
+)w2c_template"
+R"w2c_template(#else
+)w2c_template"
+R"w2c_template(#define W2C_COLD_FUNC
+)w2c_template"
+R"w2c_template(#endif
+)w2c_template"
+R"w2c_template(
+W2C_COLD_FUNC static bool func_types_eq_slowpath(
+)w2c_template"
+R"w2c_template(    const wasm_rt_funcref_table_t* table,
+)w2c_template"
+R"w2c_template(    const wasm_rt_func_type_t expected_type,
+)w2c_template"
+R"w2c_template(    uint32_t index) {
+)w2c_template"
+R"w2c_template(  const size_t sha256size = 32;
+)w2c_template"
+R"w2c_template(  if (index >= table->size) {
+)w2c_template"
+R"w2c_template(    // Table index out of bounds. Raise Wasm trap.
+)w2c_template"
+R"w2c_template(    TRAP(CALL_INDIRECT);
+)w2c_template"
+R"w2c_template(    return false;
+)w2c_template"
+R"w2c_template(  }
+)w2c_template"
+R"w2c_template(  const wasm_rt_funcref_t* const func_ref = &table->data[index];
+)w2c_template"
+R"w2c_template(  if (!expected_type || !func_ref->func || !func_ref->func_type ||
+)w2c_template"
+R"w2c_template(      memcmp(expected_type, func_ref->func_type, sha256size) != 0) {
+)w2c_template"
+R"w2c_template(    // Type check failed. Raise Wasm trap.
+)w2c_template"
+R"w2c_template(    TRAP(CALL_INDIRECT);
+)w2c_template"
+R"w2c_template(    return false;
+)w2c_template"
+R"w2c_template(  }
+)w2c_template"
+R"w2c_template(  return true;
 )w2c_template"
 R"w2c_template(}
 )w2c_template"
 R"w2c_template(
-#define CHECK_CALL_INDIRECT(table, ft, x)                \
+#define CHECK_CALL_INDIRECT(table, ft, x)                       \
 )w2c_template"
-R"w2c_template(  (LIKELY((x) < table.size && table.data[x].func &&      \
+R"w2c_template(  (LIKELY((x) < table.size && ft == table.data[x].func_type) || \
 )w2c_template"
-R"w2c_template(          func_types_eq(ft, table.data[x].func_type)) || \
-)w2c_template"
-R"w2c_template(   TRAP(CALL_INDIRECT))
+R"w2c_template(   func_types_eq_slowpath(&(table), ft, x))
 )w2c_template"
 R"w2c_template(
 #define DO_CALL_INDIRECT(table, t, x, ...) ((t)table.data[x].func)(__VA_ARGS__)
--- a/src/template/wasm2c.declarations.c
+++ b/src/template/wasm2c.declarations.c
@@ -73,15 +73,37 @@ static inline void wasm_rt_segue_write_base(void* base) {
 
 #define UNREACHABLE TRAP(UNREACHABLE)
 
-static inline bool func_types_eq(const wasm_rt_func_type_t a,
-                                 const wasm_rt_func_type_t b) {
-  return (a == b) || LIKELY(a && b && !memcmp(a, b, 32));
+#if defined(__clang__) || defined(__GNUC__)
+#define W2C_COLD_FUNC __attribute__((noinline, cold))
+#elif defined(_MSC_VER)
+#define W2C_COLD_FUNC __declspec(noinline)
+#else
+#define W2C_COLD_FUNC
+#endif
+
+W2C_COLD_FUNC static bool func_types_eq_slowpath(
+    const wasm_rt_funcref_table_t* table,
+    const wasm_rt_func_type_t expected_type,
+    uint32_t index) {
+  const size_t sha256size = 32;
+  if (index >= table->size) {
+    // Table index out of bounds. Raise Wasm trap.
+    TRAP(CALL_INDIRECT);
+    return false;
+  }
+  const wasm_rt_funcref_t* const func_ref = &table->data[index];
+  if (!expected_type || !func_ref->func || !func_ref->func_type ||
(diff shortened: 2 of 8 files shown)


src/template/wasm2c.declarations.c
```diff
@@ -73,14 +73,29 @@ static inline void wasm_rt_segue_write_base(void* base) {
...
+#define W2C_COLD_FUNC __attribute__((noinline, cold))
+#elif defined(_MSC_VER)
+#define W2C_COLD_FUNC __declspec(noinline)
+#else
+#define W2C_COLD_FUNC
+#endif
+
+W2C_COLD_FUNC static bool func_types_eq_slow(const wasm_rt_func_type_t a,
+                                             const wasm_rt_func_type_t b) {
+  return a && b && memcmp(a, b, 32) == 0;
```
zherczeg: I see that this const is coming from the old code, but why is it 32? This number is used frequently, might be worth to introduce a const.

src/template/wasm2c.declarations.c
```diff
@@ -73,14 +73,29 @@ static inline void wasm_rt_segue_write_base(void* base) {
...
+}
+
 static inline bool func_types_eq(const wasm_rt_func_type_t a,
                                  const wasm_rt_func_type_t b) {
-  return (a == b) || LIKELY(a && b && !memcmp(a, b, 32));
+  if (LIKELY(a == b)) {
```
zherczeg: Is this a pointer comparison? It seems a struct comparison... The `a && b` also seems interesting. If `!a` and `!b` then `a == b` is true, isn't it? Perhaps the `!a` could point to a static data, which could eliminate all `a && b` checks.

src/prebuilt/wasm2c_source_declarations.cc
```diff
@@ -137,20 +137,48 @@ R"w2c_template(
...
+)w2c_template"
+R"w2c_template(}
+)w2c_template"
+R"w2c_template(
 static inline bool func_types_eq(const wasm_rt_func_type_t a,
 )w2c_template"
 R"w2c_template(                                 const wasm_rt_func_type_t b) {
 )w2c_template"
-R"w2c_template(  return (a == b) || LIKELY(a && b && !memcmp(a, b, 32));
+R"w2c_template(  if (LIKELY(a == b)) {
```
zherczeg: I don't understand why this is faster. The compiler should generate the same code.

This looks like the best form, although I think the generated code is still the same.
`return LIKELY((a == b) || (a && b && !memcmp(a, b, 32));`

zherczeg (commented):
I know very little about wasm2c, so I am not the best reviewer for this part. I have experience with code optimizations though.

zherczeg:
If Garbage Collection will ever be part of wabt, do you plan to support it in wasm2c? Type checking will be affected by GC.

src/template/wasm2c.declarations.c
shravanrn: The existing code has `wasm_rt_func_type_t` is typedef to a `const char*`, so this is a pointer comparison.

I think your comment does highlight another optimization here: `a` will always be specified by wasm2c generated code and so will never be null. `b` however comes from a table index and can be null in the event that the indirect call index picks an uninitialized slot within the table size. So I can remove `a is not null` check as well.

I agree that the naming could probably use some cleanup to indicate this is a ptr type and not a struct. I can follow up to rename this in a separate PR, since this PR has performance improvement changes.

src/template/wasm2c.declarations.c
shravanrn: The 32 is because the slow path is doing a comparison on SHA-256 hashes of the function type. I believe this is the only instance of this `32`, and we can't really change this value, as changing this would entail changing the hash used in the c-writer.cc (codegenerator file of wasm2c), and so is a more global change anyway. So I don't think we need to add a const to the header etc. I can add a local const variable to make the intent clear.

src/prebuilt/wasm2c_source_declarations.cc
shravanrn: The difference is the hint `W2C_COLD_FUNC` to the compiler in the form of attributes to the function  `func_types_eq_slow`. By preventing inline, the memcmp (which involves generating wide AVX instructions on modern x86-64 platforms) is not inlined and the call site of the call indirect is more compact. This seems to have a visible impact in the benchmarks on XML parsing.

shravanrn:
> If Garbage Collection will ever be part of wabt, do you plan to support it in wasm2c? Type checking will be affected by GC.

This is not on my todo list for the forseeable future and I haven't heard anything from this other contributors about this. Maybe @keithw or @sbc100 know more.

After the performance improvements, my next two features are probably uvwasi support and threads support, although I don't really have a timeline on these as this will depend on when I am not too busy with other responsibilities.

keithw:
(1) I think it's awesome that you're diving into these performance corners! I really liked your talk yesterday.

(2) I am a little surprised by these results given that (at the time I did this call_indirect optimization) I had it benchmarked down to individual instruction counts (https://github.com/WebAssembly/wabt/pull/2120) and the penalty from cross-module call_indirect was basically zero (when the linker is able to merge the string constants, so it's just a pointer equality test) or 1-1.5 cycles (when the CPU has to do the actual 256-bit equality test).

See discussion on https://github.com/WebAssembly/wabt/pull/2120 for more (I certainly had to refresh my memory).

Do you think what you're seeing now is consistent with that or has this bitrotted somehow? And, wrt Firefox, is it possible to see if the linker is merging the string constants (as I would have expected)? Or is the idea that it **is** merging the string constants, so in practice it **is** just a 64-bit equality test, but just "having the AVX2 256-bit equality test" inline is what's blowing out the icache in production?

shravanrn:
> but just "having the AVX2 256-bit equality test" inline is what's blowing out the icache in production?

@keithw Yup, it's the icache blowout. I don't think my tests exercise the `memcmp` part of the code at all and Firefox only uses a single Wasm module (so merges the strings).  The exact benchmark for XML parsing involves calling back into the application through an indirect call each time an XML element or attribute is encountered during the parse. This makes the call indirect path extremely hot (millions of calls for a large XML document/SVG file). I don't believe this is bitrot or the original discussion was incorrect with respect to the compiler/linker/microbenchmark aspects, but I think what I learned was that the micro-architectural aspects in a large application/Wasm module (macrobenchmark) seem to have non trivial performance impact.

shravanrn:
> but just "having the AVX2 256-bit equality test" inline is what's blowing out the icache in production?

@keithw Caveat to the above. I suppose the slowdown could be a mix of other micro-architectural effect that I haven't checked as well beyond just the icache (e.g., speculative execution prefers the non branching path, runs AVX instructions followed by a rollback which could be extra slow :shrug: etc.). But, either way, the fix is the same --- move the AVX instructions more clearly into a slow/cold code path

zherczeg:
There is something I don't understand. `All` function types should be known at compile time. Hence wasm_rt_func_type_t should simply be a type index (unsigned int). There should be no HASH comparison or anything. Functions could have type descriptors though, but they should not be needed for any type comparison.

shravanrn:
> There is something I don't understand. `All` function types should be known at compile time. Hence wasm_rt_func_type_t should simply be a type index (unsigned int). There should be no HASH comparison or anything. Functions could have type descriptors though, but they should not be needed for any type comparison.

@zherczeg I believe some of the logic was documented in https://github.com/WebAssembly/wabt/pull/2120

The main issue is how you can assign a unique number for all the function types you will encounter, and there are a few options that almost work, but most of these end up having issues of performance or compatibility

1. Fresh incrementing integer --- Start a counter at 0. As you see a new type give this type a function type of the current value of counter. The problem is that the function types assigned are not synchronized outside of the current module, and each module has its own idea of what type id 1 ones. This would break the multi-module cases for wasm2c.

2. Each unique function type has a fixed numeric mapping --- E.g., `(u32, u32) -> u32` is always 111 or `(f32, f32) -> f32` is always 222 etc. The problem with this approach is that given that each function can have an arbitrary number of parameters, but the numeric we pick will have a limit ===> this places artificial cap on how general a function this can represent. Additionally, when looking at different mapping schemes with a numeric of `uint64_t` and accounting for all the different Wasm types along with things like multiple returns, the cap becomes rather small and the design is not very future proof.

3. Function types are assigned by a shared registry at runtime --- function indexes are never set at compile time, but the Wasm2c runtime layer maintains a runtime version of the counter in Option 1 that is shared across modules. This option has runtime costs because a call-indirect Wasm instruction now has to query the registry at runtime prior to executing the CALL_INDIRECT macro, and a hashmap lookup at each call indirect would dwarf any of the overheads we are trying to fix.

I think all this let @keithw converge on the design he implemented in wasm2c

4. Each unique function type has a variable size string mapping --- Effectively something like `(u32, u32) -> u32` or `(f32, f32) -> f32` would be a string we compare for type equality. However, doing an unbounded string equality has its own performance concerns, so hashing this with SHA256 to a fixed 32-byte array was chosen as the best approach. (The hash technically introduces a negligible chance for collision, but the odds of this actually happening are astronomical, so this is not a real concern). Finally the 32-byte array was preserved as a string so that the natural string de-duplication in C's compile and link, would allow us to implement a pointer fast path of direct pointer comparison.

zherczeg:
The problem with collision is that if it ever happens, you cannot fix it.

What about option 1 with a header file, where all types are in a generated header file? The header is parsed/extended as necessary.

shravanrn:
> What about option 1 with a header file, where all types are in a generated header file? The header is parsed/extended as necessary.

This would require compilation of all modules to occur
- in a shared environment where this header file can be shared and extended amongst the modules
- to be sequentialized in some way and will likely mean anyone doing this has to modify their build system 

The current design has the advantage that any Wasm module can be compiled independently and will continue to work. 

Also, I do want to be cautious that we don't detour this PR into a full redesign of call-indirect. If anyone is interested in doing the work for that, I am happy to discuss designs for that in a proposed PR/bug for that. Right now, I just want to keep the changes minimal so we improve performance with the design we have in place

**Note** Cleaned up the design a bit more.

zherczeg (approved):
The code looks reasonable. LGTM

PR #2808, opened 2026-08-07
wasm2c: Optimize force_read performance

shravanrn:
wasm2c's `FORCE_READ_*` macros allow spec compliance when using guard pages by preventing elimination of potentially trapping loads, but introduces several low-level performance issues in the generated code.

1. It prevents load followed by a modification to be prevented from being combined into a “load with modification”. This occurs in one the libraries sandboxed by Firefox. The library loads a byte and extends it to 32-bits before processing it --- all in tight loop. Since wasm2c's load byte corresponds to the following code

    ```
    u8 result;
    memcpy(&result, &mem[addr], sizeof(u8)); 
    __asm__("" ::"r"(result)); 
    return (u32) result;
    ```
  
    the generated assembly first performs a load with 32-bit load shifted to a byte, sends the byte result to the asm block, and then performs a widening. However a faster option here, is to just read the byte whie sign extending it, and then consume the sign extended result. This only requires moving the last cast prior to the asm. This simple change speeds up this  library by 3%.

2. The force_read currently uses the general purpose register "r" for force_read_float. This forces C compilers to emit an expensive float-register to int-register conversion on most architectures. The result of this is roughly an extra 7% overhead when running a sandboxed libsoundtouch which is used by Firefox to adjust the playback speed of audio files.

### DIFF
--- a/src/prebuilt/wasm2c_atomicops_source_declarations.cc
+++ b/src/prebuilt/wasm2c_atomicops_source_declarations.cc
@@ -29,9 +29,11 @@ R"w2c_template(        (_Atomic volatile t1*)MEM_ADDR(mem, addr, sizeof(t1)),   
 )w2c_template"
 R"w2c_template(        memory_order_relaxed);                                                \
 )w2c_template"
-R"w2c_template(    force_read(result);                                                       \
+R"w2c_template(    t3 ret = (t3)(t2)result;                                                  \
 )w2c_template"
-R"w2c_template(    return (t3)(t2)result;                                                    \
+R"w2c_template(    force_read(ret);                                                          \
+)w2c_template"
+R"w2c_template(    return ret;                                                               \
 )w2c_template"
 R"w2c_template(  }                                                                           \
 )w2c_template"
@@ -117,9 +119,11 @@ R"w2c_template(    wasm_rt_memcpy(&result, MEM_ADDR_MEMOP(mem, addr, sizeof(t1))
 )w2c_template"
 R"w2c_template(                   sizeof(t1));                                             \
 )w2c_template"
-R"w2c_template(    force_read(result);                                                     \
+R"w2c_template(    t3 ret = (t3)(t2)result;                                                \
 )w2c_template"
-R"w2c_template(    return (t3)(t2)result;                                                  \
+R"w2c_template(    force_read(ret);                                                        \
+)w2c_template"
+R"w2c_template(    return ret;                                                             \
 )w2c_template"
 R"w2c_template(  }                                                                         \
 )w2c_template"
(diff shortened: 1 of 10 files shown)


shravanrn:
@sbc100 If you have a chance this is the other PR that I am really hoping to land soon as it has a large performance delta. Most of the others that I will submit shortly are much more targeted/minor optimizations. Please have a look if possible.

wasm2c/wasm-rt.h
```diff
@@ -253,6 +253,17 @@ extern "C" {
...
 
+/**
+ * This macro, if defined, allows the embedder to permit elimination of unused
+ * memory loads. Note, this a non conformant configuration, i.e., this does not
+ * respect Wasm's specification, as Wasm requires all loads (even the eliminated
+ * loads to trap), whereas this configuration could eliminate an out-of-bound
+ * load, and thus allow the Wasm module to not trap. Use with caution.
+ */
+#ifndef WASM_RT_NONCONFORMING_GUARD_PAGES_NOFORCEREAD
+#define WASM_RT_NONCONFORMING_GUARD_PAGES_NOFORCEREAD 0
```
sbc100: In the past there has been push back against even including support for any kind of non-conforming feature like this.  So unless you maybe split this out into separate change if you really want to land it and we can discuss there?

Also, regarding the name how about `WASM_RT_NONCONFORMING_ALLOW_OOB_READ_ELIMINATION`?

sbc100:
Is the PR description up-to-date?  I'm not sure how the patch solves the two issues outlined?

wasm2c/wasm-rt.h
shravanrn: Happy to change the name, and can split out the change for discussion. But short version: yes, we want this. The performance delta is too significant to not provide an option for this. Please note that the high priority is for this nonconforming discussion, not the other change, so we can resume the discussion in the new PR

shravanrn:
> Is the PR description up-to-date? I'm not sure how the patch solves the two issues outlined?

Yup, its up to date. The two changes apart from the non-conforming change are 
- Move force_read a little lower, as it helps clang combine a load and a type cast into one operation
- Specify the current force_read inline assembly constraints on floats, so that float data is not moved to GPRs unnecessary

shravanrn:
> Is the PR description up-to-date? I'm not sure how the patch solves the two issues outlined?

Sorry, my bad the second change had gotten lost at some point due to a rebase. I have pushed the updated PR now and removed the NONCONFORMING change. 

(No rush on this PR: the NONCONFORMING PR which I will upload shortly is more time sensitive)

sbc100 (approved):

PR #2809, opened 2026-08-07
Fix assertion parsing typed refs in GC struct/array fields (#2750)

### LINKED ISSUE #2750: wat2wasm/wast2json: assertion failure (!IsReferenceWithIndex()) parsing a typed reference in a GC struct/array field
## Summary

The WAT text parser aborts with an assertion failure when a GC `struct` or `array` field has a **typed reference** element type (`(ref N)` / `(ref null N)`). `WastParser::ParseField` builds the field's `Type` with the single-argument `Type(Type::Enum)` constructor, which discards the type index and asserts `!IsReferenceWithIndex()`.

This is reachable from any tool/library entry point that parses WebAssembly text via `ParseWatModule` / `ParseWastScript`: `wat2wasm`, `wast2json`, `spectest-interp`, and any application linking `libwabt`.

- Debug/assert builds: `abort()` → denial of service (CWE-617, reachable assertion).
- Release (`NDEBUG`) builds: the assertion is compiled out and the field silently gets a `Type` with `type_index_ = 0` — the referenced type index is **lost** (incorrect parse result / type confusion).

Tested on `main` @ `03a00a1`, Ubuntu 24.04, clang 18.1.3.

## Steps to reproduce

```console
$ echo '(module (type (struct (field (ref 0)))))' | wat2wasm --enable-gc --enable-function-references -
wat2wasm: .../include/wabt/type.h:75: wabt::Type::Type(wabt::Type::Enum): Assertion `!IsReferenceWithIndex()' failed.
Aborted (core dumped)

$ echo '(module (type (array (ref 0))))' | wat2wasm --enable-gc --enable-function-references -
Aborted (core dumped)
```

The same inputs also crash `wast2json --enable-all`.

### Required feature flags
Both `--enable-gc` (for `struct`/`array`) and `--enable-function-references` (for the `(ref N)` syntax) are required.

### What does / does not trigger it

| Input inside `(module (type ...))` | flags | result |
|---|---|---|
| `(struct (field (ref 0)))` | gc + function-references | **abort** |
| `(array (ref 0))` | gc + function-references | **abort** |
| `(struct (field (mut (ref 0))))` | gc + function-references | **abort** |
| `(struct (field (ref 0)))` | gc only | clean parse error (ok) |
| `(struct (field i32))` | gc + function-references | ok |
| `(struct (field funcref))` (non-indexed ref) | gc + function-references | ok |

Only **indexed** typed references crash; non-reference and non-indexed reference types are fine.

## Affected tools / functions
`libwabt` — `WastParser::ParseField` (reached by `ParseWatModule` and `ParseWastScript`). Tools: `wat2wasm`, `wast2json`, `spectest-interp`.

## Root cause

`src/wast-parser.cc`, `WastParser::ParseField()` → lambda `parse_mut_valuetype` (lines ~1785–1792):

```cpp
Var type;
field->type = Type(type.opt_type());   // single-arg ctor drops the type index
```

`ParseValueType` may yield a typed reference (`Type::Ref` / `Type::RefNull`) whose `opt_type()` `IsReferenceWithIndex()`. Passing it to the single-arg constructor triggers the assert in `include/wabt/type.h:74`:

```cpp
Type(Enum e) : enum_(e), type_index_(0) { assert(!IsReferenceWithIndex()); }
```

The sibling code in the same file already handles this correctly with the two-argument constructor:

```cpp
// src/wast-parser.cc:1031  (ParseValueTypeList)
out_type_list->push_back(Type(type.opt_type(), kInvalidIndex));
```

So `ParseField` is simply missing the index argument that the rest of the parser uses.

## Suggested fix

Preserve the reference's type index instead of dropping it — either mirror line 1031:

```diff
-   field->type = Type(type.opt_type());
+   field->type = Type(type.opt_type(), kInvalidIndex);
```

(applied to both occurrences), or, better, resolve the parsed `Var` to a fully-typed reference the way other call sites do (`VarToType(...)`, used for table element types), so that `(ref N)` keeps its actual target index `N`. The minimal change stops the crash; the `VarToType`-based change additionally fixes the silent index loss in release builds.

## Discovery
Found via libFuzzer + ASAN/UBSAN with a custom `wast2json` harness mirroring `src/tools/wast2json.cc`.

ANAMASGARD:
Fixes #2750

## Summary
- Fix `WastParser::ParseField` to preserve typed reference indices in GC struct/array fields
- Defer `VarToType` for named refs until struct field storage is stable (avoids dangling pointers in `resolve_ref_types_`)
- Add `regress-2750.txt` with non-zero index `(ref 1)` coverage
- Add `--enable-gc` to `run-roundtrip.py` for GC roundtrip tests

## Test plan
- [x] Issue reproducers (struct/array/mut `(ref 0)`)
- [x] Non-zero index `(ref 1)`
- [x] Named deferred ref `(ref $t1)` via func type
- [x] `test/run-tests.py regress-2750`
- [x] Full test suite
------

### DIFF
--- a/include/wabt/wast-parser.h
+++ b/include/wabt/wast-parser.h
@@ -262,7 +262,7 @@ class WastParser {
   Result ParseCatchInstrList(CatchVector* catches);
   Result ParseCatchExprList(CatchVector* catches);
   Result ParseGlobalType(Global*);
-  Result ParseField(Field*);
+  Result ParseField(Field*, Var* out_type, bool* defer_named_ref);
   Result ParseFieldList(std::vector<Field>*);
 
   template <typename T>
--- a/src/wast-parser.cc
+++ b/src/wast-parser.cc
@@ -1790,7 +1790,12 @@ Result WastParser::ParseTypeModuleField(Module* module) {
       Error(loc, "array type not allowed");
     }
     auto array_type = std::make_unique<ArrayType>(name);
-    CHECK_RESULT(ParseField(&array_type->field));
+    Var type_var;
+    bool defer_named_ref = false;
+    CHECK_RESULT(ParseField(&array_type->field, &type_var, &defer_named_ref));
+    if (defer_named_ref) {
+      VarToType(type_var, &array_type->field.type);
+    }
     field->type = std::move(array_type);
   } else {
     return ErrorExpected({"func", "struct", "array"});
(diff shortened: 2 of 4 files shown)


zherczeg:
There is a GC patchset which reworks the whole codebase. It is more than a year old, but still actively maintained: #2607 #2618 #2622

sbc100:
Yeah, I guess it makes sense to wait on the larger patchset.

PR #2810, opened 2026-08-07
Fix validation of typed select with empty result types (#2708)

### LINKED ISSUE #2708: Missing Type Validation for select Instruction
`wasm-interp` fails to reject an invalid WebAssembly module involving the `select` instruction.

### Test case
[select.wasm.txt](https://github.com/user-attachments/files/25575081/select.wasm.txt)

### Environment

* OS: Ubuntu 20.04
* CPU: amd64
* wasm-interp version: `6ca912cf`
* Commands:
```bash
wabt/bin/wasm-interp --enable-all test.wasm --run-export=run
```

### Actual behavior

`wasm-interp` incorrectly accepts this invalid module (exit code 0). The output is:

```
run() => i32:42
```

### Expected behavior

The runtime should reject this module during validation because the module violates the specification.

```
Exception: CompileError: WebAssembly.Module doesn't parse at byte 11: select invalid result arity for, in function at index 0 (evaluating 'new WebAssembly.Module(wasm_code)')

```

ANAMASGARD:
Fixes #2708

## Summary
- Reject `SelectT` with zero result types in the binary reader instead of treating it as untyped `select`
- Reject WAT `select (result)` with an empty result list during validation
- Add regression tests for the binary and WAT cases

## Test plan
- [x] `python3 test/run-tests.py binary/bad-select-empty-result parse/expr/bad-select-empty-result spec/select`
- [x] `./out/wabt-unittests`
- [x] `python3 test/run-tests.py` (full suite)
- [x] Manual repro: invalid `SelectT` + empty type vector rejected by `wasm-interp` and `wasm-validate`

### DIFF
--- a/include/wabt/ir.h
+++ b/include/wabt/ir.h
@@ -701,8 +701,13 @@ class SelectExpr : public ExprMixin<ExprType::Select> {
  public:
   SelectExpr(const Location& loc = Location())
       : ExprMixin<ExprType::Select>(loc) {}
-  TypeVector result_type;
-  bool has_result_type = false;
+  // Untyped select is represented by {Type::Void}. An empty result_type
+  // represents an explicit typed select with zero result types and is invalid.
+  TypeVector result_type{Type::Void};
+
+  bool IsUntyped() const {
+    return result_type.size() == 1 && result_type[0] == Type::Void;
+  }
 };
 
 class TableInitExpr : public ExprMixin<ExprType::TableInit> {
--- a/include/wabt/ir.h
+++ b/include/wabt/ir.h
@@ -702,6 +702,7 @@ class SelectExpr : public ExprMixin<ExprType::Select> {
   SelectExpr(const Location& loc = Location())
       : ExprMixin<ExprType::Select>(loc) {}
   TypeVector result_type;
+  bool has_result_type = false;
 };
 
 class TableInitExpr : public ExprMixin<ExprType::TableInit> {
--- a/src/binary-reader-ir.cc
+++ b/src/binary-reader-ir.cc
@@ -1212,6 +1212,7 @@ Result BinaryReaderIR::OnReturnExpr() {
 
 Result BinaryReaderIR::OnSelectExpr(Index result_count, Type* result_types) {
   auto expr_ptr = std::make_unique<SelectExpr>();
+  expr_ptr->has_result_type = result_count > 0;
   expr_ptr->result_type.assign(result_types, result_types + result_count);
   return AppendExpr(std::move(expr_ptr));
 }
--- a/src/binary-reader-ir.cc
+++ b/src/binary-reader-ir.cc
@@ -1212,8 +1212,11 @@ Result BinaryReaderIR::OnReturnExpr() {
 
 Result BinaryReaderIR::OnSelectExpr(Index result_count, Type* result_types) {
   auto expr_ptr = std::make_unique<SelectExpr>();
-  expr_ptr->has_result_type = result_count > 0;
-  expr_ptr->result_type.assign(result_types, result_types + result_count);
+  if (result_count == 0) {
+    expr_ptr->result_type = {Type::Void};
+  } else {
+    expr_ptr->result_type.assign(result_types, result_types + result_count);
+  }
   return AppendExpr(std::move(expr_ptr));
 }
 
--- a/src/binary-reader-ir.cc
+++ b/src/binary-reader-ir.cc
@@ -1212,9 +1212,7 @@ Result BinaryReaderIR::OnReturnExpr() {
 
 Result BinaryReaderIR::OnSelectExpr(Index result_count, Type* result_types) {
   auto expr_ptr = std::make_unique<SelectExpr>();
-  if (result_count == 0) {
-    expr_ptr->result_type = {Type::Void};
-  } else {
+  if (result_count != 0) {
     expr_ptr->result_type.assign(result_types, result_types + result_count);
   }
   return AppendExpr(std::move(expr_ptr));
--- a/src/validator.cc
+++ b/src/validator.cc
@@ -590,8 +590,7 @@ Result Validator::OnReturnCallRefExpr(ReturnCallRefExpr* expr) {
 
 Result Validator::OnSelectExpr(SelectExpr* expr) {
   if (expr->has_result_type && expr->result_type.empty()) {
-    validator_.PrintError(expr->loc,
-                          "invalid arity in select instruction: 0.");
+    validator_.PrintError(expr->loc, "invalid arity in select instruction: 0.");
     result_ |= Result::Error;
     return Result::Ok;
   }
--- a/src/validator.cc
+++ b/src/validator.cc
@@ -589,16 +589,16 @@ Result Validator::OnReturnCallRefExpr(ReturnCallRefExpr* expr) {
 }
 
 Result Validator::OnSelectExpr(SelectExpr* expr) {
-  Index result_count;
-  Type* result_types = nullptr;
-
   if (expr->result_type.empty()) {
     validator_.PrintError(expr->loc, "invalid arity in select instruction: 0.");
     result_ |= Result::Error;
-    result_count = 0;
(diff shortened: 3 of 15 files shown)


include/wabt/ir.h
```diff
@@ -702,6 +702,7 @@ class SelectExpr : public ExprMixin<ExprType::Select> {
   SelectExpr(const Location& loc = Location())
       : ExprMixin<ExprType::Select>(loc) {}
   TypeVector result_type;
+  bool has_result_type = false;
```
sbc100: Can't we use the zero-length result_type of imply that the expr has not result type?

sbc100:
Does this mean we are lacking a test in the spec repo for this?

ANAMASGARD:
@sbc100  the spec repo already has tests for this in the GC proposal files (like struct/array fields with (ref N)). WABT just doesn’t run those GC spec tests yet — they’re marked unimplemented in our test script. So the tests exist upstream, but we weren’t exercising them in WABT, which is why I added regress-2750.txt for this parser bug.

include/wabt/ir.h
ANAMASGARD: @sbc100  empty result_type can mean either “untyped select” or “invalid typed select with no types” — has_result_type distinguishes those, so I don’t think we can drop it without losing that check. (Not part of this PR though.)

sbc100: Do does `has_result_type = false` correspond to "untyped select".

Maybe a comment here explaining the distinction and why this member is needed?

test/binary/bad-select-empty-result.txt
```diff
@@ -0,0 +1,26 @@
...
+section(EXPORT) { count[1] str("run") func_kind 0 }
+section(CODE) {
+  count[1]
+  func {
+    locals[0]
+    i32.const 42
+    i32.const 100
+    i32.const 1
+    0x1c
+    0x00
```
sbc100: What are these two magic numbers?  

I think you can document by using the `name_of_thing[0x1x]` syntax maybe?

sbc100:
> @sbc100 the spec repo already has tests for this in the GC proposal files (like struct/array fields with (ref N)). WABT just doesn’t run those GC spec tests yet — they’re marked unimplemented in our test script. So the tests exist upstream, but we weren’t exercising them in WABT, which is why I added regress-2750.txt for this parser bug.

But what about the non-GC tests.. if we can write tests for this where in wabt that don't depend on GC types then presumably one could write the same non-GC upstream in the spec repo?

sbc100:
@tlively does this looks like a test case that is missing upstream in the spec maybe?

test/parse/expr/bad-select-empty-result.txt
```diff
@@ -0,0 +1,16 @@
...
+  (func
+    i32.const 42
+    i32.const 100
+    i32.const 1
+    select (result)))
+(;; STDERR ;;;
+out/test/parse/expr/bad-select-empty-result.txt:8:5: error: invalid arity in select instruction: 0.
+    select (result)))
+    ^^^^^^
+out/test/parse/expr/bad-select-empty-result.txt:8:5: error: type mismatch at end of function, expected [] but got [i32, i32, i32]
```
zherczeg: Even if the results are not consumed, the condition i32 should be.

include/wabt/ir.h
zherczeg: What is the purpose of "untyped select"? This likely rare case could be represented by the void type.
Btw if "supporting untyped select" is not part of the PR, we should not introduce this.

ANAMASGARD:
@sbc100  I am sorry I mixed this up with another issue earlier . There is already an $arity-0 test in select.wast, but it uses nop operands, so WABT can fail for a stack mismatch instead of the missing arity check. A test with valid operands would catch this bug directly. Happy to follow up upstream if needed.

test/binary/bad-select-empty-result.txt
ANAMASGARD: @sbc100  thanks for the feedback  I have replaced the raw bytes with select_t[0x1c] and result_count[0x00] to make the test clearer.

test/parse/expr/bad-select-empty-result.txt
ANAMASGARD: @zherczeg  I have Fixed it . We now report the arity error but still run OnSelect, so the condition and operands are consumed correctly. I also updated the test with drop.

include/wabt/ir.h
ANAMASGARD: @sbc100 & @zherczeg  Thanks for the feedback .
 I removed has_result_type and now use {Type::Void} to represent the existing untyped select. An empty result_type now represents invalid typed select (result), and I added a comment documenting the distinction.

src/binary-reader-ir.cc
```diff
@@ -1212,7 +1212,11 @@ Result BinaryReaderIR::OnReturnExpr() {
 
 Result BinaryReaderIR::OnSelectExpr(Index result_count, Type* result_types) {
   auto expr_ptr = std::make_unique<SelectExpr>();
-  expr_ptr->result_type.assign(result_types, result_types + result_count);
+  if (result_count == 0) {
+    expr_ptr->result_type = {Type::Void};
```
zherczeg: Do you need {}? `expr_ptr->result_type = Type::Void;` should work

src/validator.cc
```diff
@@ -589,8 +589,21 @@ Result Validator::OnReturnCallRefExpr(ReturnCallRefExpr* expr) {
...
-                                 expr->result_type.data());
+  Index result_count;
+  Type* result_types = nullptr;
+
+  if (expr->result_type.empty()) {
+    validator_.PrintError(expr->loc, "invalid arity in select instruction: 0.");
+    result_ |= Result::Error;
+    result_count = 0;
+  } else if (expr->IsUntyped()) {
+    result_count = 0;
```
zherczeg: `Index result_count = 0;` Could simplify this code.

src/wat-writer.cc
```diff
@@ -977,7 +977,12 @@ Result WatWriter::ExprVisitorDelegate::OnReturnCallRefExpr(
 
 Result WatWriter::ExprVisitorDelegate::OnSelectExpr(SelectExpr* expr) {
   writer_->WritePutsSpace(Opcode::Select_Opcode.GetName());
-  if (!expr->result_type.empty()) {
+  if (expr->IsUntyped()) {
+    // no result annotation
```
zherczeg: I usually don't prefer empty statements.

`if (expr->result_type.empty()) { ... } else if (!expr->IsUntyped()) { ... }`

test/parse/expr/bad-select-empty-result.txt
```diff
@@ -0,0 +1,14 @@
+;;; TOOL: wat2wasm
+;;; ERROR: 1
+(module
+  (func
+    i32.const 42
+    i32.const 100
+    i32.const 1
+    select (result)
+    drop))
```
zherczeg, line 10: Should not this be `drop drop`? Only the `i32.const 1` should be consumed if I remember correctly.

zherczeg (commented):
I like this code much better.

src/binary-reader-ir.cc
ANAMASGARD: @zherczeg since SelectExpr already defaults to {Type::Void}, I simplified it to only assign when result_count != 0.

src/validator.cc
ANAMASGARD: Thanks @zherczeg  I have simplified this by initializing result_count to 0.

src/wat-writer.cc
ANAMASGARD: Makes sense. Updated it to avoid the empty branch.

test/parse/expr/bad-select-empty-result.txt
ANAMASGARD: @zherczeg You're right. Only the i32 condition should be consumed here. I updated the recovery logic and changed the test to drop drop.

zherczeg (approved):
LGTM

PR #2813, opened 2026-08-08
wasm2c: reject contradictory imports instead of aborting

### LINKED ISSUE #2751: wasm2c: abort (UNIMPLEMENTED contradictory import declaration) on a valid module with same-named imports of different kinds
## Summary

`wasm2c` aborts via `UNIMPLEMENTED("contradictory import declaration")` in `CWriter::ComputeUniqueImports` (`src/c-writer.cc:1773`) when a **valid** module imports the same `(module, field)` name twice with **different external kinds** (e.g. once as a function and once as a global). Duplicate import names are permitted by the core WebAssembly binary format and pass validation, so this is reachable with **default features** (no `--enable-*` flag required).

- Result: `abort()` → denial of service (CWE-617, reachable assertion / unhandled `UNIMPLEMENTED`).

Tested on `main` @ `03a00a1`, Ubuntu 24.04, clang 18.1.3.

## Steps to reproduce

```console
$ cat dup.wat
(module
  (import "e" "x" (func))
  (import "e" "x" (global i32)))

$ wat2wasm dup.wat -o dup.wasm        # default features; produces a valid 30-byte module
$ wasm2c dup.wasm -o dup.c
Aborted (core dumped)
```

`wat2wasm` accepts and validates the module (duplicate import names with differing kinds are legal); `wasm2c` then aborts while generating C.

## Affected tools / functions
`libwabt` — `CWriter::ComputeUniqueImports` (reached by `CWriter::BeginInstance` → `WriteModuleInstance` → `WriteCHeader` → `WriteC`). Tool: `wasm2c`. No feature flags required.

## Root cause

`src/c-writer.cc:1762` `CWriter::ComputeUniqueImports`:

```cpp
const auto iterator_and_insertion_bool = import_map.emplace(
    modname_name_pair(import->module_name, import->field_name), import);
if (!iterator_and_insertion_bool.second) {
  if (iterator_and_insertion_bool.first->second->kind() != import->kind()) {
    UNIMPLEMENTED("contradictory import declaration");   // <-- abort on valid input
  } else {
    fprintf(stderr, "warning: duplicate import declaration ...");
  }
}
```

The same-kind duplicate case is handled with a warning, but the different-kind case calls `UNIMPLEMENTED`, which aborts the process on attacker-controlled input.

## Suggested fix

Replace the `UNIMPLEMENTED` abort with a graceful error return (e.g. via the existing error-reporting path), so `wasm2c` rejects the module instead of crashing. If contradictory imports are genuinely unsupported, returning `Result::Error` with a clear message is the appropriate behavior for untrusted input.

## Discovery
Found via libFuzzer + ASAN/UBSAN with a custom `wasm2c` harness mirroring `src/tools/wasm2c.cc` (`ReadBinaryIr → ValidateModule → GenerateNames → ApplyNames → WriteC`), including the tool's feature gate so that only tool-reachable feature configurations are exercised.

ANAMASGARD:
Fixes #2751

## Summary
- Replace `UNIMPLEMENTED("contradictory import declaration")` with a stderr diagnostic and `Result::Error`
- Preflight `ComputeUniqueImports()` at the start of `WriteModule()` so codegen never runs on unlinkable modules
- Add `test/wasm2c/contradictory-imports.txt`

## Test plan
- [x] Manual: `wasm2c dup.wasm` exits 1 with error message (no abort)
- [x] `python3 test/run-tests.py test/wasm2c/contradictory-imports.txt`
- [x] `python3 test/run-tests.py test/wasm2c/`
- [x] `./out/wabt-unittests`

image

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -378,7 +378,7 @@ class CWriter {
   void Write(const FuncTypeExpr&);
   void WriteTagDecls();
   void WriteTags();
-  void ComputeUniqueImports();
+  Result ComputeUniqueImports();
   void BeginInstance();
   void WriteImports();
   void WriteTailCallWeakImports();
@@ -1810,7 +1810,7 @@ void CWriter::WriteTags() {
   }
 }
 
-void CWriter::ComputeUniqueImports() {
+Result CWriter::ComputeUniqueImports() {
   using modname_name_pair = std::pair<std::string, std::string>;
   std::map<modname_name_pair, const Import*> import_map;
   for (const Import* import : module_->imports) {
@@ -1821,7 +1821,13 @@ void CWriter::ComputeUniqueImports() {
         modname_name_pair(import->module_name, import->field_name), import);
     if (!iterator_and_insertion_bool.second) {
       if (iterator_and_insertion_bool.first->second->kind() != import->kind()) {
-        UNIMPLEMENTED("contradictory import declaration");
+        fprintf(stderr,
+                "error: contradictory import declaration: \"%s\".\"%s\" is "
+                "imported as both a %s and a %s\n",
+                import->module_name.c_str(), import->field_name.c_str(),
+                GetKindName(iterator_and_insertion_bool.first->second->kind()),
+                GetKindName(import->kind()));
+        return Result::Error;
       } else {
         fprintf(stderr, "warning: duplicate import declaration \"%s\" \"%s\"\n",
                 import->module_name.c_str(), import->field_name.c_str());
@@ -1838,6 +1844,7 @@ void CWriter::ComputeUniqueImports() {
   for (const auto& node : import_map) {
     unique_imports_.push_back(node.second);
   }
+  return Result::Ok;
 }
 
 void CWriter::BeginInstance() {
@@ -1845,8 +1852,6 @@ void CWriter::BeginInstance() {
     Write("typedef struct ", ModuleInstanceTypeName(), " ", OpenBrace());
     return;
   }
-
-  ComputeUniqueImports();
 
   // define names of per-instance imports
   for (const Import* import : module_->imports) {
@@ -6193,6 +6198,9 @@ void CWriter::WriteCSource() {
 Result CWriter::WriteModule(const Module& module) {
   WABT_USE(options_);
   module_ = &module;
+
+  CHECK_RESULT(ComputeUniqueImports());
+
   WriteCHeader();
   WriteCSource();
   return result_;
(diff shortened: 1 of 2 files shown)


sbc100 (approved):

sbc100:
Seems reasonable to me

src/c-writer.cc
shravanrn, line 1873: Doesn't unique_imports_ have to be populated before this code is run? With this PR's removal of `ComputeUniqueImports()` from this function, it looks like this would no longer be the case

shravanrn (changes requested):

src/c-writer.cc
ANAMASGARD: @shravanrn  You're right that it wasn't obvious from BeginInstance() alone.

unique_imports_ was already filled in by the ComputeUniqueImports() call at the top of WriteModule(), before any header codegen runs. I put the call back in BeginInstance() too, and made ComputeUniqueImports() clear its state on entry so it's safe to call from both places.

shravanrn: This seems a bit clunky. I think the right approach would be to not add multiple calls to `ComputeUniqueImports` and for ` if (Failed(ComputeUniqueImports())) {` to set the `_result` variable and return. You may need to add some checks to `_result` in the call stack of `ComputeUniqueImports` to ensure the error path is reasonably handled

shravanrn (changes requested):

ANAMASGARD:
@shravanrn I have removed the WriteModule preflight and kept a single ComputeUniqueImports() in BeginInstance(). On failure it now sets result_ and returns, with Failed(result_) checks in WriteModuleInstance, WriteCHeader, and WriteModule so we don't keep emitting after an error.

src/c-writer.cc
```diff
@@ -1794,7 +1794,12 @@ void CWriter::WriteTags() {
   }
 }
 
-void CWriter::ComputeUniqueImports() {
+Result CWriter::ComputeUniqueImports() {
+  unique_imports_.clear();
```
shravanrn: You can remove the unnecessary code change to clears.

shravanrn (approved):

shravanrn:
@ANAMASGARD Feel free to rebase and land. (Let me know if you need me to click the land button)

src/c-writer.cc
ANAMASGARD: DONE

ANAMASGARD:
@shravanrn thanks for review I have  removed the clears, rebased onto latest main . 
Ready for you to land now .

shravanrn (approved):

PR #2816, opened 2026-08-10
interp: assert stop_on_first_error in ReadBinaryInterp

ANAMASGARD:
## Summary
- Assert that `ReadBinaryInterp` is only used with `stop_on_first_error=true`.
- Documents that collect-all-errors mode is unsupported for the interp reader.

Follow-up to #2812 / #2747 ( @sbc100  ) .

## Test plan
- [x] Debug: `stop_on_first_error=false` hits the assert
- [x] `wabt-unittests` (135 passed)

### DIFF
--- a/src/interp/binary-reader-interp.cc
+++ b/src/interp/binary-reader-interp.cc
@@ -742,21 +742,11 @@ Result BinaryReaderInterp::OnExport(Index index,
 
   std::unique_ptr<ExternType> type;
   switch (kind) {
-    case ExternalKind::Func:
-      type = func_types_[item_index].Clone();
-      break;
-    case ExternalKind::Table:
-      type = table_types_[item_index].Clone();
-      break;
-    case ExternalKind::Memory:
-      type = memory_types_[item_index].Clone();
-      break;
-    case ExternalKind::Global:
-      type = global_types_[item_index].Clone();
-      break;
-    case ExternalKind::Tag:
-      type = tag_types_[item_index].Clone();
-      break;
+    case ExternalKind::Func:   type = func_types_[item_index].Clone(); break;
+    case ExternalKind::Table:  type = table_types_[item_index].Clone(); break;
+    case ExternalKind::Memory: type = memory_types_[item_index].Clone(); break;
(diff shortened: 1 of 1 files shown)


src/interp/binary-reader-interp.cc
```diff
@@ -742,11 +742,21 @@ Result BinaryReaderInterp::OnExport(Index index,
...
+      break;
+    case ExternalKind::Memory:
+      type = memory_types_[item_index].Clone();
+      break;
+    case ExternalKind::Global:
+      type = global_types_[item_index].Clone();
+      break;
+    case ExternalKind::Tag:
+      type = tag_types_[item_index].Clone();
+      break;
```
sbc100: Revert this chunk (maybe this block needs to be in `// clang-format off` .. bit doesn't need to be part of this change).

sbc100 (approved):

src/interp/binary-reader-interp.cc
ANAMASGARD: Done @sbc100  Thanks for the review

ANAMASGARD:
CIFuzz is still failing because it uses the google/oss-fuzz harness (`projects/wabt/read_binary_interp_fuzzer.cc`) with
`stop_on_first_error=false`, not `fuzzers/read_binary_interp_fuzzer.cc` from this repo (which already passes `true`).

That hits the assert added here on purpose — no further wabt change needed for this. Fix is to set the 4th `ReadBinaryOptions` arg to `true` in oss-fuzz: https://github.com/google/oss-fuzz

Sir @sbc100 should I open a PR there directly, or file an issue first please give feedback ???

sbc100:
It looks like maybe fuzzers/read_binary_interp_fuzzer.cc need updating here.  I'll send a PR to https://github.com/google/oss-fuzz/ to change it hedre.

sbc100:
https://github.com/google/oss-fuzz/pull/15983

ANAMASGARD:
Sir @sbc100 can you please review once again and give feedback if anymore revisions are needed or not for this PR ?

PR #2817, opened 2026-08-10
Assert that ReadBinaryIr is not used with skip_function_bodies

### LINKED ISSUE #2534: options skip_function_bodies can not work with ReadBinaryIr
wabt::ReadBinaryOptions options;
        options.skip_function_bodies = true;

wabt::Result errorCode = wabt::ReadBinaryIr("", wasmData, fileSize, options, &errors,
                                                    &module);

will got an assert 
at function [Result BinaryReaderIR::EndFunctionBody(Index index)]
 of file
 [binary-reader-ir.cc]

the reason is: PushLabel at BinaryReaderIR::BeginFunctionBody, but not consume it when 'skip_function_bodies' turns on.

I think it should add a Step names SkipExpr or trans options on BeginFunctionBody.

BTW, i got another problem, when parse a big wasm in Android, it will crash randomly , but same code work well in MacOS.
i will provide detail later.

Nishuuzz:
`ReadBinaryIr` fails on any module that has a function in it once `skip_function_bodies` is set. Reading sqlite (built with emscripten) both ways:

```
skip_function_bodies=0 -> ok
skip_function_bodies=1 -> FAILED
    function 34 missing end marker
```

34 is just its first non-imported function, so it gives up on the first body it reaches.

`BeginFunctionBody` pushes a label for the function, but with bodies skipped the reader steps straight over the body and the `end` that would pop that label never arrives. `EndFunctionBody` then sees it still sitting on the stack and reports the body as missing its end marker.

The option was only ever meant for the objdump reader, which doesn't build a module, so this declares it unsupported here instead of teaching the IR reader to work around it, along the lines of #2816. Nothing in the tree passes it to `ReadBinaryIr` today, so nothing starts asserting.

Fixes #2534.

### DIFF
--- a/src/binary-reader-ir.cc
+++ b/src/binary-reader-ir.cc
@@ -1894,6 +1894,11 @@ Result ReadBinaryIr(const char* filename,
                     const ReadBinaryOptions& options,
                     Errors* errors,
                     Module* out_module) {
+  // BinaryReaderIR does not support skipping function bodies; that option is
+  // for readers such as objdump that do not build a module. Skipping the
+  // bodies here leaves each function's label unclosed, because the end marker
+  // that would pop it is never read.
+  assert(!options.skip_function_bodies);
   BinaryReaderIR reader(out_module, filename, errors);
   return ReadBinary(data, &reader, options);
 }


sbc100:
I think `skip_function_bodies` was really just designed for the objdump binary reader. 

Maybe better to simply declare `skip_function_bodies` incompatible with the IR binary reader.

sbc100:
(Something like https://github.com/WebAssembly/wabt/pull/2816 maybe)

Nishuuzz:
That makes sense, switched it over — the diff is now just the comment and the assert in ReadBinaryIr, following #2816. Dropped my change to the reader and the test that went with it, and retitled accordingly. Nothing in the tree passes the option to ReadBinaryIr, so nothing starts tripping it.

One thing worth flagging rather than quietly dropping: the person who filed #2534 was using it to skip bodies when loading large modules, so asserting closes that off rather than giving them a path. Not arguing against it — the option clearly was not designed for this reader — just noting it in case you would rather it grew a supported route one day. Either way #2534 can be closed as not supported here.

PR #2818, opened 2026-08-10
Treat exn as a heap type so (ref exn) is non-nullable

chicoxyzzy:
`exnref` is the nullable valtype (`0x69`). The non-nullable form is `(ref exn)` (`0x64 0x69`), the same split as `funcref` vs `(ref func)`.

Before this, the lexer knew the `exn` keyword but both `exn` and `exnref` were the same `Type::ExnRef` valtype. `(ref exn)` failed to parse (`unexpected token "exn", expected a numeric index or a name`), and the binary reader rejected `0x64 0x69` as a heap type.

This change:

- Treats `ExnRef` as a non-typed reference with nullability (`ReferenceOrNull` vs `ReferenceNonNull`), like `FuncRef` / `ExternRef`.
- Parses `(ref exn)` and `(ref null exn)`.
- Reads and writes the typed-ref encoding; `exnref` / `(ref null exn)` still emit `0x69`.
- Applies the same nullability subtyping, so `(ref exn)` is a subtype of `exnref` and not the other way around.

`catch_ref` still produces nullable `exnref`. The pinned exception-handling testsuite uses `(block (result exnref) (try_table (catch_ref …)))` and checks those types exactly, so flipping that would fail every official `throw_ref` / `try_table` test.

Tests: `test/parse/expr/exn-ref.txt`, `exn-ref-as-non-null.txt`, `bad-exn-ref-nullable.txt`, `test/roundtrip/exn-ref.txt`. Also ran parse, roundtrip, dump, `test/spec/exception-handling`, and `test/spec/function-references`.

### DIFF
--- a/include/wabt/type.h
+++ b/include/wabt/type.h
@@ -60,7 +60,7 @@ class Type {
     I32U = 7,  // Not actually specified, but used internally with load/store
   };
 
-  // Used by FuncRef / ExternRef
+  // Used by FuncRef / ExternRef / ExnRef
   enum GenericReferenceType : uint32_t {
     ReferenceOrNull = 0,
     ReferenceNonNull = 1,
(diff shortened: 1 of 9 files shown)


sbc100:
@zherczeg has been doing a lot work on reference types (GC types), which has been stuck in the review queue.    Not sure how this fits in (or not) with that work?

zherczeg:
This is part of #2607 . It also includes support for `(ref noexn)`, `nullnoexn`, and the other types.

PR #2819, opened 2026-08-11
[interp] Fix out-of-bounds read tracing call_ref and return_call_ref

aizu-m:
`Istream::Read` never initialises `Instr::kind` for `call_ref/return_call_ref` though both are emitted into the istream, so under `--trace` the tracer decodes them with the previous instruction's kind and `Pick()` reads past the end of the value stack; give them `Imm_0_Op_1` (no immediate, one operand: the callee reference) like the other reachable opcodes.

### DIFF
--- a/test/interp/tracing-call-ref.txt
+++ b/test/interp/tracing-call-ref.txt
@@ -1,0 +1,40 @@
+;;; TOOL: run-interp-spec
+;;; ARGS*: --enable-function-references --enable-tail-call
+;;; ARGS1: --trace
+(module
+  ;; call_ref/return_call_ref carry no istream immediate, so Istream::Read has to
+  ;; give them an InstrKind explicitly; otherwise the tracer decodes them with the
+  ;; previous instruction's kind and picks operands that are not on the stack. The
+  ;; store in front leaves a two-operand kind behind to exercise that path.
+  (type $v (func))
+  (memory 1)
+  (elem declare func $t)
+  (func $t)
+  (func (export "call")
+    ref.func $t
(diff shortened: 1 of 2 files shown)


test/interp/tracing-call-ref.txt
```diff
@@ -0,0 +1,40 @@
...
+    (i32.store (i32.const 0) (i32.const 0))
+    return_call_ref $v))
+(invoke "call")
+(invoke "retcall")
+(;; STDOUT ;;;
+#0.   16: V:0  | ref.func $0
+#0.   24: V:1  | i32.const 0
+#0.   32: V:2  | i32.const 0
+#0.   40: V:3  | i32.store $0:0+$0, 0
+#0.   56: V:1  | call_ref ?
```
zherczeg, line 28: Is the '?' valid?

zherczeg (commented):
Good catch!

test/interp/tracing-call-ref.txt
aizu-m: Yes, that's expected. call_ref and return_call_ref carry no operand types in opcode.def (the params are all `___`), so TraceSource::Pick works out num_operands as 0, treats the type as Void, and the switch on instr.op has no case for them, so it returns "?". Any opcode without recorded param types traces the same way. The value under it is the callee funcref, the tracer just has no static type to print for it. Showing funcref:N there would mean recording the operand type, which is the TODO already sitting above Pick. The point of this patch is only that the kind is set at all, so Pick(1) reads the real top of stack instead of a stale two-operand kind.

zherczeg (approved):
LGTM

PR #2822, opened 2026-08-13
wasm2c: optimize performance of float handling

shravanrn:
Optimize performance of float handling in wasm2c

- The `sqrt` and `sqrtf` calls currently use the libc functions. The libc version of this function goes through the PLT and has extra operations to do with setting `errno`. However, using the `__builtin_elementwise_sqrt` (available only in clang), reduces this to a single instruction in x86-64. `__builtin_sqrt` works on gcc/clang but requires the caller to pass in the  `-fno-math-errno` flag. 
- Currently f32 constants are emitted without a "f" suffix, i.e., it is emitted as doubles. Unfortunately, with `-frounding-math`, this means that double to float conversion is done at runtime instead of compile time.

Together these 2 changes shave off another 3% when running a sandboxed libsoundtouch which is used by Firefox to adjust the playback speed of audio files.

### DIFF
--- a/src/c-writer.cc
+++ b/src/c-writer.cc
@@ -792,6 +792,21 @@ static void NormalizeFloatRadix(char* buffer) {
   }
 }
 
+// 32-bit Floating point in C expects a suffix "f" and requires constants like 1
+// to be written as 1.0f (not 1f)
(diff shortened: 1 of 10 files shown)


sbc100 (approved):
Nice!

PR #2824, opened 2026-08-16
Propagate parse errors out of instruction and expression lists

### LINKED ISSUE #2820: Heap use-after-free in WastParser::ResolveTargetTypeVector on malformed text input
## What happens

The wast parser keeps raw `TypeVector*` pointers to block signatures in `resolve_type_vectors_` and only dereferences them once the whole module has been parsed. If a block expression is destroyed on an error path in between, its entry is left dangling and `ResolveTargetTypeVector` reads freed memory.

It's a read of 8 bytes out of a freed 232-byte heap region (`std::vector::size()` on the freed `TypeVector`), reached from ordinary malformed text input. No unusual environment or heap grooming needed — it reproduces every time on a 104 byte file.

## Impact

Any tool or embedder that parses untrusted `.wat` / `.wast` text. Under ASAN it aborts; in a release build it silently reads whatever now occupies the freed block, and then treats the value as a vector length while resolving type names. libwabt is embedded in other programs, so this isn't only a CLI concern.

## Affected

`wat2wasm`, `wast2json` and `wat-desugar` — all three reproduce. Library-wise it's `ParseWatModule` / `ParseWastScript`, i.e. anything going through `WastParser::ParseModuleFieldList`.

## Features required

`--enable-all` reproduces. The relevant one is `--enable-function-references`, since the block signature has to contain a `(ref $t)` for a deferred resolution to be registered at all.

## Reproduce

Tested at afa2acc (main), built with `-fsanitize=address,undefined`, gcc 15.2.

`uaf.wast`, 104 bytes, md5 `8c7d98fae641711ee561e69bc6de4249`:

```wat
(func $n(param $r(ref $t))(result i32)(call_ref $t(block $l(result(ref $t))(n(l g(t f)n l(local.get $r))
```

```console
$ wat2wasm --enable-all uaf.wast -o /dev/null
```

It is deliberately malformed and truncated — that's the point, the block has to fail to parse after its signature has been registered. It came out of a fuzzer and I could not reduce it to something tidier while keeping it reproducing, sorry.

## Trace

```
ERROR: AddressSanitizer: heap-use-after-free on address 0x78da5b9e00f0
READ of size 8 at 0x78da5b9e00f0 thread T0
    #0 std::vector<wabt::Type>::size() const
    #1 wabt::WastParser::ResolveTargetTypeVector(...)  src/wast-parser.cc:1446
    #2 wabt::WastParser::ParseModuleFieldList(wabt::Module*)  src/wast-parser.cc:1481
    #3 wabt::WastParser::ParseModule(...)  src/wast-parser.cc:1296
    #4 wabt::ParseWatModule(...)  src/wast-parser.cc:4195
    #5 ProgramMain(int, char**)  src/tools/wat2wasm.cc:143

freed by thread T0 here:
    #1 wabt::BlockExprBase<(wabt::ExprType)9>::~BlockExprBase()  include/wabt/ir.h:785
    #2 wabt::intrusive_list<wabt::Expr>::clear()  include/wabt/intrusive-list.h:596
    #3 wabt::intrusive_list<wabt::Expr>::~intrusive_list()  include/wabt/intrusive-list.h:288
    #4 wabt::WastParser::ParseExprList(...)  src/wast-parser.cc:3445
    #5 wabt::WastParser::ParseExpr(...)  src/wast-parser.cc:3457
    #6 wabt::WastParser::ParseInstr(...)  src/wast-parser.cc:2330
    #7 wabt::WastParser::ParseInstrList(...)  src/wast-parser.cc:2289
    #8 wabt::WastParser::ParseTerminatingInstrList(...)  src/wast-parser.cc:2309
```

## Why #2805 doesn't cover this

#2805 fixed the same class of bug by remembering the sizes of the three resolve lists in `ParseModuleField` and truncating them again if the field fails to parse.

Here the field doesn't fail. The block dies further down, inside `ParseExprList`, and the expression parsers recover through `Synchronize` and carry on, so the enclosing `func` field still returns `Result::Ok` and the rollback in `ParseModuleField` never runs. The stale entry survives to the resolve loop at the end of `ParseModuleFieldList`.

I have not sent a patch because I am not confident where the rollback belongs. Copying the `ParseModuleField` pattern down to the expression level would be easy to get subtly wrong in the other direction — dropping a resolution for a block that actually was retained would silently lose a type name rather than crash, and the test suite would not notice. Happy to attempt it if you'd like, or leave it to you given you've just been through this code.

## How it was found

Fuzzing the text front end with ASAN, seeded from the spec testsuite. Worth noting that `fuzzers/wat2wasm_fuzzer.cc` calls `ParseWatModule` with a default-constructed `Features`, so no `--enable-*` syntax is ever exercised, and it never parses whole `.wast` scripts. Enabling all features and feeding it scripts turned up this plus two assertion failures (`!IsReferenceWithIndex()` and `Succeeded(result)`) within a few minutes. I'll file those separately.

Nishuuzz:
`wat2wasm`, `wast2json` and `wat-desugar` read freed memory on malformed text input — #2820. The parser holds raw `TypeVector*` pointers to block signatures in `resolve_type_vectors_` and only dereferences them once the whole module has parsed, so a block destroyed on an error path in between leaves a dangling entry behind.

The diagnosis is @zherczeg's, from the issue thread: `ParseModuleField` was returning Ok for a field that had in fact failed. Instrumenting it shows exactly that — the field records 8 errors and still succeeds:

```
[DBG] ParseExprList: ParseExpr failed, synchronising and returning Ok anyway
[DBG] field -> Ok, errors 0 -> 8
```

`ParseInstrList` and `ParseExprList` synchronise past a bad sub-expression and then return `Result::Ok` unconditionally, so the failure never reaches `ParseModuleField` and the rollback added in #2805 never runs.

So this just passes the error back. Both loops still recover and carry on, so every error is still reported — they only remember that something was dropped. `ParseModuleField` then fails as it should, and #2805's rollback discards the stale resolutions.

## The test churn

20 expected outputs get shorter, which is most of the diff: 470 lines removed against 43 added.

Previously, having reported the real error, the parser would carry on and emit a second redundant one — nearly always `unexpected token ), expected EOF` pointing at the end of the module. That no longer appears:

```
 out/test/parse/expr/bad-load-align.txt:3:25: error: unexpected token align=foo, expected ).
 (module (func (i32.load align=foo (i32.const 0))))
                         ^^^^^^^^^
-out/test/parse/expr/bad-load-align.txt:3:50: error: unexpected token ), expected EOF.
-(module (func (i32.load align=foo (i32.const 0))))
-                                                 ^
```

I checked the direction of every one of those diffs: they only ever remove lines. No test gains an error and no valid module starts being rejected. Still, it does touch spec test expectations, so shout if you would rather it were done differently.

## Test

`test/parse/bad-block-ref-result-discarded.txt` is the reproducer from the issue. It is deliberately malformed, and it only *aborts* under a sanitizer, so realistically it is the asan CI jobs that would catch a regression rather than a normal run.

Full suite is green locally against a debug ASAN/UBSAN build with all submodules checked out: 2495 tests, 0 failures.

Fixes #2820.

### DIFF
--- a/src/wast-parser.cc
+++ b/src/wast-parser.cc
@@ -2282,6 +2282,11 @@ Result WastParser::ParseResultList(TypeVector* result_types,
 
 Result WastParser::ParseInstrList(ExprList* exprs) {
   WABT_TRACE(ParseInstrList);
+  // Keep going after a bad instruction so the rest of the errors get reported,
+  // but remember that one was dropped. Returning Ok here would tell the caller
+  // the field parsed cleanly when part of it was thrown away, and anything the
+  // discarded expressions registered for later would outlive them.
+  Result result = Result::Ok;
   ExprList new_exprs;
   while (true) {
     auto pair = PeekPair();
@@ -2289,19 +2294,21 @@ Result WastParser::ParseInstrList(ExprList* exprs) {
       if (Succeeded(ParseInstr(&new_exprs))) {
         exprs->splice(exprs->end(), new_exprs);
       } else {
+        result = Result::Error;
         CHECK_RESULT(Synchronize(IsInstr));
       }
     } else if (IsLparAnn(pair)) {
       if (Succeeded(ParseCodeMetadataAnnotation(&new_exprs))) {
         exprs->splice(exprs->end(), new_exprs);
       } else {
+        result = Result::Error;
         CHECK_RESULT(Synchronize(IsLparAnn));
       }
     } else {
       break;
     }
   }
-  return Result::Ok;
+  return result;
 }
 
 Result WastParser::ParseTerminatingInstrList(ExprList* exprs) {
@@ -3433,15 +3440,17 @@ Result WastParser::ParseBlock(Block* block) {
 
 Result WastParser::ParseExprList(ExprList* exprs) {
   WABT_TRACE(ParseExprList);
+  Result result = Result::Ok;
   ExprList new_exprs;
   while (PeekMatchExpr()) {
     if (Succeeded(ParseExpr(&new_exprs))) {
       exprs->splice(exprs->end(), new_exprs);
     } else {
+      result = Result::Error;
       CHECK_RESULT(Synchronize(IsExpr));
     }
   }
-  return Result::Ok;
+  return result;
 }
 
 Result WastParser::ParseExpr(ExprList* exprs) {
--- a/test/parse/bad-block-ref-result-discarded.txt
+++ b/test/parse/bad-block-ref-result-discarded.txt
@@ -1,0 +1,32 @@
+;;; TOOL: wat2wasm
+;;; ARGS: --enable-function-references
+;;; ERROR: 1
+;; The block registers its result type for later resolution, then fails to
+;; parse and is discarded. Resolution used to run over the freed block.
+(func $n(param $r(ref $t))(result i32)(call_ref $t(block $l(result(ref $t))(n(l g(t f)n l(local.get $r))
+(;; STDERR ;;;
+out/test/parse/bad-block-ref-result-discarded.txt:6:76: error: unexpected token (, expected ).
+...result i32)(call_ref $t(block $l(result(ref $t))(n(l g(t f)n l(local.get $r))
+                                                   ^
+out/test/parse/bad-block-ref-result-discarded.txt:6:77: error: unexpected token n.
+...result i32)(call_ref $t(block $l(result(ref $t))(n(l g(t f)n l(local.get $r))
+                                                    ^
+out/test/parse/bad-block-ref-result-discarded.txt:6:79: error: unexpected token l.
+...result i32)(call_ref $t(block $l(result(ref $t))(n(l g(t f)n l(local.get $r))
+                                                      ^
+out/test/parse/bad-block-ref-result-discarded.txt:6:81: error: unexpected token g.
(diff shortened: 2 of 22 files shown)


zherczeg:
I support this change. @sbc100 what do you think? Do you know what was the reason of the original concept?

Nishuuzz:
I went digging through the history for this, since it's worth knowing whether we'd be undoing something deliberate.

It goes back to the original recursive descent rewrite, #591 (3d3920f6, 2017). That commit's own message describes what it added as "Basic error synchronization; plenty of room for improvement here", so as far as I can tell returning `Ok` here was never a considered decision about what these functions should report — it's just where the first cut landed, and it was flagged as rough at the time.

It was harmless for a long time too. `ParseModuleFieldList` recovers the same way, nothing consumed the result, and the errors were recorded in `errors_` regardless, so the tools still failed correctly. The `Ok` was effectively unobserved.

What changed is #2805, which added `resolve_ref_types_` / `resolve_type_vectors_` and made `ParseModuleField` failing the trigger for discarding the stale entries. That is the first thing that actually depends on this result being accurate, and because these two loops still reported `Ok`, the rollback never ran — which is #2820.

So my reading is that the original concept isn't really being undone, it's being caught up with: #2805 made the return value load-bearing, and this makes it truthful.

sbc100:
lgtm!

Do we still have code for saving an restoring `resolve_type_vectors_` etc?  Can than now be removed maybe?

sbc100 (approved):

PR #2825, opened 2026-08-16
[interp] Fix operand types when tracing table.set

aizu-m:
wasm-interp --trace, on a valid module that stores through a funcref table:

    interp.h:596: assert `t == type || type == ValueType::Any' failed
        Thread::TraceSource::Pick(Index, Instr)   interp.cc:2992
        Istream::Trace(...)                        istream.cc:909

table.set has two operands, an i32 index and a reference value, but opcode.def leaves the value operand Void because its type is taken from the table. Pick estimates the operand count by stripping trailing Void params, so it counts one instead of two and the operand-to-type mapping slips by one: the i32 index is handed the table element type and read back through Value::Get<Ref>(). Under NDEBUG the assert is gone but the trace still prints the index as a bogus funcref.

Set table.set to two operands so the existing GetTableElementType fallback lands on the value operand and the index keeps its i32 type. table.get, table.grow and table.fill already map correctly.

Repro:

    wasm-interp --enable-all --trace --dummy-import-func -r store_fn -a i32:0 mod.wasm

### DIFF
--- a/src/interp/interp.cc
+++ b/src/interp/interp.cc
@@ -2948,6 +2948,13 @@ std::string Thread::TraceSource::Pick(Index index, Instr instr) {
       break;
     }
   }
+  // table.set stores a reference whose type comes from the table, so opcode.def
+  // leaves that top operand's param type Void. The loop above then stops at the
+  // i32 index below it and undercounts, swapping the operand types so the i32
+  // index is read back as a reference. It always has two operands.
+  if (instr.op == Opcode::TableSet) {
+    num_operands = 2;
+  }
   auto type = index > num_operands
                   ? Type(ValueType::Void)
                   : instr.op.GetParamType(num_operands - index + 1);
--- a/src/interp/interp.cc
+++ b/src/interp/interp.cc
@@ -2948,20 +2948,15 @@ std::string Thread::TraceSource::Pick(Index index, Instr instr) {
       break;
     }
   }
-  // table.set stores a reference whose type comes from the table, so opcode.def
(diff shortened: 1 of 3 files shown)


src/interp/interp.cc
```diff
@@ -2948,6 +2948,13 @@ std::string Thread::TraceSource::Pick(Index index, Instr instr) {
       break;
     }
   }
+  // table.set stores a reference whose type comes from the table, so opcode.def
+  // leaves that top operand's param type Void. The loop above then stops at the
```
sbc100: Should we instead update the `opcode.def`?  I'm not sure if there is a some non-void type that we could put there that would make more sense?    Something that signifies a polymorphic type?   Or is that what Void means?

aizu-m: Yes, opcode.def is the better place, done. Void does double duty in that table: trailing Voids mean the operand doesn't exist, while an interior Void means it exists but its type isn't statically known. table.grow and table.fill already rely on the second meaning and Pick resolves it from the table. table.set is the one opcode whose polymorphic operand comes last, so its Void got stripped as a missing operand. I've switched that slot to Type::Any, which type.h already reserves for this sort of thing, and taught Pick to resolve Any the same way as an interior Void, so the special case is gone. The only other readers of the param columns are the type checker's memory ops and the c-writer's binary ops, so nothing else picks up the change. Full suite still passes.

aizu-m:
Rebased. Any update?

sbc100 (approved):

PR #2826, opened 2026-08-16
wasm2c: missing guard force_read error with escape hatch

shravanrn:
This PR ensures that when guard pages are used the force_read macro is either
- explicitly defined for the given platform
- explicitly opted out with `WASM_RT_NONCONFORMING_ALLOW_OOB_READ_ELIMINATION`

rather than silently failing to define it for unknown platforms

This change also allows the embedder to opt to disable force_reads (this is off by default, since it is non conforming). This is independently a performance optimization when Wasm conformance is not required as it speeds up several sandboxing workloads by a large amount

- Sandboxed XML parsing with libExpat by 5%
- Sandboxed Audio speed adjustment with libSoundTouch by 10%

wasm2c is not the first compiler to add the option for an embedder to disable dead reads. Currently Wamr and aWsm enable this by default and don't have a way to disable it, while Wasmer enables it by default and has a switch to disable it. I think wasm2c should remain conformant and not go that far, but it can provide an off-by-default option for this. The actual feature is only about 5 lines so is easy enough to maintain.

### DIFF
--- a/src/template/wasm2c.declarations.c
+++ b/src/template/wasm2c.declarations.c
@@ -165,42 +165,6 @@ static inline uint64_t checked_add_u64(uint64_t a, uint64_t b) {
 #if WASM_RT_MEMCHECK_GUARD_PAGES
 #define MEMCHECK_DEFAULT32(mem, local_memory_size, a, t) \
   WASM_RT_CHECK_BASE(mem);
-
-// When using guard pages, reads have to be immediately consumed so that OOB
-// trap checks are applied in the right place, and not optimized away.
-#ifdef __GNUC__
-#define FORCE_READ_INT(var) __asm__("" ::"r"(var));
-
-#if defined(__i386__) || defined(__x86_64__)
-#if defined(__SSE__)
-#define FORCE_READ_FLOAT_CONSTRAINT "x"
-#else  // No SSE, old x87
-#define FORCE_READ_FLOAT_CONSTRAINT "f"
-#endif
-#elif (defined(__arm__) || defined(__aarch64__)) && defined(__ARM_FP) && \
-    ((__ARM_FP & 0x4) != 0)
-#ifdef __aarch64__
-#define FORCE_READ_FLOAT_CONSTRAINT "w"
-#else
-#define FORCE_READ_FLOAT_CONSTRAINT "t"
-#endif
-#elif defined(__mips__) && defined(__clang__) && !defined(__mips_soft_float)
-// Clang before v18 uses hard floats which has a different constraint
-#define FORCE_READ_FLOAT_CONSTRAINT "f"
-#elif (defined(__powerpc__) || defined(__ppc__) || defined(__PPC__)) || \
-    !(defined(_SOFT_FLOAT) || defined(__NO_FPRS__))
-#define FORCE_READ_FLOAT_CONSTRAINT "f"
-#elif defined(__riscv) && defined(__riscv_flen) && (__riscv_flen >= 32)
-#define FORCE_READ_FLOAT_CONSTRAINT "f"
-#elif defined(__s390__) || defined(__s390x__)
-#define FORCE_READ_FLOAT_CONSTRAINT "f"
-#else
-#define FORCE_READ_FLOAT_CONSTRAINT "r"
-#endif
-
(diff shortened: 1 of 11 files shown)


shravanrn:
@sbc100 I have split out the change as requested. As noted above, this is a change I definitely want to land and use soon due to the performance implications.

shravanrn:
@sbc100 @keithw Just following up. Did y'all have thoughts about landing this? Any chance one of you can review?

test/wasm2c/hello.txt
sbc100: Hmm.. so we end up using this no-op version when WASM_RT_NONCONFORMING_ALLOW_OOB_READ_ELIMINATION?

The fact that we have this no-op version already makes it seem like we already use it on some platforms?

Does that mean we have some platforms that are non-conforming by default today?   Perhaps we should refactor this first so that we get a #error by default and the no-op version if-and-only-if WASM_RT_NONCONFORMING_ALLOW_OOB_READ_ELIMINATION?

shravanrn: @sbc100 A couple of questions here

1. Assuming we do make this change, are we comfortable that the guard page implementation is basically disallowed for any platform that we don't recognize? 

2. Specifically do we want to disallow guard pages in cases where we don't have a force_read implementation i.e., when
  a) compiling with gcc/clang on an unknown CPU architecture
  b) using a non gcc/clang compiler 

**Caveat:** I don't have the full context that @keithw has here. Iirc, there may have been some case such as "visual studio doesn't support the asm force_read, but also doesn't need it because its optimizer doesn't employ the dead load elimination, so the noop is fine".

I don't personally have any opinion/preference here, happy to go with whatever folks like. @sbc100 Let me know!

sbc100: Yes, I think we should error rather than generate a possibly non-forming output.

Thats another good reason to add `WASM_RT_NONCONFORMING_ALLOW_OOB_READ_ELIMINATION` though since it adds an escape hatch for such users.

shravanrn: Gotcha. I can make this change. 

Just to document it for the future, I'll note that the alternate way to enforce FORCE_READ is to cast a loaded pointer to the type like `volatile __attribute__((may_alias, aligned(1))) u32 *` and read (this removes the memcpy in the current i32_load). This is marginally slower than the asm_block approach but works on all architectures for gcc/clang. Unfortunately, since other compilers like msvc doesn't support these attributes, this is still not a general solution.

shravanrn:
@sbc100 I have updated the commit and description to be a reflect that this is general force_read usage cleanup. Please have a look when you get a chance.

keithw (approved):
This all lgtm, with the understanding that we are going to clean up this forest of preprocessor macros in some future version of wasm2c. :-)

src/template/wasm2c.declarations.c
```diff
@@ -220,12 +184,39 @@ static inline uint64_t checked_add_u64(uint64_t a, uint64_t b) {
   WASM_RT_CHECK_BASE(mem);          \
   RANGE_CHECK(mem, a, sizeof(t));
 
-#ifndef FORCE_READ_INT
+// When using guard pages, reads have to be immediately consumed so that OOB
+// trap checks are applied in the right place, and not optimized away. This is
+// done using FORCE_READ_INT/FORCE_READ_FLOAT
+#if WASM_RT_MEMCHECK_BOUNDS_CHECK || \
+    WASM_RT_NONCONFORMING_ALLOW_OOB_READ_ELIMINATION
```
sbc100, line 191: Kind of a shame out macro are so long this kind of things needs a continuations..

sbc100 (approved):

src/template/wasm2c.declarations.c
shravanrn: Agreed :(

shravanrn:
> This all lgtm, with the understanding that we are going to clean up this forest of preprocessor macros in some future version of wasm2c. :-)

Yup, I think we can start on this incrementally. I think the way to tackle this is probably start the split from macros into a compile time flag one-by-one starting with the guard-page vs bounds-check emission. This is mostly cleanly separable --- the only funny edge case is non-64k wasm memory sizes (the code right not does malloc even if mmap is requested which is very confusing). I'll cleanup the macro version of that first in a separate PR, before we start with splitting things into compile time flags

PR #2829, opened 2026-08-19
Support (module definition ...) constructs

zherczeg:
Related patch: #2666


sbc100:
Does this replace #2666?  Is it an alternative implementation?

include/wabt/ir.h
```diff
@@ -1397,17 +1398,21 @@ class ScriptModuleMixin : public ScriptModule {
 class TextScriptModule : public ScriptModuleMixin<ScriptModuleType::Text> {
  public:
   const Location& location() const override { return module.loc; }
+  bool IsDefinition() const override { return is_definition; }
```
sbc100: Any reason to just just make `is_definition` a public field on `ScriptModule`?   i.e. do we need to use a virtual method here?

sbc100 (commented):
Nice!   A surprisingly small patch to make this work.

sbc100 (approved):

zherczeg:
Yes this is a replacement.

include/wabt/ir.h
zherczeg: No. I will change this.

zherczeg: The code is even simpler this way.

src/prebuilt/lexer-keywords.cc
```diff
@@ -1,4 +1,4 @@
-/* C++ code produced by gperf version 3.2.1 */
+/* C++ code produced by gperf version 3.3 */
```
sbc100: I guess we do not pin the exact version that we use here?

sbc100 (approved):

src/prebuilt/lexer-keywords.cc
zherczeg: I don't know. Shall I use an older one?

zherczeg: I have downloaded and compiled a 3.2.1 gperf from here: https://ftp.gnu.org/gnu/gperf/

Anyway the README says this:
```
## Adding new keywords to the lexer

If you want to add new keywords, you'll need to install
[gperf](https://www.gnu.org/software/gperf/). Before you upload your PR, please
run `make update-gperf` to update the prebuilt C++ sources in `src/prebuilt/`.
```

If we want to use a specific gperf version, we should add its version here.

sbc100: Sorry, yes this change seems fine. I was just wondering if we used a pinned version, for consistency, but its seems we don't.  Not a huge deal i think,

zherczeg: Newer versions likely fix some issues. I think the only rule we should use is that we never go back to older gperf. Otherwise file updates will be chaos.

PR #2835, opened 2026-08-25
[interp] Fix wasm-interp stack cleanup for tail calls

KKiiim:
## Summary
Fix incorrect value-stack cleanup when lowering `return_call` and `return_call_indirect` in `wasm-interp`.

In `wasm-interp`, all call frames share a single flat value stack. When performing a tail call, `InterpDropKeep` is emitted to drop the current frame's locals and intermediate stack values while keeping the arguments to be passed to the callee.

`GetReturnCallDropKeepCount` calculates `drop_count` based on the validator's current type-stack height. Previously, `validator_.OnReturnCall*` was invoked before `GetReturnCallDropKeepCount`, which immediately popped the call arguments from the type stack. Consequently, `drop_count` was undercounted by the number of call arguments, leaving dead stack values behind that could corrupt subsequent operations (e.g., being incorrectly observed as the caller's return value).

## Fix
1. In `OnReturnCallExpr` / `OnReturnCallIndirectExpr`, if the function/type index is out of bounds, delegate directly to `validator_.OnReturnCall*` to report the standard validation error.
2. On valid paths, compute `GetReturnCallDropKeepCount` and `GetCatchCount` **before** invoking `validator_.OnReturnCall*`, ensuring the call arguments are still present on the validator's type stack during cleanup calculation.

## Tests
- Added `return-call-stack-cleanup.txt` and `return-call-indirect-stack-cleanup.txt` covering both direct and indirect tail calls with unconsumed stack values.
- Verified against existing interpreter, validation, and spec test suites.

### DIFF
--- a/src/interp/binary-reader-interp.cc
+++ b/src/interp/binary-reader-interp.cc
@@ -1236,10 +1236,17 @@ Result BinaryReaderInterp::OnCallRefExpr(Type sig_type) {
 }
 
 Result BinaryReaderInterp::OnReturnCallExpr(Index func_index) {
-  // Validation below rejects an invalid index before cleanup is emitted.
-  Index param_count = func_index < func_types_.size()
-                          ? func_types_[func_index].params.size()
-                          : 0;
+  // Don't use OnReturnCall to check the index here: it consumes the call
+  // operands from the type stack. That will cause drop/keep counts to be
+  // incorrect.
+  if (func_index >= func_types_.size()) {
+    validator_.PrintError(GetLocation(),
+                          "function variable out of range: %" PRIindex
+                          " (max %" PRIindex ")",
+                          func_index, static_cast<Index>(func_types_.size()));
+    return Result::Error;
+  }
+  Index param_count = func_types_[func_index].params.size();
 
   Index drop_count, keep_count, catch_drop_count;
   // Validation consumes the tail-call operands, so compute cleanup first.
@@ -1247,6 +1254,8 @@ Result BinaryReaderInterp::OnReturnCallExpr(Index func_index) {
       GetReturnCallDropKeepCount(param_count, 0, &drop_count, &keep_count));
   CHECK_RESULT(
       validator_.GetCatchCount(label_stack_.size() - 1, &catch_drop_count));
+  // The validator must be run after we get the drop/keep counts, since it
+  // will change the type stack.
   CHECK_RESULT(
       validator_.OnReturnCall(GetLocation(), Var(func_index, GetLocation())));
   istream_.EmitDropKeep(drop_count, keep_count);
@@ -1268,10 +1277,18 @@ Result BinaryReaderInterp::OnReturnCallExpr(Index func_index) {
 
 Result BinaryReaderInterp::OnReturnCallIndirectExpr(Index sig_index,
                                                     Index table_index) {
-  // Validation below rejects an invalid index before cleanup is emitted.
-  Index param_count = sig_index < module_.func_types.size()
-                          ? module_.func_types[sig_index].params.size()
-                          : 0;
+  // Don't use OnReturnCallIndirect to check the index here: it consumes the
+  // call operands from the type stack. That will cause drop/keep counts to be
+  // incorrect.
+  if (sig_index >= module_.func_types.size()) {
+    validator_.PrintError(GetLocation(),
+                          "function type variable out of range: %" PRIindex
+                          " (max %" PRIindex ")",
+                          sig_index,
+                          static_cast<Index>(module_.func_types.size()));
+    return Result::Error;
+  }
+  Index param_count = module_.func_types[sig_index].params.size();
 
   Index drop_count, keep_count, catch_drop_count;
   // +1 to include the index of the function.
@@ -1280,6 +1297,8 @@ Result BinaryReaderInterp::OnReturnCallIndirectExpr(Index sig_index,
       GetReturnCallDropKeepCount(param_count, +1, &drop_count, &keep_count));
   CHECK_RESULT(
       validator_.GetCatchCount(label_stack_.size() - 1, &catch_drop_count));
+  // The validator must be run after we get the drop/keep counts, since it
+  // will change the type stack.
   CHECK_RESULT(validator_.OnReturnCallIndirect(
       GetLocation(), Var(sig_index, GetLocation()),
       Var(table_index, GetLocation())));
--- a/src/interp/binary-reader-interp.cc
+++ b/src/interp/binary-reader-interp.cc
@@ -319,6 +319,7 @@ class BinaryReaderInterp : public BinaryReaderNop {
                             Index* out_keep_count);
   Result GetReturnDropKeepCount(Index* out_drop_count, Index* out_keep_count);
   Result GetReturnCallDropKeepCount(const FuncType&,
+                                    size_t orig_type_stack_size,
                                     Index keep_extra,
                                     Index* out_drop_count,
                                     Index* out_keep_count);
@@ -456,13 +457,17 @@ Result BinaryReaderInterp::GetReturnDropKeepCount(Index* out_drop_count,
   return Result::Ok;
 }
 
-Result BinaryReaderInterp::GetReturnCallDropKeepCount(const FuncType& func_type,
-                                                      Index keep_extra,
-                                                      Index* out_drop_count,
-                                                      Index* out_keep_count) {
+Result BinaryReaderInterp::GetReturnCallDropKeepCount(
+    const FuncType& func_type,
+    size_t orig_type_stack_size,
+    Index keep_extra,
+    Index* out_drop_count,
+    Index* out_keep_count) {
   Index keep_count = static_cast<Index>(func_type.params.size()) + keep_extra;
-  CHECK_RESULT(GetDropCount(keep_count, 0, out_drop_count));
-  *out_drop_count += validator_.GetLocalCount();
+  *out_drop_count =
+      (orig_type_stack_size >= keep_count ? orig_type_stack_size - keep_count
+                                          : 0) +
+      validator_.GetLocalCount();
   *out_keep_count = keep_count;
   return Result::Ok;
 }
@@ -1236,25 +1241,18 @@ Result BinaryReaderInterp::OnCallRefExpr(Type sig_type) {
 }
 
 Result BinaryReaderInterp::OnReturnCallExpr(Index func_index) {
-  // Do not unconditionally call OnReturnCall here: it consumes the call
-  // operands from the type stack, which would make subsequent drop/keep count
-  // calculations incorrect. We only call it here to report a standard
-  // validation error and bail out when the function index is out of bounds.
-  if (func_index >= func_types_.size()) {
-    return validator_.OnReturnCall(GetLocation(),
-                                   Var(func_index, GetLocation()));
-  }
+  // Capture type stack size before validation sets the block to unreachable.
+  size_t orig_type_stack_size = validator_.type_stack_size();
+  CHECK_RESULT(
+      validator_.OnReturnCall(GetLocation(), Var(func_index, GetLocation())));
+
   FuncType& func_type = func_types_[func_index];
 
   Index drop_count, keep_count, catch_drop_count;
-  CHECK_RESULT(
-      GetReturnCallDropKeepCount(func_type, 0, &drop_count, &keep_count));
+  CHECK_RESULT(GetReturnCallDropKeepCount(func_type, orig_type_stack_size, 0,
+                                          &drop_count, &keep_count));
   CHECK_RESULT(
       validator_.GetCatchCount(label_stack_.size() - 1, &catch_drop_count));
-  // The validator must be run after we get the drop/keep counts, since it
-  // will change the type stack.
-  CHECK_RESULT(
-      validator_.OnReturnCall(GetLocation(), Var(func_index, GetLocation())));
   istream_.EmitDropKeep(drop_count, keep_count);
   istream_.EmitCatchDrop(catch_drop_count);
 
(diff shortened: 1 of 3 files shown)


src/interp/binary-reader-interp.cc
```diff
@@ -1236,18 +1236,18 @@ Result BinaryReaderInterp::OnCallRefExpr(Type sig_type) {
 }
 
 Result BinaryReaderInterp::OnReturnCallExpr(Index func_index) {
-  CHECK_RESULT(
-      validator_.OnReturnCall(GetLocation(), Var(func_index, GetLocation())));
-
-  FuncType& func_type = func_types_[func_index];
+  Index param_count;
+  if (!validator_.GetFuncParamCount(func_index, &param_count)) {
```
sbc100: Why pull this information from the validator if we have our own func_types_ array?  

IIUC that validator is not normally the source of this kind of information is it?

src/interp/binary-reader-interp.cc
```diff
@@ -1236,18 +1236,18 @@ Result BinaryReaderInterp::OnCallRefExpr(Type sig_type) {
...
+    return validator_.OnReturnCall(GetLocation(),
+                                   Var(func_index, GetLocation()));
```
sbc100: Can you add a comment as to why we have this early return when param count is zero?   Can't the generic code below handle zero params?

KKiiim:
Yes. Thanks for your advice.

Current change: Compute and emit the correct `drop_keep` cleanup before validation consumes the tail-call arguments.

This bug can be reproduced using the function `f`. `f` should return `0`, not `777`.

```
(module
  (type $i_i (func (param i32) (result i32)))
  (import "host" "print" (func $imported (type $i_i)))

  (func $tail (result i32)
    i32.const 777
    i32.const 0
    return_call $imported)

  (func (export "f") (result i32)
    (local i32)
    call $tail
    local.get 0
    return)
)
```

KKiiim:
@sbc100 Hi, could you help to take a look. Thanks!

src/interp/binary-reader-interp.cc
```diff
@@ -1236,18 +1236,17 @@ Result BinaryReaderInterp::OnCallRefExpr(Type sig_type) {
...
+  // Validation consumes the tail-call operands, so compute cleanup first.
   CHECK_RESULT(
-      GetReturnCallDropKeepCount(func_type, 0, &drop_count, &keep_count));
+      GetReturnCallDropKeepCount(param_count, 0, &drop_count, &keep_count));
   CHECK_RESULT(
       validator_.GetCatchCount(label_stack_.size() - 1, &catch_drop_count));
-  // The validator must be run after we get the drop/keep counts, since it
-  // will change the type stack.
   CHECK_RESULT(
       validator_.OnReturnCall(GetLocation(), Var(func_index, GetLocation())));
```
sbc100: So the old code had two calls to `validator_.OnReturnCall`..  and the new code only has one.   Was this just a bug in the old code?  Are you sure its OK to change like this?

src/interp/binary-reader-interp.cc
```diff
@@ -1269,20 +1268,18 @@ Result BinaryReaderInterp::OnReturnCallExpr(Index func_index) {
...
                                                     Index table_index) {
-  CHECK_RESULT(validator_.OnReturnCallIndirect(
-      GetLocation(), Var(sig_index, GetLocation()),
-      Var(table_index, GetLocation())));
-
-  FuncType& func_type = module_.func_types[sig_index];
+  // Validation below rejects an invalid index before cleanup is emitted.
+  Index param_count = sig_index < module_.func_types.size()
+                          ? module_.func_types[sig_index].params.size()
+                          : 0;
```
sbc100: So is valid for `OnReturnCallIndirectExpr` to be called with an invalid `sig_index` here?

Should we instead just return an error here `sig_index` is not in range?  

Or maybe we can move `validator_.OnReturnCallIndirect` up to the top of the function would mean we return an early error.

IIUC BinaryReaderInterp no longer need to support continuing past validation errors.

KKiiim: Yes. Return error early for `sig_index`  will be more straight-forward. I fix it. Thanks.

src/interp/binary-reader-interp.cc
KKiiim: Yes, I added a comment to explain it. The root cause is the first `OnReturnCall` was originally added to fix malformed-module issue. But it is incorrect. `OnReturnCall` can't be used to check the index since it consumes the call operands from the type stack.

src/interp/binary-reader-interp.cc
```diff
@@ -456,11 +456,11 @@ Result BinaryReaderInterp::GetReturnDropKeepCount(Index* out_drop_count,
   return Result::Ok;
 }
 
-Result BinaryReaderInterp::GetReturnCallDropKeepCount(const FuncType& func_type,
+Result BinaryReaderInterp::GetReturnCallDropKeepCount(Index param_count,
                                                       Index keep_extra,
                                                       Index* out_drop_count,
                                                       Index* out_keep_count) {
-  Index keep_count = static_cast<Index>(func_type.params.size()) + keep_extra;
+  Index keep_count = param_count + keep_extra;
```
sbc100: I think we can revert this change now?

src/interp/binary-reader-interp.cc
```diff
@@ -1269,20 +1277,28 @@ Result BinaryReaderInterp::OnReturnCallExpr(Index func_index) {
...
 Result BinaryReaderInterp::OnReturnCallIndirectExpr(Index sig_index,
                                                     Index table_index) {
-  CHECK_RESULT(validator_.OnReturnCallIndirect(
-      GetLocation(), Var(sig_index, GetLocation()),
-      Var(table_index, GetLocation())));
-
-  FuncType& func_type = module_.func_types[sig_index];
+  // Don't use OnReturnCallIndirect to check the index here: it consumes the
+  // call operands from the type stack. That will cause drop/keep counts to be
+  // incorrect.
```
sbc100: I still don't quite understand this comment, but that change seems correct.

I guess the problem was that were were calling `validator_.OnReturnCallIndirect` twice in the old codepath and that is wrong/broken?

src/interp/binary-reader-interp.cc
```diff
@@ -1236,14 +1236,22 @@ Result BinaryReaderInterp::OnCallRefExpr(Type sig_type) {
...
 
 Result BinaryReaderInterp::OnReturnCallExpr(Index func_index) {
-  CHECK_RESULT(
-      validator_.OnReturnCall(GetLocation(), Var(func_index, GetLocation())));
-
-  FuncType& func_type = func_types_[func_index];
+  // Don't use OnReturnCall to check the index here: it consumes the call
+  // operands from the type stack. That will cause drop/keep counts to be
+  // incorrect.
+  if (func_index >= func_types_.size()) {
```
zherczeg: This should be a validation error.

src/interp/binary-reader-interp.cc
```diff
@@ -1269,20 +1277,28 @@ Result BinaryReaderInterp::OnReturnCallExpr(Index func_index) {
...
                                                     Index table_index) {
-  CHECK_RESULT(validator_.OnReturnCallIndirect(
-      GetLocation(), Var(sig_index, GetLocation()),
-      Var(table_index, GetLocation())));
-
-  FuncType& func_type = module_.func_types[sig_index];
+  // Don't use OnReturnCallIndirect to check the index here: it consumes the
+  // call operands from the type stack. That will cause drop/keep counts to be
+  // incorrect.
+  if (sig_index >= module_.func_types.size()) {
```
zherczeg: This is also sounds like a validation error.

zherczeg (commented):
I don't understand the description. These are return calls, there is nothing on the stack after they are executed. Is this a call frame problem? I would like a bit more information what is happening here.

zherczeg:
`This bug can be reproduced using the function f. f should return 0, not 777.` - this is also unclear to me. Normally, neither 0 nor 777 should be returned. The imported function should return with something, and the rest should be dropped.

zherczeg:
Current code:
https://github.com/WebAssembly/wabt/blob/main/src/interp/binary-reader-interp.cc#L1239
The OnReturnCallExpr calls `validator_.OnReturnCall` twice. The first one was added to avoid some fuzzer bugs. Both modify the type stack. I don't understand why this is correct.

I don't know the code, but I suspect the problem is the following. The original code called `validator_.OnReturnCall` after calling `GetReturnCallDropKeepCount`. However, this depends on a validated `func_index`. A fuzzer found this, then the first `validator_.OnReturnCall` was added at the beginning of the function, as it should be. I think the solution is removing the second `validator_.OnReturnCall`, and `GetReturnCallDropKeepCount` should use the results of the called function, not the params.

zherczeg:
Btw you could also check `OnReturnCallRefExpr`. The catch update is missing there, so the code is probably just wrong.

KKiiim:
@zherczeg Thanks for the review and questions! Let me clarify what is happening here:

1. **Why `InterpDropKeep` is needed for `return_call` (stack / call frame cleanup):**
   In `wasm-interp`, all call frames share a single flat value stack (`values_`). When a function performs a tail call, caller-allocated locals and unconsumed temporary operands on the stack below the outgoing arguments need to be discarded, while keeping the callee arguments on top of the stack. `InterpDropKeep` shifts the kept arguments down to overwrite the current frame's locals/temporaries before transferring control.

2. **Why `keep_count` must be the callee's `params`, not `results`:**
   At the point of executing a `return_call`, the callee has not been executed yet. The values currently on top of the value stack are the *arguments/parameters* to be passed into the callee (plus the table index for indirect calls). If we kept `results` instead, any arguments beyond the result count would be dropped before entering the callee.

3. **Why `validator_.OnReturnCall` cannot be called before computing `drop_count`:**
   `validator_.OnReturnCall` mutates the validator's `type_stack_` by popping the call parameters and marking the block unreachable. If we validate first, `type_stack_size()` is already decremented by `param_count`, causing `GetReturnCallDropKeepCount` to undercount `drop_count` and leave dead stack values behind (which was the bug fixed in this PR).
   Therefore, we must compute `GetReturnCallDropKeepCount` while the operands are still on the type stack, followed by `validator_.OnReturnCall`.

4. **Regarding `OnReturnCallRefExpr`:**
   Good catch. `OnReturnCallRefExpr` is currently missing full `drop_keep` and `catch_drop` handling in the interpreter. But that's not the purpose of the current bug fix.

src/interp/binary-reader-interp.cc
KKiiim: Agreed, you're right.

I've updated the code so that out-of-bounds indices directly invoke `validator_.OnReturnCall(...)` (and `validator_.OnReturnCallIndirect(...)`) to report the standard validation error, rather than manually formatting and printing errors in the reader.

src/interp/binary-reader-interp.cc
KKiiim: Call `validator_.OnReturnCallIndirect` twice in the old codepath  is wrong. I update the description.

zherczeg:
In general, the validator should run first, because everything is assumed to be correct after it. Please don't forget the validator is not only used by this project, and changing this general rule can cause confusion.

Ideally, something like this should happen:
1) Run the validator. Then it consumes the params of the called function from the type stack, and replace them with the results.
2) If validation passes, you know the results are on the type stack, and you can compute the number of unnecessary values using the result size (that was my recommended change).
3) The number of unnecessary values are the same before the call, so remove them. Also drop the catches.

KKiiim:
Thanks for the feedback! I understand the design principle that validation should ideally precede lowering. However unlike a regular `call`, `return_call` does not push `results` onto the type stack. Instead, `TypeChecker::OnReturnCall` marks the block unreachable and immediately clears the type stack via `SetUnreachable()`. 

At runtime, the interpreter stack also holds the outgoing `params` (not `results`), which `InterpDropKeep` needs to keep while dropping values below them. 

Because `OnReturnCall` pops those params and resets the type stack, we must compute the drop/keep count before `OnReturnCall` mutates the stack. If the index is invalid, we bail directly into `OnReturnCall` so validation still handles the error.

zherczeg:
Maybe an example could help. Stack below grows left from right, `t(i)`  represents removable values, `p(i)` represents parameters, `r(i)` represents results. Lets assume the called function has 3 parameters, and two results.

After the validation the type stack changes the following way:
`[t0, t1, t2, p0, p1,p2]` -> `[t0, t1, t2, r0, r1]`

Since you know the function has two results, and the type stack size is 5, you can compute that 3 (=5-2) values will never be used (and should be discarded). You also know, this is also true, before the call is executed. Copy the values back, and the stack now contains [p0, p1,p2]. Then do the call.

zherczeg:
An idea: store the `validator_.GetLocalCount()` in a local variable before the validation. This operation does not need any checks. Then perform the validation. Then the function is validated, so the dropped values can be computed. Needs some code refactor.

src/interp/binary-reader-interp.cc
```diff
@@ -456,13 +457,17 @@ Result BinaryReaderInterp::GetReturnDropKeepCount(Index* out_drop_count,
...
+    size_t orig_type_stack_size,
+    Index keep_extra,
+    Index* out_drop_count,
+    Index* out_keep_count) {
   Index keep_count = static_cast<Index>(func_type.params.size()) + keep_extra;
-  CHECK_RESULT(GetDropCount(keep_count, 0, out_drop_count));
-  *out_drop_count += validator_.GetLocalCount();
+  *out_drop_count =
+      (orig_type_stack_size >= keep_count ? orig_type_stack_size - keep_count
+                                          : 0) +
```
KKiiim, line 469: Same to `GetDropCount`.
  // The keep_count may be larger than the type_stack_count if the typechecker
  // is currently unreachable. In that case, it doesn't matter what value we
  // drop, but 0 is a reasonable choice.

zherczeg (approved):
LGTM

KKiiim:
@sbc100 Hi, could you take a look again. Looking forward to a new release, thanks

KKiiim:
@sbc100 @zherczeg Hi maintainers. Could you release a new version? Thanks.

sbc100:
https://github.com/WebAssembly/wabt/pull/2859

PR #2836, opened 2026-08-25
Support u64 limits for wasm 3.0

zherczeg:
https://github.com/WebAssembly/memory64/blob/main/proposals/memory64/Overview.md#structure

The `limits` structure uses u64 values, regardless the table/memory is i32.

This is also true for WebAssembly 3.0:
https://webassembly.github.io/spec/core/binary/types.html#binary-limits

Hence `(memory 0x1_0000_0000)` and `(table 0x1_0000_0000 funcref))` are not-malformed, only invalid now. This patch tries to support this behavior.

### DIFF
--- a/src/binary-reader.cc
+++ b/src/binary-reader.cc
@@ -640,9 +640,12 @@ Result BinaryReader::ReadTable(Limits* out_elem_limits) {
            "memory64 not allowed");
   ERROR_UNLESS(unknown_flags == 0, "malformed table limits flag: %d", flags);
 
-  CHECK_RESULT(ReadU32OrU64Leb128(&initial, is_64, "table initial elem count"));
+  CHECK_RESULT(ReadU32OrU64Leb128(&initial,
+                                  options_.features.memory64_enabled(),
+                                  "table initial elem count"));
   if (has_max) {
-    CHECK_RESULT(ReadU32OrU64Leb128(&max, is_64, "table max elem count"));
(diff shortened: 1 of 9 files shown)


src/binary-reader.cc
```diff
@@ -640,9 +640,12 @@ Result BinaryReader::ReadTable(Limits* out_elem_limits) {
...
   ERROR_UNLESS(unknown_flags == 0, "malformed table limits flag: %d", flags);
 
-  CHECK_RESULT(ReadU32OrU64Leb128(&initial, is_64, "table initial elem count"));
+  CHECK_RESULT(ReadU32OrU64Leb128(&initial,
+                                  options_.features.memory64_enabled(),
+                                  "table initial elem count"));
   if (has_max) {
-    CHECK_RESULT(ReadU32OrU64Leb128(&max, is_64, "table max elem count"));
+    CHECK_RESULT(ReadU32OrU64Leb128(&max, options_.features.memory64_enabled(),
+                                    "table max elem count"));
```
sbc100, line 648: Should we just always read this as u64 now?   

I guess the natural followup is should have features like memory64 be always-enabled these days?  (i.e. now that its stage5?)

sbc100 (approved):

sbc100 (approved):

zherczeg:
Sorry, I just saw that there was an update with enable flags. I updated the tests, and now everything green.

sbc100 (approved):

PR #2837, opened 2026-08-25
Enable supported stage 5 proposals by default

sbc100:
Enable the following Stage 5 / finished WebAssembly proposals by
default:
- `exceptions` (Exception handling)
- `tail-call` (Tail call)
- `annotations` (Custom annotation syntax)
- `memory64` (64-bit memory)
- `multi-memory` (Multiple memories)
- `extended-const` (Extended constant expressions)
- `relaxed-simd` (Relaxed SIMD)

Updates default flags in `include/wabt/feature.def` and dependency
propagation in `src/feature.cc`. Updates `--disable-*` / `--enable-*`
command line flags across all tools, test runner scripts, and
`README.md`. Regenerates man pages and documentation.


dschuff (approved):
I think we could/should even enable stage 4 proposals. those are clear for implementation by browsers, so I don't see any reason to have them disabled by default in wabt.

PR #2840, opened 2026-08-28
Improve annotation parsing

zherczeg:


zherczeg:
This test runs with these changes:
https://github.com/WebAssembly/testsuite/blob/d76759e746f3564a03f6106ae19679742f2a1831/proposals/wasm-3.0/annotations.wast

src/wast-lexer.cc
```diff
@@ -98,8 +103,7 @@ Token WastLexer::GetToken() {
           return BareToken(TokenType::Eof);
         } else {
           ReadChar();
-          ERROR("unexpected char");
-          continue;
+          return TextToken(TokenType::Reserved);
```
zherczeg, line 108: The `;` is a reserved character in 3.0
https://webassembly.github.io/spec/core/text/lexical.html#characters

It was not in 2.0, but annotations cannot be checked in lexer.
https://www.w3.org/TR/2025/CRD-wasm-core-2-20250616/#tokens%E2%91%A0

This change should not cause any issues, just yield a different error.

src/wast-lexer.cc
```diff
@@ -176,16 +180,23 @@ Token WastLexer::GetToken() {
...
+
+        if (IsKeyword(c)) {
           return GetKeywordToken();
-        } else if (IsIdChar(PeekChar())) {
+        } else if (IsIdChar(c)) {
           return GetReservedToken();
+        } else if (c == ',' || c == ';' || c == '[' || c == ']' || c == '{' ||
+                   c == '}') {
+          ReadChar();
+          return TextToken(TokenType::Reserved);
```
zherczeg: Same story as above.

test/parse/bad-annotation_id1.txt
```diff
@@ -0,0 +1,14 @@
+;;; TOOL: wat2wasm
+;;; ERROR: 1
+(module
+  (@a (@))
+  (@))
```
zherczeg, line 5: Each test contains a valid and an invalid case.

src/wast-lexer.cc
```diff
@@ -77,7 +77,15 @@ Token WastLexer::GetToken() {
...
           return BareToken(TokenType::Eof);
         } else if (MatchString("(@")) {
-          GetIdChars();
+          assert(token_start_ + 2 == cursor_);
+          if (PeekChar() == '"') {
+            // Ignore the "(@" prefix
+            return GetStringToken(TokenType::LparAnn);
+          }
+
+          while (IsIdChar(PeekChar())) {
```
zherczeg, line 85: `GetIdChars()` accepts `"`, while we should not.

zherczeg:
Unfortunately this patch adds a lot of minor changes to improve annotation support. Errors remain errors, but their description changes. Error sequences as well, so a stray `;` is now an  invalid token, and not an ignored unknown character.

To keep the old operation we could add a `bool enable_annotations` to the lexer. The question is, do we really need to keep the old operation?

I removed the empty file support, since I could not figure out a good way to handle it. The problem is:
`(module quote "(@a \09)")` 
Such spec tests are valid empty modules. However, we got an error.

An option to fix it is adding a `bool`, which is set to `true` if annotations are encountered. Then the file is not considered empty. I don't really like it, but no better idea at the moment.

@sbc100 what do you think about these changes?

src/wast-lexer.cc
```diff
@@ -331,7 +344,8 @@ Token WastLexer::GetStringToken(TokenType token_type) {
   while (in_string) {
     switch (ReadChar()) {
       case kEof:
-        return BareToken(TokenType::Eof);
+        token_start_ = saved_token_start;
```
zherczeg, line 351: The old operation could be kept with an `if (token_type == TokenType::LparAnn)` if needed. I don't know if it is needed.

zherczeg:
@sbc100 Some good news: if the annotation support is fixed, and the gc patches are applied, the whole wasm 3.0 spec test suite is passed.

sbc100 (approved):

PR #2842, opened 2026-09-01
wasm-interp: remove WASIp1 support

Sn0wyDay:
## Why

The interpreter's partial WASIp1 integration has lifecycle limitations. Following the discussion in this PR, this replaces the original missing-instance guard with the agreed removal of WASIp1 support from `wasm-interp`.

Agreed scope: https://github.com/WebAssembly/wabt/pull/2842#issuecomment-5670774069
Maintainer confirmation: https://github.com/WebAssembly/wabt/pull/2842#issuecomment-5684857081

## Changes

- Remove WASI initialization, import binding, and `_start` execution from `wasm-interp`, along with `--wasi`, `--env`/`-e`, `--dir`/`-d`, and trailing WASI program arguments.
- Remove `interp-wasi.cc`, `wasi_api.def`, the installed `interp-wasi.h` header, and the obsolete WASI tests and test-runner configuration.
- Update CLI help, the man page, and its HTML version; add a regression test covering rejection of the removed options and trailing arguments.
- Preserve the uvwasi submodule and provide `BUILD_UVWASI=ON` for the standalone wasm2c Dhrystone benchmark. The explicit `uvwasi_a` target builds uvwasi/libuv independently of `BUILD_TOOLS`; they are not linked into `wasm-interp`.
- Document the CLI/header changes and benchmark build instructions. Existing `WITH_WASI=ON` configurations emit an explicit warning and continue without interpreter WASI support.

## Validation

Local Linux x86-64 validation with GCC 13.3 and CMake 3.31.10:

- Release build and full end-to-end suite: **1,995 passed, 0 failed** (including round-trip cases).
- Unit tests: **133 passed, 2 skipped**; C API examples: **10 passed**.
- Separate Debug build with `WERROR=ON`, `WITH_WASI=ON`, and `BUILD_UVWASI=ON`: passed; interpreter/help tests: **87 passed**. Confirmed the legacy-option warning is emitted.
- Built `uvwasi_a` and `uv_a` with `BUILD_TOOLS=OFF`, `BUILD_TESTS=OFF`, and `BUILD_LIBWASM=OFF`.
- Compiled Dhrystone to Wasm with WASI SDK 20, translated it with the updated `wasm2c`, linked it with the preserved uvwasi/libuv libraries using GCC, and ran it successfully; its result checks reported `O.K.`.
- Clean Release installation excludes `interp-wasi.h` and preserves the normal interpreter headers. Both interpreter binaries contain no uvwasi/WASI-integration symbols; enabling `BUILD_UVWASI` does not add uvwasi to the default build.
- Changed Python file passes flake8 and Ruff; changed C++ lines pass clang-format; `git diff --check` passes.

An additional Release build with `WERROR=ON` encountered an existing `unused-but-set-variable` warning in the unchanged `src/tools/spectest-interp.cc` (variable `ok`). The normal Release build and the separate Debug build with `WERROR=ON` passed. macOS/Windows and other CI configurations have not been run locally.

### DIFF
--- a/src/interp/interp-wasi.cc
+++ b/src/interp/interp-wasi.cc
@@ -635,7 +635,12 @@ std::unordered_map<Instance*, WasiInstance*> wasiInstances;
                      Trap::Ptr* trap) {                                     \
     Instance* instance = thread.GetCallerInstance();                        \
     assert(instance);                                                       \
-    WasiInstance* wasi_instance = wasiInstances[instance];                  \
+    auto iter = wasiInstances.find(instance);                               \
+    if (iter == wasiInstances.end()) {                                      \
+      *trap = Trap::New(thread.store(), "WASI instance is not running");    \
+      return Result::Error;                                                 \
+    }                                                                       \
+    WasiInstance* wasi_instance = iter->second;                             \
     if (wasi_instance->trace_stream) {                                      \
       wasi_instance->trace_stream->Writef(                                  \
           ">>> running wasi function \"%s\":\n", #NAME);                    \
--- a/test/wasi/run-export.txt
+++ b/test/wasi/run-export.txt
@@ -1,0 +1,13 @@
+;;; TOOL: run-interp-wasi
+;;; ARGS1: --run-export=foo
+
+(import "wasi_snapshot_preview1" "args_sizes_get"
+  (func $args_sizes_get (param i32 i32) (result i32)))
+(memory (export "memory") 1)
+(func (export "foo")
+  (call $args_sizes_get (i32.const 0) (i32.const 4))
+  drop)
+(func (export "_start"))
+(;; STDOUT ;;;
+foo() => error: WASI instance is not running
+;;; STDOUT ;;)


src/interp/interp-wasi.cc
```diff
@@ -635,7 +635,12 @@ std::unordered_map<Instance*, WasiInstance*> wasiInstances;
                      Trap::Ptr* trap) {                                     \
     Instance* instance = thread.GetCallerInstance();                        \
     assert(instance);                                                       \
-    WasiInstance* wasi_instance = wasiInstances[instance];                  \
+    auto iter = wasiInstances.find(instance);                               \
+    if (iter == wasiInstances.end()) {                                      \
+      *trap = Trap::New(thread.store(), "WASI instance is not running");    \
+      return Result::Error;                                                 \
+    }                                                                       \
+    WasiInstance* wasi_instance = iter->second;                             \
```
zherczeg, line 643: This looks like a perf reduction.

Since wasi support should be enabled before other things, would it be better to move the `WasiRunStart` before `RunAllExports`?

https://github.com/WebAssembly/wabt/blob/main/src/tools/wasm-interp.cc#L424

Sn0wyDay: Thanks for the suggestion. I tested moving `WasiRunStart()` before the export calls.

That does establish the intended `_start` ordering, but it does not prevent the crash by itself. `WasiRunStart()` removes the entry from `wasiInstances` before returning, so a subsequent exported function that calls a WASI import still encounters a missing instance and segfaults.

I also found an ordering-independent reproducer: a WebAssembly start section can call a WASI import during `InstantiateModule()`, before `WasiRunStart()` can be called. The current code terminates with SIGSEGV, while the missing-instance check returns a controlled `WASI instance is not running` error.

I can add that case as a regression test and mark the failure branch with `WABT_UNLIKELY` to minimize the normal-path cost. To keep this PR focused, I suggest retaining the defensive check here and handling full `--wasi` plus `--run-export` lifecycle support separately.

zherczeg: Am I understand correctly, that wasi functions only alive during the execution of the `start_` function? As for me, wasi should start before the module is executed, and alive to the end. Otherwise calling the imports will crash.

Sn0wyDay: Yes, that is correct. Currently, `WasiInstance` is a local variable in `WasiRunStart()`. It is registered immediately before `_start` is called and removed immediately afterward, so WASI imports are only usable during `_start`.

I tested a local prototype that keeps the WASI context alive for the complete execution phase:

- Call `WasiRunStart()` before `RunAllExports()` and `RunSpecificExports()`.
- Store the `WasiInstance` as an owned object so it remains valid after `WasiRunStart()` returns.
- Remove it only after all requested exports have finished.
- Skip `_start` in `RunAllExports()` because it has already been invoked.
- Retain a controlled missing-context trap for calls made during `InstantiateModule()`.

With the original code, the `--wasi --run-export=foo` reproducer terminates with SIGSEGV. With the full-lifetime prototype, `_start` runs first and `foo` successfully calls the WASI import. All seven WASI tests, including the new regression cases, pass.

One remaining detail is that a WebAssembly start section runs during `InstantiateModule()`, before the current implementation can obtain the instantiated exported memory. Supporting WASI imports at that stage would require a deeper initialization redesign; the defensive check currently turns that case into a controlled error instead of a crash.

I can update this PR with the tested lifetime change if this matches the intended design.

zherczeg: @sbc100 what shall we do? The problem is that wasi support is currently limited to the execution of `start_` function. This cause all kinds of issues (crashes), when exports are explicitly executed. There are two alternatives:
1) Keep the limitation, and trap when a wasi import is executed outside of `start_` function.
2) Keep wasi alive during the execution of the whole module

I would prefer 2).

sbc100:
Can we instead remove the wasi support from the interpreter?   wasip1 is very old and this point anyway and I'm not sure how good this implementation ever really was.

Sn0wyDay:
Yes, removing the interpreter’s WASIp1 support sounds reasonable. I’d scope that to the `wasm-interp` integration (`--wasi`, `--env`, and `--dir`), the `interp-wasi` sources and installed header, and the related tests and documentation.

One dependency needs to be handled separately: `wasm2c/benchmarks/dhrystone` also uses the bundled uvwasi/libuv build, so removing the interpreter integration should not remove uvwasi blindly. Homebrew and MSYS2 also currently build WABT with `-DWITH_WASI=ON`, so I’ll make the CLI and build-option compatibility change explicit.

Does that scope match what you had in mind?

sbc100:
SGTM

Sn0wyDay:
Updated in c818b0ee7384ce1720c0cf84eac5e18597d4eae2 to remove the interpreter's WASIp1 support as agreed. This supersedes the original missing-instance-guard approach.

- Removed the interpreter WASI implementation and installed header, `--wasi`, `--env`/`-e`, `--dir`/`-d`, trailing WASI arguments, and the obsolete tests. Updated help/man pages and added a test for rejection of the removed options.
- Preserved uvwasi/libuv for the standalone wasm2c Dhrystone benchmark behind `BUILD_UVWASI=ON`. `WITH_WASI=ON` now emits an explicit migration warning; it does not restore interpreter WASI support.
- Added compatibility notes and benchmark build instructions.

Validation on Linux x86-64 (GCC 13.3, CMake 3.31.10):
- Release full suite: **1,995 passed**.
- Unit tests: **133 passed, 2 skipped**; C API examples: **10 passed**.
- Separate Debug build with `WERROR=ON`, `WITH_WASI=ON`, and `BUILD_UVWASI=ON`: passed; interpreter/help tests: **87 passed**.
- Built uvwasi/libuv with tools disabled, then compiled Dhrystone with WASI SDK 20, translated it with wasm2c, linked it with the preserved libraries, and ran it successfully with its result checks reporting `O.K.`.
- Verified installed headers, absence of WASI integration in both interpreter binaries, formatting, and Python lint.

An additional Release `WERROR=ON` attempt hit the existing unused-but-set `ok` variable warning in unchanged `spectest-interp.cc`; the normal Release build and Debug `WERROR=ON` build passed.

README.md
```diff
@@ -117,6 +117,18 @@ CMake documentation for more information.
...
+
+`wasm-interp` no longer provides WASIp1 host functions. The `--wasi`,
+`--env` (`-e`), and `--dir` (`-d`) options and trailing WASI program arguments
+have been removed, along with the `wabt/interp/interp-wasi.h` header.
+Ordinary WebAssembly interpretation and `wasm2c` are unchanged.
+
+Remove `-DWITH_WASI=ON` from existing build configurations. It now produces a
+CMake warning and does not enable WASI support. The bundled uvwasi dependency is
+still available separately with `-DBUILD_UVWASI=ON` for the
+[wasm2c Dhrystone benchmark](wasm2c/benchmarks/dhrystone/README.md).
```
sbc100, line 130: I'm not sure this needs to go in the top level README.   Maybe just skip this?

sbc100 (commented):
@shravanrn can you take a look at the wasm2c/benchmarks/dhrystone/ changes?

wasm2c/benchmarks/dhrystone/README.md
```diff
@@ -0,0 +1,28 @@
+# Dhrystone benchmark
+
+This standalone benchmark translates a WASIp1 module with `wasm2c` and supplies
+its own host functions using uvwasi. It does not use `wasm-interp`.
```
shravanrn, line 4: Remove "it does not use...". Documentation should state what the tool does not what features are removed

wasm2c/benchmarks/dhrystone/README.md
```diff
@@ -0,0 +1,28 @@
...
+uvwasi/libuv dependencies from the WABT repository root:
+
+```sh
+git submodule update --init --recursive
+cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DBUILD_UVWASI=ON
+cmake --build build --target wasm2c wasm2c-copy-to-bin uvwasi_a
+```
+
+The explicit `uvwasi_a` target also builds `uv_a`. These optional libraries are
+not built by the default target and are not linked into `wasm-interp`.
```
shravanrn, line 16: Same as other comment

README.md
```diff
@@ -117,6 +117,18 @@ CMake documentation for more information.
 since the build produces an executable called `wasm2c` which conflicts with the
 `wasm2c` directory.
 
+### WASI compatibility
+
+`wasm-interp` no longer provides WASIp1 host functions. The `--wasi`,
```
shravanrn, line 122: Same as other comment

shravanrn (changes requested):

PR #2846, opened 2026-09-06
ci: remove unused Windows ninja install and modernize CMake invocation

kevmoo:
- Remove the unused `choco install ninja` step on `windows-latest` (added in `a08df5c9` but never used since Windows CMake defaults to Visual Studio).
- Modernize CMake configuration across all platforms by using `-S . -B out` instead of manually creating `out/` and setting `working-directory: out`.

Split out from #2717.


sbc100:
IIRC the main thing this change does is convert the windows build to use Ninja.  Previously we were using the cmake default which on windows means Visual Studio.  

Are we sure we want to use Ninja on Windows rather than Visual Studio.. its seems fine I guess (since the underlying compiler is still MSVC), but most windows users will use Visual Studio I imagine?

In any case the `Fix the Windows build step missing the -G Ninja` part of the description seem wrong.  Can you update that?

kevmoo:
Good catch! The only reason I originally added `-G Ninja` on Windows was because `choco install ninja` was already present in `build.yml` (from `a08df5c9`), so I assumed Ninja was intended on Windows too.

Keeping the default Visual Studio generator on Windows makes much more sense since that's what Windows users actually use. Updated this PR to:
- Keep the default Visual Studio CMake generator on Windows (reverting the Ninja/MSVC workarounds).
- Remove the unused `choco install ninja` step on Windows.
- Keep the `-S . -B out` CMake modernization across all platforms.

.github/workflows/build.yml
```diff
@@ -45,22 +45,16 @@ jobs:
     - name: install ninja (osx)
       run: brew install ninja
       if: matrix.os == 'macos-latest'
-    - name: install ninja (win)
-      run: choco install ninja
-      if: matrix.os == 'windows-latest'
-    - name: mkdir
-      run: mkdir -p out
     - name: tool versions
```
sbc100: Make rename this to cmake version?

sbc100 (approved):

sbc100:
lgtm.

Can you update the PR description to be the one you want to use when this lands. i.e. maybe remove the "Keep the default Visual Studio.." part?

.github/workflows/build.yml
kevmoo: Done! Renamed to `cmake version`.

kevmoo:
Updated the step name to `cmake version` and updated the PR description for landing—thanks!

PR #2848, opened 2026-09-07
Fix validating imported global refs

zherczeg:
Found by George Vasiliades


zherczeg:
@Poseidonas fix for a bug you have found.

Poseidonas:
Thanks, Zoltan — and thanks for taking it upstream. That reference-type index check is an easy one to miss, so it makes sense the same gap turns up across implementations; catching it at the toolkit level is where it matters most. Glad it landed with a regression test, and happy to help however is useful.

test/parse/bad-ref-globals.txt
```diff
@@ -0,0 +1,18 @@
+;;; TOOL: wat2wasm
+;;; ARGS: --enable-function-references
+;;; ERROR: 1
+(module
+  (global (import "mod" "val") (ref 9997))
+  (global (ref null 9998) (ref.null 9998))
+)
+(;; STDERR ;;;
+out/test/parse/bad-ref-globals.txt:5:42: error: reference 9997 is out of range (max: 0) in globals
+  (global (import "mod" "val") (ref 9997))
```
zherczeg, line 10: Note: imported / not imported globals display the error location differently.

This could be fixed by passing `loc` instead of `GetLocation()`:
https://github.com/WebAssembly/wabt/blob/main/src/wast-parser.cc#L1865

There are other similar cases though.

zherczeg:
This issue is likely my fault. I did it when I implemented the function references proposal.

sbc100:
Perhaps  at `refs` to the PR title?  (Since this only applies to reference types, correct?)

sbc100 (approved):

PR #2850, opened 2026-09-08
Fix kHugeVal on AIX

### LINKED ISSUE #2849: `HUGE_VAL` and `HUGE_VALF` are not constant expressions on AIX
Even though `HUGE_VAL` and `HUGE_VALF` are defined to be described as constant expressions, on AIX they lean on reinterpret casts to a static variable, preventing them from being constant expressions. This is the only issue that prevents wabt from building on AIX.

```
[ 18%] Building CXX object CMakeFiles/wabt.dir/src/literal.cc.o
In file included from /opt/freeware/lib/gcc/powerpc-ibm-aix7.2.0.0/12/include/c++/cmath:45,
                 from /home/tcr/wabt/src/literal.cc:30:
/home/tcr/wabt/src/literal.cc:66:37: error: 'reinterpret_cast' is not a constant expression
   66 |   static constexpr float kHugeVal = HUGE_VALF;
      |                                     ^~~~~~~~~
/home/tcr/wabt/src/literal.cc:79:37: error: 'reinterpret_cast' is not a constant expression
   79 |   static constexpr float kHugeVal = HUGE_VAL;
      |                                     ^~~~~~~~
make[2]: *** [CMakeFiles/wabt.dir/build.make:373: CMakeFiles/wabt.dir/src/literal.cc.o] Error 1
```

TaylorRichberger:
HUGE_VAL and HUGE_VALF on AIX are not constexpr.

Fixes: #2849


src/literal.cc
```diff
@@ -63,7 +63,13 @@ struct FloatTraitsBase<float> {
   using Uint = uint32_t;
   static constexpr int kBits = sizeof(Uint) * 8;
   static constexpr int kSigBits = 23;
-  static constexpr float kHugeVal = HUGE_VALF;
```
sbc100, line 66: How about instead we do something like:

```
#ifdef AIX
// Can't use constexpr on AIX because ...
static const float kHugeVal = HUGE_VALF;
#else
static constexpr float kHugeVal = HUGE_VALF;
#endif
```

Also, is there chance AIX might fix this?   Can you link to a bug?

TaylorRichberger: Yes, that should work, with the obvious caveat that it prevents the value from being constexpr. The define is `_AIX`.

sbc100: SGTM

TaylorRichberger:
@sbc100  I tried the const instead of constexpr change, and that fails with this error:

```
/tmp/selfgz120590983692/sard/deps/wabt/src/literal.cc:70:22: error: 'constexpr' needed for in-class initialization of static data member 'const float wabt::{anonymous}::FloatTraitsBase<float>::kHugeVal' of non-integral type [-fpermissive]
   70 |   static const float kHugeVal = HUGE_VALF;
      |                      ^~~~~~~~
In file included from /opt/freeware/lib/gcc/powerpc-ibm-aix7.2.0.0/12/include/c++/cmath:45,
                 from /tmp/selfgz120590983692/sard/deps/wabt/src/literal.cc:30:
/tmp/selfgz120590983692/sard/deps/wabt/src/literal.cc:70:33: error: 'reinterpret_cast' is not a constant expression
   70 |   static const float kHugeVal = HUGE_VALF;
      |                                 ^~~~~~~~~
/tmp/selfgz120590983692/sard/deps/wabt/src/literal.cc:88:23: error: 'constexpr' needed for in-class initialization of static data member 'const double wabt::{anonymous}::FloatTraitsBase<double>::kHugeVal' of non-integral type [-fpermissive]
   88 |   static const double kHugeVal = HUGE_VAL;
      |                       ^~~~~~~~
/tmp/selfgz120590983692/sard/deps/wabt/src/literal.cc:88:34: error: 'reinterpret_cast' is not a constant expression
   88 |   static const double kHugeVal = HUGE_VAL;
      |                                  ^~~~~~~~
```

I could go with the numeric_limits in a guard for AIX, so there is no change for non-AIX platforms.

TaylorRichberger:
I've gone and made that change, so the rest of the platforms stay the same (though I'm now noticing that `FloatTraitsBase<double>::kHugeVal` was a float and not a double before, which seems like a mistake to me).

src/literal.cc
```diff
@@ -63,7 +63,16 @@ struct FloatTraitsBase<float> {
   using Uint = uint32_t;
   static constexpr int kBits = sizeof(Uint) * 8;
   static constexpr int kSigBits = 23;
+#ifdef _AIX
+  // AIX defines HUGE_VAL and HUGE_VALF as reinterpret_cast expressions, which
+  // can't be constexpr.
+  static constexpr float kHugeVal = std::numeric_limits<float>::has_infinity
+                                        ? std::numeric_limits<float>::infinity()
+                                        : std::numeric_limits<float>::max();
```
sbc100, line 71: Can't we be sure that `std::numeric_limits<float>::has_infinity` is true? 

i.e. can't you just do `std::numeric_limits<float>::infinity()`?

Also, is HUGE_VALF defined to be the same thing as ::infinity?

sbc100: Alternatively can you just use `INFINITY` from `cmath.h`?

TaylorRichberger: > Can't we be sure that std::numeric_limits<float>::has_infinity is true?

Practically speaking, usually, and we're unlikely to hit a situation where it's not true here. The standard allows for non-IEEE 754 floating point numbers, which are allowed to not have infinity. `infinity()` is "Only meaningful if `std::numeric_limits<T>::has_infinity == true`". (https://en.cppreference.com/cpp/types/numeric_limits/infinity shows`HUGE_VALF` for floating types, but the standard doesn't say anything about that; it just says the value is not meaningful when `has_infinity == false`).

> Also, is HUGE_VALF defined to be the same thing as ::infinity?

On platforms with standard IEEE 754 floats, yes. HUGE_VALF is infinity if it's present, otherwise it's the largest finite number (i.e. the same as `std::numeric_limits<T>::max()`). I figured it made the most sense to replicate that.

> Alternatively can you just use INFINITY from cmath.h?

`INFINITY` is not defined when floating points don't have infinity.

But most of these are indeed edge cases. I was erring on the side of pedantic correctness.

sbc100: I don't think we need to worry about the edge cases too much.   Lets just assuming `INFINITY` exists and use that on AIX.  i can't imagine there are going to be other AIX users with a different cmath.h right?

TaylorRichberger: `INFINITY` also doesn't work on AIX for the same reason:

```sh
-bash-5.3$ OBJECT_MODE=64 g++-13 -oinftest ./inftest.cxx -std=c++11
In file included from /opt/freeware/lib/gcc/powerpc-ibm-aix7.2.0.0/13/include/c++/cmath:47,
                 from ./inftest.cxx:1:
./inftest.cxx:4:31: error: 'reinterpret_cast' is not a constant expression
    4 | static constexpr double inf = INFINITY;
```

I'd be happy enough to just use `std::numeric_limits<T>::infinity()` though.

sbc100: sgtm

sbc100 (approved):

PR #2852, opened 2026-09-09
Handle zero-sized literal output buffers

### LINKED ISSUE #2851: Zero-sized literal output buffers cause out-of-bounds access
## Description and impact

The shared implementation behind `WriteFloatHex()` and `WriteDoubleHex()`, as well as `WriteUint128()`, subtracts one from the output size while truncating a formatted value. When a caller supplies `size == 0`, that subtraction wraps to `SIZE_MAX`, which leads to an invalid copy length or array index.

With an AddressSanitizer and UndefinedBehaviorSanitizer build, the float and double wrappers abort with `AddressSanitizer: negative-size-param: (size=-1)`, while `WriteUint128()` aborts on an index of `18446744073709551615`. Without sanitizer checks, these accesses are undefined behavior and can crash a process using libwabt.

## Reproduction

Minimal library-level reproduction; run the program once with each of the three calls enabled individually:

```cpp
#include "wabt/literal.h"

int main() {
  char output = 'x';
  wabt::WriteFloatHex(&output, 0, 0);
  // wabt::WriteDoubleHex(&output, 0, 0);
  // wabt::WriteUint128(&output, 0, {0, 0, 0, 0});
}
```

## Affected surface

The affected surface is the public libwabt functions `WriteFloatHex()`, `WriteDoubleHex()`, and `WriteUint128()`. Their public declarations do not document a `size > 0` precondition. I did not find a bundled CLI call site that supplies a zero-sized buffer; the current in-tree call sites use nonzero fixed-size buffers.

## Required WebAssembly features

None. No `--enable` flag is needed.

## Candidate fix and validation

I have a small fix that returns before accessing the output for `size == 0`, plus GoogleTests covering all three public functions in the regular unit-test target. The two focused regressions, all 137 WABT unit tests, a three-function ASan/UBSan harness, and the full `WERROR=ON` CMake build and `check` target pass on macOS with AppleClang.

I did not run the exhaustive 32-bit `hexfloat_test` input space.

Sn0wyDay:
Fixes #2851.

The shared writer behind `WriteFloatHex()` and `WriteDoubleHex()` underflows `size - 1` when called with a zero-sized output buffer. `WriteUint128()` has the same zero-size problem when it indexes `size - 1`. This can produce an invalid copy length or array index.

Scope:

- Current in-tree callers of `WriteFloatHex()` and `WriteDoubleHex()` pass nonzero fixed-size buffers.
- `WriteUint128()` has no current non-test in-tree caller.
- I found no `.wat`/`.wasm` or bundled CLI input path that can make `size` zero.
- WABT does not document every declaration in the installed headers as a stable public API. This is therefore a defensive library-helper robustness fix, not an input-triggerable security fix.

Return immediately for `size == 0` in both implementations and add regressions covering all three functions.

Validation:

- Focused zero-size regressions: 2/2 pass.
- Three-function ASan/UBSan zero-size harness: pass; all output sentinels preserved.
- WABT unit tests: 137/137 pass.
- Full `WERROR=ON` CMake build and `check` target: pass on macOS with AppleClang.


zherczeg:
Are these user functions? Or just internal functions?

Sn0wyDay:
Those are public libwabt C++ functions: all three are declared in the installed include/wabt/literal.h, so an embedding application can call them. They are not directly exposed through the WABT command-line tools, though. All current in-tree production call sites pass nonzero fixed-size buffers, and I found no .wat/.wasm input path that can make size zero. So the affected case is a library caller explicitly passing size == 0, rather than an input-triggerable issue in the bundled tools.

sbc100:
If these function are not used we should just delete them instead of fixing them.

Sn0wyDay:
Thanks. I rechecked the usage on current `main`.

`WriteFloatHex()` and `WriteDoubleHex()` are each used at five non-test call sites in `wat-writer.cc`, `binary-reader-objdump.cc`, and `binary-reader-stats.cc`. Removing them causes the build to fail.

`WriteUint128()` has no current non-test in-tree caller. However, all three functions are declared in the installed `include/wabt/literal.h` and present in the installed static library, so removing them could break external source or link compatibility. I found no public downstream caller for `WriteUint128()`, but I cannot rule out private or unindexed users.

My preference is therefore to keep the small zero-size guards. If you prefer removing the unused exposed function despite that compatibility risk, I can delete only `WriteUint128()` and its tests while retaining the Float/Double guards.

zherczeg:
The `include/wabt` expose the entire library, and that includes many helper functions. You can misuse at least half of them to create crashes.
Is there a rule for what is public API, and what is internal?

Sn0wyDay:
Good question. I could not find a documented rule that distinguishes stable public API from internal helpers. The build installs the full `include/wabt` tree, but that alone does not establish that every declaration is a supported, stable API. My earlier wording was therefore too strong, and I updated the PR description.

The concrete usage is: `WriteFloatHex()` and `WriteDoubleHex()` have in-tree production callers, all with nonzero fixed-size buffers; `WriteUint128()` currently has only test callers. I also found no CLI or input-triggered path to `size == 0`, so this PR is only a small library-helper robustness change.

I have not removed `WriteUint128()` without a clear project policy or an explicit request. If removing that unused helper is the preferred direction, I can revise the PR accordingly.

Sn0wyDay:
@zherczeg I updated the PR and linked issue wording to reflect the scope discussed here: these are installed-header helpers, not established stable public API, and there is no known bundled CLI or input-triggered path. The code and tests are unchanged; CI is green. Could you take another look when convenient?

Issue #13, opened 2015-12-21
br-block-named.txt is incorrect


kg:
Specifically:

```
          (block $inner     ;; 0
            (br $inner)
            (br $outer)))))))
```

->

```
0000013: 01                                         ; OPCODE_BLOCK
0000014: 02                                         ; num expressions
0000015: 06                                         ; OPCODE_BR
0000016: 00                                         ; break depth
0000017: 00                                         ; OPCODE_NOP
0000018: 06                                         ; OPCODE_BR
0000019: 04                                         ; break depth
000001a: 00                                         ; OPCODE_NOP
```

The block has 2 child expressions but four expressions are emitted. When parsing this produces ```br 0``` followed by ```nop``` and then the function body has a trailing ```br 4```, ```nop``` which is not only an incorrect decoding but invalid (due to the break depth).

kg:
A form of this also appears in br-loop-inner-expr but it's different - not every ```br``` produces a ```nop```. So even if this is intended, it is unclear how to parse it correctly.

binji:
in v8-native binary format, the br operation is defined as always taking an expression value, so I believe this is correct. `(br <depth>)` just uses nop for the expression value, but in `br-loop-inner-expr` line 8 uses `(i32.const 4)` as the expression value so
```
      (if (i32.const 3)
        (br $exit (i32.const 4)))
```
is translated as
```
0000013: 03                                         ; OPCODE_IF
0000014: 09                                         ; OPCODE_I8_CONST
0000015: 03                                         ; u8 literal
0000016: 06                                         ; OPCODE_BR
0000017: 01                                         ; break depth
0000018: 09                                         ; OPCODE_I8_CONST
0000019: 04                                         ; u8 literal
```

kg:
Why does br take a value? We should call this out in the design, because it is not stated. I do see it in ml-proto.

binji:
It's referenced here: https://github.com/WebAssembly/design/blob/master/AstSemantics.md#branches-and-nesting
> Branches that exit a block, loop, or tableswitch may take a subexpression that yields a value for the exited construct.

jcbeyler:
And you can find examples in the testsuite in labels.wast and switch.wast

PR #2695, opened 2026-02-13
Implement the basics of component model

zherczeg:
This is an experimental (draft) patch. I don't know how much it fits into the concept of wabt. Wabt has the necessary helpers though.

Components are a layer (layer 1) above module layer (layer 0). It has no official specification, only explainers. It has a binary and text format similar to modules.

PR #2853, opened 2026-09-11
Validate WASI import arity before binding host functions

aizu-m:
ASan on `wasm-interp --wasi`, running a valid module that under-declares a WASI import:

    ERROR: AddressSanitizer: heap-buffer-overflow ... READ of size 8
      #1 wabt::interp::Value::CheckType(wabt::Type) const          interp.h:596
      #2 unsigned long long wabt::interp::Value::Get<...>() const  interp-inl.h:457
      #3 (anon)::WasiInstance::clock_time_get(...)                 interp-wasi.cc:198
      #4 (anon)::clock_time_get(...)                               wasi_api.def:26
      ...
      #13 wabt::interp::Thread::DoCall(...)                        interp.cc:2209
      #19 wabt::interp::WasiRunStart(...)                          interp-wasi.cc:750

Found this fuzzing `wasm-interp --wasi` with hand-built modules.

`WasiBindImports` binds each import to its host callback by name and keeps the
func type the module declared. There is a `// TODO(sbc): Validate signatures`
right next to it. `Thread::DoCall` then sizes the `params`/`results` vectors
from that declared type, but every callback indexes `params[]`/`results[]` by
the arity of the real WASI function.

So a module that imports `clock_time_get` as `(i32) -> i32` instead of
`(i32 i64 i32) -> i32` still validates, and the call reads `params[1]` past a
one-element vector. The `*_get` handlers also do `results[0].Set<...>()` on an
empty `results`, so a mis-declared result count is an out-of-bounds write as
well. Both are reachable from an untrusted module through `wasm-interp --wasi`.

Fix carries each function's `(num_params, num_results)` in `wasi_api.def` and
rejects a mismatched import before `HostFunc::New`, so the arity check fails at
bind time rather than at the out-of-bounds access. Arity only, deliberately:
some callbacks read i64 params as u32 (`path_open` rights, `fd_seek` offset),
so checking the types here would reject conforming modules. The declared arity
is fixed by the WASI ABI, so valid modules are unaffected.

Test declares `clock_time_get` with the wrong arity and expects the bind-time
rejection; the existing `test/wasi` cases still pass. `wasm-interp --wasi`
needs `-DWITH_WASI=ON`.


zherczeg:
In #2842, it was recommended to remove the entire wasi support form wabt. The 0.1 is outdated, and newer ones require component support (#2695).

aizu-m:
Read #2842. It deletes `src/interp/interp-wasi.cc` and `wasi_api.def` outright, so `WasiBindImports` and every callback that indexes `params[]`/`results[]` go with it. That removes the out-of-bounds path entirely, which is a cleaner result than this patch.

One observation: #2842 is still open, and until it lands the read/write stays reachable on main through `wasm-interp --wasi`. So I'll keep this open as a fallback and close it once the removal merges. If you'd rather not carry it in the meantime, I'll close it now.

PR #2855, opened 2026-09-14
wasm2c: Re-enable s390x CI with some tests disabled (Fix #2655)

### LINKED ISSUE #2655: test/wasm2c/spec/memory64/memory_copy.txt is failing on s390x

shravanrn:
This repurposes the existing slow marker to give test runners the option to skip slow tests which is used in s390x CI since some tests timeout in Github Actions


keithw (approved):
lgtm

PR #2857, opened 2026-09-15
wasm2c: (Security) ensure tailcalls use initialized instance references for all cases

shravanrn:
This PR fixes a case in the tail-call implementation in wasm2c where the `instance_ptr` is not initialized before being passed to a callee --- on a `ReturnCall` to a non-imported function. As an added defense in depth `instance_ptr` is explicitly initialized to zero to ensure such bugs would result in an unexploitable crash in the future.


test/wasm2c/tail-calls.txt
```diff
@@ -1094,7 +1094,7 @@ struct wasm_multi_id w2c_test_infiniteloop(w2c_test* instance, u32 var_p0, f64 v
   var_d1 = var_p1;
   static_assert(sizeof(struct wasm_multi_id) <= 1024);
   {
-    void *instance_ptr_storage;
+    void *instance_ptr_storage = 0;
```
zherczeg, line 1097: Why setting it 0 instead of `instance`? Then the line below is unnecessary.
And why 0 instead of `NULL` or `nullptr`? I know 0 is accepted, but looks strange.

sbc100 (commented):
lgtm % @zherczeg's questions.

Also, can we at a test?  Or is there some reason why this is tricky to test?

keithw:
Yeah, I'm a little confused why none of the tail-call tests would trigger this. Are there really zero test cases where a return_call is made to a function in the same module?? If so, that seems like what we need to be fixing...

test/wasm2c/tail-calls.txt
shravanrn: Just a stylistic choice. `nullptr` is C++ only.

shravanrn:
> Also, can we at a test?

> Yeah, I'm a little confused why none of the tail-call tests would trigger this. Are there really zero test cases where a return_call is made to a function in the same module?? If so, that seems like what we need to be fixing...

@keithw @sbc100 This will trigger only when a tail call into the same module also uses memory. Which is a combination I don't think is covered in the test suite. I have a test floating around somewhere that can trigger this, but I haven't merged this in yet. I can add that test in a separate commit, but I would like to land the security fix asap, as it may take a me a couple of days before I get free time to work on this again

sbc100 (approved):

keithw:
Yeah, I wouldn't add the test here -- it should be upstream.

sbc100:
We do quite often add tests here, even when there is a plan to push them upstream eventually.

lgtm either way though

PR #2861, opened 2026-09-16
wasm2c: (Security) fail closed when table allocation runs out of memory

shravanrn:
Cleanup of #2858 preserving @sxvncry authorship on commit


shravanrn:
Will follow up separately on test integration for this along with for #2857 later this week

@sbc100 Please approve as soon as reasonable. We are trying to sync wasm2c in Firefox with the latest version here, and I'd love to get this fix in too (It doesn't affect Firefox, but not including upstream fixes mean there end up being a lot of false security reports filed on using out of date versions)

sbc100 (approved):

sxvncry:
Thanks for merging the fix. Is there anything else I need to do from my side regarding this issue?

shravanrn:
> Thanks for merging the fix. Is there anything else I need to do from my side regarding this issue?

@sxvncry Nope, we're all set. If you have time to contribute a test too, that is always appreciated :)  If not no worries, I will probably get to it in the next few days!

sxvncry:
Hi @shravanrn, one quick follow-up regarding the security issue: since I originally reported #2854 and the fix was merged through #2861, is there any security disclosure or reporter attribution process I should be aware of?

shravanrn:
> Hi @shravanrn, one quick follow-up regarding the security issue: since I originally reported #2854 and the fix was merged through #2861, is there any security disclosure or reporter attribution process I should be aware of?

@sxvncry Nope, unfortunately wasm2c does not have a formal process for security bugs or acknowledgments at the moment. It would be nice to have a better systeen. But for now, it is up to any dependent projects to stay up to date and assess if bugs affect them (For instance, I know this bug does not affect Firefox since Firefox does not compile wasm files from adversarial sources with wasm2c)

PR #2862, opened 2026-09-18
Fall back to strtof/strtod on OpenBSD.

glandium:
b7ad849395485e6adf9ab2bfad8226741038690d switched from strtof/strtod to strtof_l/strtod_l, which are not available on OpenBSD.


landryb:
fwiw i've tested the patch in the context of https://phabricator.services.mozilla.com/D327199 and it fixes the build on OpenBSD.

I've briefly looked and i think the same fix might be needed for NetBSD - FreeBSD seems to have xlocale ?

sbc100:
Seems reasonable.

How does locale-specific number parsing work on OpenBSD then?   Does it simply not work?

sbc100 (approved):

landryb:
> How does locale-specific number parsing work on OpenBSD then? Does it simply not work?

i dunno the specifics/not a specialist of this area but i think there's none.

PR #2864, opened 2026-09-18
Reject named reference value types without function-references

aizu-m:
With exceptions on but typed function references off (the default feature set), PeekMatchRefType() admits a `(ref ...)` value type, so a named typed reference such as `(ref $t)` in a param, result, local or tag-param list reaches the `is_name()` branch of ParseValueTypeList and ParseBoundValueTypeList and trips `assert(function_references_enabled())` (an abort in assert-enabled builds, reachable from untrusted .wat through wat2wasm/wast2json with no flags), so both sites now raise the same `value type not allowed` parse error the parser already uses for other feature-gated value types instead of asserting, leaving valid `--enable-function-references` modules and the `(ref 0)`/element-type paths unchanged.


zherczeg:
Why rejecting named refs only? Without typed function references `(ref ...)` should not work.

PR #2865, opened 2026-09-21
Support empty quoted modules with annotations

zherczeg:


zherczeg:
This change allows running the following tests:
https://github.com/WebAssembly/testsuite/blob/main/annotations.wast#L32

This change does not change the handling of empty modules in the library.

src/wast-parser.cc
```diff
@@ -3931,6 +3931,18 @@ Result WastParser::ParseModuleCommand(Script* script, CommandPtr* out_command) {
       std::unique_ptr<WastLexer> lexer = WastLexer::CreateBufferLexer(
           filename, qsm->data.data(), qsm->data.size(), &errors);
       auto result = ParseWatModule(lexer.get(), &m, &errors, options_);
+
+      // Empty modules with annotations are accepted.
+      if (errors.size() == 1 && errors[0].error_level == ErrorLevel::Warning &&
+          errors[0].message == "empty module") {
```
sbc100, line 3937: This seems a little hacky to generate the warning and then remove it.    Is this worth the benefit?  I.e. do we have such modules in the wild that we want to support without warnings?

zherczeg: What do you suggest instead?
This affect spec tests (Quoted modules) only.

sbc100: Is there some way we could simple not generate the "empty module" warning at all in the case that annotation is present?  I assume that is not simple otherwise you would have done it already?

zherczeg: Yes. The next token functions silently consumes the unknown annotations. Adding a bool which is set if an annotation is encountered seems the easiest solution, but this is a warning, and probably useful in the general case. Warnings converted to errors by the spec test executor for some reason. We could also ignore warnings though.

sbc100: I think the empty module warning does seem reasonable, so maybe finding a way to suppress it in the spec runner is the best way to do.     Presumably its OK if the spec tests contain warnings?

zherczeg: What are the purpose of warnings btw? We only have 3 types of warnings at the moment: empty module, empty script, and custom section reading error in binary reader.

As for me warnings represent those cases, which are valid according to the specification, but still useful to tell the user a possible issue. From this perspective, spec tests might contain warnings, and they are still valid.

To reduce human error, I would only suppress as few warnings as possible. Contents of the spec tests should be ok, but we might make errors when processing them. Warnings could help to reveal that.

